Issue #17, opened 2017-02-15
de_web_plugin.h:30: Error: Undefined interface


mahowi:
Compilation of deconz-rest-plugin (v2.04.35) fails with
```
de_web_plugin.h:30: Error: Undefined interface
Makefile.Release:224: recipe for target 'release/moc_de_web_plugin.cpp' failed
make[1]: *** [release/moc_de_web_plugin.cpp] Error 1
make[1]: Leaving directory '/usr/local/src/deconz-rest-plugin
Makefile:34: recipe for target 'release' failed
make: *** [release] Error 2

```
There were no errors when compiling previous versions.

mahowi:
I have found the cause for the failure. It was commit a86a620c2a121f080242bc0073955ce584c8d4f6 from Nov. 13. When hard-coded /usr/include is disabled, deconz.h isn't found anymore.
I have enabled it again and compilation works. :smile:

PR #18, opened 2017-02-15
add /usr/include to INCLUDES; revert commit a86a620

mahowi:
Reverted commit a86a620c2a121f080242bc0073955ce584c8d4f6 (see issue #17 )


manup:
Hey Mahowi,
thanks for looking into this.

After researching a bit it seems that the Qt MOC compiler doesn't look into the default system search paths like /usr/include and hence can't figure out where the header files are located.

Issue #19, opened 2017-03-11
UPnP discovery of deCONZ REST plugin


ebaauw:
In response to M-SEARCH, the deCONZ REST plugin identifies itself as `SERVER: FreeRTOS/7.4.2, UPnP/1.0, IpBridge/1.8.0` ([upnp.cpp, line 128](https://github.com/dresden-elektronik/deconz-rest-plugin/blob/master/upnp.cpp#L128)).  However in the ssdp:alive message, it uses `SERVER: FreeRTOS/6.0.5, UPnP/1.0, IpBridge/0.1` ([line 89](https://github.com/dresden-elektronik/deconz-rest-plugin/blob/master/upnp.cpp#L89)).

The Philips Hue bridge includes a line `hue-bridgeid: 001788FFFExxxxxx` (the bridge's mac address on the IP network), which makes it easier to identify the bridge.  Could deCONZ use something similar?

manup:
It should be possible by using SSDP header field extensions as described in the [spec](http://www.upnp.org/specs/arch/UPnP-arch-DeviceArchitecture-v1.1.pdf).

We'll figure something out.

ebaauw:
Thanks!

ebaauw:
I've reopened this issue, as I've found some more issues with deCONZ discovery, notably inconsistencies between the info provided through UPnP, through the dresden electronik portal, through the XML description and through the `/config` resource.  See the Wiki page on [Bridge Discovery](https://github.com/dresden-elektronik/deconz-rest-plugin/wiki/Bridge-Discovery).

I would really like deCONZ to support the following:
- An unauthenticated GET for `/config`;
- Inclusion of `modelid` and `bridgeid` in the `/config` response (both authenticated and unauthenticated);
- Consistent values of `modelid` (`modelNumber`), `bridgeid` (`id`, `serialNumber`), `apiversion` across UPnP, the portal, the XML description, and the `/config` resource.

Also, there's a space missing behind the colon of the `GWID.phoscon.de` header field.  I'm not sure if that's required, but all the other header lines have one.

ebaauw:
I still have issue with UPnP in v2.04.40:

When doing a search for rootdevices, deCONZ responds with:
```
HTTP/1.1 200 OK
CACHE-CONTROL: max-age=100
EXT:
LOCATION: http://192.168.76.17:80/description.xml
[...]

PR #26, opened 2017-04-15
Add some REST config properties to match Hue

LunaSela:
Add some REST config properties to match the properties of Hue bridge

NOTE: "UTC" is upper case on Hue, should be on deCONZ too as property names might get deserialized incorrectly on some frameworks if the case is incorrect

### DIFF
--- a/rest_configuration.cpp
+++ b/rest_configuration.cpp
@@ -240,6 +240,10 @@ void DeRestPluginPrivate::configToMap(const ApiRequest &req, QVariantMap &map)
     bool ok;
     QVariantMap whitelist;
     QVariantMap swupdate;
+    QVariantMap devicetypes;
+    QVariantMap portalstate;
+    QVariantMap internetservices;
+    QVariantMap backup;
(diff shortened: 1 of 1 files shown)


ebaauw:
Cool!  I'm particularly happy about `bridgeid` and `modelid`.

See also issue #19 and the [Bridge Discovery](https://github.com/dresden-elektronik/deconz-rest-plugin/wiki/Bridge-Discovery) page I've created in the Wiki

LunaSela:
Thanks, I hope i figured out "bridgeid" correctly. 

I am considering adding unauthenticated config later too. At the moment this missing feature seems the next step that prevents me from adding deCONZ bridge to an app

manup:
Hi, thanks, I'll merge the pull request later.

As for the unauthenticated addition please wait with it for now, this is already in the works.
It will return a very limited attribute set, to not breach stuff to *evil* websites which scan local networks.

manup:
PR is merged I've made some white-space adjustments and added the use of shield mac address, this allows the implementation of 'replacesbridgeid' in case the shield must be replaced.

Issue #41, opened 2017-05-12
https://dresden-light.appspot.com/discover doesn't work over IPv4


ebaauw:
When accessed over IPv4,  a GET of `https://dresden-light.appspot.com/discover` returns an empty array, `[]`, even though a deCONZ gateway has been registered.

To reproduce:
```bash
curl -4 -s -H "Content-Type: application/json" "https://dresden-light.appspot.com/discover"
```
returns
```json
[]
```

At the same time,
```bash
curl -6 -s -H "Content-Type: application/json" "https://dresden-light.appspot.com/discover"
```
returns (my formatting and masking)
```json
[
  {
    "internalipaddress": "192.168.xxx.xxx",
    "internalport": 80,
    "id": "b8:27:eb:xx:xx:xx",
    "name": "pi",
    "macaddress": "b8:27:eb:xx:xx:xx"
  }
]
```

PR #44, opened 2017-05-21
Gateway Discovery

ebaauw:
See issues #19, #41.

Versions:
- Bumped gateway version to 2.04.43;
- Bumped API version to 1.0.1;
- API version in `de_web.pro` instead of hard coded.

Bridge ID:
- Added variable `gwBridgeId` for consistent use in `config.bridgeid`, UPnP discovery, appspot discovery, and `description.xml`.
-  `gwBridgeId` is set to the gateway device’s ZIgBee address, after this device has been discovered.

UPnP discovery:
- Fixed issue #19 (space after `GWID.phoscon.de:` and `upnp:rootdevice`);
- Don’t respond to M-SEARCH and don’t announce before `gwBridgeId` has been set.

appspot discovery:
- Don’t announce before `gwBridgeId` has been set;
- Note that deCONZ only sends `mac` to appspot, which exposes this value as both `macaddress` and `id`.  I think `macaddress` should be dropped (the meethue portal only exposes `id` and `internalipaddress`).

`description.xml`:
- Added `{{BRIDGEID}}` placeholder for `description_in.xml`;
- `descriptionXml` created after `gwBridgeId` has been set;
- `description_in.xml` needs to be updated, but is not included with the REST API plugin development package:
```xml
<?xml version="1.0"?>
<root xmlns="urn:schemas-upnp-org:device-1-0">
  <specVersion>
    <major>1</major>
    <minor>0</minor>
  </specVersion>
  <URLBase>http://{{IPADDRESS}}:{{PORT}}/</URLBase>
  <device>
    <deviceType>urn:schemas-upnp-org:device:Basic:1</deviceType>
    <friendlyName>{{GWNAME}} ({{IPADDRESS}})</friendlyName>
    <manufacturer>dresden elektronik</manufacturer>
    <manufacturerURL>https://www.dresden-elektronik.de/?L=1</manufacturerURL>
    <modelDescription>dresden elektronik Wireless Light Control</modelDescription>
    <modelName>Wireless Light Control</modelName>
    <modelNumber>deCONZ</modelNumber>
    <modelURL>https://www.dresden-elektronik.de/funktechnik/solutions/wireless-light-control/webapp/?L=1</modelURL>
    <serialNumber>{{BRIDGEID}}</serialNumber>
    <UDN>uuid:{{UUID}}</UDN>
    <gatewayName>{{GWNAME}}</gatewayName>
    <presentationURL>index.html</presentationURL>
    <iconList>
      <icon>
        <mimetype>image/png</mimetype>
        <height>48</height>
        <width>48</width>
        <depth>24</depth>
        <url>hue_logo_0.png</url>
      </icon>
      <icon>
        <mimetype>image/png</mimetype>
        <height>120</height>
        <width>120</width>
        <depth>24</depth>
        <url>hue_logo_3.png</url>
      </icon>
    </iconList>
  </device>
</root>
```

GET `/api/<apikey>/config`:
- Only include `bridgeid` in response after `gwBridgeId` has been set;
- `swversion` returns full semver gateway version (incl. dots);
- Added `starterkitid` (Hue API v1.18.0);

GET `/api/config`:
- Added unauthenticated request for basic configuration:
```json
{
  "apiversion": "1.0.1",
  "bridgeid": "00212EFFFFxxxxxx",
  "datastoreversion": "60",
  "factorynew": false,
  "mac": "b8:27:eb:xx:xx:xx",
  "modelid": "deCONZ",
  "name": "pi",
  "replacesbridgeid": null,
  "starterkitid": "",
  "swversion": "2.04.43"
}
```
- Only include `bridgeid` in response after `gwBridgeId` has been set.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,5 @@
+Makefile
+Makefile.Debug
+Makefile.Release
+release
+ui_de_web_widget.h
--- a/de_web.pro
+++ b/de_web.pro
@@ -57,7 +57,7 @@ GIT_COMMIT = $$system("git rev-list HEAD --max-count=1")
 
 # Version Major.Minor.Build
 # Important: don't change the format of this line since it's parsed by scripts!
-DEFINES += GW_SW_VERSION=\\\"2.04.42\\\"
+DEFINES += GW_SW_VERSION=\\\"2.04.43\\\"
 DEFINES += GIT_COMMMIT=\\\"$$GIT_COMMIT\\\" \
 
 # Minimum version of the RaspBee firmware
--- a/de_web.pro
+++ b/de_web.pro
@@ -58,6 +58,7 @@ GIT_COMMIT = $$system("git rev-list HEAD --max-count=1")
 # Version Major.Minor.Build
 # Important: don't change the format of this line since it's parsed by scripts!
 DEFINES += GW_SW_VERSION=\\\"2.04.43\\\"
+DEFINES += GW_API_VERSION=\\\"1.0.1\\\"
 DEFINES += GIT_COMMMIT=\\\"$$GIT_COMMIT\\\" \
 
 # Minimum version of the RaspBee firmware
--- a/de_web_plugin.cpp
+++ b/de_web_plugin.cpp
@@ -7892,6 +7892,7 @@ void DeRestPlugin::idleTimerFired()
         d->gwDeviceAddress.setExt(d->apsCtrl->getParameter(deCONZ::ParamMacAddress));
         d->gwDeviceAddress.setNwk(d->apsCtrl->getParameter(deCONZ::ParamNwkAddress));
         d->gwBridgeId.sprintf("%016llX", (quint64)d->gwDeviceAddress.ext());
+        d->initDescriptionXml();
     }
 
     if (!pluginActive())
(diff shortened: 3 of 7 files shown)


manup:
Thanks well done! I reviewed the code, looks good so far.

The unauthenticated request might become configurable as optional in future, in order to prevent websites from doing evil things on the internet (thinking of a global "paranoid" security mode for deCONZ).

Preventing the internet announcement when gwBridgeId is not set is tricky, because the software update mechanism in the Web App depends on it (new versions are notified in the response). A missing gwBridgeId could happen for various reasons like: RaspBee firmware broken, no RaspBee or ConBee attached or a misconfigured UART. The update and discovery should not be blocked in these cases, therefore something like an fallback mechanism should be added. Maybe setting the gwBridgeId to the Raspberry mac address or the last known RaspBee or ConBee address.

ebaauw:
> The unauthenticated request might become configurable as optional in future, in order to prevent websites from doing evil things on the internet (thinking of a global "paranoid" security mode for deCONZ).

Better enable SSL first.  Without it, the whole `/api/<apikey>` security is a joke.

> Preventing the internet announcement when gwBridgeId is not set is tricky, because the software update mechanism in the Web App depends on it [...] therefore something like an fallback mechanism should be added

When using a fallback value, appspot returns multiple entries, so there must be a way for an app to recognise this by using a default value, instead of the mac adddress of the Raspberry.  Appspot returns an internal server error when sending `"mac": ""`, but `"mac": "0000000000000000"` seems to work.

PR #45, opened 2017-05-21
Gateway Discovery (2)

ebaauw:
Don’t block internet announcement when gwBridgeId hasn’t been set, see
#44.


ebaauw:
Never mind.  I'll create a new pull request.

PR #46, opened 2017-05-21
Gateway DIscovery (2)

ebaauw:
Follow-up for #44.

> Preventing the internet announcement when gwBridgeId is not set is tricky, because the software update mechanism in the Web App depends on it (new versions are notified in the response). A missing gwBridgeId could happen for various reasons like: RaspBee firmware broken, no RaspBee or ConBee attached or a misconfigured UART. The update and discovery should not be blocked in these cases, therefore something like an fallback mechanism should be added. Maybe setting the gwBridgeId to the Raspberry mac address or the last known RaspBee or ConBee address.

`gwBridgeId` is now persisted in the database when the gateway device has been discovered.  It's initialised to `0000000000000000` on startup, but it should be restored before the REST API advertises itself or responds to commands.  Using the same logic for appspot, UPnP, GET `/api/<apikey>/config`, GET `/api/config`, and `description.xml`, all now use the value of `gwBridgeId` unconditionally.  This would only be `0000000000000000` after a new installation or gateway reset before the gateway has been discovered for the first time.

### DIFF
--- a/database.cpp
+++ b/database.cpp
@@ -530,6 +530,14 @@ static int sqliteLoadConfigCallback(void *user, int ncols, char **colval , char 
             }
         }
     }
+    else if (strcmp(colval[0], "bridgeid") == 0)
+    {
+      if (!val.isEmpty())
+      {
+          d->gwConfig["bridgeid"] = val;
+          d->gwBridgeId = val;
+      }
+    }
     return 0;
 }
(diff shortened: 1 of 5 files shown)


manup:
Cool, I'll put it on our test machines, should work fine here too :)

PR #107, opened 2017-08-11
Enhancement CLIPLighLevel and ZHALight sensors

ebaauw:
- Add support for CLIPLightLevel sensor;
- Add support for `state.dark` and `state.daylight` to ZHALight and CLIPLightLevel sensors;
- Add event notification for `state.dark` and `state.daylight` (only when these values are changed);
- Only allow updates to `state.lightlevel`, as `state.lux`, `state.dark`, and `state.daylight` are computed.
- TODO: `config.tholddark` and `config.tholdoffset`.  For now these are hard-coded to 12000 and 4000.

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,19 +24,19 @@ Packages for Qt4 and Raspbian Wheezy are available but not described here.
 ##### Install deCONZ and development package
 1. Download deCONZ package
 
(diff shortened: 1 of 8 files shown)


manup:
Awesome, again many thanks! :)
With all the recent additions I really need to schedule an update to the documentation soon.

Issue #30, opened 2017-04-23
Lights API differences with Hue API


ebaauw:
I've finished my compatibility testing of the `/lights` resource against the Hue bridge API, see the [Wiki](https://github.com/dresden-elektronik/deconz-rest-plugin/wiki/Lights).  I've found the following differences, for which I would need to make changes to my Hue application:
1. deCONZ exposes `state.bri` for _On/off plug-in units_ (and I assume _On/off lights_) that don't support brightness.
2. deCONZ exposes `state.ct` for _Colour lights_ that don't support colour temperature;
3. deCONZ exposes `state.hue`, `state.sat`, and `state.xy` for _Colour temperature lights_, that don't support colour;
4. deCONZ exposes `state.effect` for lights that don't support colour (_On/off plug-in units_, _Dimmable lights_, _Color temperature lights_);
5. Setting `"bri": 0` turns off the light.  On the Hue API, this sets the light to minimal brightness.  Note that setting `"bri": 254` in deCONZ doesn't turn the light on.
6. deCONZ seems to convert the values of the corresponding ZigBee attributes for `state.hue`, `state.sat`, and `state.xy`, where the Hue API uses the raw values;
7. deCONZ exposes `manufacturer` instead of `manufacturername`;

manup:
Thanks a lot for the detailed tests! The falsely exposed bri, ct, hue, sat, xy and effect attributes will be fixed soon, similar to the sensor rewritten resource system in the last beta the lights resources will also be refactored.

Turing off the lights shut only be done when "on": false is given, so that will be fixed too.

> deCONZ seems to convert the values of the corresponding ZigBee attributes for state.hue, state.sat, and state.xy, where the Hue API uses the raw values;

That's more complicated, right now I don't think we will change that until it's causing errors.

> deCONZ exposes manufacturer instead of manufacturername

Tricky, we need to check if apps already using our *version* before changing this, won't be ncie to break stuff.

LunaSela:
Although it might make code less readable, it might me sensible to make behaviour switchable per API key.

So deCONZ specific apps still would get "manufacturer" and others would get "manufacturername".

Or we could add both. And deprecate the "old" one. As there are certainly more apps which expect "manufacturername", doing something might be a benefit.

manup:
Yes I think the deprecate way is the most smooth way to adjust the API

ebaauw:
Thanks @manup!

[...]

PR #115, opened 2017-08-13
ZHALightLevel and more...

ebaauw:
- Use sensor `type` ZHALightLevel instead of ZHALight.
- Also issue web socket notification for `config.battery` on read/report, when value hasn’t changed.
- Setup attribute reporting for color temperature 0x0300/0x0007 of lights and binding for Trådfri remote (commit 471679a - sorry wrong comment, but I can't seem to change it).  The color temperature works: I see dbg-info level 2 messages and the expected settings can be read back from the deCONZ GUI.  I'm not too sure about the Trådfri remote bindings - I don't see any info message for binding of cluster 0x0005.
- Bug fixes for mac address mask; Don't re-create IKEA remote rules when rules exist, but disabled.
- Change light `manufacturer` attribute to `manufacturername` (issue #30, item 7).

### DIFF
--- a/bindings.cpp
+++ b/bindings.cpp
@@ -1488,7 +1488,7 @@ void DeRestPluginPrivate::bindingToRuleTimerFired()
 
                 case ILLUMINANCE_MEASUREMENT_CLUSTER_ID:
                 {
-                    if (i->type() == "ZHALight")
+                    if (i->type() == "ZHALightLevel")
                     {
                         sensor = &(*i);
                     }
--- a/bindings.cpp
+++ b/bindings.cpp
@@ -595,6 +595,14 @@ bool DeRestPluginPrivate::sendConfigureReportingRequest(BindingTask &bt)
             reportableChange8bit = 1;
         }
     }
+    else if (bt.binding.clusterId == COLOR_CLUSTER_ID)
+    {
+        dataType = deCONZ::Zcl16BitUint;
+        attributeId = 0x0007; // color temperature
+        minInterval = 1;
+        maxInterval = 300;
+        reportableChange16bit = 1;
+    }
     else
     {
         return false;
@@ -709,6 +717,7 @@ void DeRestPluginPrivate::checkLightBindingsForAttributeReporting(LightNode *lig
         {
         case ONOFF_CLUSTER_ID:
         case LEVEL_CLUSTER_ID:
+        case COLOR_CLUSTER_ID:
         {
             DBG_Printf(DBG_INFO_L2, "create binding for attribute reporting of cluster 0x%04X\n", i->id());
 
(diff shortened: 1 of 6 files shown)


donnib:
@ebaauw i can't check it atm but does the events for remotes also give the battery status or are battery events separate only ?

ebaauw:
Separate events.

deCONZ doesn't combine multiple attributes into one event yet (with the exception of a sensor's `state.lastupdated`).  Even if it would, the sensors send a separate attribute reporting message for battery (as it's on a different cluster, so the updates arrive at a different time from the updates for presence, light level, temperature, or button event.

donnib:
@ebaauw thank you

donnib:
I forgot to ask, would it make sense to have battery for a CLIP Sensor ? I would use it for test purpose while i develop.

ebaauw:
Yes it would, not just for testing, but also if you use the CLIP sensor as a proxy for an external battery-powered sensor (that isn't connected to the ZigBee network).  If you include `"config": {"battery":100}` in the POST body while creating the CLIP sensor, the created resource should contain it and allow PUTs to change it.

Same for `reachable`.

donnib:
ahh cool, i didn't know that, will try that, i had only tried to do a `PUT` on `battery` and obviously it failed.

snozzlebert:
> Also issue web socket notification for ```config.battery``` on read/report, when value hasn’t changed.

Why? If the value isn't changed I won't expect ```"e":"changed"```. I think it is confusing.
Now I get this from my dimmer switch:
```
2017-08-16 08:03:45 {"config":{"battery":100},"e":"changed","id":"1","r":"sensors","t":"event"}
2017-08-16 08:08:46 {"config":{"battery":100},"e":"changed","id":"1","r":"sensors","t":"event"}
2017-08-16 08:13:47 {"config":{"battery":100},"e":"changed","id":"1","r":"sensors","t":"event"}
2017-08-16 08:18:48 {"config":{"battery":100},"e":"changed","id":"1","r":"sensors","t":"event"}
2017-08-16 08:23:49 {"config":{"battery":100},"e":"changed","id":"1","r":"sensors","t":"event"}
```

ebaauw:
You might be right...  and it becomes even more confusing when `config.duration` and `config.sensitivity` come into play.

My thought was to receive web socket events when the sensor sends an attribute reporting command to the gateway, so you could use these to build up a history of values.  By changing the attribute reporting setting, you could configure how often you'd get these events.  Also, somehow this seemed more consistent with the behaviour for the Hue motion sensor for the state attributes (of course in that case, `state.lastupdated` has changed even when `state.presence`, `state.lightlevel` or `state.temperature` haven't).

PR #126, opened 2017-08-20
API version 1.0.2: new config attributes

ebaauw:
Update API version to 1.0.2;
- Add `config.alert` for Hue dimmer switch, Hue motion sensor (all resources), TRÅDFRI remote, TRÅDFRI motion sensor — not yet functional;
- Add `config.sensitivity` and `config.sensitivitymax` (read-only) to Hue motion sensor (ZHAPresence resource) — not yet functional;
- Add `config.ledindication` and `config.usertest` to Hue motion sensor (all resources) — not yet functional;
- Make `config.battery` read-only for ZigBee sensors;
- Check for valid values when updating through API:
  - CLIPTemperature’s `state.temperature` (>= -27315);
  - CLIPHumidity’s `state.humidity` (0 .. 10000);
  - CLIPPressure’s `state.pressure` (>= 0).

Workaround for TRÅDFRI battery percentage, which doesn’t conform to ZigBee standard.

Code consistency:
- Data types for resources: RStateHumidity, RStatePressure, RStateLux;
- Check for Hue motion sensor (model is “SML001”)  and Hue dimmer switch (model starts with “RWL02”).


manup:
I'll add my changes on top of it now, hopefully no conflicts :)

ebaauw:
> hopefully no conflicts

Don't think there were?

tomky:
I'm wondering if we can make config.battery writeable? I have a sensor (lumi.sensor_switch) which never updates its battery status and therefore config.battery is null. I would like to change it to some positive value so HomeKit won't always reporting low battery warning.

PR #128, opened 2017-08-20
Implement config.alert for sensors

ebaauw:
Implement `config.alert` for sensors:
- Hue motion sensor;
- TRÅDFRI motion sensor;
- TRÅDFRI remote.

Remove `config.alert` from Hue dimmer switch, as it doesn’t support _Identify_ (even though it has a 0x0003 server cluster).

Change timings for light `state.alert` to match Hue bridge:
- `select`: 2s (Hue lights don’t respond when set to 1s);
- `lselect`: 15s.

### DIFF
--- a/database.cpp
+++ b/database.cpp
@@ -1808,8 +1808,6 @@ static int sqliteLoadAllSensorsCallback(void *user, int ncols, char **colval , c
                 sensor.fingerPrint().inClusters.push_back(VENDOR_CLUSTER_ID);
                 sensor.setNeedSaveDatabase(true);
             }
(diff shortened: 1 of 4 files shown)


manup:
Thanks, I guess these can be really useful in larger installations to find a device .

ebaauw:
I also use this as a confirmation when "disabling" the motion sensor when going to sleep (through a CLIPGenericStatus sensor, set by holding the _Off_ button on the dimmer switch or by Siri's _Goodnight_ scene).  I don't want my bedroom lights to turn on when I turn in my sleep, but I still want to register motion.

PR #134, opened 2017-08-23
Read/write of manufacturer specific attributes

ebaauw:
Add `manufacturerCode` as optional attribute to `DeRestPluginPrivate::readAttributes()` and `DeRestPluginPrivate::writeAttribute()`.

I want to try and set the Hue dimmer switch mode by issuing
```c++
writeAttribute(sensor, sensor->fingerPrint().endpoint, BASIC_CLUSTER_ID, 0x0031, 0x000b, VENDOR_PHILIPS);
```
However, I'm not sure what's the best place to add it.  Would it be in `DeRestPluginPrivate::delayedFastEnddeviceProbe()`?  There's already a check for the Hue dimmer switch near the end, to get the group.

### DIFF
--- a/database.cpp
+++ b/database.cpp
@@ -1808,8 +1808,6 @@ static int sqliteLoadAllSensorsCallback(void *user, int ncols, char **colval , c
                 sensor.fingerPrint().inClusters.push_back(VENDOR_CLUSTER_ID);
                 sensor.setNeedSaveDatabase(true);
             }
-            item = sensor.addItem(DataTypeString, RConfigAlert);
-            item->setValue(R_ALERT_DEFAULT);
         }
         else if (sensor.modelId() == QLatin1String("SML001")) // Hue motion sensor
         {
--- a/de_web_plugin.cpp
+++ b/de_web_plugin.cpp
@@ -2741,9 +2741,6 @@ void DeRestPluginPrivate::addSensorNode(const deCONZ::Node *node, const SensorFi
     {
         sensorNode.setManufacturer(QLatin1String("Philips"));
 
-        item = sensorNode.addItem(DataTypeString, RConfigAlert);
-        item->setValue(R_ALERT_DEFAULT);
-
         if (modelId.startsWith(QLatin1String("RWL02"))) // Hue dimmer switch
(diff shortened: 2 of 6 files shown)


manup:
The delayedFastEnddeviceProbe() method is only used during initial setup phase.

I think to make the write possible after a configuration through REST API should mark the sensor as *where is work todo, update attribute xyz* and then shortly after the sensor is awake the next time, due to an received report, the write can happen.

If I remember correctly Philips also has this *there is work todo* bookkeeping for their devices.

ebaauw:
> The delayedFastEnddeviceProbe() method is only used during initial setup phase.

I think that's good enough for this attribute?  It only needs to be set when the Hue dimmer is paired to the gateway.  I don't think it makes sense to expose it through the REST API.

> I think to make the write possible after a configuration through REST API should mark the sensor as where is work todo, update attribute xyz and then shortly after the sensor is awake the next time, due to an received report, the write can happen.

This is probably the pattern we need for `config.sensitivity` and `config.duration`.  The `writeAttribute()` would then be called from `DeRestPluginPrivate::updateSensorNode()`?

As an aside: when I send an _Identify_ (through setting `config.alert`) to the Hue motion sensor, it always reacts - even when its dormant.  The IKEA motion and remote only react once they wakeup (by detecting motion or by pressing a button).  I also find that reading attributes from the deCONZ GUI almost always works for the Hue motion sensor, but for the Hue dimmer switch and the IKEA devices only when they're woken up.  Maybe the Hue motion sensor sleeps more lightly than the other devices?  That would also explain why polling `config.duration` for the Hue motion sensor worked?  Is their a way to force-read the _Power Descriptor_ from the deCONZ GUI?  I wonder how the Hue motion sensor reports _Power Mode_.

> If I remember correctly Philips also has this there is work todo bookkeeping for their devices.

The added `config.pending` to their API in v1.15:
> Array of config parameters which is not yet committed to sensor.  As long as the atrribute is listed here, the configuration attribute value listed on GET does not take effect and might return to previous value.  A subsequent PUT on listed atrribute might return error 10.

PR #137, opened 2017-08-24
Hue dimmer fix

ebaauw:
Say the magic word to the Hue dimmer switch during setup so it no longer initiates a touchlink when holding the _On_ button.  Note: you need to delete and re-pair the dimmer.

It almost seems too simple to be true, but it seems to work.  I continue to get buttonevents and the group from the web app still seems to work.


manup:
Thanks, a small description what the magic does would be nice above the call :)

ebaauw:
But then it wouldn't be magic, would it?  ;-)

Issue #36, opened 2017-05-04
REST API doesn't find Ubisys D1 switch endpoints


ebaauw:
Today I installed the Ubisys D1 - I had to pull a neutral wire from the ceiling light point to the switch location.  The light endpoint (indeed ZHA) is recognised immediately by deCONZ.  
I added the light to the Trådfri remote group and now control my dining table lamp through that remote - very cool.

However, the two Dimmer switch endpoints of the D1 don't appear in the deCONZ REST API (v2_04_40).  I wired two pulse switches to their inputs, just to see how they'd work.  Not sure I'm gonna really need them, but I would be cool if they worked.  The first input is still bound to the light itself (factory default) and works as documented.

manup:
Yes, currently only the *lights* functionality of the Ubisys devices are supported. However all Ubisys devices have excellent support for attribute reporting and bindings, so exposing them to the realtime API should be possible. The whole switch configuration was refactored to be semi-generic, this should help to integrate the D1. No ETA but I put it on the list for a rainy weekend :-)

ebaauw:
Thanks, I'll be doing my rain dance then ;-)

ebaauw:
Somehow, the factory default binding (from the first Dimmer Switch endpoint to the Dimmable Light endpoint) got removed.  I was able to restore it manually using the deCONZ GUI, for the On/Off as well as for the Level Control cluster.  I also created binding for the second Dimmer Switch endpoint.

I cannot seem to control any other lights by creating bindings from the Dimmer Switch endpoints, though.  Would Ubisys be using custom commands (like the Trådfri remote for the Previous and Next buttons), that aren't recognised by the other lights?

ebaauw:
Any progress yet?

ebaauw:
PR #139.

I was trying to create one ZHASwitch resource with different button event ranges for the different _Dimmer Switch_ endpoints.  This would be exposed to HomeKit as two buttons with _Short Press_ and _Long Press_.  
However, I didn't get there.  Now it's two resources, one for each _Dimmer Switch_ endpoint, with button events 1002, and 1001/1003 on the first endpoint and 2002 and 2001/2003 on the second endpoint.  The distinction between _DimUp_ and _DimDown_ is lost with the current scheme (and when using one resource).

[...]

PR #141, opened 2017-08-25
buttonevent value for Stop

ebaauw:
Send the correct buttonevent value for _Stop_ to match the previous command, so 2003 after a 2001 (_Step Up_ or _Move Up_) and 3003 after a 3001 (_Step Down_ or _Move Down_):
- Rename `Sensor::previousButton` to `Sensor::previousDirection`, which better explains its purpose;
- Implement `previousDirection` for any switch using the 0x0008 cluster;
- Fix button map for the ubisys D1 (see #36);
- Adjust button maps for other switches to leverage `previousDirection`.

@manup, please review carefully:
- If I understand the code correctly, this PR should fix the x003 button events for the other switches as well, but I don't have any of these, so I cannot test this;
- The `deLightingSwitchMap` button map seems to have an inconsistent use of direction between the various modes and buttons.  Not sure if this is intentional or a bug;
- I daren't touch the `bjeSwitchMap` (also because comment "might be button x as well"  was missing), but I think it would merit to be adjusted as well.


manup:
>The deLightingSwitchMap button map seems to have an inconsistent use of direction between the various modes and buttons. Not sure if this is intentional or a bug;

Both! :) the bug is in the switch firmware therefore the button map is intention, so I'm afraid I can't merge this commit as it is. The *"might be button 2 as well"* comment means the related button can't be distinguished from the incoming command alone and the previous command must be taken into consideration, this will be added to the button handler in the next days.

>I daren't touch the bjeSwitchMap (also because comment "might be button x as well" was missing), but I think it would merit to be adjusted as well.

Not sure here, I'll check this next week in office, the Busch Jaeger switches can be configured in various ways.

ebaauw:
> The "might be button 2 as well" comment means the related button can't be distinguished from the incoming command alone and the previous command must be taken into consideration

Yes, I understood that.

> this will be added to the button handler in the next days.

This is what I already added in this PR, using the same scheme I used for the Trådfri remote's _Previous_ and _Next_ buttons (where the 0x09 command doesn't include an Up/Down direction parameter either).  The previous command is saved in `sensor->previousDirection`, and the `param` field in the button map is compared to this saved value instead of to the (missing) parameter of the ZCL command.

manup:
>This is what I already added in this PR, using the same scheme I used for the Trådfri remote's Previous and Next buttons (where the 0x09 command doesn't include an Up/Down direction parameter either). The previous command is saved in sensor->previousDirection, and the param field in the button map is compared to this saved value instead of to the (missing) parameter of the ZCL command.

Ah I see that's cool, but the button map for the dresden elektronik Lighting Switch shouldn't be changed, since it it never sends the command added in the PR — the parameter for the `stop` command (0x03) is always 0.

    { Sensor::ModeTwoGroups,        0x01, 0x0008, 0x03, 1,    S_BUTTON_2 + S_BUTTON_ACTION_LONG_RELEASED,  "Dimm stop" },

ebaauw:
> but the button map for the dresden elektronik Lighting Switch shouldn't be changed

I copied it and changed the copy, to distinguish between buttonevent 1003 and 2003.
```c++
{ Sensor::ModeTwoGroups,        0x01, 0x0008, 0x03, 0,    S_BUTTON_1 + S_BUTTON_ACTION_LONG_RELEASED,  "Dimm stop" },
{ Sensor::ModeTwoGroups,        0x01, 0x0008, 0x03, 1,    S_BUTTON_2 + S_BUTTON_ACTION_LONG_RELEASED,  "Dimm stop" },
```

> the parameter for the stop command (0x03) is always 0.

Actually, there is no parameter, as it is for the Stop with on/off command (0x07) used by the ubisys D1.  For these commands, the field from the button map isn't compared to the command parameter, but to the saved `previousDirection`:
```c++
            else if (ind.clusterId() == LEVEL_CLUSTER_ID &&
                       (zclFrame.commandId() == 0x03 ||  // stop
                        zclFrame.commandId() == 0x07) )  // stop (with on/off)
            {
                ok = false;
                if (buttonMap->zclParam0 == sensor->previousDirection) // direction of previous move/step
                {
                    sensor->previousDirection = 0xFF;
                    ok = true;
                }
            }
```

manup:
Ha ok I totally missed that :) brilliant I'll try it out tomorrow

manup:
Just testet it all events worked very well for the first endpoint, I had to adjust the second endpoint to to match the commands. 

https://github.com/dresden-elektronik/deconz-rest-plugin/commit/e74e57c4d97ca9474aea467b299b86239c582023

PR #148, opened 2017-08-27
Attribute reporting for Hue motion sensor config

ebaauw:
Still trying to get `config.sensitivity`, `config.senstivitymax`, `config.usertest` and `config.ledindication` to work.  This is as far as I got this weekend.

- Added the attributes to `updateSensorNode()`;
- Read them in `delayedFastEnddeviceProbe()` - not sure if this is a good idea: they're read multiple times (as in many) on device discovery;
- Added the _Basic_ cluster to `checkSensorBindingsForAttributeReporting()`, but somehow the binding isn't configured.  Don't know why not;
- Refactored `sendConfigureReportingRequest` to allow for:
  - Reporting config for multiple attributes per cluster;
  - a reportableChange of 0 (as issued by the Hue bridge);
  - manufacturer-specific attributes;
- Double-checked the used parameters against what Philips Hue bridge uses, notably _Battery Percentage_ (0x0001/0x0021).

I'm running production with this change now, it doesn't seem to have broken anything (although I only deleted / repaired a single motion sensor).

### DIFF
--- a/bindings.cpp
+++ b/bindings.cpp
@@ -576,6 +576,8 @@ bool DeRestPluginPrivate::sendConfigureReportingRequest(BindingTask &bt, Configu
         zclFrame.writeToStream(stream);
     }
 
+    DBG_Printf(DBG_INFO_L2, "configure reporting for 0x%016llX, attribute 0x%04X/0x%04X\n", bt.restNode->address().ext(), bt.binding.clusterId, rq.attributeId);
+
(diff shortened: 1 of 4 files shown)


manup:
Thanks a lot the refactoring looks very clean. I've added two guards to the Philips specific attributes https://github.com/dresden-elektronik/deconz-rest-plugin/commit/0d6f9325e9894409a3749a66a61cb4459110f324

PR #190, opened 2017-09-10
Fix `config.duration` for Hue motion sensor

ebaauw:


manup:
Shouldn't the attribute be rewritten if different?

The write is mainly there for the FLS-NB device (sensor light), therefore we can also whitelist the write attribute just for this device.

ebaauw:
> Shouldn't the attribute be rewritten if different?

I don't think it should, or you wouldn't be able to update the attribute from the GUI.  When updating the ZigBee attribute 0x0406/0x0010 from the deCONZ GUI, the REST attribute `config.duration` is updated.  When updating the REST attribute, the ZigBee attribute is updated.

> The write is mainly there for the FLS-NB device (sensor light), therefore we can also whitelist the write attribute just for this device.

Thanks, I wanted to do this (hence the awful != check), but I didn't know for sure.

PR #192, opened 2017-09-10
Hue motion sensor config - continued

ebaauw:
Add Basic cluster to Hue motion sensor fingerprint, so attribute reporting for ledindication and usertest will be setup.

Once attribute reporting has been setup, `config.ledindication` and `config.usertest` are indeed updated in all three resources (as I hoped).

For the Hue motion sensor, check attribute reporting for Occupancy Sensing cluster on sensitivity instead of occupied state (in the hope that attribute reporting for sensitivity is more likely to be setup).

### DIFF
--- a/bindings.cpp
+++ b/bindings.cpp
@@ -963,8 +963,14 @@ void DeRestPluginPrivate::checkSensorBindingsForAttributeReporting(Sensor *senso
         }
         else if (*i == OCCUPANCY_SENSING_CLUSTER_ID)
         {
-            val = sensor->getZclValue(*i, 0x0000); // occupied state
-            // val = sensor->getZclValue(*i, 0x0030); // sensitivity
+            if (sensor->modelId() == QLatin1String("SML001")) // Hue motion sensor
+            {
+                val = sensor->getZclValue(*i, 0x0030); // sensitivity
+            }
+            else
+            {
+                val = sensor->getZclValue(*i, 0x0000); // occupied state
+            }
         }
         else if (*i == POWER_CONFIGURATION_CLUSTER_ID)
         {
--- a/database.cpp
+++ b/database.cpp
@@ -1824,6 +1824,11 @@ static int sqliteLoadAllSensorsCallback(void *user, int ncols, char **colval , c
         }
         else if (sensor.modelId() == QLatin1String("SML001")) // Hue motion sensor
         {
+            if (!sensor.fingerPrint().hasInCluster(BASIC_CLUSTER_ID))
+            {
+                sensor.fingerPrint().inClusters.push_back(BASIC_CLUSTER_ID);
+                sensor.setNeedSaveDatabase(true);
+            }
             if (!sensor.fingerPrint().hasInCluster(POWER_CONFIGURATION_CLUSTER_ID))
             {
                 sensor.fingerPrint().inClusters.push_back(POWER_CONFIGURATION_CLUSTER_ID);
--- a/de_web_plugin.cpp
+++ b/de_web_plugin.cpp
@@ -3484,7 +3484,7 @@ void DeRestPluginPrivate::updateSensorNode(const deCONZ::NodeEvent &event)
                                 }
                                 updateSensorEtag(&*i);
                             }
-                            else if (i->modelId() != QLatin1String("SML001") && ia->id() == 0x0010) // occupied to unoccupied delay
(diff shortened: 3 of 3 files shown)


manup:
> For the Hue motion sensor, check attribute reporting for Occupancy Sensing cluster on sensitivity instead of occupied state (in the hope that attribute reporting for sensitivity is more likely to be setup).

Is this really safe, if no reports are received for occupied state reporting won't be reestablished?

ebaauw:
> Is this really safe, if no reports are received for occupied state reporting won't be reestablished?

Frankly, I don't know - I've only tested this on my test Pi with one Hue motion sensor.  The reporting configuration seems more stable than the binding, so I guess the chances are slim that sensitivity is being reported while occupancy state is not.  I'm OK if you want to skip this commit.

More fundamental, I don't think the reporting configuration should be linked to the binding response, as there's now multiple attributes reported for the same binding.  I was also thinking, for battery-powered devices, maybe the binding and reporting configuration should be called from `processZclAttributes()`, like writing the attributes, to ensure the device is awake.  However, that's too big a change for me to oversee the consequences.

ebaauw:
I think the attributes need to be defined in the ZCLDB for deCONZ to accept the reporting (cf. the Xiaomi smart cube).

```xml
	<cluster id="0000" name="Basic">
		<description>Attributes for determining basic information about a device, setting user device information such as description of location, and enabling a device.</description>
		<server>
			<attribute-set id="000" description="Basic Device Information">
				...
			</attribute-set>
			<attribute-set id="001" description="Basic Device Settings">
				...
			</attribute-set>
			<attribute-set id="003" description="Hue motion sensor settings">
				<attribute id="0032" name="User Test" type="bool" access="rw" default="0" vendor="0x100b" required="o">
				</attribute>
				<attribute id="0033" name="LED Indication" type="bool" access="rw" default="0" vendor="0x100b" required="o">
				</attribute>
			</attribute-set>
		</server>
	</cluster>

	<cluster id="0406" name="Occupancy sensing">
		<description><!-- TODO --></description>
		<server>
			<attribute-set id="000" description="Occupancy sensor information">
				...
			</attribute-set>
			<attribute-set id="001" description="PIR configuration">
				...
			</attribute-set>
			<attribute-set id="002" description="Ultrasonic configuration">
				...
			</attribute-set>
			<attribute-set id="003" description="Hue motion sensor configuration">
				<attribute id="0030" name="Sensitivity" type="u8" range="0x00,0xfe" access="rw" required="o" vendor="0x100b" default="0x00">
					<description>The sensitivity of the motion detection.  Lowest is 0, highest is specified by Sensitivity Max.</description>
				</attribute>
				<attribute id="0031" name="Sensitivity Max" type="u8" range="0x00,0xfe" access="r" required="o" vendor="0x100b" default="0x00">
					<description>The maximum value for Sensitivity.</description>
				</attribute>
			</attribute-set>
		</server>
	</cluster>
```

manup:
> Frankly, I don't know - I've only tested this on my test Pi with one Hue motion sensor. The reporting configuration seems more stable than the binding, so I guess the chances are slim that sensitivity is being reported while occupancy state is not. I'm OK if you want to skip this commit.

Mine reports the occupancy state very reliable but the setup is hasn't changed for a long time. I'll merge the PR, maybe later on both attributes should be checked for reports.

> More fundamental, I don't think the reporting configuration should be linked to the binding response, as there's now multiple attributes reported for the same binding. I was also thinking, for battery-powered devices, maybe the binding and reporting configuration should be called from processZclAttributes(), like writing the attributes, to ensure the device is awake. However, that's too big a change for me to oversee the consequences.

Configuration should happen only happen after bindings are established (if remember correctly) but I agree there is lot to improve to respect sleeping end-devices. I'm not yet sure how to the code should be refactored to achieve this.

> I think the attributes need to be defined in the ZCLDB for deCONZ to accept the reporting (cf. the Xiaomi smart cube).

Thanks I'll add it to ZCLDB so they are available in the next release.

ebaauw:
> maybe later on both attributes should be checked for reports.

Yes, and then take into account the different values for the Max Report Interval.  Best refactor the code to put the attributes in a list, cf. the button commands for the switches...

> Configuration should happen only happen after bindings are established (if remember correctly)

Both are needed before any report is sent, but the order doesn't seem to matter (except maybe for the IKEA lights).  I haven't check the ZigBee documentation, though.

Issue #194, opened 2017-09-13
Duplicate events from Hue Motion sensor


donnib:
Hi,
I am not sure this is known already, i get following all the time from the Hue Motion sensor : 
```
{"e":"changed","id":"3","r":"sensors","state":{"lastupdated":"2017-09-13T14:15:23","presence":false},"t":"event"}
{"e":"changed","id":"3","r":"sensors","state":{"lastupdated":"2017-09-13T14:15:24","presence":false},"t":"event"}
```

wvuyk:
I am seeing this regularly too. For some reason the field lastupdated changed without obvious reason

`json {"e":"changed","id":"19","r":"sensors","state":{"lastupdated":"2017-09-14T07:32:59","presence":false},"t":"event"} 
json {"e":"changed","id":"19","r":"sensors","state":{"lastupdated":"2017-09-14T07:33:01","presence":false},"t":"event"} 
`
The sensor is reporting every period of time to show it is alive, that is the repetition we get. But why does lastupdated change? Maybe something changed that is not passed through the websocket?

manup:
Exactly, this should be the result of the reports send by the sensor. Lastupdated just tells when the value was updated the last time, hence the last report, which is correct since `lastupdated` doesn't automatically mean `lastchanged`. Maybe we should add an `lastchanged` to provide this information as well?

wvuyk:
Manup,
Not necessarily for me, maybe only an indication that the current websocket call is a report?
If I check the log of last night here, I see that sensors have been reporting once per 5 minutes. And last updated is adjusted accordingly. What is weird here is that both donnib and me are seeing this report in 2 seconds after each other. why was this not the "normal" 5 minutes

ebaauw:
I'm seeing the same, occasionally.  There's a second notification for the ZHAPresence sensor, 1 or 2 seconds after the regular report, where the sensor reports it's state every 5 minutes.  I suspect lastupdated is somehow updated wrongly, when the attribute for config.duration is read, which is now triggered when receiving a report from the sensor, to ensure the sensor is awake.  I need to anlyse my deCONZ log to be sure - I'm traveling right now, so this will have to wait until the weekend.

I don't think introducing lastchanged is a good idea - it's a characteristic of the attribute, not of the state as a whole.  You'd need four lastchange attributes for a LightLevel sensor: one for lightlevel, one for lux, one for dark, and one for daylight.

wvuyk:
[...]

PR #198, opened 2017-09-16
web socket notification enhancements

ebaauw:
Added new configuration setting, `/config/websocketnotifyall`, which defaults to `true`.  When set, all `state` or `config` attributes are included in the web socket notification; when set to `false` only the changed attributes are included.  See issue #194.

Updated API version to 1.0.4.

This PR also fixes that `state.lastupdated` was updated when `config.duration`, `config.ledindication`, `config.sensitivity`, `config.sensitivitymax`, or `config.usertest` was updated.  However, no event was issued for `state.lastupdated`, so the updated `state.lastupdate` wasn't reflected in the REST API.  I don't think this solves the duplicate web socket events from issue #194.

### DIFF
--- a/bindings.cpp
+++ b/bindings.cpp
@@ -639,9 +639,10 @@ bool DeRestPluginPrivate::sendConfigureReportingRequest(BindingTask &bt)
             {
                 rq = ConfigureReportingRequest();
                 rq.dataType = deCONZ::Zcl8BitUint;
-                rq.attributeId = 0x0030; // sensitivity
-                rq.minInterval = 5;      // value used by Hue bridge
-                rq.maxInterval = 7200;   // value used by Hue bridge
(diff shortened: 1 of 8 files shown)


manup:
>Added new configuration setting, /config/websocketnotifyall, which defaults to true. When set, all state or config attributes are included in the web socket notification; when set to false only the changed attributes are included. See issue #194.

Thanks @ebaauw, I like that it is configurable now, but also would prefer if it becomes more explicit, so it can be extended in future and it should be attached to the apikey since different apps may require different configuration (the websocket authentification and reference to apikey is missing currently).

**default**
```
/websocket/subscribe

[ "/sensors/changes", "/lights/changes", ... ]
```

**and more verbose**

```
/websocket/subscribe

[ "/sensors/*", "/lights/all", ... ]
```

Issue #205, opened 2017-09-20
config.battery not sent by websocket (2.04.77)


markbeee:
Since using 2.04.77 I get no battery status by websockets any more from my IKEA motion sensors (firmware: 1.2.214).
I also get a lot more event notifications when the sensor detetcts motion now. Before there was only one event for detecting motion.

manup:
Normally the battery status should be updated after some time, after starting deCONZ it may take a while. If it isn't updated even after an hour the binding seems to be inactive. I'll check if this is handled in deCONZ, so the binding will be recreated if necessary.
 
The increased event count are due the changes that all incoming commands are forwared, not only if attributes (here presence) change. The IKEA sensor sends multiple presence packets which cause these events.

@ebaauw has created a new config flag, so that the behaviour can be changed, see PR https://github.com/dresden-elektronik/deconz-rest-plugin/pull/198

This will be available in 2.04.78

ebaauw:
> Since using 2.04.77 I get no battery status by websockets any more from my IKEA motion sensors (firmware: 1.2.214).

Indeed, checking my logs, I see plenty of battery notifications from the IKEA motion sensor to the gateway, but no corresponding web socket notifications.  This is unrelated to the IKEA motion sensor firmware (or even the IKEA motion sensor - I see the same for the IKEA remote, the Hue motion sensor, and the Hue dimmer switch).  It was broken (unintentionally, I think) in v2.04.76, when combining all config or state attributes into one web socket notification.  Now, it only sends a web socket notification when the value has changed (as before v2.04.65).  I'll try and fix this coming weekend.

> I also get a lot more event notifications when the sensor detetcts motion now. Before there was only one event for detecting motion.

You get a notification on the web socket when the IKEA motion sensor sends a ZigBee message that it has detected motion.  With the new firmware, it seems to do this more often, every 2-3 seconds while there is motion.  
Note that the IKEA motion sensor does _not_ send a ZigBee message when it no longer detects motion (unlike the Hue motion sensor).  deCONZ automatically sets `state.presence` to `false` (without updating `state.lastupdated`), 60 seconds (`config.duration`) after motion was last reported (regardless of the setting of the timer dial).  Also, deCONZ resets `config.duration` to 60 seconds shortly after you change it.  I mean to fix this when I find the time (see https://github.com/dresden-elektronik/deconz-rest-plugin/issues/23#issuecomment-325131934).

markbeee:
[...]

Issue #23, opened 2017-03-30
IKEA Trådfri Remote (and motion sensor)


jsve:
IKEA has launched some more products in their Trådfri line. I have gotten the lights to work, but not remotes. They also have motion sensors now, which I guess will not work either. 

de_web_plugin.cpp has a comment about IKEA remote not being supported yet. Is this because of the web-plugin of the core of deCONZ? Is there some work being put in to make this compatible?

I have managed to get the remote to show up in the deCONZ GUI and it has that blue indicator-dot that blinks when I push buttons. So something seems to be working at least. The logs (using --dbg-aps=2) shows the proper cluster being activated when i push the respective button, but there is nothing indicating if it was (for instance) brightness up or down that was pushed.

manup:
Hi, in a few days we will release a new beta update which adds basic support for the Trådfri Remote in the REST API and realtime event notification in a new WebSocket interface. The button events can be used in rules to control lights and groups. We've made some tests with the Remote (firmware from 2016) and so far we haven't managed to find a stable way to join it into the gateway network — by pressing the setup button 4 times, sometimes it works, sometimes not. Hopefully the firmware will improve with the wider release.



We haven't testet the motion sensor yet, due the lack of a device, but expect support in the next few releases.

jsve:
Awesome! My experiences with the remote is similar. Somehow it just pairs after a lot of tries. Not sure, but it seems to help if i spam the buttons shortly after the pair (or at the same time). Might just be a coincidence though.

Guess thats fine if you know about the issue. I think it would be OK if it was available in the web-gui as an option, but with the disclaimer that it might take a few tries.

Is there any way for me to test the beta when it comes out?

manup:
Sure, you can download beta versions from https://www.dresden-elektronik.de/rpi/deconz/
the source code will be updated here too. Right now only stable version is kept here but we will provide a new branch (beta?) with the next beta release wich is in sync with the binary version.

Be aware that the next beta will only advance the REST API, not the WebApp itself. So using the IKEA switch and others is possible but a bit hacky.

manup:
The beta is released now, with basic support for IKEA Trådfri motion sensor and remote control

https://github.com/dresden-elektronik/deconz-rest-plugin/releases

yene:
Is it technically possible to talk directly to a light, or is always a remote required?

manup:
Yes every light can be controlled directly through `/lights` REST API.
[...]

PR #206, opened 2017-09-22
Fix web socket notifications for `config.battery`

ebaauw:
Issue #205

### DIFF
--- a/bindings.cpp
+++ b/bindings.cpp
@@ -1254,7 +1254,8 @@ void DeRestPluginPrivate::checkSensorGroup(Sensor *sensor)
         }
     }
     else if (sensor->modelId() == QLatin1String("RB01") ||
-             sensor->modelId() == QLatin1String("RM01"))
+             sensor->modelId() == QLatin1String("RM01") ||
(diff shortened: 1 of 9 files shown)


manup:
Thanks, just a small note: the setValue() should be placed after the if condition, so that the save database will be triggered if needed?

ebaauw:
You're right, of course.  Thanks!

ebaauw:
Commit 715454e fixes `config.duration` and introduces `state.dark` for the IKEA motion sensor (see #205 and https://github.com/dresden-elektronik/deconz-rest-plugin/issues/23#issuecomment-325131934):
- `config.duration` is updated (only) when presence is detected from the dial setting on the sensor.  Only when the value has changed, a web socket notification for config is sent;
- `config.duration` is read-only (cannot be updated through the REST API).
- `state.dark` is updated when presence is detected.  When the dial is set to Day, it’s value is et to `true`; when the dial is set to Night, it’s value reflects the motion sensor’s measured light level.
- The `state.any_on` of the linked group and `state.on` of the lights in that group are longer set to `true` when `state.dark` is `false`, while detecting presence.

PR #207, opened 2017-09-22
Fix automatic resetting of `state.presence` to `false`

ebaauw:
Fix automatic resetting of `state.presence` to `false`:
- Now the reset of `state.presence` to `false` takes place `config.duration` seconds (+/- 1 for rounding errors) after `state.presence` was last set to `true`;
- To disable automatic resetting, set `config.duration` to 0;
- Don’t reset `state.presence` automatically for motion sensors that do attribute reporting on the Occupancy Sensing cluster (like the Hue motion sensor), under the assumption that they’ll issue a message when motion is no longer detected;
- Also reset `state.presence` automatically for CLIPPresence sensors (when `config.duration` > 0);
- `config.duration` initialised to `0` instead of `60`.

Refactored the logic:
- Loop over all sensor resources in `checkSensorStateTimerFired()` instead of just checking one sensor per second;
- Added `sensor->durationDue` so loop only needs to do simple check per sensor;
- Set `sensor->durationDue` when setting `state.presence` (only for sensors where `state.presence` needs to be reset automatically).

Fixed a segmentation fault when setting `config.duration` of a CLIPPresence sensor through the REST API.

### DIFF
--- a/database.cpp
+++ b/database.cpp
@@ -1780,7 +1780,7 @@ static int sqliteLoadAllSensorsCallback(void *user, int ncols, char **colval , c
             item = sensor.addItem(DataTypeBool, RStatePresence);
             item->setValue(false);
             item = sensor.addItem(DataTypeUInt16, RConfigDuration);
-            item->setValue(60);
+            item->setValue(0);
             if (sensor.modelId() == QLatin1String("SML001")) // Hue motion sensor
             {
                 item->setValue(0);
--- a/de_web_plugin.cpp
+++ b/de_web_plugin.cpp
@@ -2131,15 +2131,18 @@ void DeRestPluginPrivate::checkSensorButtonEvent(Sensor *sensor, const deCONZ::A
                 if (item)
                 {
                     item->setValue(true);
-                    if (item->lastSet() == item->lastChanged())
-                    {
-                        Event e(RSensors, RStatePresence, sensor->id(), item);
-                        enqueueEvent(e);
-                    }
+                    Event e(RSensors, RStatePresence, sensor->id(), item);
+                    enqueueEvent(e);
                     updateSensorEtag(sensor);
                     sensor->updateStateTimestamp();
                     sensor->setNeedSaveDatabase(true);
                     enqueueEvent(Event(RSensors, RStateLastUpdated, sensor->id()));
+
+                    ResourceItem *item2 = sensor->item(RConfigDuration);
+                    if (item2 && item2->toNumber() > 0)
+                    {
+                        sensor->durationDue = QDateTime::currentDateTime().addSecs(item2->toNumber());
+                    }
                 }
                 break;
             }
@@ -7886,6 +7889,11 @@ void DeRestPluginPrivate::handleOnOffClusterIndication(TaskItem &task, const deC
                     Event e(RSensors, RStatePresence, s.id(), item);
                     enqueueEvent(e);
                     enqueueEvent(Event(RSensors, RStateLastUpdated, s.id()));
+                }
+                item = s.item(RConfigDuration);
+                if (item && item->toNumber() > 0)
+                {
+                    s.durationDue = QDateTime::currentDateTime().addSecs(item->toNumber());
                 }
             }
         }
--- a/de_web_plugin_private.h
+++ b/de_web_plugin_private.h
@@ -1318,7 +1318,6 @@ public:
     size_t sensorIter;
     size_t lightAttrIter;
     size_t sensorAttrIter;
-    size_t sensorCheckIter;
     std::vector<Group> groups;
     std::vector<LightNode> nodes;
     std::vector<Rule> rules;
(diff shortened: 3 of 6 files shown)


manup:
Thanks a lot for all the work on the sensors, I like the improved handling for the IKEA sensor duration. But I'm afraid that it can't be merged into master yet. If I understand correctly the *automatic off* will after a merge only work for sensors which use the `on with timed off` ZCL command. If the presence is triggered by ZCL attribute reporting or IAS zone cluster, the presence will only be set to false if the sensor sends this state (which might fail in some cases).

This problem can be fixed by adapting the `durationDue` for these sensors as well, which would be quite cool.

I'd also like to keep the `sensorCheckIter` to process only *1..N* sensors per eventloop cycle, I admit it seems to be over-optimizing but the *checkSensorStateTimerFired()* function will grow over time and we already have networks running with >120 presence/illuminance sensors. I guess you might have removed the iterator due inaccurate delays while processing the *automatic off*? We can speed up/down the timer interval with `checkSensorsTimer->setInterval(0..X)`, while work (presence) is active to assure near realtime processing. This should scale in future no matter how complex the sensors become.

ebaauw:
> If I understand correctly the automatic off will after a merge only work for sensors which use the on with timed off ZCL command. If the presence is triggered by ZCL attribute reporting or IAS zone cluster, the presence will only be set to false if the sensor sends this state (which might fail in some cases).

No, I setup `durationDue` everywhere I could find `RStatePresence`, except when handling the attribute report for the _Occupancy Sensing_ cluster.  So it should work for IAS zone sensors.  I don't have any of these, so I haven't been able to test this.  I don't know of any other sensors using _Occupancy Sensing_, and not sending a report when motion is no longer detected, but we could check for the Hue motion sensor and set `durationDue` for others.

>  I guess you might have removed the iterator due inaccurate delays while processing the automatic off? 

"Inaccurate" is a bit of an understatement.  `checkSensorStateTimerFired()` fires once a second and it only checked one sensor at a time.  I have 102 sensor resources (incl. 40 CLIP sensors), so the IKEA motion sensor was reset somewhere between 0 and 102 seconds after `config.duration` expired.  For all intents and purposes: randomly.

> I'd also like to keep the sensorCheckIter to process only 1..N sensors per eventloop cycle, I admit it seems to be over-optimizing but the checkSensorStateTimerFired() function will grow over time and we already have networks running with >120 presence/illuminance sensors.
> We can speed up/down the timer interval with checkSensorsTimer->setInterval(0..X), while work (presence) is active to assure near realtime processing. This should scale in future no matter how complex the sensors become.

You would need to set the timer to fire once every 10ms in my case, to clear `state.presence` when the lights turn off (+/- 1 second).  More often with more sensors...

I (briefly) considered creating a separate array with only the sensors with `durationDue` to loop only over those, but that seemed a bit too much work.  I did move all logic (except checking `durationDue` against the current time) to setting `durationDue` to keep the loop as lean as possible.

In NodeJS I would not use an interval timer at all.  Instead, I would create a one-time timeout per sensor when `state.presence` is set (and cancel and re-create it when `state.presence` is set again), but I'm not sure how to do that in Qt5.

manup:
> No, I setup durationDue everywhere I could find RStatePresence, except when handling the attribute report for the Occupancy Sensing cluster.

That's what I mean, if a sensor only reports the presence state via attribute reporting (OSRAM motion sensor, Xiaomi?) the automatic off won't happen. I couldn't resist and out of interest added my approach, see last commits :)

> So it should work for IAS zone sensors. I don't have any of these, so I haven't been able to test this.

The IAS sensors are handled in ias_zone.cpp, the durationDue is added there now.

The check function now adapts dynamically the interval speed (100 vs. 1000 ms) and processes 10 sensors in one run, hopefully this helps with timing problems and is scale-able for future additions.

In theory the timer can be turned off entirely if no processing is needed, but for now I'll let it run at 1 second interval to not break things by accident.

ebaauw:
> I couldn't resist and out of interest added my approach, see last commits :)
> The check function now adapts dynamically the interval speed (100 vs. 1000 ms) and processes 10 sensors in one run, hopefully this helps with timing problems and is scale-able for future additions.

Cool! This seems to work for the IKEA motion sensor.  I'll do some more testing with CLIPPresence tomorrow.

Only now, the Hue motion sensor's presence is cleared before it reports no occupancy (typically it adds like 10 seconds to the delay).  I still think `durationDue` shouldn't be setup for sensors that support _PIR Occupied to Unoccupied Delay_ (and send a command for no occupancy).

Alternatively, we could simply ignore this attribute for the Hue motion sensor and handle `config.duration` in deCONZ (like e.g. `config.tholddark`).  That wouldn't work when using the Hue motion sensor standalone (controlling a group directly), but I don't think we managed to get that working anyway?

> The IAS sensors are handled in ias_zone.cpp, the durationDue is added there now.

I didn't notice that, indeed.  I thought they'd be handled by `checkSensorButtonEvent()`.

https://github.com/dresden-elektronik/deconz-rest-plugin/blob/bb50d2643c7ddec25834d7f6b5b24f5bcd139c95/de_web_plugin.cpp#L3499

I really like that you base `durationDue` on `lastSet()` instead of on `currentDateTime()`.  I'd like to change that in the other places as well.  I added an additional 500ms for the CLIPPresence, since the timing seemed wrong, but maybe using `lastSet()` fixes this as well.

Issue #216, opened 2017-10-08
IKEA Trådfri Dimmer


ebaauw:
Got myself a Trådfri dimmer switch as well - it was lying next to the colour bulb at Ikea screaming "Buy me!".

Did the firmware update to 1.2.248 using deCONZ on my test RaspBee.  Had some issues connecting it to my production RaspBee running v2.04.80 afterwards.  I would show in the deCONZ GUI, but the descriptor wouldn't be read.  Had a similar issue with the new remote that came with the colour bulb: button presses would register in deCONZ, but reading the Basic cluster from the GUI would fail (blinking red in deCONZ GUI).  Sniffing showed no Read Attribute commands being sent at all?!  Further sniffing revealed it was connected to a Philips Hue light.  I powered off the Philips Hue lights in the room, forcing both switches to connect to the RaspBee directly.  After that, the descriptor and Basic cluster were read without issues and the REST API plugin created the resource for the dimmer.

I expected the dimmer would be using the magnetic contact, but it actually works without the base unit: it seems to contain a gyroscope or something similar.  It's extremely verbose when attached to the base unit, especially in a vertical position (as if the magnet is interfering with the gyroscope) .  It will probably help it I fixate the base unit to a wall (or fridge door or whatever), but I doubt whether it can be used in practice.

Sniffing, I find that it actually uses two discrete values for rate with the Move and Move with On/Off commands: 0x46 for a small move, and 0xC3 for a larger move.  Also, there seems to be a fixed 1:1 relation between Move (with On/Off) and Stop.  I would suggest to simply ignore the Stop and maybe use the rate to distinguish between release (after press) and release after hold.  Also, I would use additional buttons for On and Off (based on the Move to Level with On/Off commands when turning the dimmer abruptly).  There's a bug in the REST API plugin that both an abrupt right turn as well as an abrupt left turn result in buttonevent 1000.  Probably because the check for cmd 0x04 (Move to Level with On/Off) is missing from https://github.com/dresden-elektronik/deconz-rest-plugin/blob/0ebbae77a3144218832af1eb8917c517aee872de/de_web_plugin.cpp#L2163  Finally, the DimUp and DimDown buttons are reversed - for the other switches DimUp comes before DimDown.  

So I suggest the buttonmap would be (similar to the Hue dimmer):

cmd 0x04, level 0xFF -> 1002
cmd 0x05, direction up (0x00), rate 0x46 -> 2002
[...]

PR #217, opened 2017-10-08
IKEA Trådfri dimmer

ebaauw:
Issue #216


manup:
Looks good, I'll test it with the older firmware hopefully commands are all the same

Issue #224, opened 2017-10-13
Empty action.body when creating rules in v2.04.81


ebaauw:
When creating rules under v2.04.81 (with the latest commits), `action.body` seems to be ignored and remains empty.  It isn't stored to the database either.  It seems to happen to all rules, irrespective to the type of action (group, sensor, light).

It's been a while since I last updated my rules, so I'm not sure if this is issue is introduced by v2.04.81 or by an earlier change.  Existing rules had been working without any issues.  When I patch the database, adding the body, the rule works as expected.

For example see the following transcript:
```
$ ph_rules_boottime 107 103
pi2: debug: gateway command: curl -s -X POST -H "Content-Type: application/json" -d '{
    "name": "Boot Time",
    "status": "enabled",
    "conditions": [
    {
        "address": "/sensors/103/state/lastupdated",
        "operator": "dx"
      },
    {
      "address": "/sensors/107/state/flag",
      "operator": "eq",
      "value": "false"
    }
  ],
    "actions": [
    {
        "address": "/sensors/107/state",
        "method": "PUT",
        "body": {"flag": true}
      },
    {
        "address": "/groups/0/action",
        "method": "PUT",
        "body": {"on": false}
      }
  ]
  }' "http://pi2/api/xxxxxxxx/rules"
pi2: debug: gateway response: [{"success":{"id":"1"}}]
pi2: /rules/1: "Boot Time"
$ ph_get /rules/1
pi2: debug: gateway command: curl -s -H "Content-Type: application/json" "http://pi2/api/xxxxxxxx/rules/1"
pi2: debug: gateway response: {"actions":[{"address":"/sensors/107/state","body":{},"method":"PUT"},{"address":"/groups/0/action","body":{},"method":"PUT"}],"conditions":[{"address":"/sensors/103/state/lastupdated","operator":"dx"},{"address":"/sensors/107/state/flag","operator":"eq","value":"false"}],"created":"2017-10-13T13:19:08","etag":"c31af637d0b1faa3a6dad2dcfdc158bf","lasttriggered":"none","name":"Boot Time","owner":"xxxxxxxx","periodic":0,"status":"enabled","timestriggered":0}
{
  "actions": [
    {
      "address": "/sensors/107/state",
      "body": {},
      "method": "PUT"
    },
    {
      "address": "/groups/0/action",
      "body": {},
      "method": "PUT"
    }
  ],
  "conditions": [
    {
[...]

Issue #223, opened 2017-10-13
deCONZ versioning scheme incompatible with semver


ebaauw:
The deCONZ versioning scheme doesn't use valid Semantic Versioning ([semver](http://semver.org)) versions, as there's a leading zero in the the minor version (e.g. 2.**0**4.81).

Would it be possible to ditch the leading zero?  I'm reluctant to change this because of the comment https://github.com/dresden-elektronik/deconz-rest-plugin/blob/4e943ff403920c68e2aa1c17486e8e1e221e7867/de_web.pro#L59

Alternatively, could we use the API version to match the swversion, but as a valid semver string?

manup:
>Alternatively, could we use the API version to match the swversion, but as a valid semver string?

The API version was supposed to reflect the hue API version, I know some apps like iConnectHue will check this attribute, so maybe keeping (and note to myself: *maintaining it*) would be better.

I'll check the scripts and webapp "version compare functions" if they would break when the leading zero is removed from `swversion`attribute. Breaking things won't be an option since a lot of installations out there are only updated via WebApp.

Changing `swversion` could be done for api response only, so that scripts and defines stay at current format.

ebaauw:
> Changing swversion could be done for api response only, so that scripts and defines stay at current format.

That sounds like a good plan.

> The API version was supposed to reflect the hue API version, I know some apps like iConnectHue will check this attribute, so maybe keeping (and note to myself: _maintaining it_) would be better.

We discussed this earlier (https://github.com/dresden-elektronik/deconz-rest-plugin/issues/20#issuecomment-290807573) - that won't be easy.

Issue #222, opened 2017-10-11
Hue Apps don‘t find my Raspbee Gateway


Krocko:
I have testet various Hue Apps, but no one can find my Raspbee Gateway. Must i change some Settings?

ebaauw:
Obviously, the deCONZ gateway doesn’t announce itself to the meethue portal.  Also, it announces itself to UPnP differently than the Hue bridge, so Hue apps won’t find it without (minor) changes.  You might try and configure the IP address in the app instead of doing discovery (if the Hue app supports that).  I only tried the official Hue app, but that doesn’t like deCONZ, even when I hard-configured the IP address.

Krocko:
Mhhh, ok. But i heard from some German Users, that they control the Raspbee with an Hue App. Also on the German Support side is written, that we can use 3rd Party Apps.

mark-tech:
I control the Raspbee with Hue App, hueManiac and Alexa using an hue emulator from github as proxy. Spent some time to setup because lack of documentation but in the end the result is awesome.

Krocko:
@mark-tech: Sounds good. Can you please explain it?

mark-tech:
i use one rpi with diyhue on port 80 and deconz on port 8080. I encored some issues that are not documented but in the end i figured out by myself what was the problem and once my setup was working everything was fine. I remember no hue application detect the emulator if deconz was running in the same time. Need to stop deconz and restart the emulator for every new device you want to pair, then start again deconz and restart again the emulator. You will not find this in documentation.

ebaauw:
With PR #225, the Hue app connects to the deCONZ gateway as if it were a Hue bridge.  Note that there's still differences between the Hue API and the deCONZ API, so expect some issues, notably around scenes.

ebaauw:
And the Hue app crashes on my production gateway, probably because I have lights, groups and sensors with IDs > 64.  Or more than 200 rules.

mariusmotea:
@mark-tech feel free to open issues for any problems you have. Regarding ssdp discover, deconz also use this service to announce is presence on the network which enter in conflict with hue emulator. It cannot be disabled so the only workaround is to start the hue emulator with deconz stopped. I will update the wiki to cover also this aspect.

Krocko:
>And the Hue app crashes on my production gateway, probably because I have lights, groups and sensors with IDs > 64. Or more than 200 rules.

[...]

PR #225, opened 2017-10-13
Minor enhancements and bug fix

ebaauw:
Make `swversion` semver compliant, see Issue #223.

Make deCONZ appear as a Hue bridge to the Hue app, see #222.

Fix body when creating a rule, see #224.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@ Makefile.Debug
 Makefile.Release
 release
 ui_de_web_widget.h
+.qmake.stash
--- a/README.md
+++ b/README.md
@@ -60,7 +60,21 @@ Packages for Qt4 and Raspbian Wheezy are available but not described here.
 
 Software requirements
 ---------------------
-* Raspbian Jessie and Qt5
+* Raspbian Jessie or Stretch and Qt5
+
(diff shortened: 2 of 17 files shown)


manup:
> Fix body when creating a rule, see #224.

Very nice this solves a very old problem that sometimes (always) device and sensor rules created by the WebApp and Phoscon App didn't work out of the box. By looking at the code it's seems to be directly related to the rule creation; only at the next update rule it works.

>Make swversion semver compliant, see Issue #223.

I've checked the version compare functions, they use JS *parseInt()* function for each version part x.y.zz and therefore should be safe.

ebaauw:
> I've checked the version compare functions, they use JS parseInt() function for each version part x.y.zz and therefore should be safe.

Meaning we could change the #define instead of patching the API output?

manup:
>Meaning we could change the #define GW_SW_VERSION instead of patching the API output?

I'd like to keep the #define for some time, because bash build script do parse it and generate the debian package version and other stuff from it.

>Note: I did not change the API output for ApiVersion_1_DDEL. That would change as well, should we change the #define.

At least for the old WebApp that would be fine too.

PR #330, opened 2017-12-15
Implement group state.all_on

ebaauw:
Implement group `state.all_on` and update API version to 1.0.5.

Fix inconsistency in websocket notifications.


manup:
Hi thanks for the group.all_on implementation very welcome. :)

I can't merge the PR yet since the sensor.id change would break the Phoscon App sensor wizards, it expects the `event.sensor.id` attribute in the `added` and `remove` events.

It's fine to add `event.id` but `event.sensor.id` must be available too to not break running setups.

ebaauw:
> It's fine to add event.id but event.sensor.id must be available too to not break running setups.

OK, I'll change that.

Issue #287, opened 2017-11-24
Gateway Cascade


ebaauw:
I've got two Pi-s, one production and one test, both with a RaspBee running deCONZ and homebridge-hue on the same (wired) subnet.  This afternoon I upgraded both to v2.04.91.  Since then, I'm seeing some disturbing log messages that I haven't seen before.  On my production Pi:

> 15:00:00:001 API error 1, api/2c34bcf4-c/groups, unauthorized user
15:00:05:002 API error 1, api/2c34bcf4-c/config, unauthorized user
15:00:10:002 API error 1, api/2c34bcf4-c/groups, unauthorized user
15:00:15:001 API error 1, api/2c34bcf4-c/config, unauthorized user
15:00:20:002 API error 1, api/2c34bcf4-c/groups, unauthorized user
15:00:25:002 API error 1, api/2c34bcf4-c/config, unauthorized user
15:00:30:002 API error 1, api/2c34bcf4-c/groups, unauthorized user
15:00:35:001 API error 1, api/2c34bcf4-c/config, unauthorized user
15:00:40:002 API error 1, api/2c34bcf4-c/groups, unauthorized user
15:00:45:002 API error 1, api/2c34bcf4-c/config, unauthorized user
15:00:50:001 API error 1, api/2c34bcf4-c/groups, unauthorized user

and 

> 15:00:00:605 unhandled http status code in connected state 403 switch to offline state
15:00:10:606 unhandled http status code in connected state 403 switch to offline state
15:00:20:604 unhandled http status code in connected state 403 switch to offline state
15:00:30:604 unhandled http status code in connected state 403 switch to offline state
15:00:40:605 unhandled http status code in connected state 403 switch to offline state
15:00:50:603 unhandled http status code in connected state 403 switch to offline state

These message have been appearing every 10 seconds.

Same on my test pi:

> 15:00:00:607 API error 1, api/5728cec5-0/groups, unauthorized user
15:00:05:607 API error 1, api/5728cec5-0/config, unauthorized user
15:00:10:607 API error 1, api/5728cec5-0/groups, unauthorized user
15:00:15:609 API error 1, api/5728cec5-0/config, unauthorized user
15:00:20:607 API error 1, api/5728cec5-0/groups, unauthorized user
15:00:25:607 API error 1, api/5728cec5-0/config, unauthorized user
15:00:30:607 API error 1, api/5728cec5-0/groups, unauthorized user
[...]

Issue #84, opened 2017-08-04
state.bri_inc not yet implemented?


ebaauw:
When I try and PUT /api/username/lights/1/state with a body of `{"bri_inc":254}` (or any other value), deCONZ returns nothing and the light's brightness doesn't change.  PUTting `{"ct_inc":512}` (or any other value) works as expected.  
Same for group requests.

manup:
Correct, `ct_inc` was implemented to support IKEA tradfri remote color temperature buttons. The `bri_inc` is on the todo list.

... need better error messages for yet unsupported features.

wvuyk:
same for hue_inc and sat_inc? Both were used in my plugin as well

manup:
Yes currently only `ct_inc` is implemented, the others will follow but there is no ETA yet.

ebaauw:
I started sniffing the commands the Hue bridge sends on PUTting these attributes, see https://github.com/dresden-elektronik/deconz-rest-plugin/wiki/Philips-Hue-sniffing.

Issue #29, opened 2017-04-23
Scene API


sfeigl:
The scene REST API of the Hue seems to be independent of groups and have their own REST collection.

Are there any plans that deCONZ will add this too? (additionally or replacing the scenes tied to groups)? Looking at the code I did not recognize much technical reasons why scenes must depend on groups. 

I could try to adapt the REST plugin by myself if it is not planned or already in progress.

manup:
Yes there is a planned adaption of the scenes API without groups. But it's more complicated when it seems in first place.

On ZigBee level every scene belongs to a group, this is why deCONZ uses the `/group/<id>/scene/<id>` scheme. A scene is stored *in* the lights nvram (scenes cluster) and can be recalled with a simple group command for a given scene id. For example a battery powered light switch doesn't need to know the scene details for 50 lights, it just needs to send one command, like `recall scene 1` and can go back to sleep afterwards, all lights will tune in the scene at the same time.

This has the benefit that the gateway doesn't need to be involved at all since the light switch can directly control the lights.

The adaption will either use a special group, the so called *global group* 0x0000 or a random internal group to keep the benefits of ZigBee scenes. However ZigBee scenes implementation is quite sophisticated:

* scenes must be *installed* on the lights
* changes or removal of scenes must be maintained in lights
* the gateway shall verify periodically that scenes on the lights are correct
* lights might be not reachable or busy with updates, so implementation must be robust to not assume everything works on first try
* scenes might be installed, removed or changed by other devices like a remote or light switch

There was the idea of handling all scenes just by the gateway, so when a scene is called the gateway sends the needed commands to the lights. I think this is how hue bridge handles scenes (or at least did in the past). That's fairly simple to implement but doesn't scale at all. For example we have projects with > 150 lights, sending out 400 commands just to tune in a scene would be overkill. And it looks terrible too since not all lights would react instantly but sequentially.

ebaauw:
For v1 scenes, the Hue bridge stores scene light states solely in the lights.  When recalling a v1 scene, the lights are updated immediately, but it takes the Hue bridge several minutes (in my case; 38 lights) before it reports the changed light states.  The Hue API documentation suggests that group commands are broadcast on the ZigBee network and the lights seem to update simultaneously.

[...]

PR #338, opened 2017-12-21
Gateway cascade enhancements (and some more)

ebaauw:
Several enhancements for gateway cascade functionality, see #287:
- Cascade setting `action.on` and `action.bri`, with or without `action.transitiontime` from the REST API;
- Cascade recalling scenes to the Hue bridge: support Hue bridge syntax for scene commands, see #29;
- Cascade all commands sent by the Hue dimmer switch in group mode;
- Cascade all commands sent by the IKEA Trådfri remote in group mode.

Notes:
- The Hue dimmer switch actually sends multiple _Step_ commands (~one every second) while holding the _DimUp_ or _DimDown_ button; the Trådfri remote sends a single _Move_ command and relies on the _Stop_ command (on long release) to stop the current movement;
- The Trådfri remote sends a _Toggle_ command on the _On/Off_ button.   As this is not supported by the REST API, `action.on` is set to true (false) when local group `state.all_on` is false (true);
- Currently, the _DimUp_ and _DimDown_ buttons (for both the Hue and the Trådfri) can only be cascaded to a Hue bridge, as deCONZ hasn't yet implemented `bri_inc` (see #84).

Todo:
- Cascade setting `action.ct` and `action.xy` from the REST API.

Other changes:
- Change ubisys dimmer buttonevents;
- Only expose ZHASwitch sensor resources for _Analog Input_ and _Multistate Input_ clusters for the Xiaomi Smart Cube, not for other devices.  Apparently, sometimes other Xiaomi devices advertise these clusters.


de_web_plugin.cpp
```diff
@@ -2694,13 +2694,17 @@ void DeRestPluginPrivate::addSensorNode(const deCONZ::Node *node)
 
                 case ANALOG_INPUT_CLUSTER_ID:
                 {
-                    fpSwitch.inClusters.push_back(ci->id());
+                    if (modelId == QLatin1String("lumi.sensor_cube")) {
```
manup: We may need to extend this whitelist in future, I'm not sure but I think there were other sensors using the analog/multi-state clusters too.

manup:
Thanks a lot, I'll merge the PR now but can't test it currently (back in end of January).

Maybe in future the support for proper external web hooks in rules replaces the cascades with a simpler general way to forward REST API requests to an arbitrary destination.

See also https://github.com/dresden-elektronik/deconz-rest-plugin/issues/347

Luckily the implementation will be fairly simple due QNetworkManager. :)

de_web_plugin.cpp
ebaauw: I double-checked that I introduced these clusters when adding support for the Smart Cube.  Also, the handling of the attribute values is specific to the cube.

ebaauw: Alternatively we could to blacklist the “funny” sensors (so far I’ve seen reports of lumi.sensor_ht and lumi.switch_86sw2 showing additional endpoints with these clusters).

ebaauw:
> Maybe in future the support for proper external web hooks in rules replaces the cascades with a simpler general way to forward REST API requests to an arbitrary destination.

Maybe, maybe not.  There’s something to say for webhooks, especially if they can be attached to rule actions.  You’d be able to interact with any REST api, not just the deCONZ/Hue API.  With cascade, you actually forward ZigBee commands as deCONZ/Hue REST API calls.  I’m amazed how fast this works, when the commands are sent by ZigBee switches or sensors, compared to commands sent by the gateway (looks like the gateway only cascades these when it reads back the groupcasts).

The translation of ZigBee commands to REST API calls is somewhat crude.  I’d love someday to design a more consistent and complete API, with different endpoints for (write-only) commands, `/actions/...`, read-only attributes, `/state/...`, and writeable attributes, `/config/...`.

manup:
Agree, they both have their use cases. Setup for cascades is also easier since it's just a group a forward to group b mapping.

Issue #356, opened 2018-01-09
Xiaomi Aqara Wall Switches


mathos77:
Hi @manup,

I have ordered these Xiaomi single and double key switches, which are powered by mains.
https://www.gearbest.com/alarm-systems/pp_610096.html?utm_source=email_sys&utm_medium=email&utm_campaign=shipping
https://www.gearbest.com/alarm-systems/pp_625774.html?utm_source=email_sys&utm_medium=email&utm_campaign=shipping

I was hoping to find these supported as being a light (on-off) as they should have this capability.

After joining them into deCONZ, I see this:
![image](https://user-images.githubusercontent.com/19706662/34729249-7842cea4-f55c-11e7-87de-4636c696fa8c.png)

Can we get these switches supported, as it would greatly improve my home automation, for places where I do not want a real smart light, so in this case it would become a semi-smart light.

Thanks!

Kind regards, Tom.

ebaauw:
> I was hoping to find these supported as being a light (on-off) as they should have this capability.

Looks like they do - see endpoint 0x03.

> Can we get these switches supported, as it would greatly improve my home automation, for places where I do not want a real smart light, so in this case it would become a semi-smart light.

I'd expect the REST API to create already an on/off light for this?  Did you open the network from the _Settings_ in the Web app, or from the deCONZ GUI?  The REST API will only create a resource when you open the network from the Web app.

Could you open the Cluster Info panel on the _Basic_ cluster on endpoint 0x01, _Read_ the attributes, and post a screenshot?  It might take a couple of attempts, but as this switch is mains powered it might work straight away.  We would especially need the _Manufacturer Name_ and the _Model Identifier_.  If they work like the other Xiaomi switches, you might need to press briefly the reset button, to have the switch report it's _Basic_ attributes.  
Please check both the single and double key switches - they're probably slightly different.

The switch looks very similar to the QBKG03LM (#335) and the WXKG02LM (#165), so maybe adding the name is enough to get the buttons to work with deCONZ.  You wouldn't be able to sniff the ZigBee network and see what messages the switch sends?

mathos77:
Hi @ebaauw ,

I did try both, WebUI (deCONZ plain) and deCONZ GUI. I even tried Phoscon.
Anyway, so you believe it should already have entry under lights.
It did not unfortunately.
I can switch (ON or OFF) and Toggle both switches on the Double-Key switch via Cluster Info btw.
[...]

PR #367, opened 2018-01-16
Xiaomi wall switches, see #356

ebaauw:
Fixed typo and added `lumi.ctrl_neutral1`.  Pretty sure we'll need different handling for buttons of the mains powered switches as endpoint 01 only holds Basic cluster, and endpoint 03 is the _On/Off light_ for the switch's mains output signal.


manup:
Hi @ebaauw thanks for the fix, very welcome :)

PR #378, opened 2018-01-22
More button fixes

ebaauw:
- Xiaomi smart button, non-Aqara version (`lumi.sensor_switch`): fix swap 1000 and 1002 values for `state.buttonevent`.  1000 should be press, 1002 release;
- ubisys C4 control unit: don't create ZHASwitch resources for the _Windows Covering Control_ endpoints, only for the four _Level Control Switch_ endpoints.

### DIFF
--- a/de_web_plugin.cpp
+++ b/de_web_plugin.cpp
@@ -2533,6 +2533,11 @@ void DeRestPluginPrivate::addSensorNode(const deCONZ::Node *node)
                     {
                         fpPresenceSensor.outClusters.push_back(ci->id());
                     }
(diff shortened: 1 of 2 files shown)


manup:
Thanks for the fix @ebaauw  :) will be available in 2.05.01.

jensflorian:
Does this mean all four switches of Ubisys C4 are supported by the API? The wiki only mentions a C1 device as supported which I do believe, doesn’t exists. (https://github.com/dresden-elektronik/deconz-rest-plugin/wiki/Supported-Devices)

Issue #407, opened 2018-02-09
Battery-powered vs Mains-powered devices


ebaauw:
The current deCONZ code seems to assume that end devices are battery powered and routers are mains powered.  This is not always the case.  Increasingly, mains-powered lights and switches don't have router functionality in their ZigBee stack.  The Heiman siren is a router, is mains powered and has a battery backup (value 81 for _Power Source_ attribute 0x0007 in the _Basic_ cluster), which I would want to expose.

I think we need to use something like `hasBattery()` and `isMainsPowered()` for device discovery and configuration instead of `isEndDevice()`.  Not sure where best to implement these; `isEndDevice()` is a deCONZ core function, defined in `/usr/include/deconz/node.h`.  I'm not sure how early in the discovery this would be known.  Would we base it on the _Power Source_ attribute in the _Basic_ cluster?  On the _Power Configuration_ cluster?  On the Power Descriptor?  Neither seem to read early enough.

manup:
Agree, isEndDevice() doesn't fit all devices anymore.

Usually for the `rxOnWhenIdle` flag from the NodeDescriptor can be used for that purpose. We should check in the NodeInfo panel that all the related (end) devices have this flag set to true. The REST-API already has access to it via `deCONZ::Node::nodeDescriptor().receiverOnWhenIdle()` method.

NodeDescriptor is also available very early in discovery.

ebaauw:
Thanks, @manup.  This could be the key to exposing end-device lights.

I'm afraid this won't help me for the Heiman Siren, though.  I'm wondering why you only put the _Power Configuration_ in the fingerprint for sensors from these manufacturers:
https://github.com/dresden-elektronik/deconz-rest-plugin/blob/475c614a33335e5fb3e8b3b383b42876dd7c2ccc/de_web_plugin.cpp#L2601-L2613

I would want to put in it regardless, also for the other sensor types and then expose `config.battery` if the cluster is in the fingerprint, instead of when it's an end device.
https://github.com/dresden-elektronik/deconz-rest-plugin/blob/475c614a33335e5fb3e8b3b383b42876dd7c2ccc/de_web_plugin.cpp#L2960-L2963

Or do you know of mains-powered sensors than expose _Power Configuration_, but don't have a (backup) battery?

manup:
>I would want to put in it regardless, also for the other sensor types and then expose config.battery if the cluster is in the fingerprint, instead of when it's an end device.

[...]

Issue #404, opened 2018-02-09
deCONZ enhancements for IAS devices


ebaauw:
I've got a whole bunch of Heiman sensors on loan from @Oggymator, to integrate into deCONZ and homebridge-hue (see https://github.com/dresden-elektronik/deconz-rest-plugin/issues/263#issuecomment-362527633).  They all connect to deCONZ without any issue, and I'm analysing them through the web GUI and ZigBee sniffing.  They're _IAS Zone_ devices, except for the temperature/humidity sensor, which is a _Temperature Sensor_ and the Siren, which is an _IAS Warning Device_).

I'm trying to figure out the best way to support these in deCONZ, see also earlier discussions in #152, #269, and #352), and would like your views, @manup, @rtenklooster, and, of course, any-one else.

1. The _IAS Zone_ devices obviously provide an _IAS Zone_ cluster.  They send _Zone Status Change Notification_ commands when an alarm occurs _and_ when it's cleared.  They set the _Restore reports_ bit in the _Zone Status_.  deCONZ could/should take this as a hint not to use `config.duration` to clear the alarm.
2. The sensors provide their type through  _IAS Zone Status_, but this is rather coarse grained.  Both the combustable gas and the carbon-monoxide (CO) sensor report _Gas sensor_.  This means, any REST API client (i.c. homebridge-hue) would still need to check the `manufacturername` and `modelid`, even if deCONZ would provide a `zonetype` attribute.  Would it still make sense to provide `zonetype`?
3. Next to _Alarm 1_, the sensors also set/clear the _Tamper_ and _Battery_ bits in the  _Zone Status_.  The sensors contain a tamper button which gets depressed when you remove them from their base plate, and I had some almost depleted AAA batteries lying around.  I think it makes sense to expose these through deCONZ as well.  deCONZ could introduce a sensor `type` of `ZHAStatus` which provides the _Zone Status_ bitmap as `state.status`.  `CLIPGenericStatus` can be used to simulate a `ZHAStatus` sensor.  Alternatively, deCONZ could introduce a sensor `type` of `ZHAAlarm`, with separate Boolean attributes `state.alarm`, `state.tamper`, `state.battery`.  We'd need a corresponding `CLIPAlarm` to simulate these.
[...]

Issue #311, opened 2017-12-08
Sengled Element Bulbs


mercenaryballs:
I have these bulbs paired with my Conbee. I can see them in the GUI, but they do not show up in the web app. I believe they are unsupported. I can control them in the GUI (on/off and level control). Here are the atttributes: 

![image](https://user-images.githubusercontent.com/34362886/33752311-2291a6c2-dbaf-11e7-9e23-26554e2aca68.png)
![image](https://user-images.githubusercontent.com/34362886/33752373-817f909a-dbaf-11e7-8300-dd8c6e8cfb62.png)
![image](https://user-images.githubusercontent.com/34362886/33752390-908fb16e-dbaf-11e7-9dbe-72775c2ac014.png)
![image](https://user-images.githubusercontent.com/34362886/33752419-a89f1a24-dbaf-11e7-9fe8-74f0f3b0913f.png)
![image](https://user-images.githubusercontent.com/34362886/33752469-edbc1440-dbaf-11e7-9e71-f6fa6aa6106b.png)
![image](https://user-images.githubusercontent.com/34362886/33752483-fe21d392-dbaf-11e7-88ac-3b4b903d45d4.png)
![image](https://user-images.githubusercontent.com/34362886/33752568-77b97c82-dbb0-11e7-879d-c45558ae6239.png)
![image](https://user-images.githubusercontent.com/34362886/33752573-81b32170-dbb0-11e7-9441-9dae69d796eb.png)
![image](https://user-images.githubusercontent.com/34362886/33752577-8d63dd66-dbb0-11e7-9250-f4d50bc983f2.png)
![image](https://user-images.githubusercontent.com/34362886/33752584-9a48654c-dbb0-11e7-9bad-0441866b5a91.png)
![image](https://user-images.githubusercontent.com/34362886/33752592-a663c59c-dbb0-11e7-9ee8-8074c104232a.png)
![image](https://user-images.githubusercontent.com/34362886/33752598-ae9f54ba-dbb0-11e7-97b0-69a04b2df1f4.png)
![image](https://user-images.githubusercontent.com/34362886/33752605-b965afb6-dbb0-11e7-958b-1f217f7ca077.png)
![image](https://user-images.githubusercontent.com/34362886/33752608-c23f4caa-dbb0-11e7-9d30-55639d43c282.png)
[...]

PR #408, opened 2018-02-09
Heiman sensors and more

ebaauw:
Just a first take at supporting the Heiman sensors, most still as ZHAPresence, see also #404 and #407.

@manup can you please review the changes I've marked.

### DIFF
--- a/database.cpp
+++ b/database.cpp
@@ -1834,6 +1834,8 @@ static int sqliteLoadAllSensorsCallback(void *user, int ncols, char **colval , c
             }
             item = sensor.addItem(DataTypeInt16, RStateTemperature);
             item->setValue(0);
+            item = sensor.addItem(DataTypeInt16, RConfigOffset);
+            item->setValue(0);
         }
         else if (sensor.type().endsWith(QLatin1String("Humidity")))
         {
--- a/database.cpp
+++ b/database.cpp
@@ -1963,7 +1963,10 @@ static int sqliteLoadAllSensorsCallback(void *user, int ncols, char **colval , c
             item->setValue(100);
         }
 
-        if (stateCol >= 0 && !isClip)
+        // @manup: What was the reason not to restore state for CLIP sensors?
+        if (stateCol >= 0 &&
+            sensor.type() != QLatin1String("CLIPGenericFlag") &&
+            sensor.type() != QLatin1String("CLIPGenericStatus"))
         {
             sensor.jsonToState(QLatin1String(colval[stateCol]));
         }
--- a/database.cpp
+++ b/database.cpp
@@ -1859,6 +1859,10 @@ static int sqliteLoadAllSensorsCallback(void *user, int ncols, char **colval , c
             {
                 clusterId = OCCUPANCY_SENSING_CLUSTER_ID;
             }
+            else if (sensor.fingerPrint().hasInCluster(IAS_WD_CLUSTER_ID))
+            {
+                clusterId = IAS_WD_CLUSTER_ID;
+            }
             else if (sensor.fingerPrint().hasInCluster(IAS_ZONE_CLUSTER_ID))
             {
                 clusterId = IAS_ZONE_CLUSTER_ID;
@@ -1870,7 +1874,7 @@ static int sqliteLoadAllSensorsCallback(void *user, int ncols, char **colval , c
             item = sensor.addItem(DataTypeBool, RStatePresence);
             item->setValue(false);
             item = sensor.addItem(DataTypeUInt16, RConfigDuration);
-            item->setValue(0);
+            item->setValue(60); // presence should be reasonable for physical sensors
             if (sensor.modelId() == QLatin1String("SML001")) // Hue motion sensor
             {
                 item->setValue(0);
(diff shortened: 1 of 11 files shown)


de_web_plugin.cpp
```diff
@@ -2593,22 +2599,27 @@ void DeRestPluginPrivate::addSensorNode(const deCONZ::Node *node)
...
                 }
                     break;
 
                 case POWER_CONFIGURATION_CLUSTER_ID:
                 {
-                    if (node->nodeDescriptor().manufacturerCode() == VENDOR_PHILIPS ||
-                        node->nodeDescriptor().manufacturerCode() == VENDOR_NYCE ||
-                        node->nodeDescriptor().manufacturerCode() == VENDOR_IKEA)
-                    {
+                    // @manup: is it safe to skip these tests?
```
manup: The tests were added because not all devices implement the power cluster properly (reporting doesn't work for example). For now I would like to keep them and extend the tests aka whitelist here.

de_web_plugin.cpp
```diff
@@ -2957,7 +2989,9 @@ void DeRestPluginPrivate::addSensorNode(const deCONZ::Node *node, const SensorFi
     item = sensorNode.item(RConfigReachable);
     item->setValue(true);
 
-    if (node->isEndDevice())
+    // @manup, will this work?
```
manup: For end-devices with power configuration cluster yes. Some battery powered devices don't have them, like Busch-Jaeger, JUNG and dresden-elektronik switches, these only provide battery information via classic PowerDescriptor.

Anyway it's ok to do it this way and maybe to extend it with a NodeDescriptor.rxOnWhenIdle check.

de_web_plugin.cpp
```diff
@@ -3061,6 +3106,7 @@ void DeRestPluginPrivate::addSensorNode(const deCONZ::Node *node, const SensorFi
     }
     else if ((node->nodeDescriptor().manufacturerCode() == VENDOR_OSRAM_STACK) || (node->nodeDescriptor().manufacturerCode() == VENDOR_OSRAM))
     {
+        // TODO: Heiman sensors also use VENDOR_OSRAM_STACK
```
manup: Indeed we should use data from basic cluster here.
I think VENDOR_OSRAM_STACK should really be VENDOR_MARVEL (if I remember correctly, need to check).

de_web_plugin_private.h
```diff
@@ -200,23 +203,25 @@
...
+#define VENDOR_NYCE         0x10B9
+#define VENDOR_UBISYS       0x10F2
+#define VENDOR_BEGA         0x1105
+#define VENDOR_OSRAM        0x110C
 #define VENDOR_BUSCH_JAEGER 0x112E
-#define VENDOR_BEGA 0x1105
-#define VENDOR_PAULMANN 0x119d
-#define VENDOR_NONE   0x0000
+#define VENDOR_PAULMANN     0x119D
+#define VENDOR_120B         0x120B     // @manup, where can I lookup manufacturer code?
```
manup: Wireshark source code contains  a comprehensive list:

http://cgit.osmocom.org/wireshark/plain/epan/dissectors/packet-zbee.h

de_web_plugin.cpp
ebaauw: As I mentioned in #407, the seem to whitelisted in bindings.cpp as well.

de_web_plugin.cpp
ebaauw: Cool, thanks.

de_web_plugin_private.h
ebaauw: Found 1 out of 2...

ebaauw:
@manup, may I once ask for your help?  The _Electrical Management_ cluster, 0x0B04, seems to contain only attribute 0x0000.  Where do I indicate I also want _Current Power_, 0x050B?
https://github.com/ebaauw/deconz-rest-plugin/blob/29cd7bcbbdaefb77517963fd4b05bbf0b7fad2e5/de_web_plugin.cpp#L1798-L1828

database.cpp
```diff
@@ -1963,7 +1963,10 @@ static int sqliteLoadAllSensorsCallback(void *user, int ncols, char **colval , c
             item->setValue(100);
         }
 
-        if (stateCol >= 0 && !isClip)
+        // @manup: What was the reason not to restore state for CLIP sensors?
```
manup: Not sure, but I figure that this was my lazy approach to always *initialize* these to defined values on startup.

ebaauw:
> manup merged commit 43e76b7 into dresden-elektronik:master  an hour ago

Thanks, but maybe a bit premature?  state.power doesn't show anything but 0, see above, and the Heiman sensors still are exposed as ZHAPresence (except for the door and temp/hum sensors).

manup:
I think that's ok, since it is a beta version if nothing crashes it should be ok that some features are still work in progress.

ebaauw:
OK.  
I wonder if end-device lights are now recognised, see https://github.com/dresden-elektronik/deconz-rest-plugin/issues/311.

manup:
Short note: I'm out to a short travel till Thursday.
When I'm back I can test this with an Climax on-off power plug from the office which is an end-device too.

PR #420, opened 2018-02-17
State attributes for IAS sensors

ebaauw:
- Introduced new types for `/sensors` resources:
  - ZHAAlarm (with `state.alarm`);
  - ZHACarbonMonoxide (with `state.carbonmonoxide`);
  - ZHAFire (with `state.fire`);
  - ZHAWater (with `state.water`).
- Heiman sensors now exposed using appropriate type;
- Xiaomi Aqara Leak sensor now exposed as ZHAWater;
- Add `state.lowbattery` and `state.tampered` for whitelisted IAS sensors (also ZHAOpenClose and ZHAPresence);
- Implemented `state.power` (electrical power in Watt) for "lights" with Electrical Management cluster, like Heiman and OSRAM plugs;
- Bumped API version to 1.0.8.


manup:
Hmmm tricky, maybe `state/battery` and similar flags should be `state/batteryalarm` or `alarm/battery` to distinguish flags or alarms from actual measured values?

Here are some experiments:

    state/fire                     // true, false!
    state/battery               // true, false or some measured value?
    state/tamper               // true, false!
    state/water                 // true, false or some measured value?
    state/carbonmonoxid // true, false or some measured value?

    // clear but not nice to read
    state/firealarm
    state/batteryalarm
    state/tamperalarm
    state/wateralarm
    state/carbonmonoxidealarm

    // easier to read but ... new alarm thing
    alarm/fire
    alarm/battery
    alarm/tamper
    alarm/water
    alarm/carbonmonoxide

    // I like this one, but not that good for rules (or maybe operator 'in': "fire" in state/alarm)
    state/alarm: [ ]
    state/alarm: [ "fire", "battery", "tamper", "water" ]

ebaauw:
They're all Boolean.

I'm happy to change them to `state.battery_alarm`, etc. or `state.alarm_battery`, etc. but imho with some documentation (oops...) `state.battery`, etc. would do.

I tried creating a single ZHAAlarm with an attribute for IAS Zone type, but that's too much work: need to change the fingerprint and the database table to hold the additional attribute.  Also, this would need a `switch ()` on this attribute everywhere to figure out the correct resource and state attribute.  So I'll create ZHAAlarm (generic), ZHACarbonMonoxide, ZHAFire, ZHAWater instead.

manup:
> I'm happy to change them to state.battery_alarm, etc. or state.alarm_battery, etc. but imho with some documentation (oops...) state.battery, etc. would do.

Providing comprehensive docs is a big thing this year :) 

>I tried creating a single ZHAAlarm with an attribute for IAS Zone type, but that's too much work: need to change the fingerprint and the database table to hold the additional attribute. Also, this would need a switch () on this attribute everywhere to figure out the correct resource and state attribute. So I'll create ZHAAlarm (generic), ZHACarbonMonoxide, ZHAFire, ZHAWater instead.

Yes I agree, this matches also existing sensor style:

    ZHAPresence    state/presence
    ZHALightLevel  state/lightlevel
    ZHAFire        state/fire

    ZHAWater state/water
             state/waterlevel    (could be added, if such a sensor exists)

ebaauw:
I think this is working.

Still TODO:
- Test CLIP variants;
- Sometimes ZHAAlarm is created when existing sensor fires alarm while network is open.  I suspect it's because IAS Zone Type hasn't been read and defaults to 0.  Maybe we should whitelist the sensors on Model ID instead of looking at the IAS Zone Type.

ebaauw:
Indeed, whitelisting the sensors instead of relying on IAS Zone type fixes the duplicate ZHAAlarm sensor and makes much faster and more reliable discovery.

manup:
Good work, what do you think can the PR be merged in the next release or should it be merged later?

ebaauw:
> Good work, what do you think can the PR be merged in the next release or should it be merged later?

I think it's ready.  I'm currently running production with this build.

ebaauw:
Oh and @manup, could you please include the latest general.xml?
[general.xml.gz](https://github.com/dresden-elektronik/deconz-rest-plugin/files/1734658/general.xml.gz)

manup:
> Oh and @manup, could you please include the latest general.xml?
general.xml.gz

Yes I'll merge it for 2.05.04

Issue #352, opened 2018-01-07
Smoke detector Schwaiger ZHS20 resp. HEIMAN HS1SA


zbzhs20:
Hi,

I connected a smoke detector to deCONZ, it is labeled "ZHS20" from Schwaiger GmbH (www.schwaiger.de). It seems to be a HEIMAN HS1SA, see Basic Cluster Info.
The HEIMAN smoke detector is certified, id is ZIG16075ZHA25799-24.
There is a Z-WAVE version of the HEIMAN smoke detector, so be careful when ordering.

I can get the cluster infos (sometimes I had to press the button at the device to get a response,
probably because it's a battery powered device).

Is there a way to get the battery status and alarms through the web socket?
The cherry on top would be to turn on the siren through the REST API :-).

If you need more information or if I should check anything for you please feel free to comment!

Here are the cluster infos:

![schwaiger-zhs20-attributes-1](https://user-images.githubusercontent.com/35171815/34651559-10d0239c-f3d2-11e7-855e-eca2778a2a48.png)

![schwaiger-zhs20-attributes-2](https://user-images.githubusercontent.com/35171815/34651572-2d6e0096-f3d2-11e7-9e36-b7a8a6635164.png)

![schwaiger-zhs20-attributes-3](https://user-images.githubusercontent.com/35171815/34651718-f0f514e0-f3d3-11e7-8be9-643634b7cc11.png)

![schwaiger-zhs20-attributes-4](https://user-images.githubusercontent.com/35171815/34651727-084ec3de-f3d4-11e7-91d0-40703579a470.png)

![schwaiger-zhs20-attributes-5](https://user-images.githubusercontent.com/35171815/34651728-103d02e0-f3d4-11e7-9929-2cb6047ede82.png)

![schwaiger-zhs20-attributes-6](https://user-images.githubusercontent.com/35171815/34651734-1e82b00c-f3d4-11e7-9a76-adc57056bfc8.png)

![schwaiger-zhs20-attributes-7](https://user-images.githubusercontent.com/35171815/34651735-237fbdf2-f3d4-11e7-9ee7-be4af780cfa8.png)

![schwaiger-zhs20-attributes-8](https://user-images.githubusercontent.com/35171815/34651736-2943ca80-f3d4-11e7-8674-a40bdc7e3363.png)

ebaauw:
[...]

Issue #152, opened 2017-08-30
Xiaomi Mijia (HonneyWell) smoke detector


rtenklooster:
Hi,
Just received my RaspBee today and started playing around.
I've upgraded to the latest beta in order to get my xiaomi switches and motion detectors running.

My [smoke detectors](https://www.gearbest.com/alarm-systems/pp_615081.html) are able to connect en they show up like this:
![rookmelder](https://user-images.githubusercontent.com/6507714/29889637-2af978ea-8dc5-11e7-9d51-d14e3c6bdc18.PNG)

With debugging enabled this event comes by when pushing the test button.
`20:39:01:991 APS-DATA.indication srcAddr: 0x00158d00014d315c, dstAddrMode: 2, profile: 0x0104, cluster: 0x0500, lqi: 199, rssi: -66
20:39:02:218 APS-DATA.request id: 13, addrmode: 0x02, addr: 0x0000, profile: 0x0000, cluster: 0x0031, ep: 0x00 queue: 0 len: 2
20:39:02:292 APS-DATA.confirm id: 13, status: 0x00 SUCCESS`
And: 
`20:39:05:011 APS-DATA.indication srcAddr: 0x00158d00014d315c, dstAddrMode: 2, profile: 0x0104, cluster: 0x0500, lqi: 199, rssi: -66
20:39:05:100 APS-DATA.request id: 40, addrmode: 0x02, addr: 0x16c2, profile: 0x0000, cluster: 0x0031, ep: 0x00 queue: 0 len: 2
20:39:05:189 APS-DATA.confirm id: 40, status: 0x00 SUCCESS`
What should I do next to get my smoke detector supported?

manup:
The cluster 0x0500 indicated that this sensor is a IAS Zone, I figure therefore it's easy to support it. However question is how should the sensor be represented by the API?

Either ZHAPresence or better ZHAOpenClose or maybe a new type like ZHAAlarm?

rtenklooster:
ZHAAlarm seems the most appropriate type. There is a Gas sensor as well, and a water leakage sensor. They all could use the ZHAAlarm type I think.

snozzlebert:
Presence of smoke/gas/water, so ZHAPresence?

ebaauw:
> However question is how should the sensor be represented by the API?

[...]

PR #424, opened 2018-02-18
More sensor stuff

ebaauw:
- Relax _Model Identifier_ checks for Heiman sensors, see #352.
- Xiaomi Mi smoke sensor, see #152.


rtenklooster:
Ok, I think i've pulled your latest commit:
git fetch origin pull/424/head:eb
git checkout eb
qmake && make -j2
sudo cp ../libde_rest_plugin.so /usr/share/deCONZ/plugins

Are this all steps needed? Because I don't see any changes:
image

ebaauw:
You won't see any changes in the GUI, but the REST API should now create a ZHAFire `/sensors` resource.  Open the network from the web app and read the attributes of the _Basic_ cluster.

PR #426, opened 2018-02-19
Power measurement, pt.2

ebaauw:
@manup may I once again ask for your help?  I need the raw byte value for the 0xff01 attribute to derive the battery level.  The Xiaomi devices announce this attribute as a string (0x42), but the next byte (length) is one off (at least on the flood sensor).  Because of this, Wireshark bitches about a malformed package.  Using the attached code, the bytes with bit 7 set (value > 0x80) are corrupted, probably because `QString` expects Unicode.  Also, it only retrieves the data up to the first 0x00 byte.

I'm afraid the deCONZ core needs to be enhanced to support a `rawValue()` or `payload()` method to `deCONZ::ZclAttribute`.  The payload data is there (with DBG_INFO2), but once I get to `updateSensorNode()` the zclFrame is out of scope.  I doesn't seem like a good idea to pass that all the way down to `updateSensorNode()`.

BTW, I found some SmartThings definitions that read the Xiaomi battery:
https://github.com/rinkek/Xiaomi/tree/master/devicetypes/bspranger.  It appears the sensor report the battery Voltage, which is extrapolated into percentage, based on the expected Voltage range ~2.5V to ~3.0V (kinda hacky).


de_web_plugin.cpp
```diff
@@ -4475,6 +4475,12 @@ void DeRestPluginPrivate::updateSensorNode(const deCONZ::NodeEvent &event)
 
                                 updateSensorEtag(&*i);
                             }
+                            else if (ia->id() == 0xff01) // Xiaomi magic
+                            {
+                                const char *str = ia->toString().toLatin1().constData();
```
manup: Careful with that one `.constData()` is only guaranteed to be valid as long as the object is on the stack or heap, caused me a lot of headaches in the past :)

One way to do it:

    QByteArray arr = ia->toString().toLatin1();
    const char *str = arr.constData();

Or easier:

    qPrintable(ia->toString().toLatin1().toHex(' '))

ebaauw: The QT version on my Raspberry Stretch doesn't support an argument to `toHex()`, but,  Indeed, it's easier.  However it doesn't solve the real issue, see following annotated log:
```
Feb 19 22:17:27 pi deCONZ[9921]: 22:17:26:542 	payload: 050042156c756d692e73656e736f725f776c65616b2e61713101ff42220121d10b03281e0421a8330521540006240000000000082104020a211e67641000
                                                         --------l u m i . s e n s o r _ w l e a k . a q 1 --------====XX============XX========
Feb 19 22:17:27 pi deCONZ[9921]: 22:17:26:543 Node data 0x00158d0001bb6159 profileId: 0x0104, clusterId: 0x0000
Feb 19 22:17:27 pi deCONZ[9921]: 22:17:26:543 Update Sensor 0x00158D0001BB6159 Basic Cluster
Feb 19 22:17:27 pi deCONZ[9921]: 22:17:26:543 >>>>> 0x00158D0001BB6159: Xiaomi magic: 01213f0b03281e04213f33052154
                                                                                      ====XX============XX========
```
The payload of the _Report Attributes_ shows two attributes for cluster 0x0000: _Model Identifier_ (0x0005) and 0xff01, both of type string (0x42).  The length for 0xff01 (0x22) is one byte more than actually transmitted.  
In `updateSensorNode()`, the attribute payload is corrupted (see XX) and trimmed.  
The battery voltage is 0x0bd1, or 3.025V.

manup:
In this particular case I would recommend extracting the data in the `handleZclAttributeReportIndication()` instead of `updateSensorNode()`, here you have access to the raw data.

https://github.com/dresden-elektronik/deconz-rest-plugin/blob/master/de_web_plugin.cpp#L6549

I've seen various of these .groovy handlers of SmartThings for Xiaomi, it seems each device has another byte/way to report the battery data (but I have just looked briefly at it).

ebaauw:
> In this particular case I would recommend extracting the data in the handleZclAttributeReportIndication() instead of updateSensorNode(), here you have access to the raw data.

Yes... of the zclFrame.  I would need to extract the address, cluster, command, skip the 0x0005 attribute; locate the sensor, locate the cluster in the sensor, locate the attribute in the cluster in the sensor - a whole lot of code duplication.  Or is there a way to somewhere store the data and use it in `updateSensorNode()`?

> I've seen various of these .groovy handlers of SmartThings for Xiaomi, it seems each device has another byte/way to report the battery data (but I have just looked briefly at it).

Haven't looked at all of them in full yet, but by the looks of it, this is because the 0x0005 attribute has a different length for different devices.  That's what you get when parsing the full frame instead of just the value for the 0xff01 attribute...

Also I still to catch a regular battery update, not triggered by the reset button.

ebaauw:
BTW, is there any documentation available on the C++ API for plugins (the `/usr/include/deconz` stuff)?

manup:
>Yes... of the zclFrame. I would need to extract the address, cluster, command, skip the 0x0005 attribute; locate the sensor, locate the cluster in the sensor, locate the attribute in the cluster in the sensor - a whole lot of code duplication. Or is there a way to store the extracted data and use it later in updateSensorNode()?

I'm afraid there is no cache of the raw data yet, the internal ZCL attribute parser does expects valid data which is kept in the internal QString.

Here is the related (core: closed code) snipped which parses the string. Pretty old code and ugly as I look on it :)

```C++
       quint8 len;
        stream >> len;

        if ((len > 0) && stream.atEnd())
            return false;

        int ret;
        char buf[0xff + 1];
        ret = stream.readRawData(buf, len);

        if ((ret >= 0) && (ret <= 0xff))
        {
            buf[ret] = '\0';
            setValue(QString((const char*)buf));
        }
        else
        {
            return false;
        }
```

https://www.dresden-elektronik.de/fileadmin/Downloads/Dokumente/Produkte/6_Software/deconz-cpp-doc/dd/ddb/classdeCONZ_1_1ZclAttribute.html#a88c5aa8bffd99044fede7c88d65f2049

> BTW, is there any documentation available on the C++ API for plugins (the /usr/include/deconz stuff)?

Yes  there is a doxygen documentation it contains most of the APS / ZCL stuff, hasn't been recreated in a while but the API is quite stable in the important parts for years.

https://www.dresden-elektronik.de/fileadmin/Downloads/Dokumente/Produkte/6_Software/deconz-cpp-doc/modules.html

ebaauw:
> Also I still to catch a regular battery update, not triggered by the reset button.

Gotcha!  It seems to report only the 0xff01 attribute:
```
Feb 20 09:37:15 pi deCONZ[15168]: 09:37:14:595 APS-DATA.indication srcAddr: 0x00158d0001bb6159, dstAddrMode: 2, profile: 0x0104, cluster: 0x0000, lqi: 231, rssi: 0 
Feb 20 09:37:15 pi deCONZ[15168]: 09:37:14:596 no button map for: lumi.sensor_wleak.aq1 cl: 0x0000 cmd: 0x0A pl[0]: 001
Feb 20 09:37:15 pi deCONZ[15168]: 09:37:14:596 ZCL attribute report 0x00158D0001BB6159 for cluster 0x0000, ep 0x01 
Feb 20 09:37:15 pi deCONZ[15168]: 09:37:14:596  payload: 01ff42220121d10b0328190421a8330521540006240100000000082104020a211e67641000
Feb 20 09:37:15 pi deCONZ[15168]: 09:37:14:596 Node data 0x00158d0001bb6159 profileId: 0x0104, clusterId: 0x0000
Feb 20 09:37:15 pi deCONZ[15168]: 09:37:14:596 Update Sensor 0x00158D0001BB6159 Basic Cluster 
Feb 20 09:37:15 pi deCONZ[15168]: 09:37:14:596 >>>>> 0x00158D0001BB6159: Xiaomi magic: 01213f0b03281904213f33052154
```

ebaauw:
Latest commit adds more power measurement features, see https://github.com/dresden-elektronik/deconz-rest-plugin/issues/432#issuecomment-368229496

manup:
Exciting stuff, is it ready for 2.05.07 or shall I wait with merging?

ebaauw:
Can you wait a couple more hours?  I want to see how this works in my production environmentm see https://github.com/dresden-elektronik/deconz-rest-plugin/issues/432#issuecomment-368256688.  Also, no support yet for CLIPPower.

manup:
Sure I'll compile 2.05.07 later tomorrow, due some more fixes around the corner 2.05.08 will arrive till Wednesday also, so no hurry :)

ebaauw:
The basics seem to work, also for the ubisys D1 and the OSRAM plugs, but still some rough edges:

- I entered the default values (65535, -32768) in general.xml.  After starting deCONZ, the previous values are restored from the database, but then these defaults shortly show, until all the attributes have been reported (or read from the deCONZ GUI);
- I probably need to whitelist which plugs support which attributes.  It seems to go well on sensor creation, but after a restart/restore from database, also non-supported attributes show in the REST API.  Probably need a whitelist anyway for factor (V, 0.1V, 0.01V, etc) each sensor uses;
- I still need to setup polling and/or make deCONZ setup attribute reporting.  As these are all mains powered (what else would there be to measure), a simple read would do, but I need to figure out where to do this.
- Of course, the OSRAM doesn't report anything useful - I think I'll drop support when I'm done testing.

ebaauw:
> I entered the default values (65535, -32768) in general.xml. After starting deCONZ, the previous values are restored from the database, but then these defaults shortly show, until all the attributes have been reported (or read from the deCONZ GUI);

Fixed by checking for default values when processing received values.

> I probably need to whitelist which plugs support which attributes. It seems to go well on sensor creation, but after a restart/restore from database, also non-supported attributes show in the REST API. Probably need a whitelist anyway for factor (V, 0.1V, 0.01V, etc) each sensor uses;

Done

> I still need to setup polling and/or make deCONZ setup attribute reporting. As these are all mains powered (what else would there be to measure), a simple read would do, but I need to figure out where to do this.

I figured I could leverage the poll manager, but I don't think this is working yet.  Would appreciate if you could take a look at this, @manup.

The Heiman seems the only plug (at least in my possession) that does attribute reporting.  I still need to verify if the reporting for _Current Summation Delivered_ works - this is a u48, so I had to hack a bit lacking a quint48 in Qt 5.

manup:
> I figured I could leverage the poll manager, but I don't think this is working yet. Would appreciate if you could take a look at this, @manup.

Yes poll manager is the way to go, code looks right, what is missing is the repeated trigger for the related nodes. I have some uncomitted code for that I'll see to include it in the very next days. Currently lights are polled on certain events and from the *DeRestPlugin::idleTimerFired()* method but this is pretty slow.

ebaauw:
Poll manager now works for the sensors.  The issue was _RConfigReachable_ for sensors instead of _RStateReachable_ for lights.  Took me a while to figure this out...

I'm still stuck on setting up the binding for attribute reporting, see @manup comment in bindings.cpp.

ebaauw:
Bloody hell, now I get a websocket notification every time the ZHAPower sensor is polled.  I guess I should only issue an event in case of changed value or of attribute reporting...

de_web_plugin.cpp
```diff
@@ -1181,6 +1181,16 @@ void DeRestPluginPrivate::addLightNode(const deCONZ::Node *node)
                     }
                     break;
 
+                case DEV_ID_XIAOMI_SMART_PLUG:
```
manup: I think in this case it's safer to extend the filter with `model id` to pick the Xiaomi plug.


According the 0xFFFF device identifier defined by DEV_ID_XIAOMI_SMART_PLUG this is tricky since device identifiers from *0xC000&ndash;0xFFFF* fall into the reserved range and normally shouldn't be used in HA or ZLL profile. I figure there are more devices than the Xiomi smart plug with that device id.

Hence DEV_ID_RESERVED_FFFF or some other general name might be a better for that define.

ebaauw:
That should do it.  Polling stops after I setup attribute reporting manually - nice.

There's still something odd, deCONZ continuously polls the ZHAPower clusters 0x0702 and 0x050b (like every 0.5 seconds), but the light cluster 0x0006 is polled far less frequently.

de_web_plugin.cpp
ebaauw: I wasn't sure `modelId` is already available this early in `addLightNode()`.

I figured the test on `deviceId`, _VENDOR_JENNIC_, endpoint 0x01 and `node->nodeDescriptor().receiverOnWhenIdle()` would probably be safe for now.

manup:
> There's still something odd, deCONZ continuously polls the ZHAPower clusters 0x0702 and 0x050b (like every 0.5 seconds), but the light cluster 0x0006 is polled far less frequently.

I think the *poll()* call should be moved out of the *checkSensorNodeReachable()* method since it is called quite often. Better place it in *idleTimerFired()* for now.

de_web_plugin.cpp
manup: >The combined test is on deviceId, VENDOR_JENNIC, endpoint 0x01, OnOff cluster, and node->nodeDescriptor().receiverOnWhenIdle(). I figured that would probably be safe for now.

Agree, a small comment should be sufficient in this case.

ebaauw:
Ok, so each time poll() is called, the corresponding node is polled only once.  I copied the code for lights to the sensors.  On my test system (one plug = one light, one sensor) the sensor still get polled more often, but one cluster at a time at no longer at sub second rate.

On my production system (74 nodes, 41 lights, 4 ZHAPower) the polling seems OK.

manup:
Yes poll manager is made so that it only uses one APS data request and it will proceed only after this requests is finished (success or timeout). The *idleTimerFired()* is not the best place to handle the trigger but works ok for now. I'm cooking a hopefully saner approach which will trigger the polling for lights and sensors based on some known data like when the the last attribute was queried by ZCL read attributes commands. It will also respect OTA and other running stuff.

ebaauw:
Cool.  I think I'm done for the day (still need to adapt homebridge-hue to ZHAPower).  Can you merge this PR and take it from here?

manup:
Thanks for all the good work I'm really impressed how fast everything evolved.

2.05.07 is already compiled, I'll merge the PR tomorrow for 2.05.08 and see how the polling behaves in our test network (currently  ~80 lights plus some sensors).

ebaauw:
Just woke up realising it’s odd to combine multiple clusters in a single `/sensors` resource.  We only do that for `/lights` resources.  I think I better split into ZHAConsumption for _Metering_ vs ZHAPower for _Electrical Management_.  That would also address the Xioami plugs that use two _Analog Input_ clusters, one for power (endpoint 0x02) and one for consumption (0x03).

manup:
:) yeah sounds more logical in comparison to other sensor implementations.

ebaauw:
Done, I hope in time for 2.05.08.  Again, the Xiaomi plug as yet untested.

I'm now seeing the value of _Current Summation Delivered_ in the GUI.  Note that there's still a`t` missing from the attribute name in general.xml.

manup:
2.05.08 is online for all platforms :)
The type will be fixed in next version.

manup:
> BTW, I found some SmartThings definitions that read the Xiaomi battery:
https://github.com/rinkek/Xiaomi/tree/master/devicetypes/bspranger. It appears the sensor report the battery Voltage, which is extrapolated into percentage, based on the expected Voltage range ~2.5V to ~3.0V (kinda hacky).

The repo seems to be taken offline

ebaauw:
Yeah, I found that also.  It was a fork of: https://github.com/bspranger/Xiaomi/tree/master/devicetypes/bspranger, which is still online.

PR #487, opened 2018-03-17
Fix binding for attribute reporting for mains-powered sensors.

ebaauw:
@manup, hope you can still include this one in v2.05.14.

Somehow the binding request isn't sent unless `setMgmtBindSupported(false)` is issued.  Instead of checking for `receiverOnWhenIdle()`, I just whitelisted the plugs as `endDeviceSupported`.  I suppose if needed, the management binding will be created from the light node anyways.

I verified in WireShark that the right _Configure Reporting_ commands are generated for the u48 and s24 attributes of the _Simple Metering_ cluster and that the Heiman plug accepts these.  Apparently the Heiman has only room for 5 reporting configurations, where there's 6 attributes to be reported:
- _On/Off_ (0x0006/0x0000);
- _Current Summation Delivered_ (0x0702/0x0000);
- _Instantaneous Demand_ (0x0702/0x0400);
- _RMS Voltage_ (0x0B04/0x0505);
- _RMS Current_ (0x0B04/0x0508);
- _Active Power_ (0x0B04/0x050B).

_Active Power_ and _Instantaneous Demand_ seem to report the same value (except that the latter isn't received by deCONZ, probably due to the fact that it's an s24 value).

### DIFF
--- a/bindings.cpp
+++ b/bindings.cpp
@@ -697,10 +697,12 @@ bool DeRestPluginPrivate::sendConfigureReportingRequest(BindingTask &bt, const s
             }
             else if (rq.reportableChange48bit != 0xFFFFFFFF)
             {
-                stream << rq.reportableChange48bit;
-                // since there's no quint48, we need to pad the quint32 to 48 bits
-                quint16 zero = 0x0000;
-                stream << zero;
+                stream << (qint8) (rq.reportableChange48bit & 0xFF);
+                stream << (qint8) ((rq.reportableChange48bit >> 8) & 0xFF);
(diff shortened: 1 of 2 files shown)


jurriaan:
@ebaauw Just received a Heiman smart plug, and it didn't report the on/off state until I created the binding manually. Is this the expected behaviour?

I'm on deconz 2.05.19 with latest firmware.

ebaauw:
No it’s not, but I haven’t really looked at this part, since attribute reporting for _OnOff_ was already supported for the `/lights` resource.

PR #497, opened 2018-03-19
Sensor Polling and Keen Home Vent

ebaauw:
- Also poll _Instantaneous Demand_;
- Don’t poll Daylight sensor;
- Check sensor type for polling `state.power`;
- Additional debug messages.


manup:
Hi, sorry for the delay I was at light and building fair this week.
Many thanks for he PR I'll merge it later on

PR #538, opened 2018-04-14
Combine ubisys switch endpoints into one ZHASwitch resource

ebaauw:
Similar to Xiaomi Aqara wall switch, see #356.


manup:
As I understand it 3) would be most useful. So that the switch can work either standalone by just adding lights to the related group(s) with the benefit that the setup works also when gateway is off for some reason. Beside that if the setup requires it, button events, which are extracted from incoming group commands, can be used to provide more sophisticated controls.

>The configuration is a single attribute (albeit with an array value with (an) element(s) per endpoint), so you need to generate a single ZigBee Write Attribute command. That's a lot easier with a single REST API PUT command, or a single ZHASwitch resource.

I hope it can be written in one command frame, need to check if the array data fits into one ZCL packet. The C4 has a "large" configuration array, in worst case we need to split that up into multiple array element writes.

cross ref https://github.com/dresden-elektronik/deconz-rest-plugin/issues/539

PR #559, opened 2018-04-27
More Xiaomi magic

ebaauw:
- Update `state.water` for lumi.sensor_wleak.aq1;
- Update `state.pressure` for lumi.sensor_weather.

### DIFF
--- a/de_web_plugin.cpp
+++ b/de_web_plugin.cpp
@@ -7158,6 +7158,7 @@ void DeRestPluginPrivate::handleZclAttributeReportIndicationXiaomiSpecial(const 
     quint32 lightlevel = UINT32_MAX; // use 32-bit to mark invalid and support 0xffff value
     qint16 temperature = INT16_MIN;
     quint16 humidity = UINT16_MAX;
+    qint16 pressure = INT16_MIN;
     quint8 onOff = UINT8_MAX;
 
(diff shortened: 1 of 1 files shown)


manup:
Cool thanks, I still wonder what the remaining unknown attributes in the magic reports are (04, 05, 06) :)

manup:
Here a report of my temperature/humidity sensor

```
39:28:606 0x00158D000171F3AC extract Xiaomi special
00:39:28:607 	01 battery 3015 (0x0BC7)
00:39:28:607 	04 unknown 17320 (0x43A8)
00:39:28:607 	05 unknown 14 (0x000E)
00:39:28:607 	06 unknown 4294967296 (0x0000000100000000)
00:39:28:607 	64 temperature 2058
00:39:28:607 	65 humidity 5047
```
ZigBee RSSI is normally expressed as signed 8-bit integer: -90 dB to 0 dB, the 05 attribute is 16-bit unsigned integer so it might represent something different than RSSI.

ZigBee LQI is normally a 8-bit unsigned integer 0&ndash;255. The 06 attribute is 40-bit unsigned integer, from various devices I see  different bits are set here but I have no clue what they mean. I think we need to gather more data and see if the values change over time and if there are differences for same sensor types.

> More promising are the "application" attributes 0a (all devices) and 97, 98, 99, and 9a (cube).

Absolutely, funny little devices, I'll add more prints to observe the values as decimal and hex :)

Issue #413, opened 2018-02-12
Rule triggering logic has a bug for operator "in" 


djwlindenaar:
Hi guys,
I was wondering this morning why my lights were still on, although I put a rule in to turn them off between T01:00:00/T18:30:00. I looked into the rule triggering logic and I believe there's an issue in the code.

The offending rule:
{
    "actions": [
        {
            "address": "/groups/2/action",
            "body": {
                "on": false
            },
            "method": "PUT"
        }
    ],
    "conditions": [
        {
            "address": "/config/localtime",
            "operator": "in",
            "value": "T01:05:00/T18:25:00"
        }
    ],
    "created": "2018-02-11T10:03:47",
    "etag": "ce0f498ef1c948a6e66ef2974d6ba838",
    "lasttriggered": "2018-02-11T11:03:47",
    "name": "Night Off",
    "periodic": 0,
    "status": "enabled",
    "timestriggered": 1
}

rest_rules.cpp @1348:
        else if (c->op() == RuleCondition::OpIn && c->suffix() == RConfigLocalTime)
        {
            QTime t = now.time();
            QTime rt;

            if (rule.lastTriggered().isValid())
            {
                rt = rule.lastTriggered().time();
            }

            if (c->time0() < c->time1() && // 8:00 - 16:00
                (t >= c->time0() && t <= c->time1()))
            {
                if (rt.isValid() && rt >= c->time0() && rt <= c->time1())
                {
                    if (eItem->descriptor().suffix == RConfigLocalTime)
                    {
                        return false;  // already handled
                    }
                }
}

So the rule triggered as I created it yesterday. The rule triggering code checks lasttriggered to see if it has already triggered in the targeted timeframe, but it neglects to take the date into account. So last night at 01:05:00 it checked the lasttrigered time, found 11:03:47 and concluded that this time is within time0 and time1 timeframe. Therefore it returned false //already handled... 

I guess to solve this, we should be comparing QDateTime items, by first reconstructing the datetime from time0 and time1 and then do a check for in. 

Does that make sense? 

Best regards
Daniel

ebaauw:
Hm, there was also an issue with changed schedules not firing at the new scheduled time, when they has already fired before, see #67.

djwlindenaar:
I didn't look into schedule handling, so I don't know if that's related.

ebaauw:
[...]

PR #582, opened 2018-05-11
Bug fixes rules, see #413

ebaauw:
- Set `lasttriggered` in UTC instead of localtime;
- Fix firing rule with `in` condition on `/config/localtime`;
- Don't restore `timestriggered` from database on restart (so it resets to 0).

### DIFF
--- a/rest_rules.cpp
+++ b/rest_rules.cpp
@@ -1337,26 +1337,18 @@ bool DeRestPluginPrivate::evaluateRule(Rule &rule, const Event &e, Resource *eRe
             }
 
             QDateTime dt = item->lastChanged().addSecs(c->seconds());
-            if (rule.lastTriggered().isValid() && rule.lastTriggered() > dt)
+            if (now.secsTo(dt) != 0)
             {
-                return false; // already handled
-            }
-
-            if (dt > now)
-            {
-                return false; // not time yet
+                return false;
             }
         }
         else if (c->op() == RuleCondition::OpIn && c->suffix() == RConfigLocalTime)
         {
             QTime t = now.time();
 
-            if (rule.lastTriggered().isValid() && rule.lastTriggered().secsTo(now) < 24 * 3600)
+            if (eItem->descriptor().suffix == RConfigLocalTime && t.secsTo(c->time0()) != 0)
             {
-                if (eItem->descriptor().suffix == RConfigLocalTime)
-                {
-                    return false; // already handled
-                }
+                return false; // Only trigger on start time
             }
 
             if (c->time0() < c->time1() && // 8:00 - 16:00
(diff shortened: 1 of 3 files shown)


rest_rules.cpp
```diff
@@ -1350,35 +1350,23 @@ bool DeRestPluginPrivate::evaluateRule(Rule &rule, const Event &e, Resource *eRe
         else if (c->op() == RuleCondition::OpIn && c->suffix() == RConfigLocalTime)
         {
             QTime t = now.time();
-            QTime rt;
 
-            if (rule.lastTriggered().isValid())
+            if (rule.lastTriggered().isValid() && rule.lastTriggered().secsTo(now) < 24 * 3600)
```
manup: Does this work: `lastTriggered()` is UTC and `now` is local time?

ebaauw: Funny enough it did.  Apparently the timezone is only considered when converting to/from a string.

Functionally it doesn't do what I wanted, thought, see below.

ebaauw:
The fix for `in` condition on `/config.localtime` didn't work in full.  I changed the logic to no longer consider `lasttriggered`, but compare to the actual time (using `QTime::secsTo()` to discard the milliseconds).  This assumes `evaluateRule()` get called every second for `/config/localtime` - which it seems to be the case in my testing.  Might be safer to actually use the value for `/config/localtime` rather than `now` in these comparisons.

Also added support for `not in` and changed the `ddx` logic.  I saw some funny stuff that `ddx` rules fired incorrectly on creation, due to the `lasttriggered` logic.

ebaauw:
Damn, unintentionally included the change to simplify the code for the `lumi.sensor_86sw` switches in this PR as well.  Won't do any harm, though.

rest_rules.cpp
manup: Interesting, maybe they also use the timezone info from QDateTime class and do some internal conversions.

manup:
>The fix for in condition on /config.localtime didn't work in full. I changed the logic to no longer consider lasttriggered, but compare to the actual time (using QTime::secsTo() to discard the milliseconds). This assumes evaluateRule() get called every second for /config/localtime - which it seems to be the case in my testing. Might be safer to actually use the value for /config/localtime rather than now in these comparisons.

Yes I think that would be fine */config/localtime* is updated once every second and related indexed rules should be evaluated.

>Also added support for not in and changed the ddx logic. I saw some funny stuff that ddx rules fired incorrectly on creation, due to the lasttriggered logic.

:) cool thanks, I'll check it out soon, we use ddx heavily for the Phoscon App motion sensor controls. The not in might simplify some other rules too.

ebaauw:
> we use `ddx` heavily for the Phoscon App motion sensor controls. The `not in` might simplify some other rules too.

I use `ddx` extensively as well, in the logic to determine wether I left a room (motion in room, followed by motion in adjacent room, followed by no more motion in room).  Been running my latest commit in production since yesterday - it seems to work.

I've never used `not in` myself - it's functionally equivalent to `in` with the times reversed.  I did some basic tests, and it seems to work.

As mentioned elsewhere, to simplify rules, I would like an `in` (and `not in`) for numeric values as well (e.g. `state.status`, `state.buttonevent`).  Preferably they would take a range (like `"`_n_`/`_m_`"`) or a list (like `"`_n1_`,`_n2_`,`_n3_`"`).  My use case for a list would be a single rule to set `bri_inc` to 0 on long release of a Hue dimmer switch's dim up (2003) or dim down (3003).  
I also think we would benefit from `le` (<=) and `ge` (>=) operators, to make rules easier readable.  Of course, this is of less importance when the rules are generated by a script or by an app.

PR #606, opened 2018-05-21
Updated instructions for disabling deCONZ GUI

saurabh984:
deCONZ GUI is started as a systemd service not an X11 autostart script

### DIFF
--- a/README.md
+++ b/README.md
@@ -106,6 +106,13 @@ $ sudo systemctl disable deconz-gui
 $ sudo systemctl stop deconz-gui
 ```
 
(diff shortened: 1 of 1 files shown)


manup:
Yes and no, the X11 autostart script was part in older deCONZ versions and documentations, it's still there on many systems and should therefore be removed (if existing). To stop and disable the deconz-gui service is right.

saurabh984:
@manup Updated.

manup:
Thanks, looks good, I'll merge later with the other PRs.

PR #630, opened 2018-06-08
innr SP 120 smart plug

ebaauw:
Support innr SP 120 smart plug, see #432.

@manup, the ZCLDB file general.xml needs to be updated as well, see https://github.com/dresden-elektronik/deconz-rest-plugin/issues/432#issuecomment-395757777.  I'm still not entirely sure if the setup of the binding for attribute reporting works.  The (light) _OnOff_ and _Level_ clusters work, but the (sensor) _Electrical Measurement_ and _Metering_ clusters seem more challenging.

Bug fix `gt` and `lt` operators in rule conditions.

### DIFF
--- a/bindings.cpp
+++ b/bindings.cpp
@@ -907,42 +907,82 @@ bool DeRestPluginPrivate::sendConfigureReportingRequest(BindingTask &bt)
     }
     else if (bt.binding.clusterId == METERING_CLUSTER_ID)
     {
+        Sensor *sensor = dynamic_cast<Sensor *>(bt.restNode);
+
         rq.dataType = deCONZ::Zcl48BitUint;
         rq.attributeId = 0x0000; // Curent Summation Delivered
         rq.minInterval = 1;
         rq.maxInterval = 300;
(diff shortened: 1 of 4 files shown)


manup:
Thanks, looks finally a usable EU plug is available :)

>Bug fix gt and lt operators in rule conditions.

I'll check if Phoscon sensor controls are working with that (they use gt/lt) .

ebaauw:
I've only used `gt` and `lt` for temperature sensors, to control my dumb fan (connected through a smart plug).  For light sensors, I find `state.daylight` and `state.dark` far more convenient, since you can simply update the thresholds instead of changing the rules.

manup:
We use it mostly to implement state machines with help of CLIPStatus and rules.
Like if status > -1; do something and set status to another value.

ebaauw:
Indeed.  I'm doing that as well, actually.  Didn't think of that, because the bug only manifests itself if you e.g. change value from 1 to 2, where you have a condition `gt` 0.

Issue #631, opened 2018-06-09
Compile error with latest commit (ubisys C4 config)


ebaauw:
I get a compile error with the latest commit:
```
bindings.cpp: In member function ‘void DeRestPluginPrivate::processUbisysC4Configuration(Sensor*)’:
bindings.cpp:2084:27: error: ‘ZclWriteAttributesStructuredId’ is not a member of ‘deCONZ’
     zclFrame.setCommandId(deCONZ::ZclWriteAttributesStructuredId);
```
I assume it needs a newer version of deconz-dev.2.05.xx.deb ?

**EDIT** Just commenting out the offending line seems to work around this issue for me.  I don't have a C4, so this code will probably never run in my configuration...

manup:
Sorry, forgot to upload. The -dev file is now updated and should work with latest commits.

holli73:
hello,

i have same issue un x64 linux but no new -dev file?

thanks
holli

manup:
Hey Holli,
version 2.05.29 is online now with updated -dev package.

Issue #138, opened 2017-08-24
Xiaomi Mi Smart Cube


ebaauw:
New gadget!

![untitled](https://user-images.githubusercontent.com/22179355/29677149-868aa578-88fa-11e7-900c-91bb0bf29a45.jpeg)

It sends unicast _Report Attribute_ commands to the RaspBee, endpoint 1:
- From endpoint 2 to cluster 0x0012, attribute 0x0055, uint16 on double-tap, flip, move.  The attribute value seems to indicate the kind of action, but it's going to be a hell of a job to reverse-engineer what value corresponds to what action.  I think they use different values for flipping the cube from side x to side y for every x, y combi.
- From endpoint 3 to cluster 0x000c, attribute 0xff05, uint16, and attribute 0x0055, float (0x39), on turn. 0xff05 always has value 0x01f4 (500), 0x0055 seems to indicate the angle of turning.

ebaauw:
I think I figured out the _Multistate Input_ cluster 0x0012).  The cube has six sides (duh), numbered as follows:
```
    +---+
    | 2 |
+---+---+---+
| 4 | 0 | 1 |
+---+---+---+
    | 5 |
    +---+
    | 3 |
    +---+
```
Side 5 is with the MI logo; side 3 contains the battery door.

Depending on the action, the following values for 0x0012/0x0055 are reported:
- _Shake_: `0x0000` (side on top doesn't matter)
- _90º Flip_ from side `x` on top to side `y` on top: `0x0040 + (x << 3) + y`
- _180º Flip_ to side `x` on top: `0x0080 + x`
- _Push_ while side `x` is on top: `0x0100 + x`
- _Double Tap_ while side `x` is on top: `0x0200 + x`

_Push_ works in any direction.  For _Double Tap_ you really need to lift the cube and tap it on the table twice.  
Sometimes I see value 0x0002 - I think when putting the cube on the table/single tapping, but this is not repeatable.  Maybe it's when "waking" the cube.

In the _Cluster Info_ plane in the deCONZ GUI, the 0x0012/0x0055 attribute is updated in real-time, but the 0x000c/0x0055 attribute isn't.  In the _Report Attribute_ command, this attributes is reported seconds, after the 0xff05 attribute.  Could this be a bug in the deCONZ GUI, that it discards the whole message, when it finds an unknown attribute?

manup:
[...]

Issue #610, opened 2018-05-25
Firmware 26210500 strange behaviour


stefan1957:
Hello,
The firmware 26210500 from the beta 2.05.28 is not working as expected with homebridge-hue v0.9.3. Only 40 of the 51 nodes work in HomeKit. In the phoscon app is everything working normally. I ll try to downgrade to 2.05.27 with the same result. So I think that there is something wrong with the firmware. I hope you can fix this
Gr Stefan

ebaauw:
I upgraded to v2.05.28 and 0x26210500 this morning.  One of my IKEA lights would only be found by deCONZ after power-cycling it, but otherwise I've seen nothing weird.  Not even sure if the IKEA light issue is related to the upgrade.

If things are working normally in the Phoscon app, I don't think there's an issue with deCONZ.  Not sure what you mean by "work in HomeKit".  Best to open an issue in homebridge-hue, include debug log file and debug dump file.

alexvandervegt:
@manup When will the new firmware be uploaded to:

http://dresden-elektronik.de/rpi/deconz-firmware/?C=M;O=A ?

manup:
Firmware is part of the deCONZ package, you can find in in `/usr/share/deCONZ/firmware`.

stefan1957:
@manup 
i downgrade to deCONZ_Rpi_0x261f0500.bin.GCF and my problems are solved.  So i still think, that in my case, the firmware 0x26210500 isn't oke. Btw the phoscon app is also unstable with the new firmware.

alexvandervegt:
I have a lot xiaomi wireless switches which lose the network since the upgrade

sberryman:
I just upgraded and wished I hadn't... My system is no longer pushing updates from (hue motion) sensors. They are all there and show up with the correct data for polling but nothing updates over the websocket

ebaauw:
I noticed that after recalling a scene, the light states aren’t updated.  Also, my Hue motion sensors blink red when detecting motion, and there seems to be a slight delay, before deCONZ reacts to these.  I won’t be able to test until this weekend.

@sberryman you can downgrade the firmware using GCFFlasher_internal and you can downgrade deCONZ by simply installing an older version.

manup:
@sberryman do the sensors blink blue from time to time or is there no communication at all?
I've running the latest firmware version in two setups with hue motion sensors, here they work &mdash; I need to do more tests, also with the IKEA join problems.

[...]

PR #632, opened 2018-06-09
Xiaomi cube drop action, see #138

ebaauw:
Newly discovered action for the Xiaomi cube: drop, see #138.

Fix scaling of current for Heiman smart plug.

@manup, sorry, I had to comment out a line in the handling of the C4 setup, see #631.

@manup could you please also add some more attributes to clusters 0x0702 and 0x0B04 to general.xml?  (Still waiting for general.xml to be part of this respository ;-)  Most plugs don't support these, but the Heiman does.

[general.xml.zip](https://github.com/dresden-elektronik/deconz-rest-plugin/files/2086889/general.xml.zip)

### DIFF
--- a/bindings.cpp
+++ b/bindings.cpp
@@ -979,6 +979,10 @@ bool DeRestPluginPrivate::sendConfigureReportingRequest(BindingTask &bt)
         {
             rq3.reportableChange16bit = 100; // 0.1 A
         }
+        else if (sensor && sensor->modelId() == QLatin1String("SmartPlug")) // Heiman
+        {
+            rq3.reportableChange16bit = 10; // 0.1 A
+        }
         else
         {
             rq3.reportableChange16bit = 1; // 0.1 A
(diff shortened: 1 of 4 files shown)


ebaauw:
@manup I see a lot of inconsistencies in initialising attributes on sensor pairing.  Sometimes they are initialised, sometimes they are null until the first report comes in.  I think `state` attributes shouldn't be initialised, but `config` attributes should be.  Do you agree?  Same for restoring sensors from the database?  And for CLIP sensors?  There seem to be some inconsistencies which `state` and (`config`?) attributes can be specified at sensor creation as well.

ebaauw:
I undid the disabling of updating the light states on a scene recall, see #610.  Even if we're able to fast-poll the lights after a scene recall, I think it's prudent to set the REST light state from the scene.  I'm now on 40+ lights and polling takes forever.

Issue #663, opened 2018-06-28
Hue app with deCONZ


KodeCR:
I'm trying to improve the compatibility of the deCONZ REST api with the Hue app (using the latest version on iOS). This is currently somewhat working but there are several submitted issues already, some quite old, so I'm trying to gather where we are at.

This is my attempt to list the currently known issues, and hopefully progress some of them:  
1. ~Xiaomi sensors seem to crash the Hue app~ done, #1046 
2. ~Searching for lights doesn't work #28 (I have implemented this now, will create a pull request)~ done, #664 
3. Scenes api is not implemented #29
4. Can't update deCONZ through the Hue app, it does show updates
5. Light api differences #30 (this might be sorted now, but issue is still open)
6. Philips Ambilight TVs might not work #440
7. Routines / capabilities api is missing  #584 (**partially in #1005**)
8. Rooms are implemented differently #596

~Any more?~
9. ~Update upnp and hue-bridge v2 description~, done #818 
10. https support, #769 
11. creating a group/room only works without lights, and then add them after creation

Seaha4487:
Hey and how get i to work the part with the hue app?
The app finds my raspbee but wants me to create a room and this seems not to work cause i always get an error message if i try to save the room with my hue bulb in it

Issue #28, opened 2017-04-21
Search for new devices through API


ebaauw:
When I issue a POST to `/api/<username>/lights` the deCONZ REST API responds: `[{"success":{"/lights":"Searching for new devices"}}]`.  The Wireless Light Control web interface shows that the network is open for new devices.  However, when doing a GET of `/api/<username>/lights/new`, it always returns `{"lastscan":"2012-10-29T12:00:00"}`, even when the search is active.

When I issue a POST to `/api/<username>/sensors`, the API responds: `[{"success":{"/sensors":"Searching for new devices","/sensors/duration":60}}]`.  Again, the web interface shows that the network is open.  A GET of `/api/<username>/sensors/new` returns `{"lastscan":"active"}`.  After the search is done, it returns: `{"lastscan":"2017-04-21T15:44:57"}`.

When I open the network from the Wireless Light Control web interface, `/sensors/new` reports an active scan, but `/lights/new` doesn't.  When I open the network through POST `/lights`, `/sensors/new` doesn't show an active scan.  According to Wireshark, the web interface PUTs `{"permitjoin": 60}` to `/api/<username>/config`

On the Hue bridge, `/lights/new` and `/sensors/new` always return the same value for `lastscan`, regardless whether the search is initiated from POST `/lights` or POST `/sensors`.

manup:
The POST interface `/api/<username>/lights` needs to be improved, it's already on the list.
However sensor and light search will be kept separated.

It's planned to have optional filters for the request, like searching only for a certain model identifier. Especially in larger environments there is always the risk that devices join the network which were not intended to join. The gateway should be able to take care of it and send ZDP leave requests to these devices.

manup:
Closing the oldest issues for know to tidy up the tracker and duplicates in newer issues.

PR #664, opened 2018-06-28
Implement searching for new lights based on the searching for sensors code.

### LINKED ISSUE #29: Scene API

KodeCR:
This implements searching for new lights through "GET /api/<apikey>/lights/new" and reporting them in a way that works with the Hue app. Hopefully this fixes ~~#29~~ #28. Also see #663.

I've basically copied the code for searching for new sensors. However, the sensors code seems to be quite a bit more involved. In order to track what is going on, and for consistency I renamed the sensor code from "find" to "search". Then I left out what didn't seem to apply to light nodes. It would be good for someone more knowledgeable to double check this though. One more change is that I moved the resendPermitJoinTimer to where the permitJoinTimer code is, that seemed more sensible.

Obviously this works for me, but some more testing is probably a good idea. Hope this helps.

### DIFF
--- a/bindings.cpp
+++ b/bindings.cpp
@@ -1288,7 +1288,7 @@ void DeRestPluginPrivate::checkSensorBindingsForAttributeReporting(Sensor *senso
         return;
     }
 
-    if (findSensorsState != FindSensorsActive &&
+    if (searchSensorsState != SearchSensorsActive &&
         idleTotalCounter < (IDLE_READ_LIMIT + 120)) // wait for some input before fire bindings
     {
         return;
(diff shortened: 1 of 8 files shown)


manup:
Thanks for adding this, I'll merge it later on. 

From quick look I think we can later join the two state machines into one to prevent some uncanny side effects, like starting light and sensor search side by side, and also keep permit join handling at one place.

KodeCR:
Thanks. Yes, I considered merging the state machines, but you note in #28 that you'd like to keep the light and sensor search separate for optional filters so I wasn't sure.

manup:
Took a while, it's merged now :)

>Thanks. Yes, I considered merging the state machines, but you note in #28 that you'd like to keep the light and sensor search separate for optional filters so I wasn't sure.

Merging the state machine should be fine since they are still be be separated via the different requests. Internally only one can be active at a certain time.

PR #709, opened 2018-08-04
Add initial support for ubisys J1 Window covering device

ma-ca:
Adding ubisys J1 to the supportedDevices array in de_web_plugin.cpp and WINDOW_COVERING_CLUSTER_ID to de_web_plugin_private.h. Also creating window_covering.cpp to later handle cluster attributes and commands.


manup:
Thanks for the PR, I like the new *window_covering.cpp*. The handling as light will be only a temporary solution until the more general devices endpoint arrives.

PR #718, opened 2018-08-08
Refactor MAC checks, add new IKEA prefixes

lbschenkel:
Multiple MAC prefixes have been seen in the wild for IKEA devices.
Because of this, existing checks looking for specific MAC prefixes are fragile.
Therefore, this PR changes existing MAC checks to invoke a new `checkMacVendor()` function that supports multiple prefixes per vendor. In addition, IKEA check has been amended to support the 4 different prefixes seen in the wild.

---

As discussed in https://github.com/dresden-elektronik/deconz-rest-plugin/pull/700

### DIFF
--- a/bindings.cpp
+++ b/bindings.cpp
@@ -123,10 +123,10 @@ bool DeRestPluginPrivate::readBindingTable(RestNodeBase *node, quint8 startIndex
     Resource *r = dynamic_cast<Resource*>(node);
 
     // whitelist
-    if ((node->address().ext() & macPrefixMask) == deMacPrefix)
+    if (checkMacVendor(node->address(), VENDOR_DDEL))
     {
     }
-    else if ((node->address().ext() & macPrefixMask) == ubisysMacPrefix)
+    else if (checkMacVendor(node->address(), VENDOR_UBISYS))
     {
     }
     else if (r && r->item(RAttrModelId)->toString().startsWith(QLatin1String("FLS-")))
@@ -964,12 +964,12 @@ bool DeRestPluginPrivate::sendConfigureReportingRequest(BindingTask &bt)
         rq.dataType = deCONZ::ZclBoolean;
         rq.attributeId = 0x0000; // on/off
 
-        if ((bt.restNode->address().ext() & macPrefixMask) == deMacPrefix)
+        if (checkMacVendor(bt.restNode->address(), VENDOR_DDEL))
         {
             rq.minInterval = 5;
             rq.maxInterval = 180;
         }
-        else if ((bt.restNode->address().ext() & macPrefixMask) == xalMacPrefix ||
+        else if (checkMacVendor(bt.restNode->address(), VENDOR_XAL) ||
                  bt.restNode->node()->nodeDescriptor().manufacturerCode() == VENDOR_XAL)
         {
             rq.minInterval = 5;
@@ -1072,7 +1072,7 @@ bool DeRestPluginPrivate::sendConfigureReportingRequest(BindingTask &bt)
         rq.dataType = deCONZ::Zcl8BitUint;
         rq.attributeId = 0x0000; // current level
 
-        if ((bt.restNode->address().ext() & macPrefixMask) == deMacPrefix)
+        if (checkMacVendor(bt.restNode->address(), VENDOR_DDEL))
         {
             rq.minInterval = 5;
             rq.maxInterval = 180;
@@ -1116,8 +1116,7 @@ bool DeRestPluginPrivate::sendConfigureReportingRequest(BindingTask &bt)
 
         return sendConfigureReportingRequest(bt, {rq, rq2, rq3, rq4});
     }
-    else if (bt.binding.clusterId == BASIC_CLUSTER_ID &&
-            (bt.restNode->address().ext() & macPrefixMask) == philipsMacPrefix)
+    else if (bt.binding.clusterId == BASIC_CLUSTER_ID && checkMacVendor(bt.restNode->address(), VENDOR_PHILIPS))
     {
         Sensor *sensor = dynamic_cast<Sensor*>(bt.restNode);
         if (!sensor)
@@ -1316,7 +1315,7 @@ void DeRestPluginPrivate::checkLightBindingsForAttributeReporting(LightNode *lig
             }
 
             BindingTask bt;
-            if ((lightNode->address().ext() & macPrefixMask) == deMacPrefix)
+            if (checkMacVendor(lightNode->address(), VENDOR_DDEL))
             {
                 bt.state = BindingTask::StateCheck;
             }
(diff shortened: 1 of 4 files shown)


manup:
I wouldn't remove the ikeaMacPrefix, but rather check for both mac prefix or alternatively the vendor code in order to to provide the highest chance of detection. In case of battery powered devices it might happen that the vendor code is not known at the beginning. MAC address is always known first, the vendor code only after the node descriptor has been received.

lbschenkel:
OK, that's sensible. However, since checking MAC address is important I did ask if we should have multiple prefixes for IKEA devices and my interpretation was that you were opposed to the idea? I have two prefixes in my household and my remotes are using the "new" (`0x90fd9f`) prefix and wouldn't match via the existing MAC comparison.

lbschenkel:
Just an idea: I think the code would benefit from having a single generic function like `get_vendor(*address, *zclFrame)` (either could be NULL) and returns the vendor code such as `VENDOR_IKEA`. It could be used everywhere where such checks are done and would improve the readability/maintainability of the codebase by encapsulating the matching logic and normalizing different checks/sources to the vendor codes.

I'm still a novice regarding this codebase and the ZigBee protocol, so please do say if this idea is completely crazy...

jurriaan:
fyi: My TRADFRI control outlets + on/off switch also have another prefix (`0xd0cf5e`)

embeddeddev:
Please merge, this makes sense.
Actually I'm unable to join some IKEA GU10 (w/o white spectrum).

By the way: In the commit only the extern definition of ``ikeaMacPrefix`` in the header file was removed but not the variable declaration in any source file itself, that should be unused now.

lbschenkel:
I have pushed an incomplete, proof-of-concept implementation for what I meant on https://github.com/dresden-elektronik/deconz-rest-plugin/pull/718#issuecomment-412036555 for IKEA and Dresden devices.  What do you think, @manup? I can then replace all other MAC checks in the codebase to invoke `get_vendor()` instead (and then refactor the code slightly to avoid calling the method in each `if` when performing these checks).

lbschenkel:
I actually went ahead and replaced all MAC checks that I could find. I just left TI because I'm not sure how to handle that case (or a different case of the same MAC being used by multiple vendors). I'm brainstorming some ideas.

manup:
Hi thanks, going to merge this soon, need to do more tests to check that all vendors work as expected. I think in future we can extend this to also take things like manufacturer attribute from basic cluster into account.

One small request, please use Qt C++ style for consistency with the code base:

* camel case for function and variable names
* prefer qt types like quint16
* function braces on next lines

```
quint16 addressToVendor(const quint64 &ext)
{
    const quint64 prefix = addr & macPrefixMask;
    ...
}
```

lbschenkel:
Right. TBH I did it quickly and didn't expect you to merge as-is, it was more to explain the concept. I can tidy it up, no problem.

I'm actually trying a different approach in which you the method is `is_mac_for_vendor()` (or the camel case equivalent), because in this way you can easily handle same vendor using different MACs... I'll push a new commit and then you can choose which style you believe works best.

lbschenkel:
Pushed new commit, now "inverting" the method. I prefer this, because it handles multiple vendors using the same prefixes. If you agree I can squash it with the previous commit, erasing it.

Now, with the method expressed like this, I realized that it could be further improved by populating it automatically with the data from https://github.com/dresden-elektronik/deconz-rest-plugin/blob/master/de_web_plugin.cpp#L93 so you don't need to maintain a static table. When new prefixes are seen in the wild, all you need to do is to add a new entry for the device there and it will be recognized everywhere else. However, I'm not a Qt developer and I'm rusty on modern C++, so I'll let this be an exercise for the reader... :-)

lbschenkel:
Changed this PR title and description to accurately describe the new approach.

lbschenkel:
I rechecked all my IKEA devices and I actually have 4 prefixes in my household:
`0x000b57`: all motion sensors, some 5-button remotes
`0x000d6f`: on/off remotes for power plugs
`0x90fd9f`: all lights, some 5-button remotes
`0xd0cf5e`: power plugs

Result from https://www.wireshark.org/tools/oui-lookup.html:
```
00:0B:57 SiliconL Silicon Laboratories
00:0D:6F Ember Ember Corporation
90:FD:9F SiliconL Silicon Laboratories
D0:CF:5E EnergyMi Energy Micro AS
```
I have amended the commit so all 4 prefixes are recognized when looking for IKEA.

lbschenkel:
@manup: Since there has been a lot of activity in `master` lately, please let me know when it's a good time for me to rebase this PR again and solve the conflicts. I want to avoid having to do that multiple times if possible.

manup:
I think a good time will be after the 2.05.57 release which will arrive very soon. So we can schedule PR merge for 2.05.58.

 I've noticed a few spots where mac checks might not be sufficient. I'll add a further function like `checkVendor(Resource*, quint16 vendor)` which can access further attributes like manufacturer code and name to check for a specific vendor. If that fails the function will delegate to `checkMacVendor()`. This is of course only possible if a Resource is available at the given time.

ebaauw:
I appreciate this PR and the simplification it provides.  I do worry that the deCONZ REST API code is still very complex, even after this PR.  I think going forward, we might need to redesign how to deal with whitelisting devices.

A lot of checks for mac address prefix, manufacturer, model ID, etc. are done to determine certain properties of the device (does it support attribute reporting, how often does it wake, does it send a "no more motion" message, how to scale the consumption and power values, ...), rather than identifying the device itself.  These checks are all over the code; I regularly find when adding support for a new device, I miss one of two spots.

I would like to model all these properties into a single device repository.  Instead of checking for the device through mac address prefix, manufacturer code, model ID to derive a property, the code would simply check for that property.  This would provide a single place to add support for new devices.

lbschenkel:
I completely agree with you, @ebaauw. The way the checks are done in the code seem to be very hard to maintain, to me. I share your idea that a device repository documenting capabilities/behaviour and doing feature checks in the code would be a much better model.

Having said that, this PR is not that ambitious. The initial intent was just to have the existing code that checks for IKEA MACs to recognize the additional prefixes used in the wild — and that was it. Then it got generalized to the other MAC checks. But it's just a bit of sugar, it does not fundamentally address any of the issues with the existing checks besides centralizing the mappings and being able to add new prefixes to existing manufacturers without having to change all the existing checks.

I suggest that the scope of this PR is kept as it is — just to put some sugar and improve the readability of the MAC checks. It's already a bit bigger than it should be because it also has the new IKEA prefixes (I can split that out if necessary). I would say that any further work, including adding a `Resource` param, it's better done afterwards — but I want to emphasize that `checkVendor()` was not intended to apply to anything else besides the MAC checks, and I strongly back @ebaauw suggestion that a device database is a much better model.

lbschenkel:
FYI @manup: rebased against latest master.

PR #719, opened 2018-08-08
Emulate "ct" for color lights ("xy") that lack color temperature support

lbschenkel:
As discussed in https://github.com/dresden-elektronik/deconz-rest-plugin/pull/700


manup:
Short note: I've testet the new IKEA beta firmware 1.3.009 CWS and was hoping they added the missing  CT commands, but sadly nothing new here.

Issue #534, opened 2018-04-13
Support for ubisys Shutter Control J1?


felixstorm:
Hi @manup, hi @ebaauw,

I saw from #404, #269 and #209 that you're currently thinking about opening up deCONZ for other actuators besides lights. I have quite a few ubisys J1 that I have paired with deCONZ but currently I can't really control them in any way. It would be great if general support for them could be included as well - so if I can be of any help with this please get in contact with me.

And many thanks for the great work on deCONZ!

Felix

felixstorm:
They have pretty detailed ZigBee information available here:  
http://ubisys.de/downloads/ubisys-j1-technical-reference.pdf

A few months ago I had tried to add their configuration attributes to general.xml, but as they use pretty tricky ZigBee attribute type configurations especially to configure the inputs on their devices, this did not really work out.

felixstorm:
Sorry to spam you with multiple comments, but another thing just came to my mind: A functionality to directly bind ZigBee switch inputs (like the C4s) to the J1s on the ZigBee level would also be helpful for me, but I am not sure if this would also be within the scope of #404, #269 etc.

schneidermarc:
Hi Felix,

take a look at my zcl file:
https://gist.github.com/schneidermarc/0a5362d4bd60d2301c536e37a9497e1c

My J1 works petty well on the DeCONZ GUI.  I am currently in contact with the support to clarify the possibilities to link the zcl profiles with theREST API.

Marc

felixstorm:
Great, thanks @schneidermarc - I will definitely try it out for control as soon as I have a GUI again. Currently, my deCONZ is running headlessly in a docker container, but should probably soon get an X11 UI again (see https://github.com/marthoc/docker-deconz/issues/15).

When I tried customizing the zcl file myself a few months ago, I wanted to configure buttons and calibration first and this has been implemented by ubisys with rather complicated zcl types (InputActions is an array of attributes with subtype raw binary data) and also requires writing to attributes with the "manufacturer specific" bit set - both of which deCONZ can't or at least could not do at that time, so I stopped looking in this direction. Back then I managed to do these tasks with the help of an XBee module, an abandoned ZCL library for C# and a few lines of code, but it was far from being pretty.

As I use Home Assistant now and temperatures get warmer in the norther hemisphere, automatic shutter control for sun protection has now moved up quite high on my priorities list and seeing #404 etc., I drew new hope ;-)  
[...]

PR #727, opened 2018-08-12
Add ubisys J1 to lights with states on and bri. 

ma-ca:
Add ubisys J1 to `/lights` with states _on_ and _bri_. Sending `PUT { "on": true/false }` triggers J1 to moveDown/Up. Sending `PUT { "bri": "stop" }` triggers a stop. Sending `PUT { "bri": <value> }` triggers move to lift percentage. The moving to lift percentage only works for calibrated J1. Moving up/down/stop always works.

The J1 is a shuttering device (window covering). Adding the J1 device to lights is just a temporary workaroud and will be fixed later when generic endpoint devices becomes available.

### DIFF
--- a/de_web.pro
+++ b/de_web.pro
@@ -147,6 +147,7 @@ SOURCES  = authentification.cpp \
            reset_device.cpp \
            rest_userparameter.cpp \
            zcl_tasks.cpp \
+           window_covering.cpp \
            websocket_server.cpp
 
 win32 {
--- a/de_web_plugin_private.h
+++ b/de_web_plugin_private.h
@@ -1073,6 +1073,7 @@ public:
     bool addTaskAddEmptyScene(TaskItem &task, quint16 groupId, quint8 sceneId, quint16 transitionTime);
     bool addTaskAddScene(TaskItem &task, uint16_t groupId, uint8_t sceneId, const QString &lightId);
     bool addTaskRemoveScene(TaskItem &task, uint16_t groupId, uint8_t sceneId);
+    bool addTaskWindowCovering(TaskItem &task, uint8_t cmdId, uint16_t pos, uint8_t pct);
     void handleGroupClusterIndication(TaskItem &task, const deCONZ::ApsDataIndication &ind, deCONZ::ZclFrame &zclFrame);
     void handleSceneClusterIndication(TaskItem &task, const deCONZ::ApsDataIndication &ind, deCONZ::ZclFrame &zclFrame);
     void handleOnOffClusterIndication(TaskItem &task, const deCONZ::ApsDataIndication &ind, deCONZ::ZclFrame &zclFrame);
--- a/light_node.cpp
+++ b/light_node.cpp
@@ -475,7 +475,7 @@ void LightNode::setHaEndpoint(const deCONZ::SimpleDescriptor &endpoint)
             for (; i != end; ++i)
             {
 
-                if (i->id() == LEVEL_CLUSTER_ID)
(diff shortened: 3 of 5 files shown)


ebaauw:
I think it would make more sense to send `{"bri_inc": 0}` to stop the movement.  Not sure if the J1 supports relative lift, but that could also be supported through `bri_inc`.

Just to double-check: lift percentage 0 should match off (which means closed?), so setting `{"bri": 0}` would have the same effect as setting `{"on" false}`.

ma-ca:
I have added `{ "bri_inc": 0 }` to trigger stop command. 

Sending `{ "on": false }` will move up/open because bri = 0 translates to tilt percentage = 0% meaning fully open. But I have to admit that this is confusing.  

Together with homebridge-hue the J1 is shown in HomeKit always being switched off. Pressing the button would therefore close the shutter. The brightness level allows to move to tilt percentage. For now a usable workaround.

ebaauw:
I’ll see if I can expose the J1 as a HomeKit shutter or something in homebridge-hue.

ebaauw:
HomeKit distinguishes between current position and target position of the shade.  I suppose with the right attribute reporting settings, the J1 reports back the actual position during movement?  At least, that's what my D1 does while dimming slowly.

@ma-ca can you check with `dc_eventlog` that, when the shutter is closed, PUTting `{"on":  false}`  to open the shutter results in a series of web socket events for `state.bri` that trace the current state of the shutter?

ma-ca:
@ebaauw, the J1 does, when configured accordingly, continuously report the lift percentage during movement. Cluster `0x0102` attribute `0x0008` type `0x20` unsigned 8-bit int with range 0 to 100 (hex 0x00 - 0x64).
But this PR #727 did only implement sending commands to J1 when PUTting `on` or `bri` values. Handling ZCL reporting values from `0x0102` cluster attribute `0x0008` and translating to `bri` has not been implemented yet. I can add another PR if that is needed for HomeKit to work.

So the `dc_eventlog` command does not show anything when the shutter is moving.

ebaauw:
> I can add another PR if that is needed for HomeKit to work.

I think it's needed.  The alternative would be to configure a timeout and assume the motion has ended when the timeout occurs.  I have no clue how to determine the length for this timeout, as it will depend on the height of the shutters and the speed of the motor.

Is `state.bri` updated immediately after you PUT it, or only when the J1 is next polled?  In the first case, you would need to make sure that `state.bri` reflects the actual state and prevent the update of the value on PUT.  For homebridge-hue, it would be enough if you configure the reporting manually.  If you don't, it will only be updated when deCONZ polls the J1, causing HomeKit to report the shutter is moving until then.

ma-ca:
Then I will create a new PR that updates the `state.bri` value, because in the current implementation that value is always 0.

At the moment, no matter what the current position is, in HomeKit the commands "Hey Siri, <my J1 name> on / off / 70%" does move the shutter to the positions fully closed/open/70%  just like any dimmable light. A second later in HomeKit the J1 is shown off 0% still while the shutter is moving to the desired position.

felixstorm:
@ma-ca, @ebaauw - thanks for your work! If you have a chance, it would be very helpful if also tilt state could be implemented somehow in the temporary REST API (setting and attribute reporting), but as the other light attributes (hue, sat, xy, ct) seem to all interact with each other somehow, there might not be a simple solution for that.

Unfortunately, my time is very limited at the moment (and my CPP knowledge also far from great), so I won't have a chance to contribute much here currently (apart maybe from testing).
Therefore I will probably still have to stay with my custom zha cover component in Home Assistant mentioned in https://github.com/dresden-elektronik/deconz-rest-plugin/issues/534#issuecomment-412131966 for now, but it is still great to see progress here.

Issue #593, opened 2018-05-18
Can i use a Xiaomi curtain on RaspBee?


sungokay:
I wanted to connect a Xiaomi curtain to RaspBee and it had connected and reported as a below.
But, it couldn't be used.

Can I use it on RaspBee?

Thanks SungOk.

- Information
    {
        "ExtAddress": "0x00158d00022c80b5",
        "NodeDescriptor": "AUCOXxF/ZAAAAGQAAA==",
        "NwkAddress": "0x2d46",
        "SceneX": 386.0,
        "SceneY": 360.0,
        "SimpleDescriptors": [
            "AQQBAgIBCwAABAADAAUACgACAQ0AEwAGAAEABgQIGQAKAA0AAgETAAYAAQAGBA=="
        ],
        "UserDescriptor": ""
    }

ebaauw:
Interesting, I’ve had my eyes on these for some time, but they’re not supported by deCONZ (yet).  Could you please post deCONZ GUI screenshots of:
- The node with dropped down endpoints and clusters;
- The node info panel;
- The cluster into panel of the _Basic_ cluster (after manually reading the attributes)?

sungokay:
I had attached them.

![screenshot from 2018-05-19 10-48-51](https://user-images.githubusercontent.com/38240278/40263821-dc066336-5b52-11e8-90ef-12cd9f6dd467.png)
![screenshot from 2018-05-19 10-49-16](https://user-images.githubusercontent.com/38240278/40263822-dc30b262-5b52-11e8-8f8e-c65fd810f6d6.png)
![screenshot from 2018-05-19 10-50-22](https://user-images.githubusercontent.com/38240278/40263823-dc5c5200-5b52-11e8-9325-e77f8e5dff60.png)
![screenshot from 2018-05-19 10-51-53](https://user-images.githubusercontent.com/38240278/40263825-dc8677e2-5b52-11e8-8415-7285fe9e3192.png)

ebaauw:
Thanks.  It uses manufacturer code 0x115f and it’s a ZigBee router, just as the Xiaomi plug and newer wall switches (lumi.ctrl_ln).  However, it seems to support over-the-air firmware updates, so it could be a different ZigBee stack.

From the clusters listed, I have no clue how the controller reports its state, nor which commands the controller accepts.  They could very well be using the 0x000d, 0x0013, and/or 0x0102 clusters.  They need to be defined in general.xml.

[...]

PR #746, opened 2018-08-23
window_covering updating bri and adding sat for move to tilt

ma-ca:
Add window_covering cluster updating `bri` value with lift percentage during moving. The update happens in realtime. This requires that reporting has been configured. The updateLightNode will map WINDOW_COVERING cluster attributes lift/tilt percentage to bri/sat on lightnode. 

Add move to tilt percentage for PUT `{ "sat" : value }`. 

Add modelId `lumi.curtain` in rest_lights.cpp setLightState, see issue #593.

The ubisys J1 now shows up in `/lights` with:
```
  "state": {
    "alert": "none",
    "bri": 165,
    "on": true,
    "reachable": true,
    "sat": 0
  },
```
This PR is a follow up on PR #727 and issue #534 

@ebaauw the `dc_eventlog` is now showing output while moving.

Summary: These commands are implemented.

ubisys J1 command | put command
-------- | -----------------------
move down/up | `{"on" : true/false }`
stop | `{"bri_inc" : 0 }`
move to lift pct | `{"bri" : <pct * 255 / 100> }`
move to tilt pct | `{"sat" : <pct * 255 / 100> }`

### DIFF
--- a/de_web_plugin.cpp
+++ b/de_web_plugin.cpp
@@ -2028,6 +2028,56 @@ LightNode *DeRestPluginPrivate::updateLightNode(const deCONZ::NodeEvent &event)
                     }
                 }
             }
+            else if (ic->id() == WINDOW_COVERING_CLUSTER_ID && (event.clusterId() == WINDOW_COVERING_CLUSTER_ID))
+            { // FIXME ubisys J1 begin
+                std::vector<deCONZ::ZclAttribute>::const_iterator ia = ic->attributes().begin();
+                std::vector<deCONZ::ZclAttribute>::const_iterator enda = ic->attributes().end();
+                for (;ia != enda; ++ia)
+                {
+                    if (ia->id() == 0x0008) // current CurrentPositionLiftPercentage 0-100
+                    {
+                        uint8_t level = ia->numericValue().u8 * 255 / 100;
+                        ResourceItem *item = lightNode->item(RStateBri);
+                        if (item && item->toNumber() != level)
+                        {
+                            DBG_Printf(DBG_INFO, "0x%016llX level %u --> %u\n", lightNode->address().ext(), (uint)item->toNumber(), level);
+                            lightNode->clearRead(READ_LEVEL);
+                            item->setValue(level);
+                            Event e(RLights, RStateBri, lightNode->id(), item);
+                            enqueueEvent(e);
+                            updated = true;
+
+                            // also change on-state if bri changes to/from 0
+                            bool on = (ia->numericValue().u8 > 0 ? true : false) ;
+                            ResourceItem *itemOn = lightNode->item(RStateOn);
+                            if (itemOn && itemOn->toBool() != on)
+                            {
+                                DBG_Printf(DBG_INFO, "0x%016llX onOff %u --> %u\n", lightNode->address().ext(), (uint)item->toNumber(), on);
+                                itemOn->setValue(on);
+                                Event e(RLights, RStateOn, lightNode->id(), itemOn);
+                                enqueueEvent(e);
+                                updated = true;
+                            }
+
+                        }
+                        lightNode->setZclValue(updateType, event.clusterId(), 0x0008, ia->numericValue());
+                        break;
+                    }
+                    else if (ia->id() == 0x0009) // current CurrentPositionTiltPercentage 0-100
+                    {
+                        uint8_t sat = ia->numericValue().u8 * 255 / 100;
+                        ResourceItem *item = lightNode->item(RStateSat);
+                        if (item && item->toNumber() != sat)
+                        {
+                            item->setValue(sat);
+                            Event e(RLights, RStateSat, lightNode->id(), item);
+                            enqueueEvent(e);
+                            updated = true;
+                        }
+                    }
+                }
+                break;
+            } // FIXME ubisys J1 end
         }
 
         break;
@@ -8490,6 +8540,7 @@ void DeRestPluginPrivate::nodeEvent(const deCONZ::NodeEvent &event)
         case GROUP_CLUSTER_ID:
         case SCENE_CLUSTER_ID:
         case COLOR_CLUSTER_ID:
+        case WINDOW_COVERING_CLUSTER_ID:  // FIXME ubisys J1 is not a light
             {
                 updateLightNode(event);
             }
--- a/light_node.cpp
+++ b/light_node.cpp
@@ -475,7 +475,7 @@ void LightNode::setHaEndpoint(const deCONZ::SimpleDescriptor &endpoint)
             for (; i != end; ++i)
             {
 
-                if (i->id() == LEVEL_CLUSTER_ID || i->id() == WINDOW_COVERING_CLUSTER_ID /*FIXME ubisys J1*/)
+                if (i->id() == LEVEL_CLUSTER_ID)
                 {
                     addItem(DataTypeUInt8, RStateBri);
                 }
@@ -523,6 +523,11 @@ void LightNode::setHaEndpoint(const deCONZ::SimpleDescriptor &endpoint)
                     default:
                         break;
                     }
+                }
+                else if (i->id() == WINDOW_COVERING_CLUSTER_ID /*FIXME ubisys J1*/)
+                {
+                	addItem(DataTypeUInt8, RStateBri);
+                	addItem(DataTypeUInt8, RStateSat);
                 }
             }
         }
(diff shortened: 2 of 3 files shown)


ebaauw:
Cool.  Hoping to use this for the Xiaomi Aqara curtain controller as well.  You might want to expose bri or sat conditionally, based on the covering type (which should indicate which of lift/tilt are supported).

ma-ca:
>You might want to expose bri or sat conditionally, based on the covering type (which should indicate which of lift/tilt are supported).

Your are right, I have added a check on attribute 0x0000 WindowCoveringType in order to only add bri and/or sat depending on whether lift or tilt or both are configured.

ebaauw:
Good news, this also seems to work for the `lumi.curtain` with a minor change:

In `rest_light.cpp`, instead of checking for `taskRef.lightNode->modelId().startsWith(QLatin1String("J1"))`, check for `taskRef.lightNode->type() == QLatin1String("Window covering device")` so that other _Window covering_ devices will also be supported.

Best create a `bool isWindowCoveringDevice` (right after the `hasOn` etc.) and check for that, to keep the code cleaner, cf. handling of alert for _Warning device_.

ma-ca:
>Best create a bool isWindowCoveringDevice (right after the hasOn etc.) and check for that, to keep the code cleaner

I have implemented that.

ebaauw:
Finally received the curtain rails today.  I had to set the _Reversed_ bit in _Mode_ to match the actual direction with the _Open_ and _Close_ commands (from the GUI).

Not sure if this is related, but PUTting `state.on` works reversed: setting it to `false` opens the curtains, but when the curtains are opened, `state.on` is `true` and `state.bri` is 254.  Likewise, setting `state.on` to `true` closes the curtains, but when they're closed, `state.on` is false and `state.bri` is 0.  Setting `state.bri` seems to work correctly.

ma-ca:
The commands _open/close_ and _move to lift pct_ are executed right, but the value from attribute _CurrentPositionLiftPercentage_ should be "reversed", i.e. 0% = fully closed, 100% = fully open? The current implementation is: 0% corresponds to the open position, 100% to the closed position

In PR #813 I introduce a ZHASwitch with config settings for attribute _WindowCoveringType_. The `config.mode` is already used for ubisys switch settings  `"momentary"` or `"rocker"`. 

So I could add a `config.reversed` = `true/false`. 

And when `{"reversed": true}` the bri value would become `255 - bri`. The `state.on` is always calculated from bri: `state.on = false` if `bri = 0 ` otherwise `state.on = true`.

Would that help?

ebaauw:
> The `state.on` is always calculated from `bri`

Thanks!  That explains why `state.on` and `state.bri` report consistently.

> So I could add a `config.reversed` = true/false. And when `{"reversed": true}` the `bri` value would become `255 - bri`.  Would that help?

Not sure yet.  I want to sniff first to double check that the _OnOff_ cluster doesn't confuse the REST API.  I also need to double check if the `lumi.curtain` reports _Current Position Lift Percentage_ differently when the _Reversed_ bit is set.

Since the _Mode_ attribute is on the same cluster, I think a `state.reversed` on the `/lights` resource would make more sense.  The momentary/rocket config is specific to the ubisys swicthes, and the `lumi.curtain` doesn't have any (high voltage) inputs, so a ZHASwitch resource doesn't make much sense for this device.

Shouldn't it be `254 - bri`?

ebaauw:
Did some tests, no sniffing yet.  The `lumi.curtain` reports the percentage taking into account the _Direction_ setting.  So in normal setting, issuing the _Close_ command results in 0%, in the _Reversed_ setting, issuing _Close_ results in 100%.

Note that it takes 100% to be open and 0% to be closed.  Is that in line with the J2?

ebaauw:
Did some tests, no sniffing yet.  The `lumi.curtain` reports 0% when (it thinks) the curtains are closed, and 100% when (it thinks) they’re open.  So in either setting, sending _Close_ results in 0%, and sending _Open_ results in 100%.  So we don’t need to mess with `bri` because of the _Reversed_ setting.

Does the J2 report _Open_ as 100% or 0%?

ebaauw:
> The ubisys J1 is _Open_ at 0% and _Closed_ at 100%.

Thanks @ma-ca, that would explain the behaviour.  I would blindly trust ubisys over Xiaomi, but let's have a look at the ZCL spec to be sure:
> The _CurrentPositionLiftPercentage_ attribute identifies the actual position as a percentage between the _InstalledOpenLimitLift_ attribute and the _InstalledClosedLimitLif_ attribute of the window covering from the up/open position if Closed Loop Control is enabled.

Indeed, "_from_ the up/open position", so 0% is _Open_.  So that would mean that we need to use 100 - _CurrentPositionLiftPercentage_ for the `lumi.curtain`, regardless of the _Direction_ setting, both when reading and when setting the percentage.  I'm happy to prep a PR for that myself so I can test immediately that it works.

> The existing code did allow `bri` values in the range from 0 to 255. And is works fine with HomeKit where it shows 100% when `bri` = 255, i.e. shades are closed.

The _Current Level_ attribute in the _Level Control_ cluster takes values 0 to 254; 255 means invalid value.  The REST API is a bit ambiguous here (as is the Philips Hue API), you can set `bri` to 255, but it reports back as 254.

> And is works fine with HomeKit 

In homebridge-hue, I multiply (divide) by 100 / 254 when converting REST API values to HomeKit values (or back), see https://github.com/ebaauw/homebridge-hue/blob/e1c406af9abc38e3fc15241d57e1c3fde3a5350c/lib/HueLight.js#L812 and 
https://github.com/ebaauw/homebridge-hue/blob/e1c406af9abc38e3fc15241d57e1c3fde3a5350c/lib/HueLight.js#L1102.

255 * 100 / 254 < 100.5, so 255 rounds off to 100 alright.

overdrive69:
Hi,
i have paird a Ubisys J1 with phoscon. Now i can see the device in lights.
I can turn it on and off. But shouldnt I be able to set percentage and get the actual position of the shutter?
regards

overdrive69:
Hi now i tried to set the J1 ON and OFF via a REST Client.

I can GET the information from the J1 but when i
PUT
{"on" : true}

I receive the following error


[
  {
    "error": {
      "address": "/lights/2",
      "description": "method, PUT, not available for resource, /lights/2",
      "type": 4
    }
  }
]

ebaauw:
You need to use `/lights/2/state` As URI.

overdrive69:
Thanks now the PUT works

[
  {
    "success": {
      "/lights/2/state/on": false
    }
  }
]


but the state "on" didnt changed

{
  "etag": "32ee19a3f1036e7bcaeff1752e8425f9",
  "hascolor": false,
  "manufacturername": "ubisys",
  "modelid": "J1 (5502)",
  "name": "Rollladen001",
  "state": {
    "alert": "none",
    "bri": 138,
    "on": true,
    "reachable": true
  },
  "swversion": "20191104-DE-FB0",
  "type": "Window covering device",
  "uniqueid": "00:1f:ee:00:00:00:3f:0c-01"
}

ebaauw:
It will once the blind stops moving and `state.bri` is 0.

overdrive69:
ok, but if i switch the j1 on and off in the phoscon webif the state changes....

ebaauw:
In the REST API or in the Phoscon app?

overdrive69:
both

overdrive69:
{
  "etag": "04da833b4a78e81a8141cdfc2c5f08b2",
  "hascolor": false,
  "manufacturername": "ubisys",
  "modelid": "J1 (5502)",
  "name": "Rollladen001",
  "state": {
    "alert": "none",
    "bri": 138,
    "on": false,
    "reachable": true
  },
  "swversion": "20191104-DE-FB0",
  "type": "Window covering device",
  "uniqueid": "00:1f:ee:00:00:00:3f:0c-01"
}

ebaauw:
Odd.  Did the blind actually move when you changed it from Phoscon?

I don’t have a J1 myself, but it should support attribute reporting.  It only exposes a single attribute, _Current Position Lift Percentage_.  The REST API plugin translates this to `state.bri` and to `state.on` (which is false for `bri` 0 and true otherwise).  When you set `state.on` to false, the REST API plugin updates its cache, showing the state as above.  Then it sends the _Open_ command to the J1.  The J1 should be opening the blind, probably sending multiple reports with the actual _Current Position Lift Percentage_, as the blind moves.  On receiving the report, the REST API plugin updates `state.bri` and `state.on`, setting it back to true.  When the blind is fully opened, `state.bri` becomes 0 and `state.on` is, once again set to false.

PR #774, opened 2018-09-06
Add full implementation of Time cluster server.

ma-ca:
Every ZCL read request on Time cluster 0x000A will generate a valid response for every attribute.

The daylight savings attributes DstStart, DstEnd, DstShift are calculated based on `QTimeZone::systemTimeZone()`.

@ebaauw  I have enhanced your existing `time.cpp` because I want to also support Bitron Thermostat 902010/32 which needs daylight saving time.

### DIFF
--- a/time.cpp
+++ b/time.cpp
@@ -5,7 +5,7 @@
  * Send ZCL attribute response to read request on Time Cluster attributes.
  *
  * 0x0000 Time         / UTC Time seconds from 1/1/2000
- * 0x0001 TimeStatus   / Master(bit-0)=1, MasterZoneDst(bit-2)=1
+ * 0x0001 TimeStatus   / Master(bit-0)=1, Superseding(bit-3)= 1, MasterZoneDst(bit-2)=1
  * 0x0002 TimeZone     / offset seconds from UTC
  * 0x0003 DstStart     / daylight savings time start
  * 0x0004 DstEnd       / daylight savings time end
@@ -60,7 +60,7 @@ void DeRestPluginPrivate::sendTimeClusterResponse(const deCONZ::ApsDataIndicatio
                                 deCONZ::ZclFCDisableDefaultResponse);
 
     quint32 time_now = 0xFFFFFFFF;              // id 0x0000 Time
-    qint8 time_status = 0x01;                   // id 0x0001 TimeStatus Master(bit-0)=1
+    qint8 time_status = 0x09;                   // id 0x0001 TimeStatus Master(bit-0)=1, Superseding(bit-3)= 1
     qint32 time_zone = 0xFFFFFFFF;              // id 0x0002 TimeZone
     quint32 time_dst_start = 0xFFFFFFFF;        // id 0x0003 DstStart
     quint32 time_dst_end = 0xFFFFFFFF;          // id 0x0004 DstEnd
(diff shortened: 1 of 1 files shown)


ebaauw:
Cool!  I half planned to at least add support for _Time Status_ (0x0001, which is mandatory) and return an _Unsupported atrribute_ response for the others.  Of course, full support for the others is even nicer!

I think we should set both the _Master_ and _Superseding_ bits in _Time State_, making sure the RaspBee/ConBee will always be chosen as the network time server.

Not sure if we'd need the debug messages.  If you add the _Time_ cluster to the _In Clusters_ in the _Endpoints_ tab of the _deCONZ network settings_ in the GUI, you can simply check the values in the _Cluster info_ panel.  @manup, could this cluster be added to the default configuration?

manup:
> @manup, could this cluster be added to the default configuration?

Cool stuff, sure I'll add it in the next firmware.

ma-ca:
I have added Superseding=1 to attribute TimeStatus. Attribute `0x0001` TimeStatus now has the setting: Master(bit-0)=1, Superseding(bit-3)= 1. Also setting MasterZoneDst(bit-2)=1 if systemTimeZone has daylight savings.

Note: There might be an issue with `QTimeZone::systemTimeZone()` which always returns "Etc/GMT" even though the local timezone is "Europe/Berlin". This is on Raspbian stretch.

https://github.com/dresden-elektronik/deconz-rest-plugin/blob/b2155e80cb1301d394ee2b69d2c7548cbaf3ba7e/time.cpp#L82-L86

```
$ cat /etc/timezone
Europe/Berlin
```

ebaauw:
What timezone did you set in the deCONZ REST API?  I think Phoscon still resets this to GMT.

```
$ ph get /config/timezone
"Etc/GMT-2"
```

ma-ca:
That solved it. I had not configured the timezone in the web app.

Krocko:
I have updated deCONZ to Version 2.05.39 and Firmware 0x26240500 and don‘t have this time cluster. Is there something i must edit?

Krocko:
This worked. Thank you.

ebaauw:
_Last Set Time_ seems to return a random value.  _Valid Until Time_ isn't updated when reading all attributes, but also returns a random value when reading the attribute individually.

ebaauw:
Found the bugs: the type isn't cast to uint8 for these attributes: https://github.com/dresden-elektronik/deconz-rest-plugin/blob/3d2eda2f053d33619ada4c58a577a853ca1b3aaa/time.cpp#L190

Where it is for the other attributes: https://github.com/dresden-elektronik/deconz-rest-plugin/blob/3d2eda2f053d33619ada4c58a577a853ca1b3aaa/time.cpp#L142

> Note: There might be an issue with `QTimeZone::systemTimeZone()` which always returns "Etc/GMT" even though the local timezone is "Europe/Berlin". This is on Raspbian stretch.

> What timezone did you set in the deCONZ REST API?

> That solved it. I had not configured the timezone in the web app.

Do we still need to set the timezone to Berlin when it's set to GMT?

manup:
>Do we still need to set the timezone to Berlin when it's set to GMT?

When the GMT settings are likely be caused from the Phoscon App which detects and writes timezone offset via Javascript, they should be correct.

Issue #635, opened 2018-06-09
Support for Innr Remote RC 110?


jo-me:
Hi
just saw that Innr offers a Zibee Remote, the "RC 110" which apparently can control the new innr plugs either directly or indirectly (using a bridge). 

See here: https://shop.innrlighting.com/en/shop/49/remote-control

Is the RC 110 supported by deconz?

This thing would be a good replacement for the old 443 mhz remotes and even more versatile that the available zigbee remotes - if the buttons/actions are customizable.

Thank you

ebaauw:
> Is the RC 110 supported by deconz?

Not yet, but I'll be happy to give it a go, if some-one can lend me one.  I do have an innr gateway and some innr lights (and the new innr SP 120 smart plugs!), so I would be able to sniff standalone mode as well as bridge mode.

> This thing would be a good replacement for the old 443 mhz remotes and even more versatile that the available zigbee remotes - if the buttons/actions are customizable.

Looking at the [manual](https://www.innrlighting.com/clients/innr/themes/uploads/1484932356_Manual%20for%20Remote%20Control%20-%20RC%20110.pdf), it doesn't seem they are.  See also the [quick start](http://www.innrlighting.com/clients/innr/themes/uploads/1455036012_RC%2011x%20installation%20manual%20-%20IL160209.pdf).

I doubt we'll be able to use the numbered buttons in the "Lights" setting, but probably, they would be usable in the "Scenes" setting.  I expect that, effectively, this will be a 9-button switch (On, DImUp, DimDown, 1, 2, 3, 4, 5, 6) with each button only supporting release (so `buttonevent` values 1002, 2002, 3002, 4002, 5002, 6002, 7002, 8002, and 9002).  You would be able to specify custom actions using deCONZ rules.

l-mb:
Has anyone made progress on this one? It's a relatively cheap ZigBee remote with plenty buttons, so it'd be very desirable. (I'd be mostly interested in using the button events via Home Assistant.) It also seems to support continuous press, so maybe we can get button held and long release too?

(I'm not incapable of programming low-level bits, but I've never worked with ZigBee on the protocol level. If somebody has pointers to docs I'd be grateful. Or if somebody felt like writing a patch + a few notes on how that all works, I'd be delighted to send someone a remote in exchange ;-)

kasteleman:
Not yet. Did not yet include it. But hope to find Some time for it soon. Do
you own the device and tried to include it?

[...]

PR #815, opened 2018-09-30
innr remote RC 110

ebaauw:
See #635.

Expose the innr remote RC 110 as seven (!) ZHASwitch resources: one for the _Scenes_ modus, and one for each of the six _Lights_.  This is still work in progress - to use this, you need to create bindings manually, from the _Level_ cluster of each of the seven controller endpoints to a (different) group ID.  The REST API plugin will set `config.group` and create the corresponding group after receiving the first command from that binding.

I left in comments the code to expose the remote as one ZHASwitch resource, with a list of seven groups in `config.group`.  The REST API plugin doesn't seem to like this, though, and I doubt Phoscon or the old web app can cope with this.


dghh:
Hi,
is there still manual work needed to include the RC 110 or can it be found now like the other ?sensors?
In my case nothing shows up when i push the minus and the program button on the remote

ebaauw:
Whare are you looking?

dghh:
what i did:

- open network on raspbee so that new devices can join
- push minus & program button on remote until it blinks
-  called ../sensors
- called ../groups

Nothing was shown. No new devices or groupsi did not already have.

Although tried afterwards with phoscon app. Detect sensors switches etc. Nothing happend

Would be great if you can bring this remote to work!

ebaauw:
Does the deCONZ GUI show a node for the switch?  You might want to double-check the _Node info_ panel, sometimes the new node is hidden behind another one.

dghh:
Running the raspi headless:-(
Maybe i try to use the conbee to check.
Wanted just to know whether the normal discovery via the web app or phoscon like for the other remotes is already working.
If not how do you do that manual ?binding?
What i have to use?

ebaauw:
The joining of the device to the network and the creation of the REST resources happen automatically.  I don’t think Phoscon supports this switch yet.  If you don’t create the bindings, the switch operates on all lights in the network.  You create the bindings in the deCONZ GUI.

I run my Raspberries headless as well, but start the graphical environment nonetheless, including the deconz-gui service.  I use VNC to get to it from my macbook.

dghh:
Thanks, but how do i create the bindings ? Can you please give me an example for this remote and one group/binding ?
(i can see now the remote in the deCONZ -:) )

ebaauw:
In the GUI, select the _Bind dropbox_ panel.  Drag the _Level Control_ cluster of one of the endpoints of the innr remote to the _Source_ dropbox and select _Group_ for the _Destination_.  Enter an ID for a new group (I use 0x1000 for the scenes and 0x1001 ... 0x1006 for the lights).  Press a button on the remote (to wake it) and then click _Bind_.  The GUI should respond with _success_.
screenshot 2018-10-27 at 12 24
Repeat for endpoints 3...8, corresponding to _1_ ... _6_ in _lights_ mode.  After that, the deCONZ REST API plugin should create the groups automatically, and link each group to the corresponding sensor resource (note `config.group`):
```
$ ph -H pi1 get /sensors/9
{
  "config": {
    "group": "4096",
    "on": true,
    "reachable": false
  },
  "ep": 1,
  "etag": "58d1efe333530c0889e6f03aec9c17f1",
  "manufacturername": "innr",
  "mode": 1,
  "modelid": "RC 110",
  "name": "innr remote",
  "state": {
    "buttonevent": 1002,
    "lastupdated": "2018-10-21T13:54:36"
  },
  "swversion": "0x21000e46",
  "type": "ZHASwitch",
  "uniqueid": "00:15:8d:00:01:0c:44:37-01-0006"
}
```
And vice versa (note `devicemembership`):
```
$ ph -H pi1 get /groups/4096
{
  "action": {
    "bri": 127,
    "colormode": "hs",
    "ct": 0,
    "effect": "none",
    "hue": 0,
    "on": false,
    "sat": 127,
    "scene": null,
    "xy": [
      0,
      0
    ]
  },
  "class": "Other",
  "devicemembership": [
    "9"
  ],
  "etag": "58d1efe333530c0889e6f03aec9c17f1",
  "id": "4096",
  "lights": [],
  "name": "RC 110 (Scenes)",
  "scenes": [],
  "state": {
    "all_on": false,
    "any_on": false
  },
  "type": "LightGroup"
}
```
When pressing _On/Off_, _DimDown_, or _DimUp_ in _scenes_ mode, the remote now sends a group command, rather than a broadcast.  You can assign lights to the group, if you want the remote to control the lights directly, bypassing the gateway (great for contingency, in case the gateway is down).  Or you can leave the group empty, and only use rules on the gateway triggered by `buttonevent` values.

dghh:
Hi, thanks for the detailled description!

Binding now works. It says success. But i still cannot find any group/ sensor....

I have
- resetted the remote
- with program and minus button connected
- it is shown in the deconz gui with all the end points
- opened the network again for some time
- i bind the level control clusters 
- each time it says success
As nothing showed up i also restarted the deconz,  restarted the raspberry, nothing...

I am on the .43 with the newest firmware

Did i forget a step? 
Nothing shows up in the deconz web app, or deconz rest api
No new groups or sensors

ebaauw:
If you don't see any `/sensors` resources for the remote, open the network/search for devices from the web app/phoscon, press a button (to wake the remote) and read the _Basic_ cluster attributes from the GUI.

Did you press a button after the binding?  The REST API plugin needs to see a group command, before it can create the `/groups` resource and link it to the `/sensors` resource.

> Could it be that the reason is that 0008 is not colored ? "blue" like 0000 Basic or 0004 Group ?

No.  Blue are server clusters; grey are client clusters.

PR #817, opened 2018-10-01
Split out network info and refactor/clean config map generation 

KodeCR:
This refactors some of the code to remove code duplication in configToMap and basicConfigToMap by splitting out the network info code. It looks like description.xml wasn't updated with a found ip-adress before. The downside could be the config doesn't update the ip-address if it changes after initialisation, but I'm not sure if that really worked before? Probably best to look at 7c2cf2 for the changes because the line endings messed thing up for me.

### DIFF
--- a/rest_configuration.cpp
+++ b/rest_configuration.cpp
@@ -40,7 +40,7 @@ ApiConfig::ApiConfig() :
 {
 }
 
-/*! Intit the configuration. */
+/*! Init the configuration. */
 void DeRestPluginPrivate::initConfig()
 {
     QString dataPath = deCONZ::getStorageLocation(deCONZ::ApplicationsDataLocation);
@@ -81,6 +81,10 @@ void DeRestPluginPrivate::initConfig()
     gwFirmwareNeedUpdate = false;
     gwFirmwareVersion = "0x00000000"; // query later
     gwFirmwareVersionUpdate = "";
+    gwMAC = "38:60:77:7c:53:18";
+    gwIPAddress = "127.0.0.1";
+    gwPort = (apsCtrl ? apsCtrl->getParameter(deCONZ::ParamHttpPort) : deCONZ::appArgumentNumeric("--http-port", 80));
+    gwNetMask = "255.0.0.0";
     gwBridgeId = "0000000000000000";
     gwConfig["websocketport"] = 443;
     fwUpdateState = FW_Idle;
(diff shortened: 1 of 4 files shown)


KodeCR:
Turns out the Hue app needs the bridgeid to match the MAC address otherwise it won't connect anymore after the first time, so I've added a commit for that. I guess changing the bridgeid might have some more implications, e.g. I know homebridge-hue uses it for identification. I also expect the connection to break (and require a different certificate) if you change from wifi to lan for example, so it may be a good idea to not store the bridgeid in the database anyone but just always use the mac of the active network. What are your thoughts?

manup:
Hi @KodeCR thanks for the PR the refactor is welcome but due the line ending issue this changes too much for one commit, I'm not feeling safe to merge.

Can you please make the PR with just the code changes but not the line endings?

KodeCR:
Hi @manup,

Sure, no problem. I rebased without the line endings change. You might want to change the line ending yourself to make sure it's just the line endings that changed. It's a shame git doesn't handle that more elegantly.

rest_configuration.cpp
```diff
@@ -253,6 +257,94 @@ void DeRestPluginPrivate::initTimezone()
...
+                    (ipv4 & 0xe0000000UL) != 0xc0000000UL)   // class C 110x xxxx
+                {
+                    // unsupported network
+                    continue;
+                }
+
+                QString mac = i->hardwareAddress().toLower();
+                gwMAC = mac;
+                gwBridgeId = mac.mid(0,2) + mac.mid(3,2) + mac.mid(6,2) + "fffe" + mac.mid(9,2) + mac.mid(12,2) + mac.mid(15,2);
+                if (gwConfig.contains("bridgeid") || gwConfig["bridgeid"] != gwBridgeId)
```
manup: ! missing?

if (!gwConfig.contains() ...

KodeCR: Thanks, good spot. I've amended the commit.

rest_configuration.cpp
```diff
@@ -316,6 +316,13 @@ void DeRestPluginPrivate::initNetworkInfo()
 
                 QString mac = i->hardwareAddress().toLower();
```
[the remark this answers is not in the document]
KodeCR: I was afraid this might be the case, hence the comment above, but I thought it might only be homebridge-hue that would be affected. The problem is the Philips Hue App requires/expects the bridgeid to match the lan mac-address for ssl to work. There may be more apps that expect this in the future when the hue bridge goes https-only. I don't really want to add another flag, but how about making it configurable? (see updated commit)

Issue #779, opened 2018-09-10
Trådfri Smart Plug


MrPsi:
Hi,

Do you think the new Trådfri smart plug will work with deconz right away when it is released? Or does support have to be added?
https://www.ikea.com/se/sv/catalog/products/90356166/

Maybe hard to know before it has been released? :)

ebaauw:
Controlling the plug through the `/lights` resource will probably work without any changes.  Supporting the new on/off switch and any metering the plug might provide will require changes.

maxlein:
I just bought the new Outlet and it seems that it doesn't work with the current version. Gateway shows only my motion sensor. Is it configuration that needs to be changed or code?

ebaauw:
Could you post deCONZ GUI screenshots of:
- The node, incl. the endpoints and clusters;
- the _Node info_ panel;
- The _Cluster info_ panel of the _Basic_ cluster, **after** reading the attributes;

maxlein:
I am new to deconz, so I don't know if this is the right information:
Cluster info of the basic cluster I don't understand. I only see basic cluster info on the Osram plug.

![image](https://user-images.githubusercontent.com/11796203/45599292-41f69e00-b9e9-11e8-930d-292ab19c5103.png)
![image](https://user-images.githubusercontent.com/11796203/45599315-86823980-b9e9-11e8-8d67-9d6b51f48006.png)

maxlein:
No, I have a Osram plug and an Ikea. But the ikea one doesn‘t even show up. I think I tried the old and new web gui to search for devices, but nothing but the already connected devices were found.

manup:
Does it have a reset button or something like that?

Note: IKEA devices usually only try **once** to join the network.

* If they are factory new (which should be the case here)
* Directly after they are powered (light search must be started before that)

maxlein:
No, I couldn't find one. The motion sensor has a button but not the plug, at least not on the surface.
Will have a look at it again whn I am home today.

maxlein:
So the default app you have, gives me this screen:
![image](https://user-images.githubusercontent.com/11796203/45642219-e7784300-bab7-11e8-8ea2-8e666824263d.png)
Blink worked, but I can't see the plug anywhere but on this page.
The other app finds nothing:
![image](https://user-images.githubusercontent.com/11796203/45642311-21494980-bab8-11e8-9cb8-2559a47646e4.png)

manup:
[...]

PR #821, opened 2018-10-01
Add support for Trådfri on/off switch

jurriaan:
This PR adds basic support for the Trådfri on/off switch (bundled with the IKEA plug, see #779).

With these changes the button emits 1001/1002/1003, 2001/2002/2003 button events.

@manup My remote has a different mac address then `ikeaMacPrefix` (it starts with 0xd0cf5e), maybe something needs to change there as well?

### DIFF
--- a/bindings.cpp
+++ b/bindings.cpp
@@ -1719,6 +1719,12 @@ void DeRestPluginPrivate::checkSensorBindingsForClientClusters(Sensor *sensor)
         clusters.push_back(SCENE_CLUSTER_ID);
         srcEndpoints.push_back(sensor->fingerPrint().endpoint);
     }
+    // IKEA Trådfri on/off switch
+    else if (sensor->modelId().startsWith(QLatin1String("TRADFRI on/off switch")))
+    {
+        clusters.push_back(ONOFF_CLUSTER_ID);
+        srcEndpoints.push_back(sensor->fingerPrint().endpoint);
+    }
     else if (sensor->modelId().startsWith(QLatin1String("D1")))
     {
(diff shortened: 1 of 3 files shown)


jurriaan:
@manup Interestingly I still sometimes get the message 
```
no button handler for: TRADFRI on/off switch ep: 0x01 cl: 0x0008 cmd: 0x07 pl[0]: 000
```
 which is weird since that one should be covered by my button map, right?

manup:
> Interestingly I still sometimes get the message

>no button handler for: TRADFRI on/off switch ep: 0x01 cl: 0x0008 cmd: 0x07 pl[0]: 000
> which is weird since that one should be covered by my button map, right?

Indeed the map looks all right.
I guess sometimes the ok condition is not hit, maybe when a command was missed.

image

tubalainen:
I am soooooo looking forward to you guys adding support for the IKEA On/Off remote control switch that is bundled with the smart plug! <3

Many thanks in advance for your hard and dedicated work!

jeroenpeeters:
I see this PR has made it to the master branch, but has it been released? I'm now at deconz (phoscon) 2.05.49 but I cannot get the on/off button to work. Am I missing something?

jurriaan:
This PR is in 2.05.49. However Phoscon (the UI) has no support for it. But you can just add it and react to the emitted events. I use it in combination with home-assistant and that works just fine for me.

jeroenpeeters:
Any pointers how to add it? I'm just using the Phoscon UI from a docker container.

Issue #798, opened 2018-09-23
lumi.ctrl_neutral2 switches still lost in v2.05.39


ebaauw:
My `lumi.ctrl_neutral2` end-device switches are still somewhat lost after restarting deCONZ v2.05.39:
```
$ ph get /lights/421
{
  "etag": "2b585d2c016bfd665ba27a8fdad28670",
  "hascolor": false,
  "manufacturername": "LUMI",
  "modelid": "lumi.ctrl_neutral2",
  "name": "Living Room Fan 1",
  "state": {
    "alert": "none",
    "on": false,
    "reachable": true
  },
  "swversion": "11-11-2016",
  "type": "Smart plug",
  "uniqueid": "00:15:8d:00:02:39:14:5e-02"
}
$ ph put /lights/421/state '{"on": true}'
ph put: error: 3 resource, /lights/421, not available
```

The switches are active on the network, and deCONZ is receiving and answering the requests for the _Time_ cluster.  However, that doesn't trigger the REST API plugin to make the resource available.  Reading the _Basic_ cluster attributes on endpoint 0x01 doesn't make a difference either.  Reading the _OnOff_ cluster attributes on endpoint 0x02 makes available the corresponding light resource.  The light resource corresponding to endpoint 0x03 is still unavailable, though, until I read the _OnOff_ cluster attributes on that endpoint as well.

Note that these switches are mains powered end-devices, but report _Receiver on when idle_ as false in the _Node info_ panel.  I don't think that's correct, though: reading cluster attributes always seems to  give an answer immediately.  Also note that these are the newer models without a neutral wire connection, despite the name.

manup:
>Note that these switches are mains powered end-devices, but report Receiver on when idle as false in the Node info panel. I don't think that's correct, though: reading cluster attributes always seems to give an answer immediately. Also note that these are the newer models without a neutral wire connection, despite the name.

Can you please make a screenshot of the Node Info Panel, maybe the Node Descriptor wasn't fetched yet.Hopefully the correct `rxOnWhenIdle` flag is available otherwise this will for sure cause problems in various cases.

The `rxOnWhenIdle` flag can be extracted from *Device Announce* and *Node Descriptor*, the former can be improved by putting the `macCapabilities` into Node Descriptor even if it is not fetched yet, I'm not sure this is currently done in every case.

ebaauw:
> Can you please make a screenshot of the Node Info Panel

[...]

PR #836, opened 2018-10-06
lumi.curtain and lumi.ctrl_neutral

ebaauw:
- Invert _CurrentPositionLiftPercentage_ so that 0% is _Open_ and 100% is _Closed_, in line with ZCL spec, see #746;
- Ignore _OnOff_ cluster;
- Suppress creating non-operational ZHASwitch and ZHAPresence sensors.
- Handle Xiaomi special attribute for `lumi.curtain` - update `state.bri` and `state.on` of the `/lights` resource;
- Handle Xiaomi special attribute for `lumi.ctrl_neutral` and `lumi.ctrl_ln` - update `state.on` of the `/lights` resource.
- Set node's `rx()` on receiving Xiaomi special attribute - this solves the `lumi.ctrl_neutral` being unavailable after restart, see #798.


ebaauw:
@manup, is this PR still in time for 2.05.40?  Thanks.

manup:
Thanks for the PR :)

Sure I think 2.05.40 has still some days to wait, hopefully some of the routing issues can be solved for it.

Kane610:
So max and min will be reversed?
And you will normalise to 0-100% instead of 0-255?
Will this also affect keen vents?

Issue #934, opened 2018-11-08
How do I enable Echo mode in this commit? For on/off devices


saurabh984:
https://github.com/dresden-elektronik/deconz-rest-plugin/commit/e05358f5b964d074d7558a70177ec67b952e5f6a

How do I enable this for devices like Ubysis S2?

manup:
So the Ubisys S2 doesn't show up in Echo lights?

saurabh984:
Yes @manup

PR #938, opened 2018-11-10
Amazon Echo control On/Off devices

saurabh984:
like Ubisys S1/S2


saurabh984:
Fixes https://github.com/dresden-elektronik/deconz-rest-plugin/issues/934

PR #1003, opened 2018-12-04
Add Support for Thermostat Cluster 0x0201 and Bitron Thermostat 902010/32

ma-ca:
Add Support for Thermostat Cluster `0x0201` and Bitron Thermostat 902010/32

A sensor type `ZHAThermostat` is created with the following options in state and config:

EDIT: moved writable settings from state to config. 
 
  option             | read/write | attribute | description
  -------------------|------------|-----------|---------------------
  state.on           | read only  | 0x0029    | running state on/off
  state.temperature  | read only  | 0x0000    | measured temperature
  config.heatsetpoint | read write | 0x0012    | heating setpoint
  config.scheduleron  | read write | 0x0025    | scheduler on/off
  config.offset      | read write | 0x0010    | temperature offset
  config.scheduler   | read write | (command) | scheduled setpoints
 

  Example sensor:
 ```
  /api/<apikey>/sensors/<id>/
     {
        config: {
          "heatsetpoint": 2200,
           "offset": 0,
            "scheduler": "Monday,Tuesday,Wednesday,Thursday,Friday 05:00 2200 19:00 1800;Saturday,Sunday 06:00 2100 19:00 1800;",
          "scheduleron": true
        },
        state: {
           "on": true,
           "temperature": 2190
         },
        "ep": 1,
        "manufacturername": "Bitron Home",
        "modelid": "902010/32",
         "type": "ZHAThermostat",
        ...
     }
 ```
  Rest API example commands:
```
  -X PUT /api/<apikey>/sensors/<id>/config -d '{ "heatsetpoint": 1800 }'
  -X PUT /api/<apikey>/sensors/<id>/config -d '{ "scheduleron": true }'
  -X PUT /api/<apikey>/sensors/<id>/config -d '{ "offset": 0 }'
  -X PUT /api/<apikey>/sensors/<id>/config -d '{ "scheduler": "Monday 05:00 2200 19:00 1800;" }'
                                           -d '{ "scheduler": "" }'  (send get scheduler command)
 ```
 
  Attributes (only a subset):
```
    ID      Type  Type    Description                     Default
    0x0000  0x29  int16s  Local Temperature               --
    0x0010  0x28  int8s   Local Temperature Calibration   --
    0x0012  0x29  int16s  Occupied Heating Setpoint       2000 (20 °C)
    0x0025  0x18  bit8    Programming Operation Mode      0
                          Bit#0 of this attribute controls
                          the enabling of the Thermostat Scheduler.
    0x0029  0x19  bit16   Thermostat Running State        0
                          Bit0=Heat State On
                          Bit1=Cool State On
```

Commands Received (Client to Server):
```
    Command-ID  Name
    0x00        Setpoint Raise/Lower
    0x01        Set Weekly Schedule
    0x02        Get Weekly Schedule
    0x03        Clear Weekly Schedule
 ```

Commands Generated (Server to Client):
```
    Command-ID  Name
    0x00        Current Weekly Schedule
```

   Weekly Schedule format

```
   Octets 1           1       1     2           2/0         2/0       ...   2   2/0   2/0
   Type   enum8       bit8    bit8  int16u      int16s      int16s
   Name   Number      Day of  Mode  Transition  Heat        Cool
          Transitions Week    (1,2) Time 1      Setpoint 1  Setpoint1

   Day of Week: bitmap8 [Sunday, Monday, Tuesday, Wednesday, Thursday, Friday, Saturday, Away]
                bit         0       1       2         3        4         5        6       7
```

Example:
```
   Monday, Tuesday, Thursday, Friday = 0011 0110 = 0x36
        |        |         |     ^-------^|  ||
        |        |          ---------------  ||
        |        -----------------------------|
        ---------------------------------------
```

@ebaauw would you be able to add thermostat to homebridge-hue?


ebaauw:
Not sure.  I think the thermostat should be exposed as a /lights resource.  The state attributes of a /sensors resource are not (supposed to be) updatable.  See 
https://github.com/dresden-elektronik/deconz-rest-plugin/wiki/Supported-Devices#resources

ma-ca:
I would like to have a thermostat in HomeKit that displays the current temperature and desired temperature. Also, it should be possible to modify the desired temperature and to enable/disable the scheduler. 

I did not consider a thermostat belonging to `/lights`. A thermostat state is on/off when the desired temperature is higher/lower than current temperature. To me this seems different behavior than dimmable lights.

Anyway, I am happy to change the thermostat from `/sensors` to `/lights` if this is needed for HomeKit. Just let me know.

manup:
Not happy so many non light devices become lights lately, guess it's really time to introduce the `/devices` endpoint or something similar. 

It will be hard in the future to clean it up once installations are out there which rely on a stable API.

ma-ca:
So the thermostat remains in `/sensors`. 

I have moved the writable settings `state.heatsetpoint` and `state.scheduleron` to `config.heatsetpoint` and `config.scheduleron`.

One question: The thermostat cluster only has reporting on the temperature attribute. How can I poll other attributes? The _poll_manager.cpp_ only allows router nodes. This thermostat is a battery powered end device.

manup:
Polling end-devices is tricky, first you need to find out when they poll (mac data request) the parent node for new data.

Knowing when the end-device is awake and polls the parent is crucial. Just sending a request to an end-device will likely fail since requests are stored only for max. ~7 seconds on the parent.

Do you have a sniffer available?

Kane610:
@manup I'd welcome a better class than lights for other devices. Move all currently supported devices over to this new endpoint. Just communicate it in good time to allow us to add support

ma-ca:
This Bitron Thermostat does respond very well to commands. Reading and writing attributes takes several seconds but always succeeds. I just need to find the right place in the code to read the attributes every 5 minutes. Even missing some read requests would not do any harm. I don't have a sniffer.

ebaauw:
> if this is needed for HomeKit

HomeKit doesn’t care about the resource - it doesn’t even know.  Homebridge-hue does.  For `/lights` resources, it took me a lot of effort to collect multiple characteristic changes for the same “light” into one API request to deCONZ or the Hue bridge, in order not to flood the gateway.  When recalling a HomeKit scene, there can be up to 50 changes at once.  There’s no similar logic for `/sensors` resources.

Moving the writeable `state` attributes to `config` and sticking to `/sensors` actually makes more sense than using `/lights`, as the thermostat is battery powered.  Afaik there’s no `/lights` resources for end devices that might turn their receiver off while sleeping.

ebaauw:
> This Bitron Thermostat does respond very well to commands. Reading and writing attributes takes several seconds but always succeeds.

That sounds similar to the Hue motion sensor, that seems to poll its parent every five seconds or so.  What does the thermostat report for _Receiver on when idle_ in the _Node Info_ panel?  And in the _Power Descriptor_ (after you’ve read it using the leftmost of the two circles on the right of the node)?  When it does turn the receiver off, it might be prudent to use `config.pending` as for the Hue motion sensor.

ebaauw:
> Not happy so many non light devices become lights lately, guess it's really time to introduce the /devices endpoint or something similar.

> I'd welcome a better class than lights for other devices. Move all currently supported devices over to this new endpoint. Just communicate it in good time to allow us to add support

As would I, preferably one that follows the ZigBee structure of node, endpoint, cluster, rather than having to re-assemble multiple resources for the same device using the mac address.

This would however break compatibility with Hue apps, particularly with the likes of Alexa.  I would probably remove deCONZ support from homebridge-hue and create a new homebridge-deconz.

ma-ca:
Bitron Thermostat Node Info

image

---
image

ma-ca:
> did you also add support for CLIPThermostat?

@ebaauw No, I just added a ZHAThermostat sensor type.

ebaauw:
I'm also missing a `state.lastupdated`.

ebaauw:
Ah, cool.  As I said: I don't yet have a thermostat, so I cannot see for myself :-(

bohtho:
Would the work both the Deconz, Deconz REST API, adn Homebridge-Hue developers have put in to support the Bitron and Eurotronic zigbee thermostatic radiator valves (eg. supporting clusters, ZHAThermostat and CLIPThermostat) make any other thermostatic valve based on Zigbee HA 1.2 also work with Deconz now ?

easybeat:
Sorry, does deconz rest-api support the Eurotronic Spirit Zigbee by now...I tried to find out by reading this post but it is confusing me.

Thank you for the information.

Cheers
Beat

PR #1005, opened 2018-12-04
Api handling

KodeCR:
This PR adds the scenes, info, and capabilities nodes to the api. Furthermore it:
- adds the timezones of the system to the info and capabilities nodes, the rest of the capabilities node and the scenes node is empty for now
- centralise the apikey check so it isn't missed anywhere, e.g. schedules
- update the error messages and handle everything under /api

The aim is still to slowly work towards better support for the hue-api/Hue app (#663)

### DIFF
--- a/authentification.cpp
+++ b/authentification.cpp
@@ -209,7 +209,7 @@ bool DeRestPluginPrivate::checkApikeyAuthentification(const ApiRequest &req, Api
 #endif
 
     rsp.httpStatus = HttpStatusForbidden;
-    rsp.list.append(errorToMap(ERR_UNAUTHORIZED_USER, req.path.join("/"), "unauthorized user"));
+    rsp.list.append(errorToMap(ERR_UNAUTHORIZED_USER, "/" + req.path.mid(2).join("/"), "unauthorized user"));
 
     if (req.sock)
     {
--- a/de_web.pro
+++ b/de_web.pro
@@ -140,6 +140,9 @@ SOURCES  = authentification.cpp \
            rest_sensors.cpp \
            rest_schedules.cpp \
            rest_touchlink.cpp \
+           rest_scenes.cpp \
+           rest_info.cpp \
+           rest_capabilities.cpp \
            rule.cpp \
            upnp.cpp \
            permitJoin.cpp \
--- a/de_web_plugin.cpp
+++ b/de_web_plugin.cpp
@@ -12949,12 +12949,15 @@ bool DeRestPlugin::isHttpTarget(const QHttpRequestHeader &hdr)
         {
             if ((ls[2] == QLatin1String("lights")) ||
                 (ls[2] == QLatin1String("groups")) ||
-                (ls[2] == QLatin1String("config")) ||
                 (ls[2] == QLatin1String("schedules")) ||
+                (ls[2] == QLatin1String("scenes")) ||
                 (ls[2] == QLatin1String("sensors")) ||
-                (ls[2] == QLatin1String("touchlink")) ||
-                (ls[2] == QLatin1String("resourcelinks")) ||
                 (ls[2] == QLatin1String("rules")) ||
+                (ls[2] == QLatin1String("config")) ||
+                (ls[2] == QLatin1String("info")) ||
+                (ls[2] == QLatin1String("resourcelinks")) ||
+                (ls[2] == QLatin1String("capabilities")) ||
+                (ls[2] == QLatin1String("touchlink")) ||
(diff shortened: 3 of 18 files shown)


ebaauw:
If you want the /scenes api to work like the Hue bridge, you need to implement a default group 0, add all lights to it, and create all scenes under that group.  You probably need change the group id numbering, to no longer use the ZigBee group id, but a sequence number instead.  Group 0 is just a regular group on the Hue bridge.
I think with the latest api version version, they’ve added support for scenes to other groups, but I haven’t looked into that.

Philips deprecated the /info ages ago (as of api version 1.15.0).  The timezones are reported under /capabilities.

KodeCR:
Thanks @ebaauw. The scenes api looks like quite a bit of work and I don't have too much time, but I thought it would be good to get started and (slowly) progress.
I've implemented the timeszones under capabilities as well, I know info is deprecated but my hue bridge still supports it for the latest firmware so I though it would be good to have it there at least until Philips removes it.

manup:
When I recall correctly all ZLL lights are in group 0 by default it's called the **global group** 0x0000, same goes for the **global scene** 0x00.

ebaauw:
> When I recall correctly all ZLL lights are in group 0 by default it's called the global group 0x0000, same goes for the global scene 0x00.

No, it's a regular ZigBee group which is added to each light upon pairing.  Unlike deCONZ, the group IDs in the Hue API do _not_ correspond to ZigBee group IDs.  All the scenes are stored under this group.  Upon a scene recall through `/groups/xx/actions`, the Hue bridge sends a recall for the `/groups/0` group scene, using a `/groups/xx` groupcast.  That way they can selectively recall the scene for a subset of lights.

manup:
Understand, but I'm afraid this would be no solution for deCONZ we need to be able to save a scene via groupcast and only for the specific group, one reason is many lights don't support the unicast add scene command very well. Still it's ok to just support this for the current api and see the Hue scenes api as addition.

Guess it's possible to support the Hue scenes endpoint for already existing scenes. I need to study the Hue scenes api more but a quick idea: List existing scenes like: `100a01` referring to the group `100a` scene `01`.

At least it would support 'recall, delete, get'. Modification and creation might need more handling.

ebaauw:
> Guess it's possible to support the Hue scenes endpoint for already existing scenes.

That might work with some additional trickery.  Some (most?) Hue apps only show the scenes they created themselves, so you want to report `owner` as the api key from the request.

KodeCR:
Thanks for all the useful info.

>I think with the latest api version version, they’ve added support for scenes to other groups
Yes, in api v1.28 they added the distinction between a "LightScene" and a "GroupScene", and it sounds like a deconz scene can be mapped to a GroupScene straightforwardly.

>you need to implement /groups/0, add all lights to it, and create all scenes under that group
That sounds doable, every scene created under /scenes (as a LightScene) goes to group 0 and is reported as a LightScene under /scenes (i.e. the hue way), and everything created under /groups/<id>/scenes (i.e. the deconz way, or as a GroupScene under /scenes) goes to the corresponding group and is reported under /scenes as a GroupScene.

>You probably need change the group id numbering
Why is this needed?

>When I recall correctly all ZLL lights are in group 0 by default it's called the global group 0x0000, same goes for the global scene 0x00.
Is group 0 treated special on the ZigBee side in the specification or is this just by convention? Any reason we couldn't just add all lights and corresponding scenes to group 0 on each node?

Note that I didn't particularly wanted this PR to have all of these changes. I'd rather have this one merged and (slowly) build on it with subsequent PRs.

ebaauw:
deCONZ uses the ZigBee group number for the group ID in the REST API.  So `/groups/1` refers to ZigBee group 0x0001, and `/groups/0` to ZigBee group 0x0000.  I think, technically this is not a group, but a broadcast - you don't need to tell the lights to join 0x0000.  Consequently, you cannot tell the lights to store a scene for 0x0000.

The Hue bridge uses sequential groups IDs in the REST API mapped to random ZigBee group numbers.  So `/groups/1` might refer to any ZigBee group ID - you'd need to sniff the Hue bridge traffic to find out which.  Or, if you manage to get your RaspBee to join the Hue bridge network as a router, you can query the lights from deCONZ and see which groups the Hue bridge has made them join.

`/groups/0` on the Hue bridge is special from the API perspective, but it is mapped to a regular ZigBee group, not to 0x0000.  You need to tell each light explicitly to join this group.  And because it's a regular group, you can tell the light store scenes for it.

PR #1046, opened 2018-12-17
Extend strict API mode

### LINKED ISSUE #663: Hue app with deCONZ

KodeCR:
Extend strict api mode to: normal, strict, amazon echo, philips hue app
Whitelist supported sensors for the Hue app (some such as ZHAAlarm crash the app), this fixes #663, nr 1
Only rooms can have classes (LightGroups with a class breaks Hue app)
Some consistency changes and warning fixes
Update Hue version and add software date
Mimic hue dimmer switches for some compatible other switches
Terminology

### DIFF
--- a/authorisation.cpp
+++ b/authorisation.cpp
@@ -125,25 +125,29 @@ bool DeRestPluginPrivate::allowedToCreateApikey(const ApiRequest &req, ApiRespon
     return false;
 }
 
-/*! Checks if the request is authorised to access the API.
-    \retval true if authorised
-    \retval false if not authorised and the rsp http status is set to 403 Forbidden and JSON error is appended
+/*! Authorise API access for the request.
  */
-bool DeRestPluginPrivate::checkAuthorisation(ApiRequest &req, ApiResponse &rsp)
+void DeRestPluginPrivate::authorise(ApiRequest &req, ApiResponse &rsp)
 {
     Q_UNUSED(rsp);
+
+    QHostAddress localHost(QHostAddress::LocalHost);
+    if (req.sock->peerAddress() == localHost)
+    {
+        req.auth = ApiAuthLocal;
+    }
+
+    if (req.sock == 0) // allow internal requests, as they are issued by triggering rules
+    {
+        req.auth = ApiAuthInternal;
+    }
 
     QString apikey = req.apikey();
     apiAuthCurrent = apiAuths.size();
 
     if (apikey.isEmpty())
     {
-        return false;
-    }
-
-    if (req.sock == 0) // allow internal requests, as they are issued by triggering rules
-    {
-        return true;
+        return;
     }
 
     std::vector<ApiAuth>::iterator i = apiAuths.begin();
@@ -199,7 +203,7 @@ bool DeRestPluginPrivate::checkAuthorisation(ApiRequest &req, ApiResponse &rsp)
                 apiAuthSaveDatabaseTime.start();
                 queSaveDb(DB_AUTH, DB_HUGE_SAVE_DELAY);
             }
-            return true;
+            req.auth = ApiAuthFull;
         }
     }
 
@@ -219,7 +223,6 @@ bool DeRestPluginPrivate::checkAuthorisation(ApiRequest &req, ApiResponse &rsp)
     }
 #endif
 
-    return false;
 }
 
 /*! Encrypts a string with using crypt() MD5 + salt. (unix only)
--- a/authorisation.cpp
+++ b/authorisation.cpp
@@ -35,9 +35,9 @@ void ApiAuth::setDeviceType(const QString &devtype)
     devicetype = devtype;
 }
 
-/*! Init authentification and security.
+/*! Init authentication.
  */
-void DeRestPluginPrivate::initAuthentification()
+void DeRestPluginPrivate::initAuthentication()
 {
     bool ok = false;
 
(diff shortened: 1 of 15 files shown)


rest_groups.cpp
```diff
@@ -1773,6 +1772,10 @@ bool DeRestPluginPrivate::groupToMap(const ApiRequest &req, const Group *group,
         else if (item->descriptor().suffix == RAttrClass) { map["class"] = item->toString(); }
         else if (item->descriptor().suffix == RAttrUniqueId) { map["uniqueid"] = item->toString(); }
     }
+    if (map["type"] == "LightGroup")
```
manup: When I read the Hue docs correctly when `class` is only meant for rooms, since there are more types than Room and LightGroup it might be easier to make the condition:

`if (map["type"] != QLatin1String("Room"))`

Please also prefer using QLatin1String, which prevents creating a QString object for static C strings.

http://doc.qt.io/qt-5/qlatin1string.html#details

http://doc.qt.io/qt-5/qstring.html#operator-eq-eq-1

KodeCR: Indeed, I'll change that. Thanks for the pointer to QLatin1String, I'm not too familiar with QT.

manup:
Are these changes testet yet with the related apps and Echo? I'm giving this a go at the weekend but we really need to be careful to not break the API.

I'm afraid we need to refactor some of these handlers. Not so happy that so many config related methods which were behind `handleConfigurationApi()` are now brought directly into handleHttpRequest().

I guess it should be safe to let handleConfigurationApi() return REQ_NOT_HANDLED and keep the related methods inside handleConfigurationApi()?

manup:
Cool, thanks. I'll see to check the iConnectHue App and Hue Essentials (which supports deCONZ specific features like scenes).

KodeCR:
Ok, I added an authorisation status for requests and moved the handlers back to rest_configuration. Hope that is what you had in mind. Note that I also put " PUT /api/<localuser>/config/wifi/updated" and "// PUT /api/<localuser>/config/wifi/scanresult" under localhost authorisation instead of being accessible to everyone as I thought that would be the intent, easy to change back if not.

As mentioned above, I tested this again with the Hue app (using https://github.com/KodeCR/hue-pass), and with an Echo using local access.

Thomas-Vos:
It looks like this broke the Hue Essentials app. The model ID is now reported as `BSB002` instead of `deCONZ`. This was how Hue Essentials detected it as a deCONZ bridge, and enables deCONZ specific API features like group scenes.

Hue Essentials authenticates with a devicetype like this: `Hue Essentials#DEVICEMODEL`
Could you add this as an exception and not set the api mode to ApiModeHue for Hue Essentials?

authorisation.cpp
```diff
@@ -182,13 +182,28 @@ bool DeRestPluginPrivate::checkApikeyAuthentification(const ApiRequest &req, Api
...
+            if ((!(i->useragent.isEmpty()) && i->useragent.startsWith(QLatin1String("iConnect"))) || i->devicetype.startsWith(QLatin1String("iConnectHue")))
+            {
+                req.mode = ApiModeStrict;
+            }
+            else if (i->devicetype.startsWith(QLatin1String("Echo")))
+            {
+                req.mode = ApiModeEcho;
+            }
+            else if (i->devicetype.startsWith(QLatin1String("hue_")) ||
+                     i->devicetype.startsWith(QLatin1String("Hue ")))
```
Thomas-Vos: See comment: https://github.com/dresden-elektronik/deconz-rest-plugin/pull/1046#issuecomment-451887310

manup: @Thomas-Vos Should work again with 2.05.55.

Another thing I've noticed, the Phoscon App creates resourcelinks which contain `content like this:
```
{"classid":2020,"description":null,"links":["/userparameter/2","/groups/3","/sensors/29","/sensors/49","/sensors/50","/rules/22","/rules/15","/rules/18","/rules/5","/rules/24","/rules/23","/rules/19","/rules/16","/rules/8","/rules/42"],"name":"Fluri3","owner":"4E8BDC7B8E","recycle":false,"type":"Link"}
```
Here are two problems:
1.  `description` should never be `null` but an empty string: this is fixed in 2.05.55
2. The /userparameter/2  in links[] isn't allowed by Hue Essentials and an error is thrown. This can only be addressed in the Hue Essentials app.

ebaauw:
Another non-documented API endpoint :-(.  Maybe only expose this (and include it in the resourcelink) when the `vnd.ddel.v1` header is set?

```
$ ph get /userparameter
{
  "groupssequenceleft": "[\"220\",\"230\",\"240\",\"310\",\"320\",\"270\",\"330\",\"110\",\"200\",\"210\",\"250\",\"260\",\"280\",\"300\",\"340\"]",
  "groupssequenceright": "[\"4604\",\"222\",\"3\",\"11575\",\"31396\",\"5671\",\"6252\",\"5\",\"26084\",\"9811\",\"35438\",\"19250\",\"61290\",\"4\"]"
}
$ 
```

authorisation.cpp
Thomas-Vos: @manup Thanks for your change. I have addressed your points in the upcoming release 1.1.0 which will be available soon.

Issue #1124, opened 2019-01-12
device support: iCasa switches


ebaauw:
[wifimedia.eu](https://www.wifimedia.eu/nl/brands/icasa/) kindly provided me with iCasa switches to integrate in deCONZ and homebridge-hue.

The battery-powered Pulse 2, Pulse 4S, and Pulse 8S keypads seem like regular switches.  They pair to deCONZ on first attempt.  They all look the same from a Zigbee perspective, except for the _Model Identifier_:
- Pulse 2: `ICZB-KPD12`;
- Pulse 4S: `ICZB-KPD14S`;
- Pulse 8S: `ICZB-KPD18S`.

![screenshot 2019-01-12 at 20 22](https://user-images.githubusercontent.com/22179355/51077507-e04cf480-16a7-11e9-8bc7-f3ba44151df3.png)

deCONZ receives APS indications for the_OnOff_ cluster (0x0006) on pressing the _On_ or _Off_ button; for the _Scenes_ cluster (0x0005) on pressing or holding (store scene?) a scene button; and for the _Level control_ cluster (0x0008) on holding and on releasing the _On_ of _Off_ button(except for the 4S - the light blinks instead of burns while holding the button **Edit** because of the low battery).  I haven't yet brought out the sniffer, but this seems pretty standard.

I can bind the _Power configuration_ cluster to the RaspBee, but get unsupported cluster commands when trying to read or write the report configuration for _Battery Voltage_ or _Battery Percentage Remaining_.  We might need to poll the switches for the battery percentage.  They seem to poll their parent after a button is pressed.

The battery percentages are 100 (3.2V), 35 (2.5V) and 98 (3.0V) - I think they report in 1%, like the IKEA Trådfri, instead of the standard 0.5%.

They appear to be OTA upgradable; the OTAU plugin shows them with _Version_ 0x0000000A and _Image_ 0x2FAA.

The _Power Descriptor_ does not seem valid.  I see ZDP answers to 0x8003, but the _Node Info panel_ continues to show them as _Mains powered_.  The _Node Descriptor_ seems fine, though.
![screenshot 2019-01-12 at 20 47](https://user-images.githubusercontent.com/22179355/51077792-53a43580-16ab-11e9-870a-9832c8934b72.png)

manup:
Cool, the dimmer looks very interesting have it already on my wish list since a few days :)

[...]

PR #919, opened 2018-11-02
Add ubisys Device Setup Cluster

ma-ca:
Configures ubisys manufacturer specific Device_Setup Cluster `0xFC00` on Endpoint 232 (0xE8).
 
  Implements support for:
  - ubisys S1 power switch
  - ubisys S2 power switch
  - ubisys J1 shutter control

It can be used to control the behavior of inputs (i.e. permanent switches, push-button switches, normally open vs. normally closed, on/off, level control or scene selection, etc.)

Example usage:

    -X PUT -d '{ "mode": "momentary" }' /sensor/xx/config
    -X PUT -d '{ "mode": "rocker" }' /sensor/xx/config
 
In addition to predefined configurations for momentary switches and rocker switches, this implementation allows custom configurations for a more complex move or recall scene command.

Example usage:

    -X PUT -d '{ "mode": "custom_41020006000D0306000206010D04060002" }' /sensor/xx/config
    
Writes ZCL raw data `"41020006000D0306000206010D04060002"` to attribute `0x0001` on cluster `0xFC00`

PR #1127, opened 2019-01-13
Support more switches

ebaauw:
- ubisys S1, S1-R (untested), see #919;
- iCasa keypads, see #1124;
- iCasa dimmer, see #1124.

@manup there's still two loose ends:
- I cannot get deCONZ to create the binding and attribute reporting for the iCasa dimmer.  Don't know what's going on; I reset and re-paired my Trådfri colour bulb, but no bindings for that either.  deCONZ happily polls the lights.  Even after creating the binding manually, the report configuration isn't set.  After setting that manually as well, deCONZ stops polling the lights.  This is my test network (2.05.54, self compiled plugin from latest commits), with only two lights (the iCasa dimmer and the Trådfri).
- The iCasa keypads don't support attribute reporting for the battery, so deCONZ needs to poll them.  The good news: they seem to poll their parent at any button click.  The bad news: I'm lost were in the code to implement this - please help.

### DIFF
--- a/bindings.cpp
+++ b/bindings.cpp
@@ -1264,6 +1264,9 @@ void DeRestPluginPrivate::checkLightBindingsForAttributeReporting(LightNode *lig
         else if (lightNode->modelId() == QLatin1String("NL08-0800")) // Nanoleaf Ivy
         {
         }
+        else if (lightNode->modelId() == QLatin1String("ICZB-IW11D")) // iCasa Dimmer
+        {
+        }
         else
         {
             return;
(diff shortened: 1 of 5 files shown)


ebaauw:
Or maybe I'm too impatient.  I now (>30 mins after starting deCONZ) see the first log message that the REST API plugin is checking the bindings.

ebaauw:
Thanks, thats a relief.

What about the polling for the battery?

manup:
> What about the polling for the battery?

>The good news: they seem to poll their parent at any button click.

See the other comment, this needs to be schedules on a per device basis when we can be sure the device does a mac data poll *after* a command is received. For example if these switches do mac data polling after a button was pressed this is a good time to read battery state. Maybe with a little book keeping to do this only when the last value has a certain age.

PR #1130, opened 2019-01-13
Fix read and write attributes on Thermostat and Window Covering Clusters

ma-ca:
Fix writing 8-bit bitmap attributes to Thermostat Cluster.

Fix reading multiple attributes from Window Covering Cluster.

### DIFF
--- a/thermostat.cpp
+++ b/thermostat.cpp
@@ -573,11 +573,11 @@ bool DeRestPluginPrivate::addTaskThermostatReadWriteAttribute(TaskItem &task, ui
     if (readOrWriteCmd == deCONZ::ZclWriteAttributesId)
     {
         stream << (quint8) attrType;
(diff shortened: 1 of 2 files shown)


thermostat.cpp
```diff
@@ -577,7 +577,7 @@ bool DeRestPluginPrivate::addTaskThermostatReadWriteAttribute(TaskItem &task, ui
         {
             stream << (quint8) attrValue;
         }
-        else if (attrType == deCONZ::Zcl16BitInt || attrType == deCONZ::Zcl16BitBitMap)
+        else if (attrType == deCONZ::Zcl16BitInt || attrType == deCONZ::Zcl16BitBitMap || attrType == deCONZ::Zcl8BitBitMap)
```
ebaauw: Are you sure you want to write an 8-bit bitmap as 16-bit value?!

ma-ca: That was a mistake. Fixed.

Issue #1154, opened 2019-01-20
Offset problems 


cosmo84:
Referring to #763 I have set an offset of -2 degrees for one of my Xiaomi temperature sensors. The offset itself is working, but it is not respected every few updates, see the following logs:

[1/20/2019, 6:32:26 AM] [Phillips Hue] Draußen: set homekit temperature from -0.4°C to -2.3°C
[1/20/2019, 6:42:32 AM] [Phillips Hue] Draußen: set homekit temperature from -2.3°C to -0.4°C
[1/20/2019, 7:13:28 AM] [Phillips Hue] Draußen: set homekit temperature from -0.4°C to -2.5°C
[1/20/2019, 7:38:32 AM] [Phillips Hue] Draußen: set homekit temperature from -2.5°C to -0.5°C

So it seems only every second measuring is with the offset

ebaauw:
Looking at the code, `config.offset` is ignored when updating `state.temperature` (or `state.humididy`) from the Xiaomi special report.  Note that it should not be applied when updating `config.temperature`, but likely `config.offset` wouldn’t be present anyways for those resources.

cosmo84:
Sorry, I don‘t get the point yet. Am I using the offset in a wrong way, or is it not intended to work like this at all?

ebaauw:
No, sorry, there's an omission in de REST API plugin.  Preparing a fix...

cosmo84:
All right, thanks a lot.

manup:
Please checkout if it works now with 2.05.57, closing the issue for now.

kugelkopf123:
Hey. 

Sorry that I have to reopen this Issue. Its just a Question. Where and how can I set this offset. Because I have such a Aqara TempHumidityPressure Sensor, too. It would be nice to set an offset on the temperature. But i can not find it in the Deconz GUI nor Phoscon. 
<img width="1920" alt="Bildschirmfoto 2019-03-23 um 13 26 35" src="https://user-images.githubusercontent.com/45996965/54866025-57b99700-4d6f-11e9-8e04-348ff1c6f102.png">
<img width="1907" alt="Bildschirmfoto 2019-03-23 um 13 26 13" src="https://user-images.githubusercontent.com/45996965/54866026-57b99700-4d6f-11e9-9124-c2efc36ce52e.png">

Thanks for your help!!

PR #1155, opened 2019-01-20
Bug fixes Xiaomi special report

ebaauw:
- Apply `config.offset` when updating `state.temperature` or `state.humidity`, see #1154;
- Also update etag and save to database when only `config.reachable`, `config.battery`, and/or `state.lastupdated` are updated;
- Issue event when `state.lastupdated` is updated to fire rules based on `state.lastupdated`.


ebaauw:
That was quick!

manup:
>That was quick!

Coffee makes merging quick :)

Issue #1201, opened 2019-01-31
Hue outdoor motion sensor


MattL0:
https://www2.meethue.com/en-us/p/hue-outdoor-sensor/046677541736# 

Who id willing to try theses?  :o

fsandblom77:
Hi @MattL0 
I have two of these sensors. Bought them today.
It doesn´t work :) What would you like me to try?

ebaauw:
Could you pair one with deCONZ and post the screenshots as described in https://github.com/dresden-elektronik/deconz-rest-plugin/wiki/Request-Device-Support.

From what I read, they're very much like the regular motion sensors, probably just another _Model Identifier_.

ebaauw:
Indeed, this looks all too familiar.  Could you try and read the _SW Build ID_ and _Date Code_ (double-click on the attribute to get a popup window).  Just need to whitelist the model identifier, at 17 places in the code...

fsandblom77:
Thanks! 
New C
![hue-outdoor-clusterinfo3](https://user-images.githubusercontent.com/47333826/52236264-a3e76f80-28c6-11e9-99ad-fb036ddc95bd.jpg)
luster Info screenshot..

ebaauw:
That does seem to be different (newer?) firmware.  My regular Hue motion sensors report 6.1.0.18912.

fsandblom77:
Yes, you are correct. I compared firmware version with my other regular motion sensors and I have the same version as you 6.1.0.18912 so it seems like they have a new firmware on the new outdoor motion sensor.

fsandblom77:
Just an update..
Updated to latest commit(Deconz)
I can now add the sensors from Phoscon. It is not visable in Phoscon but I can see it in Rest API. Also get presence detection, temperature and brightness.
So I'm happy because I'm using the Rest API.
Thanks @ebaauw for your effort!

ebaauw:
Also added support in homebridge-hue v0.11.14.

NedSin:
Hello together, 
is there any News or Information about an update for Phoscon who make the new Sensor Running ? 

Phoscon Version: 262F0500 

Best Greetings

SirMaggot:
Motion and temp is running just fine, but i cant get the lux to show anything else then 0.
Anyone else having the same issue?
Running DeCONZ 2.05.59 and firmware 26300500

olemr:
on .59
Just added the  HUE outdoor sensor from PWA.
[...]

PR #1218, opened 2019-02-04
Hue outdoor motion sensor

ebaauw:
_Model Identifier_ SML002 (vs SML001 for the regular motion sensor).  See #1201.


manup:
Looks like only the housing changed :)

Issue #973, opened 2018-11-24
xy values


ebaauw:
For the list: there's still a number of issues how the REST API plugin handled `xy` values:
1. ~~Hue bridge cuts the API values to four decimals; the deCONZ REST API uses up to six decimals.~~
2. ~~The API reports the ZigBee values divided by 65279 (0xfeff, the maximum allowed value for the _Current X_ and _Current Y_ attributes).  It should instead report the ZigBee value divided by 65535 (0xffff), as defined by the ZigBee standard (and as implemented by the Hue bridge).  When setting these values through the API, the API values should be multiplied by 65535 as well, but maximised to 65279 (corresponding to an maximum API value of 0.9961);~~ **Fixed in v2.05.57.**
3. ~~Some (non-Hue) lights hang when receiving a _Current Y_ value of 0 (I assume because of a divide by 0 bug in the firmware), see #450.  To workaround this, I think the REST API plugin should minimise the value to 1.~~ **Fixed in v2.05.57.**
4. ~~The web socket events contain the raw ZigBee values.  They should use the values divided by 65535 instead;~~
5. ~~The `x` and `y` values are still issues through separate web socket events.  They should be combined in a single event (preferably with the other `state` attributes);~~
6. Typically, lights support way fewer colours than in the [0.0001, 0.0001] x [0.9961, 0.9961] range.  While there is a standard for a light to report their gamut, through the primaries in the _Color control_ cluster, not all manufacturers have implemented this.  The Hue bridge API now exposes this through `capabilities.control.colorgamut`, see https://developers.meethue.com/documentation/lights-api.  It would be nice if the deCONZ REST API would as well (currently I whitelist all Hue lights in homebridge-hue for this info).  I'm not sure, however, if the Hue bridge actually retrieves this info from the light, or whether it has whitelisted the Hue lights;
7. When requesting a non-supported colour, some lights display (Hue) the closest supported colour, while others (IKEA) display a default colour instead. Hue lights report back the actual colour displayed, other lights typically report back the requested colour.  The latter results in the API not reflecting the actual colour.  The former leads to weird behaviour: the API reflects the requested colour at first, but the colour changes "spontaneously" the next time the light is polled.  
[...]

PR #1357, opened 2019-03-15
Web socket notifications for state changes.

ebaauw:
Multiple fixes in the handling of web socket notifications for `state` changes:
- Report `orientation` as an array in web socket notification (as is done in the GET response), see #748;
- Combine multiple `state` changes for a light or group into a single web socket notification (as is done for sensors), see #973 issue 5;
- Report `xy` as an array in the web socket notification (as is done in the GET response), see #973 issue 5;
- Use CIE 1931 values (between 0 and 0.9961) for `xy` in the web socket notification (as is done in the GET response), see #973 issue 4;
- Round `xy` values to four decimals (as is done by the Hue bridge), see #973 issue 1.

@manup, I propagated `lastStatePush` from `Sensor` to `RestNodeBase` and added `RestNodeBase` as super type of `Group`.

### DIFF
--- a/de_web_plugin.cpp
+++ b/de_web_plugin.cpp
@@ -6057,7 +6057,7 @@ void DeRestPluginPrivate::updateSensorNode(const deCONZ::NodeEvent &event)
                                     }
                                     else if (i->modelId() == QLatin1String("SmartPlug")) // Heiman
                                     {
-                                        current *= 100; // 0.01A -> mA
+                                        current *= 10; // 0.01A -> mA
                                     }
                                     else
                                     {
--- a/group.h
+++ b/group.h
@@ -16,15 +16,13 @@
 #include <QTime>
 #include <vector>
 #include "resource.h"
-#include "rest_node_base.h"
 #include "scene.h"
 
 /*! \class Group
 
     Represents the group state of lights.
  */
-class Group : public Resource,
-              public RestNodeBase
+class Group : public Resource
 {
 public:
     enum State
--- a/group.h
+++ b/group.h
@@ -16,13 +16,15 @@
 #include <QTime>
 #include <vector>
 #include "resource.h"
+#include "rest_node_base.h"
 #include "scene.h"
 
 /*! \class Group
 
     Represents the group state of lights.
  */
-class Group : public Resource
+class Group : public Resource,
+              public RestNodeBase
 {
 public:
     enum State
--- a/resource.h
+++ b/resource.h
@@ -240,6 +240,7 @@ public:
     int itemCount() const;
     ResourceItem *itemForIndex(size_t idx);
     const ResourceItem *itemForIndex(size_t idx) const;
+    QDateTime lastStatePush;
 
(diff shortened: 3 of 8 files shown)


Kane610:
> Report xy as an array in the web socket notification (as is done in the GET response), see #973 issue 5;
This looks like a breaking change

ebaauw:
> This looks like a breaking change

I suppose so: if you rely solely on web socket notifications, you no longer get to see `xy` values.  I think any app should be able to handle (i.e. ignore) new attributes.

I could remove one `else` from the code and (for the time being?) include in the web socket notification both the new `xy` attribute (with the normalised floating point values) as well as the old `x` and `y` attributes (with the raw integer Zigbee values).  Would that help?

ebaauw:
And maybe update the API version?  Personally, I don't use this anymore - I sort of lost track what it's supposed to mean when it was bumped from 1.0.9 to 1.16.0 in https://github.com/dresden-elektronik/deconz-rest-plugin/commit/f85529781bb09894ec07d4b021deeb4415d84979.

Kane610:
Doing it gradually would be really great. This is the first breaking change, but maybe there should be an explicit strategy for that? Maybe keep a behaviour for minimum of 2 releases or something.

Kane610:
Just to be clear; I do like this effort. It is just hard to avoid bad experiences among users, if it is just removed without giving some time to upgrade (in this case, two systems)

ebaauw:
Not to worry, @Kane610.  New commit includes both `xy` and `x` and `y` in the message.

manup:
Sorry for the delay, thanks a lot for the good work. Here are a few comments for the PR `lastStatePush`.

>@manup, I propagated lastStatePush from Sensor to RestNodeBase and added RestNodeBase as super type of Group.

I'm afraid RestNodeBase is not a good fit here, since this base class originally was intended for nodes, aka devices, only. I would love to get rid of RestNodeBase completely in mid future in favour of Resources and ResourceItems.

It would be great if you can add the `lastStatePush` timestamp in form of a ResourceItem. Since everything including groups are already Resources, this is the most general and future proof method.

ebaauw:
> It would be great if you can add the `lastStatePush` timestamp in form of a ResourceItem.

I’m not sure that would be a good idea.  We don’t want to expose `lastStatePush` through the REST API.  So we’d need additional code in rest_*.cpp to ignore this ResourceItem, on GET as well as for the web socket notifications (basically everywhere where we loop over a resource’s ResourceItems).  If you want to go this route, we’d better introduce a `hidden` or `notExposed` property for the ResourceItem.

ebaauw:
@manup, I moved `lastStatePush` from `RestNodeBase` to `Resource` and removed `RestNodeBase` as superclass for `Group`.  It's a push of the resource's state - I don't know why I didn't put it there in the first place.

ebaauw:
Added a bug fix for value of `state.current` for the Heiman smartplug.

PR #1394, opened 2019-03-29
Add rule action with address /script that executes a script 

ma-ca:
Add rule action with `"address": "/script/somevalue"` that executes a script when an action is triggered. 

The action address must have at least two strings separated with "/" for example `/script/somevalue`. The values after /script are passed as parameters to the script.

The script `deconz_rule_action.sh` is executed with the following arguments: rule id, rule name, action body, url after /script. 

The script name is hardcoded to prevent the execution of arbitrary code. The script is located in the deCONZ user home directory, for example `/home/pi` directory.

Example script deconz_rule_action.sh that executes a curl on a Philips Hue Bridge when an action is executed:

```
#!/bin/bash

echo "$(date) $@" >> /home/pi/deconz_rule.log
echo "rule id: $1" >> /home/pi/deconz_rule.log
echo "name: $2"  >> /home/pi/deconz_rule.log
echo "body: $3" >> /home/pi/deconz_rule.log
echo "args: $4" >> /home/pi/deconz_rule.log

if [ "$1" = "2" ]; then
  curl -s -X PUT -d '{"on":true}' http://192.168.1.252/api/29b7e422ee6f973bc97b3e1f3269a3/lights/9/state -o /dev/null
fi
```

Example rule that executes a script.

```
    "2": {
        "actions": [
            {
                "address": "/script/lighton",
                "body": {
                    "on": true
                },
                "method": "PUT"
            }
        ],
        "conditions": [
            {
                "address": "/sensors/12/state/dark",
                "operator": "eq",
                "value": "true"
            },
            {
                "address": "/sensors/12/state/dark",
                "operator": "dx"
            }
        ],
```

### DIFF
--- a/rest_rules.cpp
+++ b/rest_rules.cpp
@@ -12,8 +12,6 @@
 #include <QVariantMap>
 #include <QRegExp>
 #include <QStringBuilder>
-#include <QFile>
-#include <QProcess>
 #include "de_web_plugin.h"
 #include "de_web_plugin_private.h"
 #include "json.h"
@@ -841,7 +839,7 @@ bool DeRestPluginPrivate::checkActions(QVariantList actionsList, ApiResponse &rs
         }
 
         //no dublicate addresses allowed
-        const char *resources[] = { "groups", "lights", "schedules", "sensors", "rules", "script", nullptr };
+        const char *resources[] = { "groups", "lights", "schedules", "sensors", "rules", nullptr };
 
         for (int i = 0; ; i++)
         {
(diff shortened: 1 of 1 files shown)


ebaauw:
I understand what you're trying to do, but I'm not sure this is the right way to do it.  Spawning subprocesses from deCONZ could cause a lot of issues, e.g. when the script hangs or goes crazy consuming system resources.

You can probably achieve the same by setting a CLIPGenericFlag or CLIPGenericStatus from your rule, and running a small programme (like `dc_eventlog`) that picks up the web socket notification for the CLIP sensor and executes the script.  That way, the script execution can run under another user (or even on another system).  
I'm half planning to add web socket notifications for when a rule has fired, which would eliminate the need for the CLIP sensor.

ma-ca:
The script is started in a new process and without awaiting the return code. So even if the script fails or hangs, it does not affect deCONZ. 

```cpp
QObject *parent = nullptr;
QProcess *scriptProcess = new QProcess(parent);
scriptProcess->start(script, arguments);
```

ma-ca:
I have the action script enabled for more than a month without any problems. Also a long running script (with a sleep 3600) is running fine. If the rules action is triggered again while the script is still running, the same script is executed is a second process.

In general the action script is working well and provides an enhancement to the rules actions.

ma-ca:
An external URL in the actions address is not possible, or was this added recently? 
```
curl <IP>/api/<key>/rules/<id> -X PUT -d '{ "actions": [ { "address": "http://192.168.1.252/api/29b7e422ee6f973bc97b3e1f3269a3/lights/9/state", "method": "PUT", "body": { "on": true } } ], "periodic": 0 }'

[
    {
        "error": {
            "address": "http://192.168.1.252/api/29b7e422ee6f973bc97b3e1f3269a3/lights/9/state",
            "description": "Rule actions contain errors or an action on a unsupported resource",
            "type": 608
        }
    }
]
```

And looking into this, I found that it was more work to implement a call to an external URL rather that just execute a generic script.

Anyway, I will change my PR to call an external URL instead a script.

manup:
>An external URL in the actions address is not possible, or was this added recently?

Very recently :) it's in the commit https://github.com/dresden-elektronik/deconz-rest-plugin/commit/d57e88d9b5f86a19dc6b535ff533be27c2540ad5 from today.

ma-ca:
Well, then I will use Webhooks and close this PR.

PR #1471, opened 2019-04-26
Hue scenes API

KodeCR:
This PR add support for the Hue API scenes. It does so by:
- creating a group with all the lights and maps this to "group 0" to support LightScene
-- just using group id 0 did not work for all lights nor 65535, so it finds a free group from 65520 going downwards
-- removes all "group0" special casing and maps it to the above group
- update the scene class to support all new hue api properties
-- update the database to store them
-- refactor where the scene class is used
-- split out LightState and share between scenes and light nodes
- add the relevant scenes api
-- supports LightScene and GroupScene, support setting lightstates, removing, adding, modifying lights in a LightScene
-- recalling a scene for all lights is like a regular group scene recall but instead using the dedicated "group0" group
-- recalling a scene I believe always uses a groupcast so calling "group0" LightScene from a specific group should only change the light in the group and the scene
-- there is significant overlap between what is currently for scenes under groups, if this has been tested better what's now under groups can be merged in here

This addresses issues #29, #663, and the discusion in PR #1005.

I realise this is a massive PR so any testing or other feedback would be appreciated.

~There is one issue I have remaining: recalling a color-scene behaves a bit strangely. The Hue app first shows the scene color, than changes brightness to whatever it was previously set, and then updated it again to the brightness of the scene. The lights themselves do seem to behave, so it looks like it's a mismatch between the lightstate of the light and scene. I don't know well enough how this is supposed to work. In particular how the hue-sat, ct, and xy modes interact, and how and when lightstates are/should be updated. There are also various colour conversions going on, but it looks like they are not always kept in sync, which looks like is deliberate? Some help here would be useful 👍.~


manup:
Hi, sorry for the late reply. Thanks very much for the PR. Since this is a very large one, I'm afraid it will take some time until it's fully tested and can be merged into master.

We need to make sure that existing scenes made with current versions as well as API clients and apps are still working. Also that Zigbee scene creation and maintenance won't be affected.

KodeCR:
Hi @manup,

Thanks, yes I realise it is quite big. If it helps I can split it up in separate PRs?

In terms of risk of breaking current API clients, I think that is quite low however:
- group 0 now behaves like any other group, it just has all lights in it and is hidden, so that shouldn't break anything
- the scene class refactoring didn't change functionality, it only added fields so if they are properly ignored all is fine
- splitting out the lightstate shouldn't change functionality
- the old/deconz scenes api is untouched, except for moving the code recalling a scene so that it is shared
- all the rest is in the new scenes api, which is the most risky in terms of having bugs, but that doesn't matter for current clients as it's all new. In fact I already found an issue with GroupScenes so I'll add a commit to fix this later

Let me know if I can help with progressing this.

KodeCR:
I fixed the GroupScene issue, and another issue where not all of the scenes where set to the correct settings because the Hue App creates a lot of requests at the same time causing some zigbee tasks to be dropped. That in turn gave a "gateway busy" error-code which is deconz specific and crashed the Hue app leaving the scenes but with the wrong settings. I upped the max-tasks to 50 and used only Hue API compliant error-codes now.

KodeCR:
It turns out only my Osram lights behave a bit strangely when recalling a scene in the Hue app. My philips lights work perfectly.

@manup, the Hue App creates scenes with xy or ct. There are three ways to set a colour; xy, ct, and hue/sat, should the others be computed and stored in the scene as well when setting one? How does this work on the Zigbee level, is there only one representation there?

Thanks

ebaauw:
I would like to see the deCONZ REST API supporting the `scenes` resources, as exposed by the Hue API.

> If it helps I can split it up in separate PRs?

Maybe we can phase the implementation by first supporting only `GroupScene` scenes.  These were introduced in Hue API v1.28.0 and are linked to a group (other than `/groups/0`) like deCONZ scenes.

> We need to make sure that existing scenes made with current versions as well as API clients and apps are still working.

It would just be additional syntax; the functionality remains the same.

Please make sure `scenes` can be added to `resourcelinks`.

> creating a group with all the lights and maps this to "group 0" to support LightScene

The real underlying issue is that the Hue bridge assigns separate resource IDs to groups (much like the light IDs), and the deCONZ uses the Zigbee _Group ID_ as resource ID.  I think deCONZ should be using separate resource IDs as well, exposing the Zigbee _Group ID_ (in hex) under `uniqueid`.  Just as a light resource can be identified by light ID (`/lights/1`) as well as by `uniqueid` (`/lights/00:17:88:01:00:f0:71:18-0b`), a group resource could be identified by group ID (`/groups/1`) as well as by `uniqueid` (`/groups/ab:cd`).  The existing `/groups/0` functionality would still be addressable under `/groups/00:00` where `/groups/0` would be a real Zigbee group, just like on the Hue bridge.

The Hue bridge uses a random (?) string as ID for the `/scenes` resource.  deCONZ should do the same, and provide a `uniqueid`, e.g. `ab:cd-01` for Zigbee scene 0x01 under group 0xabcd.  Then `/scenes` resources can be identified by `/scenes/xxxxxxxxxx` as well as by `/scenes/ab:cd-01` and `/groups/ab:cd/scenes/1` (though I would prefer a hex string instead of a decimal number).

> I fixed the GroupScene issue, and another issue where not all of the scenes where set to the correct settings because the Hue App creates a lot of requests at the same time causing some zigbee tasks to be dropped.

Do you have a log of the API requests by the Hue app and the corresponding Zigbee requests by the Hue bridge vs the deCONZ gateway?

> After some more testing it is clearly that for ct scenes the Hue app uses xy to represent the colour which can be way off until it's updated with the actual xy from a light report, both for Osram and Philips lights.

Maybe the light states under a scene should be read-only in the REST API?  They're mainly there for the REST API plugin to update the light resources when recalling a scene (so the resources reflect the actual light state before the light is next polled).

> There is one issue I have remaining: recalling a color-scene behaves a bit strangely. The Hue app first shows the scene color, than changes brightness to whatever it was previously set, and then updated it again to the brightness of the scene. The lights themselves do seem to behave, so it looks like it's a mismatch between the lightstate of the light and scene. I don't know well enough how this is supposed to work. In particular how the hue-sat, ct, and xy modes interact, and how and when lightstates are/should be updated. There are also various colour conversions going on, but it looks like they are not always kept in sync, which looks like is deliberate? Some help here would be useful

In Zigbee you can only store the _current_ light state in a scene.  That means, updating the scene light state involves:
- Setting the light state (through multiple commands to change _On/Off_; _Current Level_; and, depending on how you set the colour, _Current X_ and _Current Y_, or _Color Temperature_, or _Current Hue_ and _Current Saturation_, or _ Enhanced Current Hue_ and _Current Saturation_;
- Sending the _Store Scene_ command.

The problem with this is that you should only send the _Store Scene_ command when the light state has settled and is reflected correctly in the REST API.  When storing a scene too early, the values cached by the REST API don't match the actual light state, resulting in all sorts of funny behaviour when recalling the scene.

There's no generic way of knowing for sure whether the light state has settled, as different light types/manufacturers handle changing and reporting light states very differently.  E.g. some lights report the target state when specifying a long transition time, others "follow" the transition; Hue lights update _Color Temperature_ when setting _Currrent X_ and _Current Y_ or v.v.; I haven't seen any other lights that do this.  Some lights pretend to support both _Color Mode_ and _Enhanced Color Mode_, but actually only report one of these correctly.  And there's the issue of IKEA lights ignoring a next command when a transition is still underway.

KodeCR:
Thanks very much for your feedback @ebaauw. I very much would like to get this in, it was quite a bit of effort.

> The real underlying issue is that the Hue bridge assigns separate resource IDs to groups (much like the light IDs), and the deCONZ uses the Zigbee Group ID as resource ID. I think deCONZ should be using separate resource IDs as well, exposing the Zigbee Group ID (in hex) under uniqueid. Just as a light resource can be identified by light ID (/lights/1) as well as by uniqueid (/lights/00:17:88:01:00:f0:71:18-0b), a group resource could be identified by group ID (/groups/1) as well as by uniqueid (/groups/ab:cd). The existing /groups/0 functionality would still be addressable under /groups/00:00 where /groups/0 would be a real Zigbee group, just like on the Hue bridge.

This PR does just that, it creates a real Zigbee group, e.g. group 65320, to act as group 0 and adds all lights to that group. It then stores this id and translates to it whenever the api uses id 0, pretty simple really. Not quite sure why any other group would need to differ between the api and zigbee side though?

>The Hue bridge uses a random (?) string as ID for the /scenes resource. deCONZ should do the same, and provide a uniqueid, e.g. ab:cd-01 for Zigbee scene 0x01 under group 0xabcd. Then /scenes resources can be identified by /scenes/xxxxxxxxxx as well as by /scenes/ab:cd-01 and /groups/ab:cd/scenes/1 (though I would prefer a hex string instead of a decimal number).

The deconz database uses 0xaaaabb as the key for a scene which is the group hex and scene hex concatenated, I've just used that unique string as the /scenes id as well. A uniqueid based on a guid should be pretty simple to add.

>Maybe we can phase the implementation by first supporting only GroupScene scenes. These were introduced in Hue API v1.28.0 and are linked to a group (other than /groups/0) like deCONZ scenes.

Because of the above, a groupscene and a lightscene are pretty much the same on the implementation side. There is just a little bit of additional code to handle translating group 0 to the actual zigbee group and to deal with calling a group 0 scene from another group. That is so little it's not worth splitting up. I could split out the group 0 stuff (basically the first commit) so we can go through that first.

>It would just be additional syntax; the functionality remains the same.

Agreed, in fact the /scenes api code is almost completely separate from the /groups/scene code and the latter has hardly been touched.

>Please make sure scenes can be added to resourcelinks.

I got this working trivially, I'll create another PR. In fact I can now fully configure a Hue Dimmer Switch and Hue Motion sensor with the Hue app and they behave exactly the same.

>Do you have a log of the API requests by the Hue app and the corresponding Zigbee requests by the Hue bridge vs the deCONZ gateway?
>Maybe the light states under a scene should be read-only in the REST API? They're mainly there for the REST API plugin to update the light resources when recalling a scene (so the resources reflect the actual light state before the light is next polled).

I don't have the log at the moment but I could recreate if you really want it, but this is what the Hue app does for each scene:
- create a new (group)scene (using the current lightstate)
- for each light in the scene use a PUT /scenes/x/lightstates/y to set the actual desired lightstate (lightstates thus shouldn't be readonly)

On the Zigbee side deconz does a StoreScene and a ModifyScene for each. Not sure if the Hue bridge does the same thing (I only have one Raspbee, which is in use :) and thus not set up to sniff).
Thus, when adding the default scenes for a room with three colour lights in the Hue app for example, this creates:
- 11 storescene tasks and
- 3x11 modifyscene tasks 
in quick succession, which was too much for deconz

>In Zigbee you can only store the current light state in a scene. That means, updating the scene light state involves:

Thanks, that makes much more sense now. I see that anything other then xy for lights is optional, so you can't rely on the light to compute the proper ct or hs. I think the scene class should compute the other two when setting one in that case.
I'm a bit confused that it is required to set a light to the appropriate colour, brightness etc. and then store the scene. If I for example would add all the default room scenes in the Hue app as I have done I thus would expect to see the light colour and brightness change 11 times when creating the scenes. I definitely didn't see that happen...

ebaauw:
I’ll see this weekend if I can sniff the Zigbee commands generated by the Hue bridge, when creating a scenes in the Hue app.

> I'm a bit confused that it is required to set a light to the appropriate colour, brightness etc. and then store the scene. If I for example would add all the default room scenes in the Hue app as I have done I thus would expect to see the light colour and brightness change 11 times when creating the scenes. I definitely didn't see that happen...

I think the Hue bridge might actually defer storing a v2 scene to the lights until it is first recalled.

> On the Zigbee side deconz does a StoreScene and a ModifyScene for each. Not sure if the Hue bridge does the same thing

I don’t think it does.

> in quick succession, which was too much for deconz

deCONZ is far more powerful than the Hue bridge, even when it’s running on a Raspberry Pi.

I think the Hue bridge handles a PUT of a scene lightstate in memory, without generating any ZigBee commands.  When recalling the scene for the first time, the bridge sends the regular commands to change the light state, followed by a _Store Scene_ and/or _Modify Scene_.  Only on subsequent recalls it issues a _Recall Scene_.

We really need to sniff the API traffic _and_ the Zigbee traffic to know for sure.

KodeCR:
>deCONZ is far more powerful than the Hue bridge, even when it’s running on a Raspberry Pi.

Yeah, I know. Sorry, I wasn't clear, I meant that deconz queues the zigbee commands and the max depth was only 20. So with the 44 tasks almost at once it would drop tasks and issue a "gateway busy" error code, which the Hue app doesn't know and subsequently crashes. So I up-ed the max-tasks to 50.

>I’ll see this weekend if I can sniff the Zigbee commands generated by the Hue bridge, when creating a scenes in the Hue app.
>We really need to sniff the API traffic and the Zigbee traffic to know for sure.

Great, thanks, that will be useful

>I think the Hue bridge might actually defer storing a v2 scene to the lights until it is first recalled.

Hmm, I meant with deconz. I just re-did the default scenes for the room I'm in and the lighstates indeed are all the same for each scenes. It double checked before and it definitely _was_ correct, so I think the Hue app does some checking and updating to fix them if incorrect.

>I think the Hue bridge handles a PUT of a scene lightstate in memory, without generating any ZigBee commands. When recalling the scene for the first time, the bridge sends the regular commands to change the light state, followed by a Store Scene and/or Modify Scene. Only on subsequent recalls it issues a Recall Scene.

This sounds sensible. Ok, I'll wait for the results from sniffing and then we can do it in the same way as the Hue bridge does.

ebaauw:
Did some preleminary sniffing with one Hue colour temperature light connected to my gen2 Hue bridge.  And some deep diving into the ZCL and ZLL specs.

- On creating a scene, the Hue bridge sends a single unicast _Enhanced Add Scene_ command (0x40).  Basically this is the same a the regular _Add Scene_ command, but with a finer-grained transitiontime (with a 0.1 second resolution).  I don't know whether this command is sent on the POST of the scene, or the PUT of the scene's light state;
- The Hue app (for the gen2 bridge) now generates `GroupScene` scenes (?!):
  ```
  "BT8rpwky0h0V18X": {
    "name": "Relax",
    "type": "GroupScene",
    "group": "1",
    "lights": [
      "48"
    ],
    "owner": "16b617c0-339c-4ef7-9c52-87e59c06b27a",
    "recycle": false,
    "locked": false,
    "appdata": {
      "version": 1,
      "data": "OylnN_r01_d01"
    },
    "picture": "",
    "lastupdated": "2019-05-28T13:11:23",
    "version": 2
  }
  ```
- They seem to be recalled using a groupcast _Recall Scene_ command, to /groups/0 (?!);
- Contrary to what I thought, the _Enhanced Add Scene_ command actually includes the light state in the payload:
  ```
  0bd1 01 0400 00 0600 01 01 0800 01 90 0003 04 7980 4b6a
  |    |  |    |  |    |  |  |    |  |  |    |  |    + Current Y: 0.4152
  |    |  |    |  |    |  |  |    |  |  |    |  + Current X: 0.5019
  |    |  |    |  |    |  |  |    |  |  |    + length: 4 bytes
  |    |  |    |  |    |  |  |    |  |  + Color Control cluster
  |    |  |    |  |    |  |  |    |  + Current Level: 144
  |    |  |    |  |    |  |  |    + length: 1 byte
  |    |  |    |  |    |  |  + Level Control cluster
  |    |  |    |  |    |  + On/Off: on
  |    |  |    |  |    + length: 1 byte
  |    |  |    |  + On/Off cluster
  |    |  |    + scene name: empty string, length 0 bytes
  |    |  + transition time: 0.4s
  |    + scene 0x01: Relax
  + group 0xd10b: /groups/0 (?!)
  ```
- Note that the deCONZ GUI doesn't support specifying a payload to the _Add Scene_ command, hence my confusion;
- Note that the payload doesn't contain the attribute ids nor their data types: only their values.  The ZLL spec states that for _Color Control_ (0x0300), the attributes are: _CurrentX_, _CurrentY_, _EnhancedCurrentHue_, _CurrentSaturation_, _ColorLoopActive_, _ColorLoopDirection_, _ColorLoopTime_.  It's allowed to leave out trailing attributes, so only _CurrentX_ and _CurrentY_ are sent.   _Even to a colour temperature light that otherwise doesn't support these attributes_ (?!).  According to the ZCL spec:
  > Since there is a direct relation between _ColorTemperature_ and XY, color temperature, if supported, is stored as XY in the scenes table.
- The _Store Scene_ command doesn't accept any state in its payload, and stores the current light state:
  > On receipt of this command, the device SHALL (if possible) add an entry to the Scene Table with the Scene ID and Group ID given in the command, and all extension field sets corresponding to the current state of other clusters on the device.
- Hue lights support up to 50 scenes.  This is also documented in their API.  The ZCL spec specifies at least 16 scenes.

I'm afraid this leads to more questions than answers:
- I don't understand why they use the group for `/groups/0` instead of the one for `/groups/1` (the room group for which the app creates the scenes);
- How would an enhanced colour temperature light (supporting both XY and colour temperature) distinguish between colour temperature and XY in an _Enhanced Add Scene_.  Or will it only recall colour temperature when the scene was updated through _Store Scene_?
- The Hue API supports 50 scenes per light, Hue lights support 50 scenes.  The ZCL specs states that 16 is the minimum.  I wonder how the Hue APIs deals with lights that support fewer scenes.

ebaauw:
I should RTFM.  A Hue bridge `GroupScene` is a scene containing a single group; a `LightScene` is a scene containing a list of lights.  The `lights` attribute is read-only for a `GroupScene` and updated automatically when lights are added or removed from the group.  Both are still stored under the group for `/groups/0`.

> - How would an enhanced colour temperature light (supporting both XY and colour temperature) distinguish between colour temperature and XY in an _Enhanced Add Scene_.  Or will it only recall colour temperature when the scene was updated through _Store Scene_?

After recalling the scene for an enhanced color light, it reports colormode `xy`, even though the scene's light state contains `ct`.

KodeCR:
Excellent, this is very useful info. So:
- creating a GroupScene will result in a StoreScene for that group
- creating a LightScene will result in a StoreScene for each light in the list
- modifying a scene/putting a lightstate will result in EnhancedAddScene for each/that light

That all makes sense. Still, I think this PR should be merged first before changing scenes more generally.(?)

ebaauw:
I haven’t yet checked the API calls that the Hue app makes (it uses SSL nowadays), but I’ve only seen unicast _Enhanced Add Scene_ commands (i.e. to a light, not to a group).  I haven’t seen any _Store Scene_ commands.  As far as I can tell, the (only?) difference between a `LightScene` and a `GroupScene` is that the Hue bridge (I think, could also be the Hue app) automatically adds the scene to the light, when the light is added to the group.

KodeCR:
Oh, since the api docs for creating a new scene say "For a given scene the current light settings of the given lights resources are stored." I assumed it would do a StoreScene.

I have the Hue app API calls going to deconz using https://github.com/KodeCR/hue-pass.

Just checked again:
1. I first deleted all light from the group which also deletes the default scenes:

```
May 30 08:34:24 _ deCONZ[8752]: 08:34:24:276 HTTP API PUT /api/_/groups/47 - 192.168.1.152
May 30 08:34:24 _ deCONZ[8752]: 08:34:24:277 Text Data: #011{"name":"Study","lights":[],"class":"Office"}
May 30 08:34:24 _ deCONZ[8752]: 08:34:24:277 ApiMode: 3
May 30 08:34:24 _ deCONZ[8752]: 08:34:24:277 [{"success":{"/groups/47/name":"Study"}},{"success":{"/groups/47/lights":[]}}]
May 30 08:34:24 _ deCONZ[8752]: 08:34:24:328 HTTP client hdr detected:
May 30 08:34:24 _ deCONZ[8752]: DELETE /api/_/scenes/0x002F07 HTTP/1.1
May 30 08:34:24 _ deCONZ[8752]: Host: 192.168.1.158
May 30 08:34:24 _ deCONZ[8752]: Accept: */*
May 30 08:34:24 _ deCONZ[8752]: Content-Type: application/json
May 30 08:34:24 _ deCONZ[8752]: #015
May 30 08:34:24 _ deCONZ[8752]: 08:34:24:329 HTTP API DELETE /api/_/scenes/0x002F07 - 192.168.1.152
May 30 08:34:24 _ deCONZ[8752]: 08:34:24:329 ApiMode: 3
May 30 08:34:24 _ deCONZ[8752]: 08:34:24:330 [{"success":{"id":"0x002F07"}}]
May 30 08:34:24 _ deCONZ[8752]: 08:34:24:334 HTTP client hdr detected:
May 30 08:34:24 _ deCONZ[8752]: DELETE /api/_/scenes/0x002F08 HTTP/1.1
...
May 30 08:34:27 _ deCONZ[8752]: 08:34:26:970 APS-DATA.indication srcAddr: 0x5c3d, srcEp: 0x01 dstAddrMode: 2, profile: 0x0104, cluster: 0x0004, lqi: 191, rssi: -67
May 30 08:34:27 _ deCONZ[8752]: 08:34:26:971 Remove from group response for light 5. Status: 0x00, capacity: 1
...
```

Notable is that I couldn't find any remove scene responses while I believe addTaskRemoveScene should have been called. Maybe the remove from group also removes all scenes for that group?

2. Next, I added the lights back, and then all default scenes:
```
May 30 08:34:50 _ deCONZ[8752]: 08:34:50:019 HTTP API POST /api/_/scenes - 192.168.1.152
May 30 08:34:50 _ deCONZ[8752]: 08:34:50:019 Text Data: #011{"type":"GroupScene","group":"47","name":"Bright","recycle":false,"appdata":{"version":1,"data":"z8utG_r47_d05"}}
May 30 08:34:50 _ deCONZ[8752]: 08:34:50:019 ApiMode: 3
May 30 08:34:50 _ deCONZ[8752]: 08:34:50:020 add store scene task, aps-req-id: 0
May 30 08:34:50 _ deCONZ[8752]: 08:34:50:020 [{"success":{"id":"0x002F01"}}]
May 30 08:34:50 _ deCONZ[8752]: 08:34:50:027 HTTP client hdr detected:
May 30 08:34:50 _ deCONZ[8752]: PUT /api/_/scenes/0x002F01/lightstates/5 HTTP/1.1
May 30 08:34:50 _ deCONZ[8752]: Host: 192.168.1.158
May 30 08:34:50 _ deCONZ[8752]: Accept: */*
May 30 08:34:50 _ deCONZ[8752]: Content-Type: application/json
May 30 08:34:50 _ deCONZ[8752]: Content-Length: 21
May 30 08:34:50 _ deCONZ[8752]: #015
May 30 08:34:50 _ deCONZ[8752]: 08:34:50:027 HTTP API PUT /api/_/scenes/0x002F01/lightstates/5 - 192.168.1.152
May 30 08:34:50 _ deCONZ[8752]: 08:34:50:027 Text Data: #011{"on":true,"bri":254}
May 30 08:34:50 _ deCONZ[8752]: 08:34:50:027 ApiMode: 3
May 30 08:34:50 _ deCONZ[8752]: 08:34:50:028 Start modify scene for 0x000B57FFFEBAE751, groupId 0x002F, scene 0x01
May 30 08:34:50 _ deCONZ[8752]: 08:34:50:028 [{"success":{"/scenes/0x002F01/on":true}},{"success":{"/scenes/0x002F01/bri":254}}]
May 30 08:34:50 _ deCONZ[8752]: 08:34:50:032 HTTP client hdr detected:
...
May 30 08:34:50 _ deCONZ[8752]: 08:34:50:411 APS-DATA.indication srcAddr: 0x5c3d, srcEp: 0x01 dstAddrMode: 2, profile: 0x0104, cluster: 0x0005, lqi: 175, rssi: -69
May 30 08:34:50 _ deCONZ[8752]: 08:34:50:412 Modified scene 1 in node 0x000b57fffebae751 status 0x00
```

Notable here is that although is says "add store scene task" and then "Start modify scene", I can only see a "Modified scene" response.

3. If I just create a new scene it behaves the same:
```
May 30 08:35:27 _ deCONZ[8752]: 08:35:27:672 HTTP API POST /api/_/scenes - 192.168.1.152
May 30 08:35:27 _ deCONZ[8752]: 08:35:27:672 Text Data: #011{"type":"GroupScene","group":"47","name":"Tst","recycle":false,"appdata":{"version":1,"data":"JkWjy_r47_d99"}}
May 30 08:35:27 _ deCONZ[8752]: 08:35:27:672 ApiMode: 3
May 30 08:35:27 _ deCONZ[8752]: 08:35:27:673 add store scene task, aps-req-id: 203
May 30 08:35:27 _ deCONZ[8752]: 08:35:27:673 [{"success":{"id":"0x002F04"}}]
May 30 08:35:27 _ deCONZ[8752]: 08:35:27:684 APS-DATA.indication srcAddr: 0x1b3e, srcEp: 0x0B dstAddrMode: 2, profile: 0x0104, cluster: 0x0006, lqi: 255, rssi: -55
May 30 08:35:27 _ deCONZ[8752]: 08:35:27:684 APS-DATA.request id: 199 erase from queue
May 30 08:35:27 _ deCONZ[8752]: 08:35:27:685 ZCL got data for node=0x1B3E, cl=0x0006, at=0x0000, status=0x00, type=0x10
May 30 08:35:27 _ deCONZ[8752]: 08:35:27:698 HTTP client hdr detected:
May 30 08:35:27 _ deCONZ[8752]: PUT /api/_/scenes/0x002F04/lightstates/5 HTTP/1.1
May 30 08:35:27 _ deCONZ[8752]: Host: 192.168.1.158
May 30 08:35:27 _ deCONZ[8752]: Accept: */*
May 30 08:35:27 _ deCONZ[8752]: Content-Type: application/json
May 30 08:35:27 _ deCONZ[8752]: Content-Length: 21
May 30 08:35:27 _ deCONZ[8752]: #015
May 30 08:35:27 _ deCONZ[8752]: 08:35:27:698 HTTP API PUT /api/_/scenes/0x002F04/lightstates/5 - 192.168.1.152
May 30 08:35:27 _ deCONZ[8752]: 08:35:27:698 Text Data: #011{"on":true,"bri":180}
May 30 08:35:27 _ deCONZ[8752]: 08:35:27:698 ApiMode: 3
May 30 08:35:27 _ deCONZ[8752]: 08:35:27:699 Start modify scene for 0x000B57FFFEBAE751, groupId 0x002F, scene 0x04
May 30 08:35:28 _ deCONZ[8752]: 08:35:27:699 [{"success":{"/scenes/0x002F04/on":true}},{"success":{"/scenes/0x002F04/bri":180}}]
May 30 08:35:28 _ deCONZ[8752]: 08:35:27:714 APS-DATA.request id: 202, addrmode: 0x01, addr: 0x002f, profile: 0x0104, cluster: 0x0005, ep: 0x01 -> 0xFF queue: 0 len: 9 tx.options 0x00
May 30 08:35:28 _ deCONZ[8752]: 08:35:27:774 HTTP client hdr detected:
May 30 08:35:28 _ deCONZ[8752]: PUT /api/_/scenes/0x002F04/lightstates/6 HTTP/1.1
...
May 30 08:35:28 _ deCONZ[8752]: 08:35:27:929 APS-DATA.indication srcAddr: 0x5c3d, srcEp: 0x01 dstAddrMode: 2, profile: 0x0104, cluster: 0x0005, lqi: 191, rssi: -67
May 30 08:35:28 _ deCONZ[8752]: 08:35:27:929 Modified scene 4 in node 0x000b57fffebae751 status 0x00
```

So it indeed seems the Hue app only explicitly sets the lightstate of lights for scenes.

I wonder what the Hue bridge does when you only do a API POST to /scenes. I believe that should correspond to a StoreScene zigbee command.

ebaauw:
> I wonder what the Hue bridge does when you only do a API POST to /scenes. I believe that should correspond to a StoreScene zigbee command.

No, it only seems to issue _Enhanced Add Scene_ commands:

A `ph post /scenes '{"name":"test", "lights":["48"], "recycle": false}'` results in a unicast
- _Enhanced Add Scene_ (0bd1, 01, 0400, 00, 0600, 01, 01, 0800, 01, 01, 0003, 04, 7980, 4b6a)

The payload reflects the actual light state, `{"on": true, "bri": 1, "ct": 447}`;

To double check, I changed the light to `{"on": false, "bri": 127, "ct": 447}` and repeated the `ph post`:
-  _Enhanced Add Scene_ (0bd1, 02, 0400, 00, 0600, 01, 00, 0800, 01, 7f, 0003, 04, 7980, 4b6a)

A `ph put /scenes/OYJG-95raJ27dWP/lightstates/48 '{"bri":147}'` results in:
- _Enhanced Add Scene_ (0bd1, 01, 0400, 00, 0800, 01, 4f)

A `ph put /scenes/OYJG-95raJ27dWP/lightstates/48 '{"on": true, "bri":1, "ct": 500}'` results in:
- _Enhanced Add Scene_ (0bd1, 01. 0400, 00, 0600, 01, 01, 0800, 01, 01, 0003, 04, 5c81, 476a)

There is no difference between creating a `LightScene` and a `GroupScene`.  A `ph post /scenes '{"name":"test", "group": "1", "type": "GroupScene", "recycle": false}'` also results in:
- _Enhanced Add Scene_ (0bd1, 03, 0400, 00, 0600, 01, 00, 0800, 01, 7f, 0003, 04, 7980, 4b6a)

There is also no difference with a v1 scene.  A `ph put /scenes/foo '{"name":"test", "lights":["48"]}'` also results in
- _Enhanced Add Scene_ (0bd1, 04, 0400, 00, 0600, 01, 00, 0800, 01, 7f, 0003, 04, 7980, 4b6a)

Haven't yet tried with multiple lights - currently have only one light paired with the Hue bridge.

Let's do some stress testing:
```
$ for scene in $(ph get -al /scenes | grep /name: | cut -d / -f 2) ; do
>   ph delete /scenes/$scene
> done
$ for i in 0 1 2 3 4 5; do
>   for j in 0 1 2 3 4 5 6 7 8 9; do
>     ph post /scenes '{"name": "Test'$i$j'", "lights": ["48"], "recycle": true}'
>   done
> done
"gfzfCNnZMAl6v8V"
"IFsnAK0us3vzOEp"
"9Il93xcXwUgpxkL"
"uT7wZxCYKcvj6U4"
"bu5JoSYhfSPiDXY"
"c79sHH1udbrwkvG"
"k1RP5dkV87IWyA1"
"ToLAWCsbgBKsDk7"
"4N0dhHrdkJ05j5n"
"WSHxPY5mXaTcrIp"
"CCVi7CfFtRYY22v"
"aqQFqa16zXa169V"
"j7VI-YFF9fZSiIE"
"Mrwx-0Khz0seUhN"
"GnvEJTfmyJ3VYwm"
"CjggGQQO-ObQ3CD"
"EqY2IoV3prVJ9ts"
"OSygOmvGzbbw6kj"
"DqZXJ9XNw9jr822"
"QHWETMhgZZ772f5"
"fsZuW49rt-FtlBk"
"q4fymcJVTScmLA5"
"TH2HOmEli1TCWOb"
"VjrRgEHnWjH9i7N"
"K7ylOAA2bL5wbYM"
"Z-4dbWjNDofIMmf"
"0xrQl4lqB1CqTN0"
"nNESgKQ-Fvm3KMm"
"Yon5rPb9BhIRjb5"
"KW6PSJM6Nyh1qrX"
"dx92f3gHtuFy8Ep"
"ESWYX39AitdsL0A"
"pQj5U36pu3EXnyu"
"b6xuprqeHovh4lc"
"iNgMohzwtZdHF5x"
"vOrNRqRjzyeATEs"
"GhPEPmvrAFL0uKm"
"1eMGyW73h4FYmjJ"
"ry5VcJNS0OqhvcU"
"oAQewYBEtirVL7T"
"3T3U9P1qB4bbcl-"
"-wOTU3yuLwZyUBG"
"HJfvipmaVe1jdo-"
"i68Ad0wYEjq6DsQ"
"yIyD0EmckzkJmoc"
"sDsvuTreWytrHPW"
"lyPn6RVAHvMplPL"
"z0bngpfH4fWh08t"
"92K7l90saLjHggn"
"yI3Pexkus18GMIK"
"4rF7s3RCipYfdDI"
"IB2CPZbYeSJV9K6"
"xfpQAoujhjZfPA2"
"xiXMha5PGUl5dL1"
"0n0fDXAihCkIPfc"
"gr3RRVo73obig4z"
"Y7HBD815wekJp6y"
"D3NmfDbI1euHg2N"
"ObA2vzv25Q63M-j"
"8MzcLfSqu1NuVbl"
$ ph get -al /scenes | grep /name:
/gfzfCNnZMAl6v8V/name:"Test00"
/IFsnAK0us3vzOEp/name:"Test01"
/9Il93xcXwUgpxkL/name:"Test02"
/uT7wZxCYKcvj6U4/name:"Test03"
/bu5JoSYhfSPiDXY/name:"Test04"
/c79sHH1udbrwkvG/name:"Test05"
/k1RP5dkV87IWyA1/name:"Test06"
/ToLAWCsbgBKsDk7/name:"Test07"
/4N0dhHrdkJ05j5n/name:"Test08"
/WSHxPY5mXaTcrIp/name:"Test09"
/CCVi7CfFtRYY22v/name:"Test10"
/aqQFqa16zXa169V/name:"Test11"
/j7VI-YFF9fZSiIE/name:"Test12"
/Mrwx-0Khz0seUhN/name:"Test13"
/GnvEJTfmyJ3VYwm/name:"Test14"
/CjggGQQO-ObQ3CD/name:"Test15"
/EqY2IoV3prVJ9ts/name:"Test16"
/OSygOmvGzbbw6kj/name:"Test17"
/DqZXJ9XNw9jr822/name:"Test18"
/QHWETMhgZZ772f5/name:"Test19"
/fsZuW49rt-FtlBk/name:"Test20"
/q4fymcJVTScmLA5/name:"Test21"
/TH2HOmEli1TCWOb/name:"Test22"
/VjrRgEHnWjH9i7N/name:"Test23"
/K7ylOAA2bL5wbYM/name:"Test24"
/Z-4dbWjNDofIMmf/name:"Test25"
/0xrQl4lqB1CqTN0/name:"Test26"
/nNESgKQ-Fvm3KMm/name:"Test27"
/Yon5rPb9BhIRjb5/name:"Test28"
/KW6PSJM6Nyh1qrX/name:"Test29"
/dx92f3gHtuFy8Ep/name:"Test30"
/ESWYX39AitdsL0A/name:"Test31"
/pQj5U36pu3EXnyu/name:"Test32"
/b6xuprqeHovh4lc/name:"Test33"
/iNgMohzwtZdHF5x/name:"Test34"
/vOrNRqRjzyeATEs/name:"Test35"
/GhPEPmvrAFL0uKm/name:"Test36"
/1eMGyW73h4FYmjJ/name:"Test37"
/ry5VcJNS0OqhvcU/name:"Test38"
/oAQewYBEtirVL7T/name:"Test39"
/3T3U9P1qB4bbcl-/name:"Test40"
/-wOTU3yuLwZyUBG/name:"Test41"
/HJfvipmaVe1jdo-/name:"Test42"
/i68Ad0wYEjq6DsQ/name:"Test43"
/yIyD0EmckzkJmoc/name:"Test44"
/sDsvuTreWytrHPW/name:"Test45"
/lyPn6RVAHvMplPL/name:"Test46"
/z0bngpfH4fWh08t/name:"Test47"
/92K7l90saLjHggn/name:"Test48"
/yI3Pexkus18GMIK/name:"Test49"
/4rF7s3RCipYfdDI/name:"Test50"
/IB2CPZbYeSJV9K6/name:"Test51"
/xfpQAoujhjZfPA2/name:"Test52"
/xiXMha5PGUl5dL1/name:"Test53"
/0n0fDXAihCkIPfc/name:"Test54"
/gr3RRVo73obig4z/name:"Test55"
/Y7HBD815wekJp6y/name:"Test56"
/D3NmfDbI1euHg2N/name:"Test57"
/ObA2vzv25Q63M-j/name:"Test58"
/8MzcLfSqu1NuVbl/name:"Test59"
```
The API happily created 60 scenes for the same light, but the lights only contains 44 scenes (0x01 .. 0x2c) (looking at the _Get Scene Membership Response_ from the bridge polling the light).  I only see 44 _Enhanced Add Scene_ commands.  Maybe the bridge silently dropped some of them?

Let's try again, now giving the bridge some more time:
```
$ for scene in $(ph get -al /scenes | grep /name: | cut -d / -f 2) ; do
>   ph delete /scenes/$scene
> done
$ for i in 0 1 2 3 4 5; do
>   for j in 0 1 2 3 4 5 6 7 8 9; do
>     ph post /scenes '{"name": "Test'$i$j'", "lights": ["48"], "recycle": true}'
>     sleep 1
>   done
> done
"TlZQ92MXjko9xDf"
"ZHykANOlDPWQwdd"
"6Wrt0lgDYzzk8fW"
"aGgSC97JmHUbHy2"
"twqf5ARcvtIVyGL"
"pI6LwBlXpGRje11"
"pkq6CqcP6kczpK5"
"wCi49uZOlVA-XmE"
"5Yc2CevGP1GknTy"
"uiPtpG4Te7FIjDs"
"Z-ZDsYxAN1ckVpg"
"WmHoGWgWvVAb5JR"
"71SCfnvwFfxTmiw"
"G1ycb2-QlZpY49k"
"75jGeIErnJ-rIbW"
"9ZUcCKkzCWPOVlx"
"QlkOclWnGrLkjLS"
"8SvCudDPOU6tKbL"
"GBmO98FU1gXbIL4"
"9-NHps2BfQa2q3R"
"D6zOtqdGWW2OqZW"
"4D5C-CqxA4Rmc8d"
"8QjYtTJfFHWwc-V"
"gpAUYfyjBEvScLV"
"ElYmCVPoa8dPjWY"
"Pz48ganTfBJ6A3-"
"kMz14JxCjEuPg7O"
"4OynjK9rHvbrrsF"
"To2jLkmODGaUmpe"
"mryOjDbRCszihVV"
"lawCRQy3mz3DwQh"
"zGGl8TLexfee5xU"
"EQ3WWPB6kyqTQRD"
"YUNKLDtSKJL1vlu"
"5RxuTJehpzO067q"
"hlhKOdvLtAHEnmC"
"cF5XChLIYNvBYD3"
"BRsrrPPkm9SLK7C"
"ME-ITKDrXlEAWau"
"LCLDPMgJXyqCUcN"
"FyWG-YmfQf0ui16"
"xkzMlE6q382EiOd"
"iMpFdhXgcyu0MWZ"
"aDMGNzAI9tNlNrH"
"qLbs3SuT73Q2iM7"
"KQrMFEST3XrR689"
"yQUIDbHdvN4i62J"
"tFUz2IWwEkPAHgd"
"BmvFCKuUkUOdqZY"
"Nwy51Fc7kz7XDxF"
"HKs27s1oobFKhEV"
"RVEzS78Piih0U45"
"KKNZBIFPz8tBV56"
"IBkUcNJavAg9KIh"
"8MguwgiYJFa87Q6"
"pyNUpfxoGbSvmfe"
"p7cCPYw86b0PofT"
"zXz-Cr-1NWp9zz1"
"2NB5ffUitciYWYR"
"IvORANYRXP7XIUQ"
```
Again, only the first 44 scenes seem to be sent to light.  The additional posts don't seem to generate any ZigBee command, not even after a couple of minutes.

Let's try to recall the scenes.  A `ph put /groups/1/action '{"scene": "TlZQ92MXjko9xDf"}'` results in:
- _Recall Scene_ (0xd10b, 0x01)

as expected.  `ph put /groups/1/action '{"scene": "IvORANYRXP7XIUQ"}'` results in:
- _Recall Scene_ (0xd10b, 0x3c)

WTF?  Tried again with `"recycle": false`, but, again, the ZigBee commands stop after scene 0x2c.  Added the default scenes in the Hue app.  No Zigbee commands.  Recalling a default scene works?!  The bridges sends a (broadcast) _Recall Scene_ followed by (unicast) _On_, _Move to Color Temperature_ and _Move to Level_ commands.

Delete all scenes again, now only creating the default scenes in the Hue app.  Now recalling a scene only results in a `_Recall Scene_`, no _On_, _Move to Color Temperature_, nor _Move to Level_.  Apparently the bridge keeps track which lightstate is actually stored in the light.

KodeCR:
Thanks again. I looked into it a bit more. Deconz is properly sending a AddScene with the appropriate payload with the lighstate. The reason I wasn't getting AddScene (modified scene) response in my test is because those were Ikea lights and I was hit by the bug that casused it to disregarded this (second) AddScene command.

I think a good design would be to track as part of the lightstate if it was stored based on the StoreScene/AddScene response. When a scene is recalled we can check for this and try again if there is space left in the scene table and else/also use "MoveTo" commands. This way if a light was unreachable when the scene was created/updated through the api the scene will be added to the light/updated on the next recall, and if there is no more space for scenes it will behave like the Hue bridge.

Issue #1509, opened 2019-05-11
Support for Luzense devices


PrimaryInc:
Hi,

I've been testing a 2 luzense switches, but not been able to pair them,
As far as I know thy are not supported in deconz at the moment.

What information can I provide for support?

wvuyk:
Check this page? https://github.com/dresden-elektronik/deconz-rest-plugin/wiki/Request-Device-Support

corvy:
Yes they are ZigBee, the only info I have is here: https://www.unilamp.no/produkt/zig-b-veggbrytere/

These remotes are meant to be used with some "acutators" to dim traditional lights and make them smart. The great thing about them is that they fit directly into the 55mm frames from Elko in Norway. I am trying to chase the importer to find out who they buy their products from and see if they have more detailed product documentation. This is bought from a sub supplier (whitelabled) and rebranded for the Nordic market. I will see if I can get more information and share when I get it.

corvy:
Here is another one of the same source, but with another brand name. 

https://en.robbshop.nl/wallremote-8v-zigbee-robb-smarrt

corvy:
I think this might be the source: https://sunricher.en.alibaba.com/product/60812056361-0/Sunricher_wireless_Zigbee_3_0_single_color_keypad_smart_switch.html

corvy:
[ZigBee Shenzen Synricher.zip](https://github.com/dresden-elektronik/deconz-rest-plugin/files/3185560/ZigBee.Shenzen.Synricher.zip)

The only information I got from the producer is this. I could send them another question to get more information if you could help me formulate something. I think this is not detailed enough. They only state that they are "ZigBee certified" but still we are not able to join the switch to the gateway.

manup:
They look like the iCasa switches:

https://www.amazon.de/iCasa-Pulse-8S-ZigBee-Keypad/dp/B07L2SLY29

Looks like these are branded version of Sunricher OEM products.

http://www.sunricher.com/zigbee-series/zigbee-remotes.html

@ebaauw added supported for iCasa switches in 2.05.58.

https://github.com/dresden-elektronik/deconz-rest-plugin/releases/tag/V2_05_58

PrimaryInc:
Hi @manup @corvy ,

Perhaps our attempts to pair switch to DeConz has been wrong? 
I've uses pair "other" switch and set switch in pair mode, pairing times out with no switch found.
Any ideas?

manup:
[...]

PR #1525, opened 2019-05-19
Don't poll iCasa Pulse keypads for battery percentage

ebaauw:
It appears these keypads report battery percentage periodically after all, slightly under every four hours. see https://github.com/dresden-elektronik/deconz-rest-plugin/issues/1509#issuecomment-493671362 and https://github.com/dresden-elektronik/deconz-rest-plugin/issues/1124.


manup:
Cool thanks, good catch

PR #1554, opened 2019-05-29
Several small api fixes

KodeCR:
- allow "type" as a resourcelink attribute on creation (even though there is only one option)
- allow schedules and scenes actions in rules
- a schedule is also valid if providing localtime instead of time as a required attribute
- also allow a sensor config as a json dictionary

Together this enables me to set up a Hue dimmer switch and a Hue motion sensor from the Hue app.

### DIFF
--- a/rest_sensors.cpp
+++ b/rest_sensors.cpp
@@ -300,7 +300,7 @@ int DeRestPluginPrivate::createSensor(const ApiRequest &req, ApiResponse &rsp)
 
     for (; pi != pend; ++pi)
     {
-        if(!((pi.key() == "name") || (pi.key() == "modelid") || (pi.key() == "swversion") || (pi.key() == "type")  || (pi.key() == "uniqueid")  || (pi.key() == "manufacturername")  || (pi.key() == "state")  || (pi.key() == "config")))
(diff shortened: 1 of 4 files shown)


rest_sensors.cpp
```diff
@@ -963,6 +958,15 @@ int DeRestPluginPrivate::updateSensor(const ApiRequest &req, ApiResponse &rsp)
         }
     }
 
+    if (map.contains("config")) // optional
```
manup: Not sure if I like this, since other parts of the API don't do this kind of deep-copy and only update respective attributes on the nesting level given by the URL. Does the Hue bridge support this?

KodeCR: Yep, the Hue app sends a {"config": {"usertest": true}} when setting/testing the sensitivity of a Hue motion sensor... I agree it's a bit inconsistent, but it does seem to be supported.

ebaauw: Haven't seen that one before (and, afaik, it's undocumented).  But the Hue API uses nested objects in the body on multiple occasions.  E.g. to reset the HomeKit server, PUT `/config` with a body of `{"homekit": {"factoryreset": true}}`.

KodeCR: Yes, it seems undocumented. As in, usertest is described as an attribute of the config object, but the docs way to change a config is indeed to have the url end in .../config. I only found this by chance when adding the motion sensor.

manup: Interesting, since the Hue bridge uses this we can add it too. Albeit I would prefer supporting just the documented ways or find a generic solution for nested requests (might be difficult).

ebaauw: API v2...

PR #1555, opened 2019-05-29
Implement "group 0" as a regular zigbee group

KodeCR:
This is mostly the first commit from #1471 so it can be reviewed and merged in stages. This:

- creates a (regular Zigbee) group with all the lights in it and translates api "group 0" to this Zigbee group
-- just using Zigbee group id 0 did not work for all lights nor 65535, so it finds a free group from 65520 going downwards
-- removes all "group0" special cases

This should cause not change in behavior. On the Zigbee side each light will now have an extra group, and turning off or on all lights via group 0 on the api side is now achieved on the Zigbee side with a groupcast to this zigbee group instead of a broadcast.

### DIFF
--- a/database.cpp
+++ b/database.cpp
@@ -1040,6 +1040,18 @@ static int sqliteLoadConfigCallback(void *user, int ncols, char **colval , char 
             }
         }
     }
+    else if (strcmp(colval[0], "group0") == 0)
+    {
+        if (!val.isEmpty())
+        {
+            uint group0 = val.toUInt(&ok);
+            if (ok && (group0 <= 65535))
+            {
+                d->gwGroup0 = group0;
+                d->gwConfig["group0"] = (uint16_t)group0;
+            }
+        }
+    }
     else if (strcmp(colval[0], "updatechannel") == 0)
     {
         if ((val == "stable") || (val == "alpha") || (val == "beta"))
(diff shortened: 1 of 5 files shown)


de_web_plugin.cpp
```diff
@@ -325,10 +325,31 @@ DeRestPluginPrivate::DeRestPluginPrivate(QObject *parent) :
     }
 
     // create default group
-    Group group;
-    group.setAddress(0);
-    group.setName("All");
-    groups.push_back(group);
+    if (gwGroup0 == 0) { // get new id and replace old group0 and get new id
```
manup: Please always use braces on the next line for if, for. So that the code base just has one code style.
```
if (gwGroup0)
{
}
```

database.cpp
```diff
@@ -1040,6 +1040,18 @@ static int sqliteLoadConfigCallback(void *user, int ncols, char **colval , char
             }
         }
     }
+    else if (strcmp(colval[0], "group0") == 0)
+    {
+        if (!val.isEmpty())
+        {
+            uint group0 = val.toUInt(&ok);
+            if (ok && (group0 <= 65535))
```
manup: Shouldn't 65535 be 0xFFF0 to align with the later code?

manup:
Thanks, I think this should work. For a later refactor we should consider to switch to a SQL table to map an REST-API group id and unqiueid to a Zigbee group id.

```
id      | uinqueid                    | Zigbee GID
--------+-----------------------------+-----------
0       |                             |  0x7432
1       | d8:5d:ef:11:a1:00:01:76-01  |  0x0001    // did exist prior to api change
2       |                             |  0x0002    // did exist prior to api change
3       |                             |  0xF040    // new random mapping

```

database.cpp
KodeCR: Yes, good catch,

de_web_plugin.cpp
KodeCR: Thanks, missed this one, it's no my usual code style.

KodeCR:
Thanks @manup for reviewing. I'll update the PR tomorrow or the day after, or feel free to make those changes yourself.

manup:
>Thanks @manup for reviewing. I'll update the PR tomorrow or the day after, or feel free to make those changes yourself.

Done :)

Issue #1534, opened 2019-05-21
Support for Develco smoke sensor


SwoopX:
Hi there,

I'm hoping to get the smoke sensor integrated in deconz. Interestingly, it was quite a hassle to have it paired so the clusters become visible but what you can pull out of it seems quite useless (similar to [#1294](https://github.com/dresden-elektronik/deconz-rest-plugin/issues/1294)). However, I managed to find the [Technical manual](https://www.develcoproducts.com/media/3374/smszb-120-technical-manual-smoke-alarm.pdf), hoping that's equally good.

Thanks!

Edit:
![3](https://user-images.githubusercontent.com/4005212/58288951-57f90180-7db5-11e9-8e40-cabd6c4f2a07.jpg)
![1](https://user-images.githubusercontent.com/4005212/58288952-57f90180-7db5-11e9-9c6c-8c0e77dfef9c.jpg)
![2](https://user-images.githubusercontent.com/4005212/58288953-58919800-7db5-11e9-9c5f-123236774144.jpg)

ebaauw:
Interesting, this looks like the first smoke detector whose siren can be activated remotely.  Where did you buy it?

SwoopX:
:) Always looking for interesting stuff. I actually bought it as new at ebay Kleinanzeigen for 40€ after some research regarding zigbee smoke detectors. Just found 2 shops really selling them. The one associated to ubisys in GER (78€) and Cozify in FI (~49€)

b2un0:
You cal also activate the siren on Xiaomi MiJia Honeywell smoke detector

https://github.com/Koenkk/zigbee2mqtt/issues/420

SwoopX:
> you can also activate the siren on Xiaomi MiJia Honeywell smoke detector

Might bentrue but as farmasmI know, those do not have necessary certifications to be officially operated in Europe. The Develco‘s do ;)

ebaauw:
> you can also activate the siren on Xiaomi MiJia Honeywell smoke detector

That's not what I read.  Apparently, they managed to activate the sensor's self-test remotely (causing it to beep), but it doesn't support the _IAS_WD_ cluster.

ma-ca:
The Bitron 902010/24 smoke detector with siren has an _IAS_WD_ cluster that can be used to remotely start the siren.

ebaauw:
[...]

Issue #1021, opened 2018-12-10
Device Integration: Add Heiman HS3SA Smoke Sensor


peer69:
I have receives the Heiman HS3SA Smoke Sensor. Is it possible to add it as a device? 

Basic Cluster: 
<img width="433" alt="image" src="https://user-images.githubusercontent.com/28559804/50969118-9fa17f80-14dd-11e9-8366-4eb6c19d05a5.png">

IAS Zone Attributes: 
<img width="426" alt="image" src="https://user-images.githubusercontent.com/28559804/50969132-ab8d4180-14dd-11e9-9828-d5e96f8112ab.png">

Node Info Panel:
<img width="447" alt="image" src="https://user-images.githubusercontent.com/28559804/50960915-c3a59680-14c6-11e9-8266-cb524a25179f.png">

If needed I will provide the debug output.

saxn-paule:
I also got two of the HS3SA-E smoke detectors. Pairing via Phoscon was possible, but it's always shown as not connected:
![image](https://user-images.githubusercontent.com/8352690/133047763-9c883679-8393-4956-8074-dda8c9c0a500.png)

Via Home-Assistant three entities are detected:
![image](https://user-images.githubusercontent.com/8352690/133048028-6b03ddfe-156e-414e-9c32-79ea85ae1801.png)

But they don't provide a status. In case of testing the alarm function nothing happens in Phoscon/HA.

Mimiix:
@peer69 Does this device work correctly for you?

peer69:
@Mimiix : Phoscon looks the same but the GUI shows connected and API also returns values that seem to be correct. So, yes. As much as its supposed to, it is working.

PR #1593, opened 2019-06-14
Add siren support to smoke sensors

ma-ca:
Add siren support to Bitron Smoke Detector with siren.

Add Develco smoke sensor to supported devices list, issue #1534.

Add siren support to Develco smoke detector.

A smoke detector has device id 0x0402 `WD Zone`. Therefore, the smoke detector was not added to type `Warning device` (device id 0x0403). But the smoke detector does have the clusters `IAS Zone (0x0500)` and `IAS WD (0x0502)`. This PR adds device type `Warning device` whenever a `IAD WD` cluster is found. 

The following REST API calls are now supported on Bitron smoke sensors and on (untested) Develco smoke sensors in order to start or stop the siren.
```
/api/<apikey>/lights/<id>/state -X PUT -d '{ "alert": "select" }'
```
```
/api/<apikey>/lights/<id>/state -X PUT -d '{ "alert": "lselect" }'
```
```
/api/<apikey>/lights/<id>/state -X PUT -d '{ "alert": "none" }'
```

```
{
    "etag": "8dc0ac55f158350cf99ed489e23fb2c1",
    "hascolor": false,
    "manufacturername": "Bitron Home",
    "modelid": "902010/24",
    "name": "Light 32",
    "state": {
        "alert": "none",
        "reachable": true
    },
    "swversion": "20150615",
    "type": "Warning device",
    "uniqueid": "00:12:4b:00:07:7f:13:44-01"
}
```
Note: The smoke detector is a battery powered end device. Before sending a command to the smoke detector, a lightNode->rx() is necessary otherwise deCONZ did mark the device as zombie and dropped the sending command.

### DIFF
--- a/de_web_plugin.cpp
+++ b/de_web_plugin.cpp
@@ -72,6 +72,7 @@ const quint64 tiMacPrefix         = 0x00124b0000000000ULL;
 const quint64 netvoxMacPrefix     = 0x00137a0000000000ULL;
 const quint64 boschMacPrefix      = 0x00155f0000000000ULL;
 const quint64 jennicMacPrefix     = 0x00158d0000000000ULL;
+const quint64 develcoMacPrefix    = 0x0015bc0000000000ULL;
 const quint64 philipsMacPrefix    = 0x0017880000000000ULL;
 const quint64 ubisysMacPrefix     = 0x001fee0000000000ULL;
 const quint64 deMacPrefix         = 0x00212e0000000000ULL;
@@ -195,6 +196,7 @@ static const SupportedDevice supportedDevices[] = {
     { VENDOR_JENNIC, "SPZB0001", jennicMacPrefix }, // Eurotronic thermostat
     { VENDOR_NONE, "RES001", tiMacPrefix }, // Hubitat environment sensor, see #1308
     { VENDOR_119C, "WL4200S", sinopeMacPrefix}, // Sinope water sensor
+    { VENDOR_DEVELCO, "SMSZB-120", develcoMacPrefix }, // Develco smoke sensor
     { 0, nullptr, 0 }
 };
 
--- a/de_web_plugin.cpp
+++ b/de_web_plugin.cpp
@@ -1377,6 +1377,7 @@ void DeRestPluginPrivate::addLightNode(const deCONZ::Node *node)
         node->nodeDescriptor().manufacturerCode() == VENDOR_JENNIC || // Xiaomi lumi.ctrl_neutral1, lumi.ctrl_neutral2
         node->nodeDescriptor().manufacturerCode() == VENDOR_EMBER || // atsmart Z6-03 switch
         node->nodeDescriptor().manufacturerCode() == VENDOR_NONE || // Climax Siren
+        node->nodeDescriptor().manufacturerCode() == VENDOR_DEVELCO || // Develco Smoke sensor with siren
         node->nodeDescriptor().manufacturerCode() == VENDOR_1233) // Third Reality smart light switch
     {
         // whitelist
--- a/de_web_plugin.cpp
+++ b/de_web_plugin.cpp
@@ -1392,6 +1392,7 @@ void DeRestPluginPrivate::addLightNode(const deCONZ::Node *node)
         bool hasServerOnOff = false;
         bool hasServerLevel = false;
         bool hasServerColor = false;
+        bool hasIASWDCluster = false;
 
         for (int c = 0; c < i->inClusters().size(); c++)
         {
@@ -1399,6 +1400,7 @@ void DeRestPluginPrivate::addLightNode(const deCONZ::Node *node)
             else if (i->inClusters()[c].id() == LEVEL_CLUSTER_ID) { hasServerLevel = true; }
             else if (i->inClusters()[c].id() == COLOR_CLUSTER_ID) { hasServerColor = true; }
             else if (i->inClusters()[c].id() == WINDOW_COVERING_CLUSTER_ID) { hasServerOnOff = true; }
+            else if (i->inClusters()[c].id() == IAS_WD_CLUSTER_ID) { hasIASWDCluster = true; }
         }
 
         // check if node already exist
@@ -1583,6 +1585,15 @@ void DeRestPluginPrivate::addLightNode(const deCONZ::Node *node)
                         {
                             // Xiaomi wall switch lumi.ctrl_neutral1, lumi.ctrl_neutral2
                             // TODO exclude endpoint 0x03 for lumi.ctrl_neutral1
+                            lightNode.setHaEndpoint(*i);
+                        }
+                    }
+                    break;
+
+                case DEV_ID_IAS_ZONE:
+                    {
+                        if (hasIASWDCluster)
+                        {
                             lightNode.setHaEndpoint(*i);
                         }
                     }
@@ -2080,6 +2091,7 @@ LightNode *DeRestPluginPrivate::updateLightNode(const deCONZ::NodeEvent &event)
             case DEV_ID_HA_WINDOW_COVERING_DEVICE:
             case DEV_ID_ZLL_ONOFF_SENSOR:
             case DEV_ID_XIAOMI_SMART_PLUG:
+            case DEV_ID_IAS_ZONE:
             case DEV_ID_IAS_WARNING_DEVICE:
             case DEV_ID_FAN:
                 break;
--- a/de_web_plugin_private.h
+++ b/de_web_plugin_private.h
@@ -251,6 +251,7 @@
 #define VENDOR_PHILIPS      0x100B // Also used by iCasa routers
 #define VENDOR_VISONIC      0x1011
 #define VENDOR_ATMEL        0x1014
+#define VENDOR_DEVELCO      0x1015
 #define VENDOR_LGE          0x102E
 #define VENDOR_JENNIC       0x1037 // Used by Xiaomi, Trust, Eurotronic
 #define VENDOR_CENTRALITE   0x104E
--- a/light_node.cpp
+++ b/light_node.cpp
@@ -476,6 +476,11 @@ void LightNode::setHaEndpoint(const deCONZ::SimpleDescriptor &endpoint)
                 else if (i->id() == FAN_CONTROL_CLUSTER_ID) {
                     addItem(DataTypeUInt8, RStateSpeed);
                 }
+                else if (i->id() == IAS_WD_CLUSTER_ID)
+                {
+                    removeItem(RStateOn);
+                    ltype = QLatin1String("Warning device");
+                }
             }
         }
 
--- a/light_node.cpp
+++ b/light_node.cpp
@@ -478,8 +478,12 @@ void LightNode::setHaEndpoint(const deCONZ::SimpleDescriptor &endpoint)
                 }
                 else if (i->id() == IAS_WD_CLUSTER_ID)
                 {
-                    removeItem(RStateOn);
-                    ltype = QLatin1String("Warning device");
+                    if (    modelId() == QLatin1String("902010/24") ||   // Bitron Smoke Detector with siren
+                            modelId() == QLatin1String("SMSZB-120"))     // Develco Smoke Alarm with siren
+                    {
+                        removeItem(RStateOn);
+                        ltype = QLatin1String("Warning device");
+                    }
                 }
             }
         }
--- a/rest_lights.cpp
+++ b/rest_lights.cpp
@@ -1295,7 +1295,7 @@ int DeRestPluginPrivate::setLightState(const ApiRequest &req, ApiResponse &rsp)
 
         bool isSmokeDetector = false;
         if (taskRef.lightNode->modelId() == QLatin1String("902010/24")  // Bitron Smoke Detector with siren
-                )
+                || taskRef.lightNode->modelId() == QLatin1String("SMSZB-120"))  // Develco smoke sensor
         {
             isSmokeDetector = true;
             taskRef.lightNode->rx();  // otherwise device is marked as zombie and task is dropped
--- a/rest_lights.cpp
+++ b/rest_lights.cpp
@@ -1293,6 +1293,14 @@ int DeRestPluginPrivate::setLightState(const ApiRequest &req, ApiResponse &rsp)
         QString alert = map["alert"].toString();
         bool isWarningDevice = taskRef.lightNode->type() == QLatin1String("Warning device");
 
+        bool isSmokeDetector = false;
+        if (taskRef.lightNode->modelId() == QLatin1String("902010/24")  // Bitron Smoke Detector with siren
+                )
+        {
+            isSmokeDetector = true;
+            taskRef.lightNode->rx();  // otherwise device is marked as zombie and task is dropped
+        }
+
         if (alert == "none")
         {
             if (isWarningDevice)
@@ -1309,7 +1317,13 @@ int DeRestPluginPrivate::setLightState(const ApiRequest &req, ApiResponse &rsp)
         }
         else if (alert == "select")
         {
-            if (isWarningDevice)
+            if (isWarningDevice && isSmokeDetector)
+            {
+                task.taskType = TaskWarning;
+                task.options = 0x12; // Warning mode 2 (fire), strobe
+                task.duration = 1;
+            }
+            else if (isWarningDevice)
             {
                 task.taskType = TaskWarning;
                 task.options = 0x14; // Warning mode 1 (burglar), Strobe
@@ -1323,7 +1337,13 @@ int DeRestPluginPrivate::setLightState(const ApiRequest &req, ApiResponse &rsp)
         }
         else if (alert == "lselect")
         {
-            if (isWarningDevice)
+            if (isWarningDevice && isSmokeDetector)
+            {
+                task.taskType = TaskWarning;
+                task.options = 0x12; // Warning mode 2 (fire), strobe
+                task.duration = 300;
+            }
+            else if (isWarningDevice)
             {
                 task.taskType = TaskWarning;
                 task.options = 0x14; // Warning mode 1 (burglar), Strobe


ebaauw:
> This PR adds device type Warning device whenever a IAD WD cluster is found.

That win’t do, you need to whitelist the sensors.  The Heiman smoke and CO sensors advertise an _IAS WD_ cluster, but it doesn’t do anything.  See #404.

ma-ca:
> That won’t do, you need to whitelist the sensors.

Why not blacklist the Heiman sensors instead? If those have a `IAS WD` cluster that should not appear in `/lights` it seems more appropriate to only filter out the Heiman sensors. Assuming that all other smoke sensors with siren and a `IAS WD` cluster work correctly.

ebaauw:
> Assuming that all other smoke sensors with siren and a IAS WD cluster work correctly.

I'm afraid a working _IAS WD_ cluster in a smoke/fire or gas sensor could be the exception, but blacklisting the Heiman sensors should also work.  I no longer have these devices, but if I read my notes correctly, only the (old model) fire sensor has this cluster:
https://github.com/dresden-elektronik/deconz-rest-plugin/blob/aa223512b0fe84c9a48ffec703f9885d17749800/de_web_plugin.cpp#L175
I don't know about the new model (#1021):
https://github.com/dresden-elektronik/deconz-rest-plugin/blob/aa223512b0fe84c9a48ffec703f9885d17749800/de_web_plugin.cpp#L181

When adding support for Heiman, I found it easier to whitelist the Heiman Siren (using its device type), rather than blacklisting the devices with a nonfunctional _IAS WD_ cluster.   Battery powered devices (or rather: devices where _Receiver On When Idle_ is false) need to be whitelisted for `/lights` resources anyway.

ma-ca:
> When adding support for Heiman, I found it easier to whitelist the Heiman Siren (using its device type), rather than blacklisting the devices with a nonfunctional IAS WD cluster. Battery powered devices (or rather: devices where Receiver On When Idle is false) need to be whitelisted for /lights resources anyway.

Right. The Bitron Smoke detector was whitelisted already (`VENDOR_NONE`). So yes, I agree, whitelisting at three places is more streamline than mixing both whitelist and blacklist.

For the record, I had to whitelist the smoke detectors in de_web_plugin.cpp, light_node.cpp and rest_lights.cpp in order to expose their siren in the `/lights` resource.

ebaauw:
Thanks.

> For the record, I had to whitelist the smoke detectors in de_web_plugin.cpp, light_node.cpp and rest_lights.cpp in order to expose their siren in the /lights resource.

I know, there's always one more place where you have to whitelist new devices.  I think we need to model the various capabilities a device can have, and only check for these capabilities in the code, instead of checking for a specific device.  Then we only need to whitelist a device (with all its capabilities) once.   If we were to use a JSON (or even XML) file for that, we should be able to add new devices (with already supported capabilities) without any coding, much like adding clusters and attributes to `general.xml`.

manup:
>I think we need to model the various capabilities a device can have, and only check for these capabilities in the code, instead of checking for a specific device. Then we only need to whitelist a device (with all its capabilities) once. If we were to use a JSON (or even XML) file for that, we should be able to add new devices (with already supported capabilities) without any coding, much like adding clusters and attributes to general.xml.

I like that capabilities approach a lot. The more devices we add it becomes clear that at least large parts of the code could be simplified and supported by such "static" data about a device. 

Is this PR ready for merging? I'm currently crafting the 2.05.66 release (...weeks after my ETA :/ )

ma-ca:
> Is this PR ready for merging? 

Yes.

SwoopX:
@ma-ca Many thanks for considering the Develco device. I'd be more than happy to support in testing, once I can free some time within the next couple of days!

SwoopX:
@manup Are ma-ca's changes already part of .66 beta release?

Anyhow, I manually compiled the plugin with ma-ca's changes an had some rather frustrating tests. Frustration mainly come from the pairing attempts, but I'd like to move that to the end to stick to the commits.

So I checked if a lights node has been created and that was obviously the case

`curl http://127.0.0.1/api/1A12C1474B/lights`

`{"1":{"etag":"81984206006d09dc6683ed492b07eb27","hascolor":false,"manufacturername":"Develco Products A/S","modelid":"SMSZB-120","name":"Light 1","state":{"alert":"none","on":false,"reachable":true},"swversion":"2016-07-05 11:26","type":"Unknown","uniqueid":"00:15:bc:00:31:00:03:2f-23"}}`

Additionally, a sensors node has been created but not shown in Phoscon

`curl http://127.0.0.1/api/1A12C1474B/sensors`

`{"1":{"config":{"configured":false,"on":true,"sunriseoffset":30,"sunsetoffset":-30},"etag":"1666aba7e5bf603019f0e5941cd91c67","manufacturername":"Philips","modelid":"PHDL00","name":"Daylight","state":{"dark":null,"daylight":null,"lastupdated":"none","status":null},"swversion":"1.0","type":"Daylight","uniqueid":"00:21:2e:ff:ff:02:2f:c7-01"},"2":{"config":{"battery":100,"on":true,"reachable":true},"ep":35,"etag":"a8ef1ff4b21947729c52bb2862e4fd6b","manufacturername":"Develco Products A/S","modelid":"SMSZB-120","name":"SMSZB-120","state":{"alarm":false,"lastupdated":"2019-06-18T19:36:45","lowbattery":false,"tampered":false},"swversion":"2016-07-05 11:26","type":"ZHAAlarm","uniqueid":"00:15:bc:00:31:00:03:2f-23-0500"}}`

I tried to check the the siren, but although I received a success message, nothing happened

`curl http://127.0.0.1/api/1A12C1474B/lights/1/state -X PUT -d '{ "alert": "select" }'`

`[{"success":{"/lights/1/state/alert":"select"}}]`

`curl http://127.0.0.1/api/1A12C1474B/lights/1/state -X PUT -d '{ "alert": "lselect" }'`

`[{"success":{"/lights/1/state/alert":"lselect"}}]`

So, based on that, no luck with the changes. Now to my "alien experience" with might be connected. It was really a nightmare to have the device join and strangely only worked in 3 stages. Initial join detected the device (in grey as sensor), but with no clusters at all (2nd circle in the GUI was missing) and with a manufacturer code of 0x0000. Trying the read the simple clusters had no effect unfortunately.
Fiddling around, doing resets, restarts and whatnot, clusters came in, but only 1/3 (IAS Zone). However, info pane for each and every cluster was empty.
After further fiddling, the device turned from grey to orange and the clusters were finally populated with data in the info pane. Interestingly, the manufacturer code was and still is 0x1135 (Dresden Elektronik). While reading the basic cluster in deconz, the manufacturer was at least updated in Phoscon and the database. What could be the reason for all this?

Due to that challenges, I started checking the code to eventually get an idea where that might come from. I stumbled over de_web_plugin_private.h, where beginning on line 366 in the declaration and the subsequent case instruction further amendments might be required?

Happy to read your thoughts.

ebaauw:
> I tried to check the the siren, but although I received a success message, nothing happened

Did you check in the deCONZ GUI that the sensor actually supports sounding the alarm over Zigbee?  As mentioned above, the Heiman smoke sensor exposes the _IAS WD_ cluster, but it doesn't react to any commands sent to it.

Go to the _Cluster Info_ panel, select the _IAS WD_ cluster of the sensor, select _Warning Mode 0_, set _Warning Duration_ to 1 and press _Exec_.  If it doesn't react, try a different _Warning Mode_.
Screenshot 2019-06-18 at 23 15

SwoopX:
Will do so tomorrow afternoon, thanks!

SwoopX:
Just checked it. The device only reacts on Siren Level 0 and 1 and Strobe...

ebaauw:
Joy, we need a different `task.options` value per device type.  More whitelisting :-(

@ma-ca, I think value 0x12 (0b 0001 00 10) is actually  _Warning mode_ 1 (burglar), _No strobe_  and _Siren level_ 2 (high).  _Warning mode_ 2 (fire) and _Strobe_ would be 0b 0010 01 00 or 0x24.

_Strobe_ and _Siren Level 0_  and _Siren Level 1_ (actually _Siren Level_ 3 (very high level sound)) would be 0b 0000 01 11 or 0x07.

The attribute is defined as a 8-bit bitmap (`bmp8`), but it's value is actually three subfields: a 4-bit enum _Warning mode_, a 2-bit enum _Strobe_, and a 2-bit enum _Siren Level_.  The deCONZ GUI doesn't support these subfields, so I mapped each bit separately (_Strobe_ only takes value 0 and 1), so bit 3 is missing from `general.xml`.

SwoopX:
Just to be clear here, I didn't mean the combination but each value individually. Sorry if that was misleading. So only Siren Level 0 -> Audio, only Siren Level 1 -> Audio and only Strobe -> Audio

ebaauw:
Ok.  This is kinda odd, because not setting any of the _Warning mode_ bits would normally stop the alarm.

ma-ca:
I tried different `task.options` and found that the Bitron smoke detector siren did only start with values `0x02` and `0x12`. The latter did also start the blinking red LED. 

I believe the lower 4 Bits are the Warning Mode.

image

ebaauw:
> I believe the lower 4 Bits are the Warning Mode.

I read this as the high 4 bits.  This is consistent with how the Heiman siren reacts, e.g. when I set just bit 2, 0x04, the Heiman siren activates the strobing warning light, but not the siren.  See also the code for `”alert”: “blink”`.  I have no experience with any other sirens or smoke sensors with a siren that can be activated remotely.

Anyway, I’m not arguing which value to send to the Bitron, and by the looks of it, the Develco seems to look at the low 4 bits as well.  Not sure why it doesn’t react to the API.

SwoopX:
> @ma-ca, I think value 0x12 (0b 0001 00 10) is actually Warning mode 1 (burglar), No strobe and Siren level 2 (high). Warning mode 2 (fire) and Strobe would be 0b 0010 01 00 or 0x24.

Jep, verified that based on the sniffer results. But that's details. I issued 0x12 via deconz GUI and it did trigger audio.

I also noted that deconz GUI might construct the attribute wrongly. Based on the PCAP and ZCL documentation...
image
...the attribute is MSB. The PCAP showed 2 valid values for warning mode 0 and 1, and unknown for warning mode 2 and 3. That as a side note.

The 1M questions for me are why is the device so extraordinary stubborn while joining and why is nothing happening via REST API. Also the sensors never show up in Phoscon :(

ebaauw:
> ZCL documentation

Which documentation is this?

> I also noted that deconz GUI might construct the attribute wrongly.

As I said before, the GUI (or rather `general.xml`) uses the subfield order used by the Heiman siren, where _Warning Mode_ is bits 4-7, _Strobe_ is bits 2-3, and _Siren Level_ is bits 0-1.  I left out bit 3, since it isn't a valid _Strobe_ value.

I don't even care which Lilliputian faction is right here, I just hate that, apparently, there's room for different interpretations by different devices, resulting in us having to whitelist which device uses which interpretation.  I think we best change `general.xml` to show the politically correct _Bit 0_, ..., _Bit 7_.

ebaauw:
BTW, WireShark seems to agree with Heiman:
```
IEEE 802.15.4 Data, Dst: 0x119a, Src: 0xaed2
ZigBee Network Layer Data, Dst: 0x119a, Src: 0x0000
ZigBee Application Support Layer Data, Dst Endpt: 1, Src Endpt: 1
ZigBee Cluster Library Frame
    Frame Control Field: Cluster-specific (0x01)
    Sequence Number: 144
    Command: Start Warning (0x00)
    Payload
        0000 .... = Warning Mode: Stop (no warning) (0)
        .... 01.. = Strobe: Use strobe in parallel to warning (1)
        Warning Duration: 0x0005
```

ma-ca:
> and why is nothing happening via REST API. 

The Develco technical reference for the smoke alarm states that the `IAS WD command start warning` only has warning modes with values 0 (Siren off) and 1 (Siren on). 

image

ebaauw:
> The Develco technical reference for the smoke alarm states that the IAS WD command start warning only has warning modes with values 0 (Siren off) and 1 (Siren on).

So we would have to send 0x01 to start the alarm, and 0x00 to stop it.

That's not entirely in line with @SwoopX's observations, though:

> Just to be clear here, I didn't mean the combination but each value individually. Sorry if that was misleading. So only Siren Level 0 -> Audio, only Siren Level 1 -> Audio and only Strobe -> Audio

This would suggest that 0x01, 0x02, and 0x04 all work.

> I issued 0x12 via deconz GUI and it did trigger audio.

Are you sure?  You would have to select _Warning Mode 0_ and _Siren Level 1_ for that.  What duration did you use?  Can you sniff the message sent by the REST API and double-check that you send the same message from the GUI?

> why is the device so extraordinary stubborn while joining

Probably because it's battery powered and it's radio is off most of the time.  Search for devices in the Phoscon app, press the test button (cover your ears first ;-) and read the _Basic_ cluster attributes in the GUI.  If the endpoints don't show: select _Read Node Descriptor_ and _Read Simple Descriptor(s)_ from the dropdown for the left of the two dots in the GUI.  You might want to run the sniffer while doing that, to see if the sensor responds.

SwoopX:
> ZCL documentation
Which documentation is this?

https://www.nxp.com/docs/en/user-guide/JN-UG-3115.pdf#page882

> As I said before, the GUI (or rather general.xml) uses the subfield order used by the Heiman siren, where Warning Mode is bits 4-7, Strobe is bits 2-3, and Siren Level is bits 0-1. I left out bit 3, since it isn't a valid Strobe value.

That‘s perfectly fine with bit 3, as it is reserved. I don‘t wanna argue here, as everything is working from deconz GUI perspective. I just want to understand how things are working there and it could very well be that I misunderstand. I‘ll explain this evening in more detail what I was after, when I cann acces my PC and then we leave it at that 🙂

Btw, I now know why the API call didn‘t trigger anything: the type is still set to **unknown**, but should be **warning device**.

@ma-ca: saw that in the technical reference as well and was puzzled  they refer to byte instead of any bit. I would have interpreted it as 0b00000000 for siren off and 0xffffffff for siren on bit that doesn‘t fit in at all. The test showed different results.

SwoopX:
Ok, this is going to be some kind of fairytale but allow me to be more precise on one of my previous statements:

> I also noted that deconz GUI might construct the attribute wrongly.

 That was a little bit too much. For reference, my interpretation of the 'start warning' payload:

image

The above quote came from not fully understanding the checkboxes in deconz GUI. At that time, I thought WD 3 checkbox would mean "3 - Emergency". Based on the PCAPs, Wireshark told me it was "8 - Unknown". Since, according to documentation, any integer values higher than 6 for the warning nibble are reserved, one could spare that particular bit in general.xml like it was done with with the bit (3) for strobe. A corresponding checkbox is also not provided. Not sure if leaving bit 7 would go through 1:1 but WD 3 checkbox would not be required in that sense. So, just an observation which doesn't cause any harm (but confused me :) ).

> The Develco technical reference for the smoke alarm states that the IAS WD command start warning only has warning modes with values 0 (Siren off) and 1 (Siren on).
> 
> So we would have to send 0x01 to start the alarm, and 0x00 to stop it.
> 
> That's not entirely in line with @SwoopX's observations, though:
> 
> Just to be clear here, I didn't mean the combination but each value individually. Sorry if that was misleading. So only Siren Level 0 -> Audio, only Siren Level 1 -> Audio and only Strobe -> Audio
> 
> This would suggest that 0x01, 0x02, and 0x04 all work.

Agreed, that's a bit weird. Maybe some inaccurate documentation... What's indeed working is 0x01, 0x02, and 0x04.

> Are you sure? You would have to select Warning Mode 0 and Siren Level 1 for that. What duration did you use? Can you sniff the message sent by the REST API and double-check that you send the same message from the GUI?

I am, just double-checked. The sniffer output is below. I cannot, however, sniff anything that comes from the API since it is not working at all. And that's the real issue I have here driving me nuts. In the database, the device still has type "unknown". In Phoscon, all I see from the device is the light bulp with the manufacturer info, nothing else. I would have also expected to see the temperature sensor in Phoscon, but there's nothing

image

ebaauw:
> Btw, I now know why the API call didn‘t trigger anything: the type is still set to unknown, but should be warning device.

That would explain the observed behaviour.

Looking at the code, the type would be set based on the _Model Identifier_ from the _Basic_ cluster which hasn't been read yet when initialising the light node.  Better not whitelist on _Model Identifier_ this early when adding a light.  Trying to delay the initialisation doesn't work out well, see #1616.

I think it's probably safe just to check for `IAS_WD_CLUSTER_ID`, or, cleaner, for `DEV_ID_IAS_ZONE`, as I've done for the Hieman's `DEV_ID_IAS_WARNING_DEVICE`.

> I would have also expected to see the temperature sensor in Phoscon, but there's nothing

Is the ZHATemperature `/sensors` resource created in the REST API?  If not, double-check that the _Basic_ cluster on the _Temperature Sensor_ endpoint 0x26 shows the same values as the _IAS Zone_ endpoint 0x23.  You might need to repeat the read while pressing the test button procedure for this endpoint.

Phoscon is not part of this repo (it's closed source), so I cannot help you there.  I'm not sure if temperature sensors need to be whitelisted there, but I wouldn't be surprised.

SwoopX:
**Bullseye!!**

@ebaauw First quick feedback: API call working now, thanks a lot!!

spookoly:
Apologies if this is too basic a question for this thread. I've just bought a Bitron smoke detector and was hoping to independently control the siren. However having paired the device as a sensor I don't see a light added in the REST API. Do I have to do something special for the corresponding dummy light to be created? I've tried restarting the hub, and removing and re-pairing the device. 
Cheers.

ebaauw:
@SwoopX sorry, I missed this earlier, but there's an issue: the ZHAAlarm and ZHAFire sensor have the same `uniqueid` value.  The ZHAAlarm shouldn't have been created at all, and all its attributes should be exposed through the ZHAFire.  The ZHAAlarm type is only there to support standalone sirens (_Zone Type_ value _Standard Warning Device_).

@spookoly Did you compile the REST API plugin yourself from the latest commit?  Otherwise you'll have to wait for the next release of deCONZ for this PR to be included.

spookoly:
@ebaauw I don't think so, I'm running v2_05_66 of deconz and v1.16.0 of the api. So I'll wait for the next release. Many thanks for all your help.

SwoopX:
> **@SwoopX** sorry, I missed this earlier, but there's an issue: the ZHAAlarm and ZHAFire sensor have the same uniqueid value. The ZHAAlarm shouldn't have been created at all, and all its attributes should be exposed through the ZHAFire. The ZHAAlarm type is only there to support standalone sirens (Zone Type value Standard Warning Device).

@ebaauw thanks for highlighting. I'm recognized that as well and was wondering but didn't pay any more attention to it as I was too obsessed with the temperature reporting not working at all. I've already forked the repo and did some changes which didn't do the trick so far. However, I just saw today (with a conbee instead of my 2531 as sniffer) that the sensor gets hammered with requests on the SW_BUILD... I'm about to slightly modify the poll manager to get rid of that, maybe that is somehow related. I'll check out the ZHAAlarm after that. Presumably explicit whitelist for Fire only?

spookoly:
I'm having trouble getting the siren of my bitron smoke detector to be independently controllable. The device is successfully added to the API and registered as both a sensor and a light:

In sensors:

"43": {
        "config": {
            "on": true,
            "reachable": true
        },
        "ep": 1,
        "etag": "9adefe0dd09dad62d14e467f7b39a22a",
        "manufacturername": "Bitron Video",
        "modelid": "902010/24A",
        "name": "902010/24A",
        "state": {
            "fire": false,
            "lastupdated": "2019-09-18T19:41:36",
            "lowbattery": false,
            "tampered": false
        },
        "swversion": "20170421",
        "type": "ZHAFire",
        "uniqueid": "00:12:4b:00:16:81:76:a6-01-0500"
    },

and in lights:

"26": {
        "etag": "8b3f49f4ed73809338ad084d8347573a",
        "hascolor": false,
        "manufacturername": "Bitron Video",
        "modelid": "902010/24A",
        "name": "Light 26",
        "state": {
            "alert": "none",
            "reachable": true
        },
        "swversion": "20170421",
        "type": "Warning device",
        "uniqueid": "00:12:4b:00:16:81:76:a6-01"
    }

However when i try and post the API call changing the alert state:

/api/<apikey>/lights/<id>/state -X PUT -d '{ "alert": "select" }'

I receive a success message but no siren:

{
        "success": {
            "/lights/26/state/alert": "lselect"
        }

Similarly trying to toggle the dummy light on and off has no effect. I've tried removing and adding the device several times, and have confirmed that in the cluster Info panel setting the the IAS WD cluster of the sensor to siren level 0 causes the siren to turn on. 

Not sure what I'm missing here, and any help is very greatly appreciated.

Many thanks.

ma-ca:
The modelid "902010/24A" is not whitelisted, but modelid "902010/24" (without A) is whitelisted.

https://github.com/dresden-elektronik/deconz-rest-plugin/blob/65db7989b6d44c722d251fbdad2b53471b331753/light_node.cpp#L475

https://github.com/dresden-elektronik/deconz-rest-plugin/blob/65db7989b6d44c722d251fbdad2b53471b331753/rest_lights.cpp#L1297

spookoly:
Drat. I should have paid more attention when I bought it. Should I create a PR to whitelist it? I've never made one before but can give it a go. Or can I modify the relevant files locally? Thanks again for your help.

kptkip:
Hi folks!

Is there a reason why "902010/24A" is not whitelisted and could it be whitelisted in future?

Thanks for response.

Issue #1606, opened 2019-06-19
Ikea OTA download


olivertheobald:
I am using the deconz docker container and i dont want to install python on my host, or inside the container. so i wrote a little shell script, which dowloads the ikea firmware files.
Should i create a pull request? Probably we could put the additional scripts inside a tools folder?
jq is already installed in the docker image.

```sh
#!/bin/bash

dir=$PWD

curl 'http://fw.ota.homesmart.ikea.net/feed/version_info.json' | jq -r '[.[] | .fw_binary_url] | .[]'  > $dir/files.txt

cd /root/otau && xargs -n 1 curl -O < $dir/files.txt

rm $dir/files.txt
```

manup:
Cool, always nice to see use of `jq` :)

I think it can be put in `https://github.com/dresden-elektronik/deconz-rest-plugin/tree/master/linux/deCONZ/usr/bin`

One suggestion `cd /root/otau` could be more genral as `cd $HOME/otau`.

PR #1608, opened 2019-06-19
Add Ikea OTA download with curl and jq

olivertheobald:
Ikea OTA download script with curl and jq. No Python needed.

Script is also tested in the [Deconz Docker image](https://hub.docker.com/r/marthoc/deconz/)

See [1606](https://github.com/dresden-elektronik/deconz-rest-plugin/issues/1606)

### DIFF
--- a/linux/deCONZ/usr/bin/deCONZ-ikea-ota-download.sh
+++ b/linux/deCONZ/usr/bin/deCONZ-ikea-ota-download.sh
@@ -1,0 +1,7 @@
+#!/bin/bash
+
+curl 'http://fw.ota.homesmart.ikea.net/feed/version_info.json' | jq -r '[.[] | .fw_binary_url] | .[]'  > $HOME/files.txt
+
+cd $HOME/otau && xargs -n 1 curl -O < $HOME/files.txt
+
+rm $HOME/files.txt


linux/deCONZ/usr/bin/deCONZ-ikea-ota-download.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+
+curl 'http://fw.ota.homesmart.ikea.net/feed/version_info.json' | jq -r '[.[] | .fw_binary_url] | .[]'  > $HOME/files.txt
```
bjorne: Nice addition. It seems this URL works with HTTPS as well, which would be preferable if it works, right?

bjorne: I see now that the firmware filenames include `.signed` and figured there might be other mechanisms to prevent mitm attacks etc., so perhaps I was a bit overzealous in my previous comment.

olivertheobald: > Nice addition. It seems this URL works with HTTPS as well, which would be preferable if it works, right?

You're right. https works as well. I just did a copy and paste without thinking about it :-).

bjorne: What I discovered is that the response JSON refers to additional resources served over (unsecured) HTTP, so it might actually make the most sense to keep the same protocol for the first request as well to avoid a false sense of security? (This would be inline with how Debian and others go about this, trusting the signatures of packages rather than the transfer mechanism to ensure integrity...)

olivertheobald: No problem. I reseted the changes.

janpse:
I'm running deconz in docker and everytime I restart the otau folder is removed. Any ideas why?

olivertheobald:
> I'm running deconz in docker and everytime I restart the otau folder is removed. Any ideas why?

This is a normal docker behaviour. You can mount the /root/otau folder to your local file system, or you can use volumes. I use volumes most of the time

Part of the docker-compose file
```
volumes:
  deconz:
    driver: local
  otau:
    driver: local

#Using volumes or -v flag
    volumes:
      - deconz:/root/.local/share/dresden-elektronik/deCONZ
      - otau:/root/otau

```

janpse:
Thank you!

captaindyson:
How exactly do you go about running this script? I am running deconz in docker, and the file does not exist in the usr/bin dir of the container. Running the lines manualy has the shell reporting 'jq: not found'. What am I missing?

olivertheobald:
> How exactly do you go about running this script? I am running deconz in docker, and the file does not exist in the usr/bin dir of the container. Running the lines manualy has the shell reporting 'jq: not found'. What am I missing?

Didnt run deconz for a long time. Which docker image are you using?

PR #1637, opened 2019-06-28
Fix #1616

### LINKED ISSUE #1616: state.bri missing for dimmable light since 2.05.66
I have a Philips LWB010 setup in deConz (which is connected to OpenHab - I am using it as a wake up light). 

Since 2.05.66 when I look at the state through postman the state looks like this (notice missing "bri"):
```json
   "state": {
        "alert": "none",
        "on": true,
        "reachable": true
    }
```

Since I am using deConz in docker I quickly siwtched to version 2.05.65 and realized that the state there is (notice available "bri"):
```json
    "state": {
        "alert": "none",
        "bri": 10,
        "on": true,
        "reachable": true
    },
```

It seems that the state is set on the light (at least it gets brighter when changing it ;-)), but it just does not get reported.

ebaauw:
See #1616.

### DIFF
--- a/light_node.cpp
+++ b/light_node.cpp
@@ -344,16 +344,6 @@ void LightNode::setHaEndpoint(const deCONZ::SimpleDescriptor &endpoint)
         isInitialized = item(RStateColorMode) != nullptr;
     }
 
-    if (manufacturerCode() == VENDOR_PHILIPS && endpoint.deviceId() == DEV_ID_HA_ONOFF_LIGHT)
-    {
-        // This could be an iCasa in-wall switch
-        if (modelId().isEmpty())
-        {
-            return; // wait until known
-        }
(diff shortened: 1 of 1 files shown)


light_node.cpp
```diff
@@ -344,16 +344,6 @@ void LightNode::setHaEndpoint(const deCONZ::SimpleDescriptor &endpoint)
         isInitialized = item(RStateColorMode) != nullptr;
     }
 
-    if (manufacturerCode() == VENDOR_PHILIPS && endpoint.deviceId() == DEV_ID_HA_ONOFF_LIGHT)
```
manup: Note to myself: We need a public database of known devices descriptors, clusters and endpoints.
For example user can opt-in to export this data to an public endpoint which fills the database.

Issue #1667, opened 2019-07-10
New Device Request: Develco magnetic switch w/temperature sensor


pivot69:
I am trying to get this Develco magnetic switch to work and I have managed to get it included in the network.

Here's is all the info I could get:

<img width="189" alt="image" src="https://user-images.githubusercontent.com/24453970/60983803-e829e080-a33a-11e9-8ab8-5d722c6c51f4.png">
<img width="467" alt="image" src="https://user-images.githubusercontent.com/24453970/60983924-1e676000-a33b-11e9-89cb-717f49be1d99.png">
<img width="450" alt="image" src="https://user-images.githubusercontent.com/24453970/60983982-37701100-a33b-11e9-83cc-b28ba827dd83.png">
<img width="453" alt="image" src="https://user-images.githubusercontent.com/24453970/60984029-48b91d80-a33b-11e9-87ad-803da34fa511.png">
<img width="468" alt="image" src="https://user-images.githubusercontent.com/24453970/60984066-5bcbed80-a33b-11e9-85b4-0729466ef78a.png">
<img width="467" alt="image" src="https://user-images.githubusercontent.com/24453970/60984127-743c0800-a33b-11e9-84ea-436eb7e669d3.png">

Most of the clusters are either invalid or with general info like this:
<img width="464" alt="image" src="https://user-images.githubusercontent.com/24453970/60984347-dd238000-a33b-11e9-8141-19a084cf8054.png">
<img width="465" alt="image" src="https://user-images.githubusercontent.com/24453970/60984397-ed3b5f80-a33b-11e9-8f89-bb8109f79eac.png">

What I really need is to have a way of reading change status of the switch, when it happens.
It would also be nice with temperature reading, and of course battery status (it seems to be wrong now, they are fresh batteries).

If there is any more info I can send you to add this device just let me know :-)
Thanks!

SwoopX:
Yay, another guinea pig :)

Could you please restart deconz and check if the node info panel changes (#1644 )?

General Develco support has already merged into the code. If you feel funny, you may compile the latest version and give it a try or you wait for the next official release.

pivot69:
I’m happy to be a lab rat, especially when I can collaborate with the scientist to get the results needed 😎

Restart the device (from deconz GUI), the service or the raspberry pi itself?

SwoopX:
:)
Deconz service should suffice...

pivot69:
Now there is more! :-)

[...]

PR #1684, opened 2019-07-17
Add support for Develco magnetic sensor #1667

SwoopX:
- #1667
- Added Develco/Wattle magnetic sensor.
- Update of Develco power binding value
- Added Develco specific attribute to basic cluster in general.xml
- Corrections from previously closed PR

### DIFF
--- a/bindings.cpp
+++ b/bindings.cpp
@@ -931,8 +931,16 @@ bool DeRestPluginPrivate::sendConfigureReportingRequest(BindingTask &bt)
     {
         rq.dataType = deCONZ::Zcl16BitInt;
         rq.attributeId = 0x0000;       // measured value
-        rq.minInterval = 10;           // value used by Hue bridge
-        rq.maxInterval = 300;          // value used by Hue bridge
+        if (sensor && sensor->modelId().startsWith(QLatin1String("SMSZB-120")))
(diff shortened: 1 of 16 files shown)


bindings.cpp
```diff
@@ -1815,7 +1819,10 @@ bool DeRestPluginPrivate::checkSensorBindingsForAttributeReporting(Sensor *senso
             {
                 val = sensor->getZclValue(*i, 0x0035); // battery alarm mask
             }
-            else if (sensor->modelId() == QLatin1String("Motion Sensor-A") || sensor->modelId() == QLatin1String("SMSZB-120"))
+            else if (sensor->modelId() == QLatin1String("Motion Sensor-A") ||
+                     sensor->modelId() == QLatin1String("SMSZB-120") ||
+                     sensor->modelId() == QLatin1String("WISZB-120") ||
+                     sensor->modelId().startsWith(QLatin1String("ZHMS101"))
```
manup: Compilation fails due missing parentheses, it should be: `sensor->modelId().startsWith(QLatin1String("ZHMS101")))`

SwoopX: Strange my local compile goes through. Nevermind, fixed it, thanks!

PR #1689, opened 2019-07-18
Implemented support for Develco Smart Plug Mini

patrikolausson:


SwoopX:
Do you get the correct manufacturer code in the basic cluster info (0x1015) for the device?

patrikolausson:
Yes. The manufacturer code is reported as 0x1015

manup:
Thanks for the PR. Can you please remove the `100755 → 100644` file mode changes from it, so it can be merged.

Issue #1747, opened 2019-08-12
Sensor working but missing in Phoscon sensor list


udoewich:
I'm working on adding support for a motion sensor and have it paired, working and visible via the API:
```
{
    "config": {
        "battery": 100,
        "duration": 90,
        "on": true,
        "pending": [],
        "reachable": true
    },
    "ep": 1,
    "etag": "27282d54132d9a50ead664fbd9ba950f",
    "manufacturername": "ORVIBO",
    "modelid": "895a2d80097f4ae2b2d40500d5e03dcc",
    "name": "Motion1",
    "state": {
        "lastupdated": "2019-08-12T21:55:33",
        "lowbattery": false,
        "presence": false,
        "tampered": false
    },
    "type": "ZHAPresence",
    "uniqueid": "00:15:8d:00:01:c1:24:8d-01-0500"
}
```
However, it is not shown in the Connected sensors list of Phoscon.  I was able to assign it to a group via Sensor control and it works fine triggering events when motion is detected.

What determines whether or not a sensor is listed under Sensors?

Ronoon:
the phoscon app is not open source therefor you have to ask @manup to integrate the motion sensor to the app

udoewich:
Right, forgot about that.  I'll look into getting my changes into shape for a pull request.

liorm0505:
Hi udoewich, 

May i know how did you able to pair the device and get the info through the API?
When i pair my device the node has no name but it still visible in the deconz app. 
Also, Requesting the api for all sensors does not respond with this sensor.  

I am running the app on Rp3 with HASSIO the deconz addon.

udoewich:
Sensors that are not listed as already supported (see the Supported Devices Wiki page for that) have to be added to the source code before they are recognized by the API and Phoscon.  deCONZ will show devices even if they have not been added to the REST API yet.

I had to make changes to the code before the ORVIBO sensors worked for me.  I'm prepping a pull request right now so hopefully there will be official support for these soon.

liorm0505:
That's great news :)
I would love to have support for this sensor in the next release.

udoewich:
Turns out that it _does_ list the sensor on Phoscon. A second sensor I added showed up just fine on the sensor listing. The first still not being listed might be due to the incomplete support on the API side at the time I added it. I'll eventually remove it from the database to make sure.

Issue #901, opened 2018-10-23
Orvibo Smoke Sensor


alonalon:
Hi,
Would appreciate if someone could look into add support for [Orvibo Smoke Sensor.](https://www.orvibo.com/en/product/smoke_sensor.html)

I run everything on headless Linux server, so can't provide any cluster info unfortunately. 
But I think it's very much like [Trust smoke detector](https://www.trust.com/en/product/71197-zigbee-smoke-detector-zsdr-850) & [Heiman HS1SA Smoke detector](http://www.heimantech.com/product/96.html)

Maybe all those three devices can share most of the code?

jonatanhost:
I have the same device connected but not recognized. Here is a screenshot of node info.

![image](https://user-images.githubusercontent.com/11466166/47442603-e3c09600-d7b2-11e8-8106-0c4a85fecedb.png)

jonatanhost:
@manup Do you guys need more information from us of this smoke sensor to implement it?

manup:
>@manup Do you guys need more information from us of this smoke sensor to implement it?

Some more screenshots as described here would be useful, especially `model identifier` from basic cluster is needed.

https://github.com/dresden-elektronik/deconz-rest-plugin/wiki/Request-Device-Support

alonalon:
@jonatanhost is it possible for you to get the info manup asks for?

jonatanhost:
> > @manup Do you guys need more information from us of this smoke sensor to implement it?
> 
> Some more screenshots as described here would be useful, especially `model identifier` from basic cluster is needed.
> 
> https://github.com/dresden-elektronik/deconz-rest-plugin/wiki/Request-Device-Support

I don't know how to get those attributes to show

alonalon:
<img width="550" alt="screenshot 2019-02-06 at 20 37 59" src="https://user-images.githubusercontent.com/3322693/52368696-59452f00-2a4f-11e9-9796-4b3824db1e49.png">

think i got it, does this help @manup ?

jonatanhost:
@manup 

Do you guys need any more information about the device or is this enough?

jonatanhost:
@manup Hi, do you have any status on this? Have two sensor myself of this I would like to use as not just dumb smoke sensors :)

manup:
Hi, sorry for the late reply. Yes one more thing, please provide a screenshot of the IAS Zone cluster after reading the attributes.

[...]

PR #1763, opened 2019-08-20
Support for ORVIBO switch, dimmer and motion sensor

### LINKED ISSUE #794: Device support request: Orvibo in-wall dimmer switch
I took a chance and bought an Orvibo in-wall Zigbee dimmer switch (US version; box says model# T10D1) from a local reseller to see if I could get it to work with Phoscon.  Alas, it does not show up with either a light or sensor scan.  It did however pair with my Conbee and shows up in deCONZ.  Here's the info:

image

Basic Cluster Info:

orvibo-deconz-02

Basic Node Info:

orvibo-deconz-03

Would love to see this added as a light (or switch?) in Phoscon.  I was already able to turn the connected lights on, off and dim them via deConz.

udoewich:
This adds support for ORVIBO in-wall switch, in-wall dimmer and motion sensor. Closes #794, #1686 and #1747.

### DIFF
--- a/de_web_plugin.cpp
+++ b/de_web_plugin.cpp
@@ -206,6 +206,7 @@ static const SupportedDevice supportedDevices[] = {
     { VENDOR_EMBER, "3AFE140103020000", konkeMacPrefix }, // Konke Kit Pro-FT Temp Humidity Sensor
     { VENDOR_EMBER, "3AFE130104020015", konkeMacPrefix }, // Konke Kit Pro-Door Entry Sensor
     { VENDOR_NONE, "RICI01", tiMacPrefix}, // LifeControl smart plug
+    { VENDOR_JENNIC, "SN10ZW", jennicMacPrefix }, // ORVIBO motion sensor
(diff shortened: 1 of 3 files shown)


alonalon:
@udoewich is it possible for you to add support for #901 also? :)

udoewich:
@alonalon 
I just pushed a small update which "should" add support for the ORVIBO smoke sensor. Obviously that's untested as I don't have one of those.

de_web_plugin.cpp
```diff
@@ -3504,6 +3559,15 @@ void DeRestPluginPrivate::addSensorNode(const deCONZ::Node *node, const deCONZ::
                             else if (modelId.isEmpty() && j->id() == 0x0005) // model id
                             {
                                 modelId = j->toString().trimmed();
+                                // replace ORVIBO model ID
```
manup: I'll merge this for 2.05.67 but suggest we refactor it later, so that optional translations or mappings of device model identifiers into readable text are kept in the device map.

PR #1790, opened 2019-08-28
Fix for value convertion

bastshoes:
Fixing value conversion for Current and Power Consumption for LifeControl smart plug.


bastshoes:
Messed up again.

PR #1793, opened 2019-08-29
Update ikea-ota-download.py

HackThatNow:
Valid for python3


manup:
Thanks Python 3 support is welcome, but now it doesn't work on Python 2 anymore.

Test on my Arch Linux box:

```
python2.7 ikea-ota-download.py 
Traceback (most recent call last):
  File "ikea-ota-download.py", line 10, in <module>
    import urllib.request
ImportError: No module named request
```

It appears that `urllib.request` is for Python 3 only:

https://stackoverflow.com/a/37517332

PR #1806, opened 2019-08-31
Update ikea-ota-download.py - python2/3 compat

ragzilla:
Resolves #1793 using try/except approach.  Alternative recommended approach is to use the Requests library, but unless there's other significant python code it's probably not worth adding the dependency.


ragzilla:
This is a pretty good resource for python2/3 compat: https://python-future.org/compatible_idioms.html

Issue #1833, opened 2019-09-08
Is there a was to Update Osram Ledvance Downlights TW?


lons81:
Hello

How can i Update my Ledvance Downlights 25W to a newer Firmware?
In every Switch on the go to cold white with 100% and dont remember the last state.

Thanks, Marcel

ebaauw:
> How can i Update my Ledvance Downlights 25W to a newer Firmware?

Afaik OSRAM/Ledvance haven't published their firmware, so I'm afraid the only answer is: through their gateway.

> In every Switch on the go to cold white with 100% and dont remember the last state.

Unlike Hue, OSRAM use a manufacturer-specific cluster for the power-on settings, see #102.  They have been for as long as I remember (over several years), so I doubt you'd actually need a firmware update for this functionality.

Unlikely that REST API will support this anytime soon, but the `general.xml` from the above commit adds support to the deCONZ GUI:
![Screenshot 2019-09-08 at 14 57](https://user-images.githubusercontent.com/22179355/64488545-79d5a180-d249-11e9-93a1-fa97791dde6c.png)

Copy `general.xml` to `/usr/share/deCONZ/zcl` and restart deCONZ.  You should then see the updates in the GUI.  Set the bulbs brightness and colour temperature to the desired values and press the button to store them as new power-on defaults.  I don't think they support powering-on the lights as off.  Their app used to demand a minimal brightness, but I don't think their lights actually enforce this.

I have only tested this on an OSRAM Classic B40 TW - LIGHTIFY with firmware V1.05.10.

lrybak:
Hi Eric (@ebaauw),
I've copy the general.xml file to the /usr/share/deCONZ/zcl but see no change in GUI. Still no power-on Defaults option. Any hint?

![image](https://user-images.githubusercontent.com/5631648/70148168-f0960580-16a5-11ea-93da-26c84bc0b049.png)

ebaauw:
You need to select the 0xfc0f cluster.

lrybak:
Where it is?

lrybak:
Forgive me, I was blind. Now I found it, thx.

7wells:
Please consider also this post over there:
https://github.com/dresden-elektronik/phoscon-app-beta/issues/213#issuecomment-701205883

Issue #1827, opened 2019-09-06
Battery levels as own entites


embeddeddev:
Unfortunalety the battery levels of devices like Philips motion sensor and Xiaomi/Aqara ambience, door/window and vibration sensors are modelled only as attributes of their sensor values/entities like temperature and humidity so they aren't directly accessible as own entities in Home Assistant and need to be integrated complicatedly and error-prone by using template sensors.

Contrary ZHA component of Home Assistent directly exports these battery levels correctly as own entities.

In addition current implementation provides the values redundantly but in fact its only one unique device attribute which should be exported as such one.

Looking forward to receive a reply.

erikced:
How to expose battery levels in HA should probably be discussed [pydeconz](https://github.com/Kane610/deconz) instead of here.

embeddeddev:
Thx for the hint!

ebaauw:
I’m all in favour of exposing the battery (or rather: the _Power Configuration_ cluster 0x0001) as a separate resource, but I’m not sure whether we should do that before API v2.

Technically we could introduce a `ZHABattery` sensor type, exposing the battery as `state.battery` with a corresponding `state.lastupdated` (a long-time wish of mine).  It would also enable us to expose the battery for devices that are currently exposed a `/lights` resources (like the IKEA FYRTUR).  On the downside, this would almost double the amount of `/sensors` resources.

@manup, what domyou think?

embeddeddev:
Hi @ebaauw, thx for your reply and your support!

Indeed it double's the number of sensors for single sensor devices, but in reality it IS a sensor w/ important information. ZHA already grabbed this issue and provides battery levels for all battery based sensors (even if no _Power Configuration_ cluster visible in deCONZ GUI!?)

Related to @erikced's reply I opened an [mirror issue on pydeconz](https://github.com/Kane610/deconz/issues/42). Now I am really unsure to which this topic has to be associated to due to my lack of technical internals.

ebaauw:
Exposing it thru the REST API is one thing, actually using it in apps or plugins to home automation systems is another.  That needs to be discussed in their respective repositories.  We’d keep the current `config.battery` around, at least for a while for backwards compatibility, and probably forever in Hue compatibility mode.

Issue #102, opened 2017-08-08
Feature request: support power-on defaults for OSRAM lights


ebaauw:
Could we support setting the power-on defaults of an OSRAM light from the deCONZ REST API?  I'd imagine a syntax like PUT /lights/xx/config/poweron with a body of `{"bri":xxx, "ct":xxx}`.

Sniffing around the OSRAM gateway, it looks like it issues a 0x01 command to the 0xFC0F cluster to store the power-on defaults (packet 691).  Before that, it issues an On command to the 0x0006 cluster (packet 683), and a 0x0A command to 0x0300, with the current colour temperature as data (packet 687).  The same sequence in 311, 315, 317.

Attached the BitCatcher log file:
[osram.dcf.gz](https://github.com/dresden-elektronik/deconz-rest-plugin/files/1208569/osram.dcf.gz)

You probably need the network key as well: CB:2D:00:A0:72:EA:FF:63:03:5E:D5:95:41:36:83:24.

Note that BitCatcher starts counting packets at 0, but WireShark at 1, so add 1 when viewing with WireShark.

ebaauw:
Please keep this on the list.  It will be very relevant this fall, with the expected new firmware for the Hue bulbs finally allowing a power-on setting.

manup:
Level control cluster attributes suggest that when I recall correctly, but I haven't tested if it works.

ebaauw:
There's two different issues (I had them confused as well):
- How the light behaves when turning it on (while it has remained powered).  This is set through the _On Level_ and _OnOff Transition TIme_ attributes of the Level cluster.  The Hue lights don't support these, and behave as if they're set to 255 (previous level) and 4 (0.4 sec transition time).  Most other lights (OSRAM, IKEA, ubisys, innr) support these, so you want to set them to the Hue values.  They're stored in volatile memory, so this settings needs to be applied regularly, or at least after a device announcement.
- How the lights behaves when powered on.  Currently, the Hue lights come on at full intensity and some unholy colour temperature.  OSRAM allows changing this through a manufacturer specific cluster.  I've not seen anything like this for other lights.   Hue have announced they will support setting another power-on default, see https://github.com/ebaauw/homebridge-hue/issues/41#issuecomment-436936220.  I'm assuming it will be through an API enhancement, probably writing some manufacturer specific attribute.

ebaauw:
> hopefully Philips will go the standard route here.

[...]

PR #1837, opened 2019-09-08
FYRTUR battery and OSRAM power-on settings

ebaauw:
**FYRTUR Battery**
- Add support for `ZHABattery` and `CLIPBattery` sensors, with `state.battery` and `state.lastupdated`, but without `config.battery`, see #1827;
- For now, a `ZHABattery` sensor resource is created only for the IKEA FYRTUR and KADRILJ smart blinds (that are otherwise only exposed as a light resource), see #1121;
- I haven't touched `config.battery` for other sensor resources;
- Not sure how much of this we want to pursue before API v2, see #1827.  The logic works for "all" devices with a _Power Configuration_ cluster (to test, remove the `if` statement from https://github.com/ebaauw/deconz-rest-plugin/blob/dc83c121cf40e6d4331765f811623631c443b5d5/de_web_plugin.cpp#L3627-L3631), with some restrictions:
  - I haven't implemented the code to update `state.battery` from _Battery Voltage_;
  - I haven't implemented the code to create a `ZHABattery` for Xiaomi devices, nor to update `state.battery` from the Xiaomi special attribute report;
  - `config.group` is populated for the `ZHABattery` sensor for devices that send group commands (wireless switches, Trådfri motion sensor).  This is done in too many places in the code to check for `ZHABattery`.
```
{
  "config": {
    "alert": "none",
    "on": true,
    "reachable": true
  },
  "ep": 1,
  "etag": "03741a300ae80dac92894b7b84e2b2e6",
  "modelid": "FYRTUR block-out roller blind",
  "name": "FYRTUR block-out roller blind ",
  "state": {
    "battery": 92,
    "lastupdated": "2019-09-08T16:04:43"
  },
  "swversion": "2.2.007",
  "type": "ZHABattery",
  "uniqueid": "00:0d:6f:ff:fe:9e:00:aa-01-0001"
}
```

**OSRAM Power-On Settings**
- Added to `general.xml` the OSRAM manufacturer-specific cluster to set the power-on defaults to the current state, see #102, #1833.


mysteryan:
The power-on default cluster appears with those commits to general.xml for bulbs identifying their manufacturer as “OSRAM”, but doesn’t show up for LEDVANCE ones.  I think it uses a similar cluster address FC01 instead of FC0F so a simple inclusion of that address should fix it.
B2264033-248A-45B7-A88A-E6071571C0CA

ebaauw:
What _Manufacturer Code_ does Ledvance use?  Also 0xbbaa?

mysteryan:
Manufacturer code: 0x1189
Copy and pasted a duplicate of your commit and substituted fc01 for fc0f, manufacturer code with 0x1189 for 0xbbaa and LEDVANCE for OSRAM for name and seems to work fine for the modified duplicate.

ebaauw:
Cool!  Could you create a PR for that?

Issue #1874, opened 2019-09-16
Rule operator `ne`


Krocko:
Is it possible to add support for the rule operator `ne`?

Krocko:
Hi @manup any plans on this?
Is it also possible to add the operator `not in`?

PR #1978, opened 2019-10-18
Various enhancements

ebaauw:
- Ignore faulty -100°C temperature in the regular and Xioami special attribute report for `lumi.weather`, see #1748.
- Add `ne` operator for rule conditions, see #1874.
- Initial support for iCasa remote, see https://www.icasa.io/afstandsbediening.  Still work in progress.

### DIFF
--- a/bindings.cpp
+++ b/bindings.cpp
@@ -1714,6 +1714,7 @@ bool DeRestPluginPrivate::checkSensorBindingsForAttributeReporting(Sensor *senso
         sensor->modelId().startsWith(QLatin1String("CSW_ADUROLIGHT")) ||
         // innr
         sensor->modelId() == QLatin1String("SP 120") ||
+        sensor->modelId().startsWith(QLatin1String("RC 110")) ||
         // Eurotronic
         sensor->modelId() == QLatin1String("SPZB0001") ||
         // Heiman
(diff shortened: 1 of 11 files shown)


manup:
Cool thanks, good stuff, the remote looks interesting.

ebaauw:
It feels very nice, but I'm running into a number of challenges supporting it in full.

The (unnumbered) _On_/_Off_ buttons send commands to all four groups, corresponding to the numbered (1-4) _On_/_Off_ buttons.  Same for the _S1_ and _S2_ buttons.  When using the default groups, it sends the commands from the last endpoint used for the numbered _On_/_Off_ buttons, causing the REST API plugin to see new group assignments.  When I bind the endpoints to groups manually, the commands are sent from the appropriate endpoint.  However, I haven't yet been able to make the REST API create these bindings, since it already sees the default groups.

manup:
>However, I haven't yet been able to make the REST API create these bindings, since it already sees the default groups.

Yes tricky perhaps a read-only group/button setup for which the bindings are created while sensor search is active will work.
* When remote joins and modelid is detected:
* Create a todo list of bindings which are to be created, which is processed one at a time
* The bindings target new dynamically created groups
* During active sensor search, no default Groups are created based on received commands

A "binding todo list" might be useful in general.

Does this make sense?

Thomas-Vos:
> Add ne operator for rule conditions, see #1874
Is there any place where this is documented? It is quite difficult to just guess the API differences between Hue and deCONZ without documentation.

Issue #2201, opened 2019-12-13
Tradfri remote (round / 5 Buttons) - Powerbutton Bug


rehmosch:
Hello, wanted to ask if someone also has this bug with the Ikea Tradfri power button (round with 5 Buttons): 

If you press the power button briefly you get the value 1002 = all right. 
If you press the power button for a long time you get the value 1001 = all good 
If you hold down the power button for a long time, you do not get the value 1003 = error 
= the error here is, that when you press the "long button" both commands 1002 and 1001 are executed simultaneously (this is the case with all other 4 buttons) not so, here I can execute at least 2 commands separately, but unfortunately not with the power button, does anyone have a solution?

For long press the Power-Button need about 6 Secondes all other Buttons about 2 oder 3 Secondes, how i can change the long press value?

 Thanks Jutta

ebaauw:
I agree, an x001 (Hold) event should always be followed by an x003 (Release after hold).  The Trådfri remote sends a command on holding the On/Off button, but not on releasing it.  Imho, the REST API plugin should have ignored the command On hold and not send the 1001.

rehmosch:
Dear ebaauw,

thank you for the answer. You think, that the devs can can fix  the REST API bug?  Or i can fix it, when I use the REST API? Sorry, im a noob ;)

ebaauw:
You could simply ignore the 1001.  I can create a PR so that the API no longer issues the 1001.  Not sure if any-one is actually using that event, though.

ebaauw:
Hm, the command that leads to the 1001 is actually sent on release after hold, not on hold.  So (before my change) the remote sends a command on press, which generates the 1002 and a command on release after hold, which generates the 1001.  It doesn't generate a command on short release (so the press is correctly mapped to 1002), nor on hold.

I'm not sure where I would stand on issuing a 1003 without a prior 1001, but that's far better than issuing a 1001 without a 1003.  However, a single press/hold/release action should not result in a 1002 followed by a 1003, as that would indicate two separate actions.  Still think ignoring the  command on release after hold is the best option.

[...]

Issue #2148, opened 2019-11-30
Rules not triggered


ebaauw:
I think there's an issue with rule triggering.  I have several rules that should trigger 10 minutes after a CLIPGenericStatus sensor has changed `status` to -2.  One of these rules actual resets the `status` to 1.  The rules seem to be triggered in random order, which is OK.  However, after the rule that changes the `status` has executed, the remaining rules are no longer triggered.  Presumably, because the conditions of the remaining rules are evaluated only after the previous rule has been executed.  I would expect the conditions of all rules to be evaluated before any rule gets executed.

I think a very long time ago, we had a similar issue with rules that would set a CLIPGenericFlag and update it.  My use case is a more complicated, since it involves multiple rules.

As a workaround, I now changed the condition for the rule that changes the `status` to be triggered 10 seconds after the other rules.

ebaauw:
Now on v2.05.72, still seeing some issues with rule triggering for time-based conditions:
- `ddx` rules seem to be triggered twice, a second apart, looking at the deCONZ log and at the `timestriggered` attribute.  No big deal for me, as the actions are idempotent;
- Rules with a single condition on `/config/localtime` `in` `W7/T21:00:00/T09:00:00` didn't get triggered yesterday (Friday).  The rules with value `W120/T21:00:00/T08:00:00` did get triggered on Thursday.  I restarted deCONZ Thursday (after installing v2.05.72), so this would have been the first trigger since deCONZ started.  I think I've seen before that rules on `localtime` don't get triggered on the day the rule is created and/or the day deCONZ starts.

ebaauw:
> - `ddx` rules seem to be triggered twice, a second apart, looking at the deCONZ log and at the `timestriggered` attribute. No big deal for me, as the actions are idempotent;

https://github.com/dresden-elektronik/deconz-rest-plugin/blob/bb47cdbc19c257e3000fdb8475aa456830e6df94/rest_rules.cpp#L1387-L1391

It seems that `QDateTime::secsTo()` ignores the microseconds part of the `QDateTime` value, instead of rounding them to the nearest second.  This means that any value in the range from -0.999 to 0.999 will evaluate to 0.  Using `QDateTime::msecsTo()` instead seems to fix this:
```c++
            QDateTime dt = item->lastChanged().addSecs(c->seconds());
            qint64 delta = now.msecsTo(dt);
            if (delta < -500 || delta >= 500)
            {
                return false;
            }
```

[...]

PR #2203, opened 2019-12-13
Various bug fixes

ebaauw:
Various bug fixes:
- Trådfri remote: don't issue 1001 button events without corresponding 1003s.  See #2201.  I'm not sure every-one would agree to this change; please see the discussion in the issue;
- Sometimes `ddx` rules are triggered twice, see #2148;
- Illegal `state.bri` and `state.sat` value of 255 for window covering devices, see #2257.

### DIFF
--- a/sensor.cpp
+++ b/sensor.cpp
@@ -157,7 +157,7 @@ static const Sensor::ButtonMap ikeaRemoteMap[] = {
 //    mode                          ep    cluster cmd   param button                                       name
     // big button
     { Sensor::ModeColorTemperature, 0x01, 0x0006, 0x02, 0,    S_BUTTON_1 + S_BUTTON_ACTION_SHORT_RELEASED, "Toggle" },
-    { Sensor::ModeColorTemperature, 0x01, 0x0005, 0x07, 2,    S_BUTTON_1 + S_BUTTON_ACTION_HOLD,           "Setup 10s" },
+    // { Sensor::ModeColorTemperature, 0x01, 0x0005, 0x07, 2,    S_BUTTON_1 + S_BUTTON_ACTION_HOLD,           "Setup 10s" },
     // top button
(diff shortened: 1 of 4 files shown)


sensor.cpp
```diff
@@ -157,7 +157,7 @@ static const Sensor::ButtonMap ikeaRemoteMap[] = {
 //    mode                          ep    cluster cmd   param button                                       name
     // big button
     { Sensor::ModeColorTemperature, 0x01, 0x0006, 0x02, 0,    S_BUTTON_1 + S_BUTTON_ACTION_SHORT_RELEASED, "Toggle" },
-    { Sensor::ModeColorTemperature, 0x01, 0x0005, 0x07, 2,    S_BUTTON_1 + S_BUTTON_ACTION_HOLD,           "Setup 10s" },
+    // { Sensor::ModeColorTemperature, 0x01, 0x0005, 0x07, 2,    S_BUTTON_1 + S_BUTTON_ACTION_HOLD,           "Setup 10s" },
```
manup: I agree with the reasoning that the HOLD action isn't a good fit here. However in current Phoscon App versions the pairing wizard for this remote requires this button event to finish the procedure. The next gen. Phoscon App wizard is happy to accept any button event, but it's not released yet.

I would suggest we use either another existing action like INITIAL_PRESS or add a new action to capture the event?

ebaauw: I was afraid some-one or something would depend on this button event.  Best leave it for now, and pick it up once new Phoscon release is out.

Imho, sending INITIAL_PRESS is just as bad as sending HOLD: client applications expect a SHORT_RELEASED or HOLD after INITIAL PRESS, just as they expect a LONG_RELEASED after HOLD.  As discussed in the issue, we could map it to LONG_RELEASED, but that's still kinda weird without an INITIAL_PRESS or HOLD.  Then again, client applications polling the Hue bridge might miss these and only see the LONG_RELEASED.

PR #2224, opened 2019-12-18
Fixed inclusion of eWeLink SmartPlug model SA-003-Zigbee

muchasuerte:
This resolve the inclusion of eWeLink SA-003-Zigbee reported as "type": "Unknown" and now as "On/Off plug-in unit"


muchasuerte:
I have tested this fix also with sonoff BASICZBR3 which is really similar to  SA-003-Zigbee, it is working fine

PR #2239, opened 2019-12-24
Add support for Aqara Opple switches - WIP

merdok:
Trying to add support for the Aqara Opple switches. See #2061 

### This is a **work in progress.**

Although the switches already work, there are still multiply issues:

1. The switches share the same vendor group with `VENDOR_115F`. I suspect this might be causing some of the issues.

2. Adding a switch does not always work for some reason. Sometimes it works on first try but sometimes i have to try multiply times. I have no idea why.

3. I can only add a switch when i search for sensors. I can also search for switches but that never succeeds, even that the switch still gets paired but is behaving strange. Also no idea why...

4. When i add a switch then always a new group is created for every switch. I have already like 50 groups right now (because of multiply times removing and re-adding the switches). Can this be somehow done that for the Aqara Opple switches always just one group is created or is the current behavior as it should be? 

5. ~~Sometimes in the config temperature comes up in the rest api. Even i blacklisted it in the database.cpp...~~ Fixed

6. The 4 button switch does not have a hold button event. No matter which button i hold, there always comes just a single press one.

7. The 6 button switch does not have a double press event. Here no matter where i double press always a single press comes.

8. Also the 6 button switch for holding sends a hold event when i hold the button and a release event when i release it. No matter if i hold the left or right button, on release always the same event comes, there is no difference between left or right release. Is that ok like that? Also the hold and release for the 5th and 6th button does not work, even that the map is correct. Maybe something with the button map?

9. I think that there may be some unnecessary places where i whitelisted the switches, but it might also be that some places are missing. 

10. Battery voltage level is not correctly reported. 

It would be good if @ebaauw and @manup could have a look at the changes and see what can be improved to fix the specified issues. I am sure that there are some things missing...

### DIFF
--- a/database.cpp
+++ b/database.cpp
@@ -3321,7 +3321,8 @@ static int sqliteLoadAllSensorsCallback(void *user, int ncols, char **colval , c
             if (!sensor.item(RStateTemperature) &&
                 sensor.modelId() != QLatin1String("lumi.sensor_switch") &&
                 !sensor.modelId().contains(QLatin1String("weather")) &&
-                !sensor.modelId().startsWith(QLatin1String("lumi.sensor_ht")))
+                !sensor.modelId().startsWith(QLatin1String("lumi.sensor_ht")) &&
+                sensor.modelId().contains(QLatin1String("86opcn01")) == false) // exclude Aqara Opple
             {
                 item = sensor.addItem(DataTypeInt16, RConfigTemperature);
                 item->setValue(0);
--- a/database.cpp
+++ b/database.cpp
@@ -3322,7 +3322,7 @@ static int sqliteLoadAllSensorsCallback(void *user, int ncols, char **colval , c
                 sensor.modelId() != QLatin1String("lumi.sensor_switch") &&
                 !sensor.modelId().contains(QLatin1String("weather")) &&
                 !sensor.modelId().startsWith(QLatin1String("lumi.sensor_ht")) &&
-                sensor.modelId().contains(QLatin1String("86opcn01")) == false) // exclude Aqara Opple
+                !sensor.modelId().contains(QLatin1String("86opcn01"))) // exclude Aqara Opple
             {
                 item = sensor.addItem(DataTypeInt16, RConfigTemperature);
                 item->setValue(0);
--- a/de_web_plugin.cpp
+++ b/de_web_plugin.cpp
@@ -3461,6 +3461,16 @@ void DeRestPluginPrivate::checkSensorButtonEvent(Sensor *sensor, const deCONZ::A
                 }
 
             }
+            else if (ind.clusterId() == COLOR_CLUSTER_ID && 
+                     (zclFrame.commandId() == 0x4c))  // Aqara Opple hold
+            {
+                ok = false;
+                if (zclFrame.payload().size() >= 1 && buttonMap->zclParam0 == zclFrame.payload().at(0)) // direction
+                {
+                    sensor->previousDirection = zclFrame.payload().at(0);
+                    ok = true;
+                }
+            }
 
             if (ok)
             {
--- a/de_web_plugin.cpp
+++ b/de_web_plugin.cpp
@@ -3426,7 +3426,8 @@ void DeRestPluginPrivate::checkSensorButtonEvent(Sensor *sensor, const deCONZ::A
                 }
             }
             else if (ind.clusterId() == COLOR_CLUSTER_ID &&
-                     (zclFrame.commandId() == 0x4b && zclFrame.payload().size() >= 7) )  // move color temperature
+                     (zclFrame.commandId() == 0x4b && zclFrame.payload().size() >= 7) && // move color temperature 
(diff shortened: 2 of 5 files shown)


database.cpp
```diff
@@ -3321,7 +3321,8 @@ static int sqliteLoadAllSensorsCallback(void *user, int ncols, char **colval , c
             if (!sensor.item(RStateTemperature) &&
                 sensor.modelId() != QLatin1String("lumi.sensor_switch") &&
                 !sensor.modelId().contains(QLatin1String("weather")) &&
-                !sensor.modelId().startsWith(QLatin1String("lumi.sensor_ht")))
+                !sensor.modelId().startsWith(QLatin1String("lumi.sensor_ht")) &&
+                sensor.modelId().contains(QLatin1String("86opcn01")) == false) // exclude Aqara Opple
```
manup: Hi, please keep the style consistent for easier reading.

`!sensor.modelId().contains(QLatin1String("86opcn01"))) // exclude Aqara Opple`

de_web_plugin.cpp
```diff
@@ -93,6 +93,7 @@ const quint64 bjeMacPrefix        = 0xd85def0000000000ULL;
 const quint64 xalMacPrefix        = 0xf8f0050000000000ULL;
 const quint64 lutronMacPrefix     = 0xffff000000000000ULL;
 const quint64 legrandMacPrefix    = 0x0004740000000000ULL;
+const quint64 aqaraOppleMacPrefix = 0x04cf8c0000000000ULL;
```
manup: According to https://wintelguy.com/index.pl the MAC address prefix `04cf8c` belongs to XIAOMI Electronics,CO.,LTD.
I think Opple is just product line (?), perhaps a name like `xiaomiMacPrefix` or `aqaraMacPrefix` is more general and future safe.

database.cpp
merdok: Will do that 👍

de_web_plugin.cpp
merdok: Well, there are already Aqara devices which use a different MAC prefix, so i guess Xiaomi uses this new MAC prefix for the Aqara Opple line?

SwoopX: Seems so. I'd prefer xiaomiMacPrefix

merdok: Xiaomi devices (including some Aqara ones) are already under the jennicMacPrefix. I have no idea why this is called like that. 
The Aqara Opple devices have this new mac prefix. 
Calling it just `xiaomiMacPrefix` or `aqaraMacPrefix ` will make it even more confusing then it already is...

SwoopX: Im fact, it's not really difficult at all.

We're here talking about the manufacturer of the transmission unit or maybe even SoC. Not too many vendors manufacture them but regardless, every one of them gets a unique MAC prefix for identification. Now, those transmission units or SoCs are used to build devices. Maybe the MAC vendor builds them or a third party vendor. Based on those possibilities, you cannot stipulate only Xiaomi uses that MAC for their devices. Maybe they sell them... To make things worse, there are also OEMs, pretty much selling the hardware but the buying company labels them as their own...

Bottom line is that you cannot do anything about the confusion. That's the reason the whilelist combinations are in the code, as of now. From my point of view, there is only one reasonable decision on how to name it.

merdok: That is true but it is not related to this. I am not trying to fix the confusion here, I am trying to prevent further confusion. The Aqara Opple devices have clearly an unique mac and should be grouped smorodiny to that.

SwoopX: I'm afraid your answer tells that you still did not get the idea. Maybe it will me more obvious if in a couple of months, e.g. IKEA devices come up with a Xiaomi MAC...

Anyhow, I stop arguing here and leave the decision up to @manup

manup: No worries :) there is kind of a system behind all this. The Jennic prefix refers to the underlying radio manufacturer MAC address prefix. Companies often start using  such MAC address prefixes before getting their own prefix (sometimes multiple).

The Jennic prefix and radios are not just used by Xiaomi, same goes for emberMacPrefix, tiMacPrefix, stMacPrefix and silabsMacPrefix. These are used by multiple companies because they are using their radios of the respective manufacturers.

Another example is the philipsMacPrefix, which refers to Philips as company (not just the Hue line) `0x001788`. I think xiaomiMacPrefix is the most adequate name and I guess there will be new Zigbee products from Xiaomi / Aqara with this prefix not necessarily belonging to the OPPLE range.

merdok: Thanks! That makes sense now. I was kind of confused by the jennic mac prefix since all other xiaomi/aqara devices currently  are using that, so i was not sure how this new switches fit in that. I will rename the Opple mac prefix to `xiaomiMacPrefix `.

JackPoint:
@merdok Thanks for all the work for this PR!
@manup Do you know when there will be a new release with this PR? I'd like to have this available in Home Assistant.

PR #2293, opened 2020-01-06
New konke motion model support

airens:


nicedevil007:
If this will be implemented to the official and dockerrelease what else would be necessary to get the PIR connected to the Conbee stick?

airens:
I already got the pir connected to Conbee, it's just doesn't exposed to the Rest, so, I can't use it in Home Assistant

PR #2384, opened 2020-01-27
Extend and refactor Scene class

KodeCR:
This is mostly the second commit from #1471 so it can be reviewed and merged in stages, cherry-picked onto the current master. This:
- extends the scene class with all fields additionally available with the Hue api and adds those to the database, and
- refactors the scene class.
This PR should not result in any functional changes to the deCONZ api or behaviour. The only change visible on the api side is that the additional fields are returned when retrieving lightstates for a scene. All previous fields should also still be included.

### DIFF
--- a/database.cpp
+++ b/database.cpp
@@ -96,6 +96,10 @@ void DeRestPluginPrivate::checkDbUserVersion()
     }
     else if (userVersion == 6)
     {
+        updated = upgradeDbToUserVersion7();
+    }
+    else if (userVersion == 7)
+    {
         // latest version
     }
     else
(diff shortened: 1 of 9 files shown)


manup:
Hey, thanks for the PR, overall it looks good and feels way cleaner compared to the former implementation. Due the amount of changes I won't merge it for the 2.05.73 since we need more testing with various lights and brands. So hopefully it will be added in 2.05.74.

KodeCR:
Thanks for the review. Ok, that’s fine, I appreciate a bit more testing being done first.

KodeCR:
Hi @manup, did you have a chance to do some testing? I'm keen to get this in before there are too many conflicts (e.g #2553 might create some). Thanks.

KodeCR:
I would really like this to be merged, I started this work over a year ago. What can I do to help? Could perhaps someone else review and test?

KodeCR:
@SwoopX and/or @ebaauw, could you perhaps review? I'm happy of course to resolve conflicts if needed?

Mimiix:
@KodeCR added backlog label so the bot stops bothering you.

SwoopX:
@KodeCR I'm afraid I won't be much of a help here since I don't use any scenes at all. However, I think it would be a pitty if there are good improvements and we potentially give them away...

Thomas-Vos:
I need these extra attributes for the next steps with Hue Essentials on a Phoscon bridge.

Issue #2304, opened 2020-01-10
Ikea Color Bulbs don't react to Alexa color changes - the second


acid115:
Hi, currently i have an issue, that my Ikea Color Bulb ignores color changes, triggered by an Amazon Echo 3.gen (through speech or app)
Error is always: "malfunction of the device". Switch on/off and Change brightness works.

The author of the issue #1866 (https://github.com/dresden-elektronik/deconz-rest-plugin/issues/1866) solved it by installing 2.05.69 and repairing the bulbs.

i am currently on 2.05.71 / 26330500 with my ConBee I
Bulb: 
IKEA of Sweden
Modell TRÅDFRI Lampe E27 CWS opal 600lm
Version 1.3.009

Yesterday i have deleted the bulb (in Phoscon and Alexa App), reset the bulb and reconnect the bulb.
the bulb got a new id, but the error didnt disappear. 

i have debugged the little bit and can see, that the amazon echo sends this command to the rest api:
"on": true
"hue": 18699 (example)
"sat": 254

but if a send this command though an browser oder curl request, the bulb do not change the color. The color is only be changed, when i send x,y values:
example:
"on": true
"bri": 200
"sat": 254
"xy": [0.2322,0.6422]

so, how can deconz handle hue-values for ikea bulbs ?
i see in the debugging infos, that the api calculates x and y in rest_lights.cpp, when my echo sends the "hue"-request and sets the x- and y-values. 
But the color will not change.
maybe there should also set the correct colormode ?
Does anyone have the same issue with an ikea color bulb ?

Pirlobeck:
Hallo,
I can confirm the same behavior when using an Alexa Echo Dot Gen2. When I  change the color of my Osram an Philips bulbs, the color change works, but I get the answer from Alexa "it looks like, that the bulb does not work proper!"

acid115:
I have found the problem and now i am looking for a solution.
I am now able to compile the plugin and can change the source.
I have tried to calculate the xy values from the given hue and sat.
There are some lines the code where the xy is calculated (Hsv2Rgb,Rgb2Xy) but the result is wrong. These are not the xy values for the given colors / hue values.
I set the wrong calculated xy-values for the light, that is ok. But the color is wrong.
So i am trying to find out the right calculation for hue - xy transformation ...

ebaauw:
[...]

PR #2402, opened 2020-02-01
correct Hsv2xy calculation if hue and sat value is set and handling of ikea color and ct  bulbs in combination with amazon alexa

acid115:
correct Hsv2xy calculation if hue and sat value is set.

changed handling for ikea bulbs in combination with amazon echo requests (echo requested hue and sat, ikea bulb only unterstand xy, but hue and sat has to been set for rest api state requests, else the echo responds with error)
correct ct (color temp) for ikea bulbs in combination with echo request.


acid115:
The issue discussion you can find here: #2304

manup:
Thanks for the PR, we have also observed the issue and it was reported recently in our support channel. I wonder though if the Amazon Echo check is needed, since basically it should affect all API users which control the IKEA lights via this method.

Can you please also heck the code formatting to only use 4 spaces indentation for consistency.

acid115:
Hi,
sorry for my late answer...
i am a little bit busy :D 
and have to learn so much about github.
i think now, that i have it under control and update my fork and sync it with the master.
i made the changes you wish.
i think, die check for Echo-Api is necessary for the "yellow-bug", because the origin wrong hue value came from the echo. But I dont have a color map...i think it is a AmazonEcho->IkeaColorBulb Problem...
the same for the correction of the colortemp for ikea bulbs in connection with amazon echo.
i think the echo is responsible for this bug, so i decided, that only the for echo requests it should correct the ct-value.
can you re-check the code ?
thx

Sneeg:
This is not only relevant for IKEA bulbs but also INNR bulbs show some weird behaviour with some colors. Also, the error "the device Light is not responding" occurs even with Philips Hue lights for me when changing colors (even though the color changes properly, but Alexa still gives this error). 
Would really love to see this fixed soon. If I can help somehow, let me know. Have many different lights here that I could test with (Innr, Osram, Philips...).

acid115:
Hmm, are you sure?
Why no one has report this major error until now? 
The echos 2nd and 3rd generation are available for at least 1 year
And color bulbs too...
So this must be a major error for one year or longer.... I cant believe this.
But i will think about it
The problem is, if your philips color bulb changed the color you requested correctly, it seems not affected from the wrong calculation of rgb, so another method/function in code is also the problem... 
Or did you test my code changes with your philps?
Important for the echo is to not throw an error, that every requested value to change (on, sat, hue, ct, bri...) has to changed in the lightnode (the values under /api/lights/id/ ) too, not only in Response, because the echo diffs the values

Sneeg:
I tested your code changes with my philips bulbs, yes. They had the same problem that "yellow" actually turned to blue without the changes. The only thing I would omit from your commit is the `ct += 65;` because that leads to slightly too bright colors, red for instance. All the other changes are definitely relevant for the other vendors, too.
I have no idea why this has not been fixed or even reported yet, maybe there are not so many people using deconz in combination with Alexa, but actually I would have expected the devs to be aware of this already, I mean, they sell a relatively expensive product (compared to the alternatives) that is advertised with Alexa support etc., but no one seems to have tested this thoroughly. It's a little disappointing tbh.

freekode:
Could you please resolve merge conflicts and continue work on this PR?
Looks like I have the same problem with IKEA bulbs

acid115:
Sorry, at the moment i have no time for resolving the conflicts.
And i dont agree with the omit of my fix for ct +=65
At First i have to merge all changes since february and i know there are some issues that causes losts of devices. And i have no time to rebuild my zigbee network.
I have some special rules integrated, that cant be configured with the webapp.
And some extra changes in code to hide some devices from my echo devices :) (just an -noecho- in the devicename). But no one needs this and it should be integrated in the device configuration webpage, but it works for me. And if someone gets access to my amazon account, he can not switch off my cam or my raspberry :)
And i want to integrate a feature so run shell scripts, if a button has been pressed (ikea switch...)

So, you can see my changes and someone maybe can merge these changes in the current code.
At the moment i say: never change a running system, because i have no time to fix it and the waf (wife acceptance factor) is very low

Issue #2257, opened 2019-12-29
Incorrect `state.bri` and `state.sat` value of 255 for window covering devices


ebaauw:
For window covering devices, a _Current Position Lift Percentage_ value of 100 is translated to a `state.bri` value of 255 (instead of 254).  Same for _Current Position Tilt Percentage_ and `state.sat`.  Setting `state.bri` to 254 in the API doesn't fully close the blind, where setting it to 255 does.  Both should be handled the same (as for lights).

See https://github.com/ebaauw/homebridge-hue/issues/586.

ebaauw:
Fixed in v2.05.73.

PR #2409, opened 2020-02-03
Support state.ontime in combination with state.alert

ebaauw:
Support `state.ontime` in combination with setting `state.alert` to `lselect` or (for Warning Devices) to `blink`.  The default duration of the warning (300s) or identification (15s) can be overwritten using `state.ontime` (in seconds).
See https://github.com/ebaauw/homebridge-hue/issues/614.


ebaauw:
Second commit is follow-up to #2257: two more places where the position from the `lumi.curtain` is translated to `state.bri`: the Xioami special attribute report and the _Analog Output_ cluster.

Issue #2305, opened 2020-01-10
Support for Lutron Aurora (Friends of Hue dimmer) 


TFenby:
Looking to either supplant or re-open #1672.

Looks like the necessary info is in that issue, but please let me know if I need to provide more!

carver7:
I would also appreciate having this switch supported:

- It's the only switch that renters can fasten onto an existing hard-wired light switch while simultaneously immobilizing it in order to prevent users from cutting off power to the light bulb socket out of habit
- It screws onto the most common in North America with no electrical modifications required, no additional zigbee switches are needed to stick to the wall, it looks like a common traditional N. American dimmer switch.

Some screenshots of a test setup in deConz with a Conbee II (if more info needed, please let me know) :

![NodeInfo](https://user-images.githubusercontent.com/18132891/72235811-cf281380-3599-11ea-908f-27e401581695.PNG)

![Cluster_Basic](https://user-images.githubusercontent.com/18132891/72235833-e2d37a00-3599-11ea-967d-c4ee5f963719.PNG)

![Cluster_PowerConfig](https://user-images.githubusercontent.com/18132891/72235845-ecf57880-3599-11ea-9fef-0037233dccd3.PNG)

![Cluster_Identify](https://user-images.githubusercontent.com/18132891/72235852-f1219600-3599-11ea-927d-f238e1d5d003.PNG)



Related requests for Lutron Aurora dimmer switch:
https://github.com/dresden-elektronik/deconz-rest-plugin/issues/1672
https://github.com/ebaauw/homebridge-hue/issues/522

TFenby:
I could also have one shipped to a dev if that'd help. :)

ebaauw:
Just to be sure: this is a wireless switch that covers the 20th century wall switch, but doesn’t need any wiring?  Does it work without being installed/attached to the wall switch?  I wouldn’t be able to install it due to my EU mains power and wall switches.

Otherwise I’ll be happy to have a look at it.  I haven’t yet gotten my hands on a ZLLRotary switch.  As mentioned in the homebridge-hue issue: not sure how to expose these to HomeKit, but if we can get web socket notifications for each rotary event, it would work quite well as a _Stateless Programmable Switch_ with automations for setting _Change Brightness_ from homebridge-hue, or _Change Volume_ from homebridge-zp.

francoiscote:
I just bought a couple of those also, so I am interested in deconz support and will try to help as much as I can.

[...]

PR #2446, opened 2020-02-13
Hue dimmer switch and Lutron Aurora Friends-of-Hue dimmer switch

ebaauw:
See #2305.

### DIFF
--- a/de_web_plugin.cpp
+++ b/de_web_plugin.cpp
@@ -206,6 +206,7 @@ static const SupportedDevice supportedDevices[] = {
     { VENDOR_120B, "WarningDevice", emberMacPrefix }, // Heiman siren
     { VENDOR_120B, "Smoke", jennicMacPrefix }, // Heiman fire sensor - newer model
     { VENDOR_LUTRON, "LZL4BWHL01", lutronMacPrefix }, // Lutron LZL-4B-WH-L01 Connected Bulb Remote
+    { VENDOR_LUTRON, "Z3-1BRL", lutronMacPrefix }, // Lutron Aurora Friends-of-Hue dimmer
     { VENDOR_KEEN_HOME , "SV01-", keenhomeMacPrefix}, // Keen Home Vent
     { VENDOR_INNR, "SP 120", jennicMacPrefix}, // innr smart plug
     { VENDOR_PHYSICAL, "tagv4", stMacPrefix}, // SmartThings Arrival sensor
@@ -609,8 +610,16 @@ void DeRestPluginPrivate::apsdeDataIndication(const deCONZ::ApsDataIndication &i
             handleIasZoneClusterIndication(ind, zclFrame);
             break;
 
-        case DE_CLUSTER_ID:
-            handleDEClusterIndication(ind, zclFrame);
+        case VENDOR_CLUSTER_ID:
+        // case DE_CLUSTER_ID:
+            if (zclFrame.manufacturerCode() == VENDOR_PHILIPS)
+            {
+                handlePhilipsClusterIndication(ind, zclFrame);
+            }
+            else // Shouldn't we check for DE manufacturer code?
+            {
+                handleDEClusterIndication(ind, zclFrame);
+            }
             break;
 
         case XAL_CLUSTER_ID:
@@ -3170,7 +3179,8 @@ void DeRestPluginPrivate::checkSensorButtonEvent(Sensor *sensor, const deCONZ::A
         checkReporting = true;
         checkClientCluster = true;
     }
-    else if (sensor->modelId().startsWith(QLatin1String("RWL02"))) // Hue dimmer switch
+    else if (sensor->modelId().startsWith(QLatin1String("RWL02")) || // Hue dimmer switch
+             sensor->modelId().startsWith(QLatin1String("Z3-1BRL"))) // Lutron Aurora Friends-of-Hue dimmer switch
     {
         checkReporting = true;
     }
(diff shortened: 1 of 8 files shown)


de_web_plugin.cpp
```diff
@@ -609,8 +610,16 @@ void DeRestPluginPrivate::apsdeDataIndication(const deCONZ::ApsDataIndication &i
...
 
-        case DE_CLUSTER_ID:
-            handleDEClusterIndication(ind, zclFrame);
+        case VENDOR_CLUSTER_ID:
+        // case DE_CLUSTER_ID:
+            if (zclFrame.manufacturerCode() == VENDOR_PHILIPS)
+            {
+                handlePhilipsClusterIndication(ind, zclFrame);
+            }
+            else // Shouldn't we check for DE manufacturer code?
```
manup: Yes could be added (0x1135), or just remove `handleDEClusterIndication()` seems like the function isn't doing much anyway :)

manup:
Cool thanks for the PR. I like that the Philips switches have their own function now. Should I merge this for 2.05.73 (release scheduled today) or better wait for 2.05.74?

I currently have no Philips Dimmer here for testing, the Aurora is ordered but hasn't arrived yet.

ebaauw:
Please merge, of possible.  Have tested the Hue dimmer extensively myself.

I'm really stuck getting the Lutron Aurora to behave like when connected to the Hue bridge.  I would appreciate any suggestions here.

de_web_plugin.cpp
ebaauw: Less is more?  Can you remove it on merge, or do you want me to do that?  I'm a bit uncomfortable messing with code I don't understand and cannot test (I don't have any DE devices).

manup:
> Please merge, if possible. Have tested the Hue dimmer extensively myself.

Merged :) 

> I'm really stuck getting the Lutron Aurora to behave like when connected to the Hue bridge. I would appreciate any suggestions here.

I'll have a look when the switch arrives here, perhaps the switch is missing some command which the Hue bridge sends but I've also seen you already tried a lot to align the behaviour. MAC address and NWK address (hue bridge is using != 0x0000) are the only thing we can't fake without breaking installations.

> Added commit for IKEA SYMFONISK controller: buttonevent values for Up vs Down were reversed; all other switches have up before down.

The team is currently playing with these for Phoscon App integration, quite nice devices.

ebaauw:
> perhaps the switch is missing some command which the Hue bridge sends but I've also seen you already tried a lot to align the behaviour.

The bridge sends regular group commands from 0xFC01, which I hadn’t seen before, see the capture.  However, not during the pairing with the dimmer.  I suspect these are Hue Entertainment related, but I’ve never looked at this (other than adding support to turn it off from homebridge-hue).

> MAC address and NWK address (hue bridge is using != 0x0000) are the only thing we can't fake without breaking installations.

The bridge is using NWK 0x0001.  I havent’t tried that yet, I would probably need to configure another RaspBee/ConBee as router, or use an XBee (for which I’m nowhere near ready to write an application).  My fear is that the dimmer somehow recognises that it’s joined a ZLL network, and uses that to trigger the behaviour.  Last time I tried, the ZLL link key wouldn’t stick in the settings, and we could no longer join the RaspBee as a router to the Hue bridge network.

> The team is currently playing with these [SYMFONISK controllers] for Phoscon App integration, quite nice devices.

I’m using one to control my Sonos speakers through deCONZ - homebridge-hue - HomeKit automations - homebridge-zp.  It’s not as smooth as using the Trådfri hub, but works remarkably well.  The Trådfri hub actually translates the ZigBee commands to SOAP/HTTP calls to the Sonos zone players, basically implementing Sonos app functionality (like homebridge-zp).  
From memory, the controller sends _Step_ and _Stop_ commands when turning the dial, with a variable rate, somehow linked to the speed and/or angle of the turn.  Would be nice to map that to ZHARelativeRotary `state.expectedrotation`, instead of exposing virtual _Right Turn_ and _Left Turn_ buttons.  This would enable rules with a `gt` or `lt` condition on `expectedrotation` to set different values to `bri_inc`, and even more sophisticated logic in home automation systems.
I also want to map the analog cluster of the Xioami magic cube to ZHARelativeRotary, so `state.buttonevent` would always follow the x00y scheme (at least for ZHASwitch; ZGPSwitch is a different story...).

> 2.05.73 (release scheduled today)

Looks like another long day for you.  I hope you’re doing OK.  Please let me know (here or by email) if I can help.  We need to find a way to materialise the C4 model for API v2, separating API plugin releases from the core programme and Phoscon, if only to give you and the team some breathing space.

manup:
>Looks like another long day for you. I hope you’re doing OK. Please let me know (here or by email) if I can help. 

Thanks really appreciate that :) the 2.05.73 is online now in the beta channel, I'll wait 1–2 days before releasing it in stable.

Things getting more relaxed in recent days, internally we've managed to fix some bottlenecks and distribute various tasks like website maintenance and device integration through the team.

So API v2 can start to evolve in the coming days :)

>We need to find a way to materialise the C4 model for API v2, separating API plugin releases from the core programme and Phoscon, if only to give you and the team some breathing space.

That's a good idea for packaging we can also utilise the automatic build process which basically builds different channels (currently HEAD aka alpha, beta, stable) based on git tags for various platforms. Currently I manually start the build server after updating the tags every 1–2 months which is too long. At least the beta branch could be updated more often. This could also be done daily or as commit hook. The whole process works only on the master branch.

I'd also like to extend PR merging powers to selected contributors so that for example you can merge your changes and tags into the official repository, if that is ok. This way HEAD and beta can have the same pace as PR's with proper packages build in a timely manner (it would also include most recent version of the Phoscon App which is updated daily).

PR #2472, opened 2020-02-19
API support for RaspBee, ConBee, ConBee II, and XBee

ebaauw:
Several enhancements:
- Add definitions for DE cluster on RaspBee, ConBee, ConBee II and Digi Drop-In-Networking cluster for XBee;
- Implement _Basic_ cluster attributes for the gateway node.
  - Fix _ZCL Version_ and 0xF000 Legrand attribute;
  - Add _Application Version_, _Stack Version_, _HW Version_;
  - Add _Manufacturer Name_: set to `dresden elektronik`;
  - Add_ Model Identifier_: set to device name;
  - Add _Power Source_: set to _DC Power_;
  - Add _SW Build ID_: set to device firmware version.
- Add a `/lights` resource for a RaspBee, ConBee, ConBee II, and XBee, cf. the Trådfri repeater.  The `state` object only contains `reachable`, as there's nothing to set for these devices, but you can now set the name for these nodes through the API.
- Cleanup use of hex vendor codes, see #2315.

With these changes, the gatway and XBee now look like:
Screenshot 2020-02-19 at 20 29
Note that I set the names through the API.  Also note I added the _Basic_ (and _Time_ clusters for the gateway through the GUI, in the _Endpoints_ tab of the _deCONZ Network Settings_ window.

The REST API resources should be created when searching for new devices through the API.  If not, read the _Basic_ cluster (on the gateway node) or read the descriptors (on the gateway and XBee).  To pair the XBee, configure it as follows (from https://www.dresden-elektronik.de/faq.html#deconz):

Abbr. | Value | Comment
-- | -- | --
ID | 0 | extended pan id will be discovered during network search
SC | 1FFE | scan on all channels for a network
ZS | 2 | ZigBee Stack Profile 2 (ZigBee PRO)
EE | 1 | Security
EO | 3 | Security
NK | 0 | Network Key will be transferred from trust center (or set here if known)
KY | 5a6967426565416c6c69616e63653039 | HA default trust center link key:

The resources look like:
```json
{
  "etag": "0d07d6a775bdd618a99ec3140cf5fdf9",
  "hascolor": false,
  "manufacturername": "dresden elektronik",
  "modelid": "ConBee II",
  "name": "ConBee II",
  "state": {
    "reachable": true
  },
  "swversion": "0x264a0700",
  "type": "Configuration tool",
  "uniqueid": "00:21:2e:ff:ff:03:d4:a7-01"
}
```
```json
{
  "etag": "06c98905e857fd0f80c52f20a6cec701",
  "hascolor": false,
  "manufacturername": "Digi",
  "modelid": "XBee",
  "name": "XBee 2",
  "state": {
    "reachable": true
  },
  "swversion": null,
  "type": "Range extender",
  "uniqueid": "00:13:a2:00:41:83:1c:4d-e6"
}
```


manup:
Ha that's an interesting one :) but what can we do with it?

ebaauw:
> Ha that's an interesting one :) but what can we do with it?

Almost the same as what we can do with the Trådfri repeater (except for making it blink): change the device name.  Helps me in identifying the right XBee or 2nd Raspbee, configured as a router.  We needed the _Basic_ cluster on the RaspBee/ConBee/ConBee II to support the Legrand devices; this kinda seemed like the logical next step.

Someday, I might actually add some real API support for the XBee, but not anytime soon.

manup:
>We needed the Basic cluster on the RaspBee/ConBee/ConBee II to support the Legrand devices; this kinda seemed like the logical next step.

Agree, perhaps later the v2 API should contain a section for these kind of devices. The XBee has quite some functionality to offer which would be nice to has it accessible via API.

Issue #1861, opened 2019-09-14
TRADFRI bulb E26 WS opal 980lm reports color mode as "xy"


glward:
The REST api seems to be reporting this bulb as color mode "xy" even though it's only a color temperature bulb:

```
{
    "ctmax": 454,
    "ctmin": 250,
    "etag": "da6d097ee7330bd24d91488002420aef",
    "hascolor": true,
    "manufacturername": "IKEA of Sweden",
    "modelid": "TRADFRI bulb E26 WS opal 980lm",
    "name": "Writing Desk Lamp",
    "state": {
        "alert": "none",
        "bri": 254,
        "colormode": "xy",
        "ct": 250,
        "on": true,
        "reachable": true
    },
    "swversion": "1.2.217",
    "type": "Color temperature light",
    "uniqueid": "00:0b:57:ff:fe:dd:47:63-01"
}
```

Setting the color temperature still works correctly, both via the REST api and the webapp, but it seems odd. Attempting to change the color mode over the REST api does not return _anything_, no errors or success:

```
glward@ares:docker/ha$ curl -H "Content-Type: application/json" -X PUT "http://IP/api/KEY/lights/8/state" -d '{"colormode": "ct"}'
glward@ares:docker/ha$
```

sub0ne:
I can confirm this. When I query my bulbs state I got this in return:

{ 
   "ctmax":65535,
   "ctmin":0,
   "etag":"1fccf86c78b24a8414433d4f4468960e",
   "hascolor":true,
   "manufacturername":"IKEA of Sweden",
   "modelid":"TRADFRI bulb E27 WS opal 1000lm",
   "name":"Kellerleuchte 1",
   "state":{ 
      "alert":"none",
      "bri":254,
      "colormode":"xy",
      "ct":100,
      "on":true,
      "reachable":true
   },
   "swversion":"2.0.022",
   "type":"Color temperature light",
   "uniqueid":"00:0d:6f:ff:fe:03:b1:1d-01"
}

Kane610:
@manup can we get a comment on this?

manup:
I think since these are are plain color temperature lights we can fix the color mode to `ct`.
It's similar to the LEDVANCE bulbs issue https://github.com/dresden-elektronik/deconz-rest-plugin/commit/b30f4f8a493cf6d65bc3581fb96bb2a7978f38d7

Looking forward to get this fixed for 2.05.70.

[...]

Issue #1688, opened 2019-07-18
RFC REST-API v2


manup:
## Why
I think it's time that we need to figure out a proper v2 REST-API to support a wider range of devices which are available today or might arrive in the future. I'm increasingly unhappy about the /lights hack which is currently used in some places to add devices which don't fit in the provided categories.

A first humble attempt to address more complex devices was the _/devices_ endpoint which allows querying all sub-devices behind one physical device referenced by its MAC address. https://github.com/dresden-elektronik/deconz-rest-plugin/commit/96daf83a87e38be60978283bb375f3365be4ef2a

## Request For Comments
To proceed I invite everyone to share ideas and wishes how a new interface may look like.
Beside this GH issue the [RFC REST-API v2](https://github.com/dresden-elektronik/deconz-rest-plugin/wiki/RFC-REST-API-v2) wiki page should be used to develop the new API.

### Initial thoughts

* The current API v1 is used by several home automation projects, therefore v2 should be implemented as a parallel interface in order to not break any v1 client.
* The v2 API should be considered completely independent from v1, it should be the best possible API and not a workaround or compatibility compromise.
* A proper security scheme should be integrated. HTTPS can't be used due certificate complexity. However using widely available libsodium we can have end-to-end encryption, authentication and identity management without much hassle.
* The internal `Resource` and `ResourceItem` C++ classes are a good common ground to build a generic API on top. These are already used in many places and simplify the implementation of the rules engine. For v2 they are likely to be extended to be more versatile.
* deCONZ and the REST-API is now nearly 10 years old with some historically grown code. As a by-product of v2 new code must be designed and implemented with a proper architecture to escape spaghetti land. The code and many processes like joining new devices could be such much better than it is right now. By now we have so much knowledge about various devices, what can go wrong and how-to workaround issues manually, that's punishment of the user and should be handled by code. The related architecture and design decision should be thoughtful developed with RFC documents.
* The implementation of various switches have shown that a general notion of capabilities might be desirable.
* The v2 interface should be as high level as possible, clients shall not be required to understand any Zigbee specific knowledge. However an optional low-level v2 interface can be offered in parallel to provide full access to Zigbee world.
* Currently Websockets are used to provide real time events, MQTT can be offered as well. Qt has native support for it so the implementation is simple (in fact there exists already a very basic prototype for it).

ebaauw:
Love where this is going!  My initial thoughts below.

[...]

Issue #629, opened 2018-06-08
Ikea Tradfri: Color temperature not changing with scenes (Phoscon)


brandungskieker:
I created two scenes in phoscon app:

Brightness: 100% / Color temperature: minimum level (cold)
Brightness: 100% / Color temperature: maximum level (warm)
and saved both.
If the bulb is already in cold color switching to (2) works fine. But not otherwise, the temperature stays in warm mode.

Setup:

Raspberry Pi 3
RaspBee Version 2.05.28 / 3.5.2018, Firmware 26210500
TRADFRI bulb E27 WS clear 950lm (1.2.217)

manup:
For some reason the IKEA bulbs don't restore the color temperature in the scene, it seems to the a light firmware issue. If it get fixed in the light firmware we can OTA update the lights.

ebaauw:
Are you sure it's the Trådfri firmware?  See https://github.com/dresden-elektronik/deconz-rest-plugin/issues/53.

manup:
We may are able to build a workaround, but there sure is a bug in the firmware.
Just testet it with my IKEA GU10 tunable white in plain deCONZ scenes cluster

scene 1)
* setup light as warm white brightness ~70%
* store scene 1 in group 1 (unicast)

scene 2)
* setup light as cool white brightness ~50%
* store scene 2 in group 1 (unicast)

On store scene command the light should store the scene locally with current settings.
But when I recall the scenes only the brightness is restored and color temperature is always cool white.

So at least the `ZCL store scene` command in the firmware has a bug.
However a scene can also be created with `ZCL add scene` command which also carries the parameters (on, bri, x, y ...) this might work, I haven't testet it yet but there is already code in the REST plugin which we can use for that.

Not also due some weird reason in the specification scenes don't have a concept of color temperature, in a scene the color temperature is always stored as x,y color. Therefore the add scene command should contain the x,y color of the color temperature (point of blackbody curve).

manup:
A few tests later: add scene doesn't work, but when instead using move to color temperature command the move to color x,y is used, the `ZCL store scene` commands works as it should.

The problem is that the move to color temperature command doesn't update the local x,y values in the light (as it should).

So as a first fix, I'll try to convert move to color temperature commands into move to color x,y if the it is a unicast command to an IKEA light (so won't effect other lights).

manup:
Verified with IKEA GU10 WS 400lm tunable white and test firmware version 1.2.22.

[...]

PR #2507, opened 2020-02-28
Fix for issue 1861

sverleysen:
Commenting out this line of code should fix this issue.
https://github.com/dresden-elektronik/deconz-rest-plugin/issues/1861

### DIFF
--- a/zcl_tasks.cpp
+++ b/zcl_tasks.cpp
@@ -382,7 +382,7 @@ bool DeRestPluginPrivate::addTaskSetColorTemperature(TaskItem &task, uint16_t ct
         bool useXy = supportsXy && !supportsCt;
         // IKEA lights need to use "xy" because move to color temperature is broken and
         // won't update x,y values resulting in broken scenes.
-        useXy = useXy || task.lightNode->manufacturerCode() == VENDOR_IKEA;
+        //useXy = useXy || task.lightNode->manufacturerCode() == VENDOR_IKEA;
 
         if (useXy)
         {


manup:
Thanks for the PR, does the ct mode then still work for the IKEA CWS color light?

sverleysen:
I don't know, I don't have these lights. This change I have done is a change of @Flurkmark. He is using this change for a while now without issues.

lbschenkel:
I do have these lights, and I had introduced that code you're commenting it out. Commenting it out breaks the `ct`->`xy` transformation, which makes `ct` not work for the CWS lights.

I'm not quite sure if I understand what is the original issue, maybe I can recommend an alternative fix and help testing it (I have both WS and CWS lights).

sverleysen:
Initially a CT light bulb is correctly processed as a CT bulb, but when changing the color temperature using the REST API (what happens when you use Home Assistant), Deconz reports the same light bulb as a XY bulb. 
Which results in not be able to get the current color temperature in Home Assistant. When changing the color_temp using the Deconz UI, the light is again a CT bulb and the color_temp attribute is available again in Home Assistant

sverleysen:
ezgif com-optimize

lbschenkel:
Thank you. I also use Home Assistant. I will try to replicate and get back, and hopefully propose some alternative.

sverleysen:
> Thank you. I also use Home Assistant. I will try to replicate and get back, and hopefully propose some alternative.

Any success in replicating the issue?

lbschenkel:
Hi, sorry for the delay. With the coronavirus and some other personal stuff going on, I couldn't take a look at it. But now I have.

I could replicate your video: When I turn on the light on with the physical switch, I do see the `color_temp` attribute. But then if I change the color temperature *of the light* (unicast), either via HASS or Phoscon, the attribute disappears. If I change it via the *group* (multicast), then it reappears.

The reason for why this happens is because IKEA lights have a firmware bug in which they don't remember the color temperature in scenes, unless the colour is set via `xy` [1][2][3]. That's why the use of `xy` is forced for all IKEA lights, but only on unicast commands.

On top of that, I introduced [4] to use the same mechanism to allow using `ct` commands on lights lacking the `ct` capability but supporting colour (like the IKEA CWS lights), by translating `ct` into `xy`.

Commenting out the line you proposed in this PR would cancel both behaviours I mentioned above, which would be a pretty annoying regression. In particular, saving scenes with IKEA lights will become impossible (unless they recently fixed this bug, but I saw no evidence of that).

Now, I totally understand this is frustrating and causes this side effect. I think this could be fixed by making deCONZ pretend it sent the original `ct` command and update the REST API status accordingly, or since it supports emulating `ct` anyway for lights that lack this capability, do the opposite translation and report a `color_temp` for all lights.

I'll need to dig into the source code again and check what's feasible. @manup, @ebaauw: any other suggestions?

[1] https://github.com/dresden-elektronik/deconz-rest-plugin/commit/4a43cb9193f0ab262bb8d6bb0f462f854e1da87c
[2] https://github.com/dresden-elektronik/deconz-rest-plugin/issues/629
[3] https://github.com/dresden-elektronik/deconz-rest-plugin/issues/53
[4] https://github.com/dresden-elektronik/deconz-rest-plugin/pull/719

ebaauw:
I'm afraid, you don't want to ask me in this case; I'm no fan of this functionality.

See my post on the API v2 (#1688) about a "raw" vs a "cooked" API:
>I like the idea of a layered architecture, where the "raw" low-level v2 API would be RESTifying the ZCL commands and attributes (cf. the level of the GUI interface and the CLI plugin), on top of which which the high-level v2 API would provide the "cooked" home automation features (caching values, rules, ...). We might even consider developing these layers as two separate repositories. On the long run, the high-level API v2 might be used with implementations of the low-level API v2, e.g. using the RaspBee/ConBee serial interface to the radio device, or using other hardware (e.g. XBee). And the Hue-compatible v1 API might eventually be based on the same low-level v2 API.

 I think the current API is more "raw" than "cooked", and should stick to exposing the lights as they are, warts and all.  Imho, introducing "cooked" features like mapping `ct` to `xy` in the current code is doing more harm than good: it causes too much confusion and is hardly maintainable, lacking a proper capability model.

lbschenkel:
> I'm afraid, you don't want to ask me in this case; I'm no fan of this functionality.
> 
> See my post on the API v2 (#1688) about a "raw" vs a "cooked" API:

I have read it back then and I actually agree with you. However, the current situation is what it is, and the 2nd cooked layer does not currently exist.

The mapping from `ct` to `xy` was introduced by @manup to work around the IKEA firmware bug: if that is disabled, scenes will simply stop working for any IKEA white spectrum lights (in the sense of not saving properly) [*]. This will be a regression big enough to be unacceptable to many users, myself included. What else can we do in the current set of circumstances?

If we can't find a way for the REST API to keep the firmware bug workaround, the only other thing I can think of is that we add a way to enable/disable "quirks" from the command line, so the set of users who don't care about scenes can disable (let's say) `quirk.ikea.ct_to_xy`.

[*] the other circumstance, when the light does not support `ct` but does `xy`, is actually not kicking in nor it is a problem in this particular situation

ebaauw:
> What else can we do in the current set of circumstances?

I appreciate the dilemma.  I’m sorry, I don’t have an answer.  If memory serves, there was a similar problem for OSRAM colour temperature lights, but they don’t accept xy commands.  I think they worked correctly when storing a scene from the current light state, but not when creating a scene while specifying the state (I need to brush up on the _Scenes_ cluster commands).

lbschenkel:
> > What else can we do in the current set of circumstances?
> 
> I appreciate the dilemma. I’m sorry, I don’t have an answer. If memory serves, there was a similar problem for OSRAM colour temperature lights, but they don’t accept xy commands. I think they worked correctly when storing a scene from the current light state, but not when creating a scene while specifying the state (I need to brush up on the _Scenes_ cluster commands).

Interesting. Unless I'm misremembering it, the IKEA lights have the opposite bug in which they don't correctly store a scene from the current light state (which is the way you do it via deCONZ), unless the current light state was set via `xy`.

*If* the lights work correctly when you use "create scene", then it's possible that the workaround could be implemented for the scene creation step instead. This could have the additional benefit of working properly even when you set the light state via group commands, contrary to what happens today.

ebaauw:
I replaced the IKEA and OSRAM bulbs on my production network with Hue bulbs (because of the routing problems), but I still have them lying around, somewhere.  I’ll see if I can do some tests over the weekend.

> which is the way you do it via deCONZ

I think deCONZ also supports the other command, when you update the scene state.  One of the hidden features in the API:
https://github.com/dresden-elektronik/deconz-rest-plugin/blob/ca72c46fd147dcf5c0b04304a96316d2c97ff477/rest_groups.cpp#L93

lbschenkel:
Off-topic P.S.: This reminds me of an old idea that I have which is to introduce a device database to deCONZ.

This device database will have entries mapping from a set of rules (MAC prefix, vendor, capabilities, firmware version, etc.) to a specific device entry. A device entry will have general information such as the device name, manufacturer, but most importantly: a set of quirks that apply.

Then in every case in the code in which it has a special case or alternative behaviour guarded by `if(macPrefix == ...)` or `if(manufacturer == VENDOR_IKEA)` could be substituted by something like `if(dev.quirks.force_ct_to_xy)`. I feel that this would immensely improve the maintainability of the codebase, plus minimize or eliminate the "whitelisting dance" that is currently done for new devices.

Because at the beginning deCONZ might not have enumerated enough data to have an ideal mapping, every time new information is discovered that results in the device entry changing (let's say: from "IKEA light" to "IKEA GU10 WS light (firmware 3.x)" then the current best entry is saved to the DB.

Device entries can be listed in order of most specific to most generic, and the first matched entry wins. So we could have a separate entry for lights with newer firmware that has less/different quirks than the same light with an old firmware. To make it manageable, we could also allow entries to specify a parent entry, so they inherit all the parent's quirks unless the device forcefully disables the quirk (example: "IKEA GU10 WS light" can inherit "IKEA WS light" which inherits from "IKEA light").

ebaauw:
Indeed, this is what I mean by capability model.  Preferably, I would be a JSON file, so new devices can be supported without code changes.  I have something like that in Homebridge Hue, but very rudimentary and still in code.  You can specify default quirks per manufacturer, overriding them per model, if needed.
https://github.com/ebaauw/homebridge-hue/blob/b8a7aaf25f2a5ef98199399c9c5f15db656bf0ee/lib/HueLight.js#L31

lbschenkel:
> Indeed, this is what I mean by capability model. Preferably, I would be a JSON file [...]

Agreed. If the DB is in a external file, it also immensely lowers the barriers for contributors. Code changes will be required only to implement entirely new quirks, but not to enable/disable existing quirks for certain devices.

ebaauw:
> I’ll see if I can do some tests over the weekend.

This turned out to be more challenging than I imaged.  Summary:
- The ZCL spec doesn't cater for _Color Temperature_ in a scene state; different lights handle this very differently;
- The _Store Scene_ command works consistenly across all _Color temperature lights_ I've tested (Philips Hue, OSRAM Lightfy, IKEA Trådfri, icasa Filament);
- Not all lights support specifying the _Color Temperature_ in the scene state through _Add Scene_;
- Reporting the scene state varies greatly per model.  Updating the light states from the cached scene state on scene recall is very challenging.  Doing a "fast poll" after recalling a scenes seems really necessary;
- Reporting the scene state from the REST API plugin is hugely inconsistent;
- The REST API plugin implementation of PUT `/groups/#/scenes/#/lights/#/state`, resulting in an _Add Scene_ command, is buggy.  I haven't encountered any obvious bugs in the light firmware (other than different lights handling _Color Temperature_ in scenes differently).

Details:

The REST API plugin is hugely inconsistent for scenes:
- When doing a GET of `/groups` or `/groups/#`, the `scenes` are retuned as an array of objects, with an `id` key to indicate the scene ID;
- When doing a GET of `/groups/#/scenes`, the return value is an object, with the scene IDs as keys.  In each object, the `lights` are returned as an array of IDs;
- When doing a GET of /groups/#/scenes/#`, the `lights` are returned as an array of objects, with `id` indicating the light (resource) ID.  The state attributes are in the object; no separate `state` object.  `x` and `y` are separate keys, instead of a combined `xy` array;
- A GET of `/groups/#/scenes/#/lights` results in a "resource not available" error;
- Updating a scene is done through a PUT of `/groups/#/scenes/#/lights/#/state`.

The _Store Scene_ command sets the scene state to the current light state.  The _Add Scene_ and _Enhanced Add Scene_ commands contain the scene state in their payload.  The _View Scene_ and _Enhanced View Scene_ commands return the scene state, but the GUI doesn't display this, probably because the state is encoded in a kinda odd way.  
It scene state doesn't list the attribute IDs nor types.  It only lists the values (in "scene order" defined by the ZCL standard) and the total length per cluster.  For _On/Off_, the length is 1 and the values is that of the _OnOff_ attribute.  For _Level Control_, the length is also 1 and the values is that of the _Current Level_ attribute.  _Color Control_ is more complex.  The defined order seems to be: _Current X_ (two bytes), _Current Y_ (two bytes), _Enhanced Hue_ (two byes), _Saturation_ (1 byte), _Color Loop Active_ (1 byte), _Color Loop Direction_ (1 byte), and _Color Loop Time_ (2 byes), for a total of 11 bytes.  Note that _Current Color Temperature_ is missing here.  According to the standard, it's OK to leave out trailing attributes.  
On _Store Scene_, all supported attributes from all clusters seem to be copied to the scene state; on _Add Scene_, only the attributes specified in the payload.  So, using _Add Scene_, it is possible to create a scene that doesn't change brightness.  Not through the REST API, nor through the deCONZ GUI, though.

Different _Color temperature lights_ seem to handle color temperature in scenes very differently.  I've been testing the following lights:

Manufacturer | Type | Firmware | Color Capabilities | XY reflects CT
--- | --- | --- | --- | ---
Philips | LTW001 | 1.50.2_r30933 | 0x0010 | Yes
OSRAM | Classic B40 TW - LIGHTIFY | V1.05.10 | 0x0010 | No
IKEA | TRADFRI bulb E27 WS�opal 980lm | 2.3.007 | 0x0010 | No
icasa | ICZB-FC | 2.4.2 | 0x0010 | Yes

Setting _Color Temperature_ from the deCONZ GUI, following by a _Store Scene_:
- The Philips light updates _Current X_ and _Current Y_ to reflect the _Color Temperature_.  It stores these in the scene state.  The scene state only contains _Current X_ and _Current Y_;
- The OSRAM light reports fixed values for _Current X_ and _Current Y_.  It stores and recalls _Color Temperature_ in the scene state alright, but it doesn't report it.  It reports the full 11 bytes for _Color Control_ in the scene state, with fixed values;
- The IKEA light is very similar to the OSRAM, except that it reports only _Color X_ in the scene state, with a fixed value;
- The icasa updates _Current X_ and _Current Y_ to reflect the _Color Temperature_.  The scene state contains _Current X_, _Current Y_, and _Enhanced Hue_.  The first two are fixed at 0, the latter holds the _Color Temperature_ value.

Issuing _Add Scene_ from the GUI is also handled differently.  deCONZ issues a payload without any state attributes.  This is valid, if I read the ZCL spec correctly, but WireShark thinks it's a malformed packet.  The lights cannot seem to agree:
- The Philips light returns a _Default Response_ with status _Malformed Command_;
- The OSRAM light returns an _Add Scene Response_ with status _Insufficient Space_;
- The IKEA light returns a _Default Response_ with status _Malformed Command_;
- The icasa light accepts the command, and returns an empty scene state on the next _View Scene_.

Issuing a PUT to `/groups/#/scenes/#/lights/#/state` results in an _Add Scene_ command, with all values specified.  The REST API plugin seems to fill in the values missing from the body from its cached scene state and/or default values - not from the current light state.  I think this is a bug: the plugin should send only the values corresponding to the attributes in the body.
- The Philips light accepts the _Current X_ and _Current Y_ values in the scene state.  On recalling the scene, it still reports a _Color Mode_ of ct;
- The OSRAM light accepts the _Current X_ and _Current Y_ values in the scene state, and reports them back in the next _View Scene_ command.  However, on recall, _Color Temperature_ is set to 153, irrespective of the values in the scene state;
- For the IKEA and icasa lights, the REST API plugin sends a malformed package: the _Color Control_ length is set to 11, but the payload only contains _Current X_ and _Current Y_.  **This appears to be a bug in the REST API plugin, rather than in the Trådfri firmware?!**  As far as I can tell, the difference with the other lights is that these lights don't support the _Hue_, _Saturation_, _Enhanced Hue_, and _Color Loop_ attributes.

lbschenkel:
Great job. I just skimmed what you read because I'm extremely busy today (and will be on this weekend), but I'll come back to this and let's see if how I can contribute in improving the status quo.

ebaauw:
Some more reading in the ZCL spec (section 5.2.2.5):
> Since there is a direct relation between _ColorTemperatureMireds_ and XY, color temperature, if supported, is stored as XY in the scenes table.

Neither OSRAM, nor IKEA, nor icasa adhere to this.  Kinda odd that they're still certified.

> Attributes in the scene table that are not supported by the device (according to the _ColorCapabilities_ attribute) SHALL be present but ignored.

I'm not sure what "present but ignored" means, but this could be read as not even the Philips Hue light adheres to the standard (since they return only _Current X_ and _Current Y_ in the scene state).

ebaauw:
> For the IKEA and icasa lights, the REST API plugin sends a malformed package: the Color Control length is set to 11, but the payload only contains Current X and Current Y. This appears to be a bug in the REST API plugin, rather than in the Trådfri firmware?! 

Looking at the code, this could only happen if the `colormode` in the scene's light state hasn't yet been set.  Looking at the database, `cm` is `"none"` for the IKEA and icasa.  The payload isn't _Current X_ and _Current Y_, but _Colorloop Active_, _Colorloop Direction_, and _Colorloop Time_.  Also, the scene name is `"Scene %u"`.  I'm not sure which command I used to create the scene initially (too much testing), but I think this might have happened when the REST API plugin read the scene from the light, instead of creating it itself (e.g. when creating the scene from the GUI).

ebaauw:
Patched the code, so it'll no longer send the malformed payload.

Back to testing the IKEA.  Indeed, setting `ct` thru the API results in a _Move to Color_ command, instead of a _Move to Color Temperature_.  The light seems happy about this; the _Current X_ and _Current Y_ attributes reflect the color, but the _Color Temperature_ attribute isn't updated.  This will confuse the hell out of API clients, homebridge-hue included.

Doing a PUT to the scene state now succeeds; the light returns only _Current X_ and _Current Y_ in the _View Scene_.  However, when recalling the scene, these values seem to be igored, and the light reverts to the colour temperature from the last _Store Scene_.  So the light seems to update its internal colour temperature on _Move to Color_, and stores it in the scene, without exposing this over ZigBee.

For the icasa, the REST API sends _Move to Color Temperature_ allright, when setting `ct`.  I patched `addTaskAddScene()` to map `ct` in the scene state to `hue` and `modifyScene()` to copy `ct` from the body to the `LightState` and set `cm` in the `LightsState` accordingly.
This seems to work, I can now PUT `/groups/#/scenes/#/lights/#/state` with a body containing `ct` to store colour temperature in a scene on the iCasa.

Setting `ct` in the scene state now also works for the LTW010, modulo some rounding / conversion errors in the ct -> xy translation.  As this is a _Color temperature light_, the `colormode` remains `ct` on scene recall.  For the LCT015 _Enhanced color light_, storing the scene works similar, but recalling it sets `colormode` to `xy`.  I cannot tell without mutilating the light and removing the diffuser, but I suspect it drives the RGB channels instead of the CW/WW channels.  I need to double-check if/how the Hue bridge handles this.

When doing a PUT to `/groups/#/scenes/#/store`, the REST API plugin issues a (group broadcast) _Add Scene_, with an empty scene state, followed by a _Store Scene_.  The _Add Scene_ is, of course, ignored by most lights.  I double-checked by sending a groupcast _Add Scene_ from the GUI, followed by a unicast _View Scene_: indeed, no scene was created.

manup:
The scene handling is really confusing in the specification. I remember when implementing it for the FLS-PP lp RGBW ballast around 2013, it wasn't clear how to treat color temperature properly since it wasn't technically part of the scene. The add scene commands only carry xy values but no hint if it relates to color or color temperature.

I can imagine that in the next API we could provide an option to recall scenes via unicast by just sending the related commands to set on/off, brightness and color based on the scene values stored in the database. Should work ok in smaller groups but scales badly for > 10'ish device groups, on the other hand it would solve the broadly broken firmware scene issues.

ebaauw:
Not a fan of "recalling" scenes through unicasts: this will be a lousy user experience, as the lights won't react simultaneously.  If you want to go this route, make sure it's exposed differently in the API from recalling a ZigBee scene.

I think the safe bet would be to stay away from _Add Scene_ and to rely solely on (unicast) _Store Scene_ commands to create and maintain scene light states.  Most lights seem to support that alright.  The only real limitation is that the 5-channel enhanced colour lights will drive the RGB channels (colormode `xy`) instead of the WW/CW channels (colormode `ct`).  There might be some odd lights that don't handle _Store Scene_ correctly: I think we should document this and advice people not to buy these.

ebaauw:
One of my Homebridge Hue users has an issue that HomeKit doesn't reflect the colour temperature of their IKEA TW bulb correctly, see https://github.com/ebaauw/homebridge-hue/issues/699.

Did some more testing/sniffing on my Trådfri TW bulb (which seems to be a different model)
Screenshot 2020-05-17 at 14 44

When sending a _Move to color temperature_ command, the light updates the _Color temperature_   attribute to the actual setting (as opposed to the requested), but not _Current x_ and _Current y_.  On a subsequent _Store scene_, only _Current x_ is reported in the scene, with the current (now bogus) value of _Current x_.  However, when recalling the scene, the colour temperature is updated and _Color temeperature_ shows the correct actual value.

When sending a _Move to color_ command, the light updates the _Current x_ and _Current y_ attributes to the actual settings, but not _Color temperature_.  On a subsequent _Store scene_, both are reported in the scene, with the current values.  Because _Color temperature_ isn't updated, the REST API plugin reverts to the old value of `state.ct` when the lights is next polled, causing API clients to show the wrong colour temperature.

I don't have any other IKEA TW lights, but it would seem that their recent firmwares have solved the issue for which the unholy hack has been introduced.
https://github.com/dresden-elektronik/deconz-rest-plugin/blob/1caed496f90b01c69be6f03886597004183152c8/zcl_tasks.cpp#L392-L394
Even without the firmware fix, I think this hack does more harm than good, as the REST API reports the wrong colour temperature.

ebaauw:
After removing the cursed line, the REST API plugin behaves properly.  A PUT of `/lights/xx/state` or `/groups/xx/action` with `{"ct": n}` works.  A PUT of `/groups/xx/scenes/xx/store` and a subsequence PUT of `/groups/xx/scenes/xx/recall` store and recall the colour temperature.  It's just not reported when GETting `/groups/xx/scenes/xx/lights`.

lbschenkel:
If recent firmwares can now save/recall `ct` in scenes properly, then I agree that this workaround is not warranted anymore. I just urge that we don't break the ability to save scenes for IKEA WS lights, otherwise this will be a huge feature regression.

lbschenkel:
@ebaauw: can you provide me the binary with the hack removed, so I don't need to compile by myself? I have 4 different models of IKEA lights, WS and CWS, I can test what happens with those.

ebaauw:
@lbschenkel What server do you run Homebridge on?  What OS?

lbschenkel:
> @lbschenkel What server do you run Homebridge on? What OS?

Sorry, should have been more specific. I didn't mean Homebridge: I meant deCONZ. I would like to replicate your tests with the workaround disabled on all my lights, to confirm that latest firmware can save/recall scenes set via `ct`.

I'm running Debian "buster" x86-64.

ebaauw:
> I didn't mean Homebridge: I meant deCONZ.

My bad.  Doing too many things in parallel.

> I'm running Debian "buster" x86-64.

Sorry, I don't run that.  I can only compile for Raspbian Stretch or Raspbian Buster on Raspberry Pi (armv7l).

ebaauw:
Included in https://github.com/dresden-elektronik/deconz-rest-plugin/commit/a4c712bf6d4cc60b376968b9f6d292b2b916755f

PR #2531, opened 2020-03-03
Improve Aqara Opple support

merdok:
Proper support for Aqara Opple switches #2061. 

Send the magic word to the switch to activate the multistate cluster. 
No need to create a group for the switch anymore. 

Now all events work: Single press, Double press, Triple press, Hold and Long release.

Thanks @ebaauw for the support.


ebaauw:
Thanks, @merdok.  I'll do a review tomorrow or this weekend; too tired right now to take a good look.

de_web_plugin.cpp
```diff
@@ -716,6 +716,10 @@ void DeRestPluginPrivate::apsdeDataIndication(const deCONZ::ApsDataIndication &i
                     {
                         sensorNode = getSensorNodeForAddressAndEndpoint(ind.srcAddress(), 0x03);
                     }
+                    else if (sensorNode->modelId().contains(QLatin1String("86opcn01"))) //Aqara Opple enable events from all multistate clusters
```
jdberry: The model ends with "86opcn01" in all cases, right? Perhaps I'm a performance nut, but it would seem more efficient to use "endsWith" rather than "contains", since presumably endsWith compares starting at the end of the string, and would fail immediately if the tail doesn't match.

merdok: Not sure if this will make a significant difference...

de_web_plugin.cpp
```diff
@@ -4286,7 +4282,8 @@ void DeRestPluginPrivate::addSensorNode(const deCONZ::Node *node, const deCONZ::
                     //         fpSwitch.outClusters.push_back(ci->id());
                     //     }
                     // }
-                    else if (node->nodeDescriptor().manufacturerCode() == VENDOR_JENNIC)
+                    else if (node->nodeDescriptor().manufacturerCode() == VENDOR_JENNIC ||
+                             modelId.contains(QLatin1String("86opcn01"))) // Aqara Opple prevent client clusters creation
                     {
                         // prevent creation of ZHASwitch, till supported
```
ebaauw: Best put a comment that client clusters aren't used now fpSwitch is created for 0xfc00 cluster.

merdok: Done!

de_web_plugin.cpp
jdberry: I think it just offends me to use a general substring search when we know very well that we only care about the suffix match. It’s useful to consider that this substring search occurs for every model, on every iteration of this code, and only succeeds for these opple switches. Consider how many failing string searches that implies...! Conversely, with the suffix match (endsWith), in most cases the common failure case occurs after looking at one character. 

This comment applies to all of the uses of contains in this PR.

Issue #2158, opened 2019-12-01
[Request device support] Support for Namron 4-channel switch (Luzense/Sunricher clone)


eckhart:
This is widely available in Norway, and fits the standard ELKO frames which are common here:

https://www.elektroimportoren.no/namron-zigbee-4-kanaler-bryter-k8/4512703/Product.html

To me it seems more or less identical to the 4-channel ones in https://github.com/dresden-elektronik/deconz-rest-plugin/issues/1509 but under a different brand name. It shows up in deCONZ with "Namron AS" as manufacturer instead of "Sunricher" and doesn't show up as a switch in Phoscon/REST API after the switch confirms successful pairing.

Below is what I have from deCONZ. I've read out all the attributes from the seperate channels, they look identical so just including one of them here. Let me know if more info is needed.

![namron_switch1](https://user-images.githubusercontent.com/1478965/69915308-e4177000-144d-11ea-96d1-8e9bd471ba8d.png)
![namron_switch3](https://user-images.githubusercontent.com/1478965/69915321-fe514e00-144d-11ea-92ba-a1f22eaea360.png)
![namron_switch4](https://user-images.githubusercontent.com/1478965/69915322-fe514e00-144d-11ea-8426-b57bc67c6c7d.png)
![namron_switch5](https://user-images.githubusercontent.com/1478965/69915323-fee9e480-144d-11ea-88c5-90003d906c61.png)
![namron_switch6](https://user-images.githubusercontent.com/1478965/69915325-fee9e480-144d-11ea-994f-9797e16367ef.png)
![namron_switch11](https://user-images.githubusercontent.com/1478965/69915341-29d43880-144e-11ea-81e9-ba81ca75dfe7.png)

corvy:
@eckhart it does not show up in groups  in the old web interface either? 

@ebaauw I guess it takes a very basic change to add support for this?

eckhart:
No, it doesn't show in the old web interface either.. Would be great to have it supported as it's a lot cheaper than other brands.

ebaauw:
Probably.  Just whitelist it in `supportedDevices` in `de_web_plugin.cpp` and assign it to the buttonmap for the 8-button switch in `sensors.cpp`.

eckhart:
@ebaauw Sorry that I haven't had the time to follow this up, would be great if no more than this is needed. Being new to this, do the devs prefer that I add, build and test this myself and then make a pull request?

[...]

PR #2589, opened 2020-03-14
Support for 4-ch Namron Switch (Luzens/Sunricher clone)

phithor:
Added support for this button
https://github.com/dresden-elektronik/deconz-rest-plugin/issues/2158


de_web_plugin.cpp
```diff
@@ -5944,6 +5946,7 @@ void DeRestPluginPrivate::updateSensorNode(const deCONZ::NodeEvent &event)
                                         i->modelId().startsWith(QLatin1String("ZGRC-KEY")) || // Sunricher wireless CCT remote
                                         i->modelId().startsWith(QLatin1String("ZG2833K")) || // Sunricher remote controller
                                         i->modelId().startsWith(QLatin1String("SV01-"))) // Keen Home vent
+                                        i->modelId().startsWith(QLatin1String("4512703")) || // Namron 4-ch remote controller
```
manup: I haven't compiled it yet, but I guess this line and the one in line 5914 should throw a compile errors?

phithor: Sorry, I just copy pasted the code of the similar switch. I have updated it below, and I think that fixes it.

Issue #2633, opened 2020-03-31
[Device Support Request] Tuya Gas Detector


timbru31:
I've bought a new Zigbee device: The Tuya (combustible) gas detector (model TS0204) - sadly, it's not yet supported.
See attached Node and Cluster info below:

![tuya2](https://user-images.githubusercontent.com/838003/78023599-753c1c00-7346-11ea-8bca-9c9a025ad157.jpg)
![tuya3](https://user-images.githubusercontent.com/838003/78023600-75d4b280-7346-11ea-824f-b275d7051ad6.jpg)
![tuya1](https://user-images.githubusercontent.com/838003/78023602-766d4900-7346-11ea-90a4-1a32fd9e582d.jpg)

---

Hardware image:
![TS0204](https://user-images.githubusercontent.com/838003/78023694-a1579d00-7346-11ea-9af7-0b63784b467c.jpg)

---

PS: Deconz GUI says it's a CO sensor, it seems the translation of the attribute `e16ZoneType` (i.e., E_CLD_IASZONE_TYPE_GAS_SENSOR (0x002B)) is wrong, since it _can_ but is not always a CO sensor.

SwoopX:
Can you also please provide a screenshot with the available clusters? Thanks.

timbru31:
Sure:
![tuya4](https://user-images.githubusercontent.com/838003/78028357-6c4f4880-734e-11ea-82bf-7662a7954e44.jpg)


I'll work on a PR, too. Hopefully it's just adjusting the bindings to whitelist the model id.

SwoopX:
Nope, it isn't ;)

timbru31:
What should be missing then? Basically it's yet another gas sensor, the clusters aren't very special (as expected)

ebaauw:
> Deconz GUI says it's a CO sensor, it seems the translation of the attribute e16ZoneType (i.e., E_CLD_IASZONE_TYPE_GAS_SENSOR (0x002B)) is wrong, since it can but is not always a CO sensor.

It advertises incorrectly as a CO sensor.  Make sure to expose it as ZHAFire instead of ZHACarbonMonoxide, see #2515.

SwoopX:
Ah, wait. Seems that I mixed it up as it doen't have the IAS WD cluster...

Don't even think any binding is required here. Probably just add them on top of de_web_plugin.cpp as the other devices and the changes mentioned by @ebaauw somewhere around line 3989.

[...]

PR #2634, opened 2020-03-31
fix: rename IAS Zone Carbon Monoxide (CO) sensor to Gas sensor

timbru31:
The value 0x002B (IAS_ZONE_TYPE_GAS_SENSOR) _can_ by a CO sensor, but also, e.g., a combustible gas sensor, too. Therefore the label in the GUI is ~wrong~ confusing.

See, e.g., http://dev.ti.com/tirex/content/simplelink_cc13x2_26x2_sdk_3_40_00_02/docs/zigbee/doxygen/zigbee/html/group___t_y_p_e___s_s___i_a_s___z_o_n_e.html#ga3ba58ff7204f89aad419a95bdeee533a

Ref #2633 (but does not close)

### DIFF
--- a/general.xml
+++ b/general.xml
@@ -2170,7 +2170,7 @@ controller device, that supports a keypad and LCD screen.</description>
 						<value name="Contact switch" value="0x0015"></value>
 						<value name="Fire sensor" value="0x0028"></value>
 						<value name="Water sensor" value="0x002a"></value>
-						<value name="Carbon Monoxide (CO) sensor" value="0x002b"></value>
+						<value name="Gas sensor" value="0x002b"></value>
 						<value name="Personal emergency device" value="0x002c"></value>
 						<value name="Vibration / Movement sensor" value="0x002d"></value>
 						<value name="Remote Control" value="0x010f"></value>


SwoopX:
You referenced the TI interpretation, the name of their function to be more precise. Besides the name, it sticks to the Zigbee standard, which is "Carbon Monoxide (CO)".
Zone Type | Alarm1 | Alarm2 | Carbon Monoxide (CO) sensor | CO indication | Cooking indication | 

Therefore, the current code is correct (although your argument is not wrong).

timbru31:
Well, that's true:
image

However, for the end user, IMHO my PR makes sense. For a moment I was assuming either I've bought the wrong sensor from AliExpress or that Tuya messed up the zone...

SwoopX:
Consequence of that change would be that people having a CO sensor could come up with a similar request to change it back. Imho, 99% of the users never take a good look at that particular information in deconz GUI. Don't get me wrong, I'm totally indifferent since I don't own such devices but also if I would.

Additionally, there's other places in deconz where you pretty much have the same situation. And there you can only chose between fire and carbon monoxide sensor. I recall a recent change having moved a Heimna gas sensor towards a fire sensor...

ebaauw:
Please stick to the ZigBee definition.  Some combustable gas sensors advertise the wrong type, the API should map these to Fire sensors, see #2515.

timbru31:
I can understand that we can agree to use the ZigBee spec name, however I disagree that the sensor is advertised wrongly - `0x002B` is correct for both natural gas and CO - the chosen wording of the Zigbee alliance (without the Alarm1 and Alarm2 additions) is just _confusing_, but not _wrong_.

ebaauw:
> `0x002B` is correct for both natural gas and CO

Says who?  Source?

Granted, the ZCL spec downloadable from the ZigBee Alliance web site is old, and probably outdated.  But unless you have a newer version, I think we should be deCONZ on offcial specs, not on some vendor documentation.

Because of the ZCL spec, I defined the `ZHACarbonMonixide` sensor type, and (arbitrarily) mapped the Heiman combustable gas sensor to `ZHAFire`.

timbru31:
I assume the most recent ZCL spec is from 2016 (https://zigbeealliance.org/developer_resources/zigbee-cluster-library/) - the file upload from 2019.
I agree that, and will close the PR, that we should keep the name.

However, my opinion is still that a gas sensor (methane, propane, butane) fit's better in the CO category (gas only) than fire category. Most vendors seems to use 0x200b for both, too (TI, [NXP](https://www.nxp.com/docs/en/user-guide/JN-UG-3077.pdf))

timbru31:
Fun fact, the "Zone Enroll Request" label is already wrong and has "Gas sensor" as the label ;)
https://github.com/dresden-elektronik/deconz-rest-plugin/blob/02b76f02385aeadbad40612ccbe1cbea273356de/general.xml#L2248

Issue #2576, opened 2020-03-12
[Device Support Request] Samsung SmartThings Outlet 2018 Model ID: GP-U999SJVLDAA


jbruening:
https://www.samsung.com/us/smart-home/smartthings/outlets/samsung-smartthings-outlet-2018-gp-u999sjvldaa/

Side of the unit also lists M/N "IM6001-OTP01"

This looks similar to https://github.com/dresden-elektronik/deconz-rest-plugin/issues/2121, but lacks the 'Simple Metering' cluster

Phoscon currently treats this as an on/off "light". It supposedly supports power monitoring as well though.

![image](https://user-images.githubusercontent.com/1760487/76475900-9e0f6800-63bd-11ea-92c8-68471502da10.png)

![image](https://user-images.githubusercontent.com/1760487/76475942-b4b5bf00-63bd-11ea-9c91-c1065f276bc3.png)

![image](https://user-images.githubusercontent.com/1760487/76476106-360d5180-63be-11ea-83b0-558e0454399f.png)

Electrical measurement cluster picture taken with a 20 inch box fan set to slowest setting of 3
![image](https://user-images.githubusercontent.com/1760487/76476211-85ec1880-63be-11ea-984c-6bb8aab63ecb.png)

SwoopX:
I can take care of that. Can you please provide a screenshot of the node info panel as well? Thanks!

jbruening:
![image](https://user-images.githubusercontent.com/1760487/76568566-51d03080-646e-11ea-9ede-21ac3a6524b4.png)

Mimiix:
It seems this issue is resolved or otherwise inactive. If it is not, please re-open!

jbruening:
Yes, it is resolved. I re-added the device in deconz (2.5.75) (which appeared to show nothing different in phoscon about it), and then restarted home assistant.  There is now a wattage sensor showing up.

![Uploading firefox_2020-06-11_12-10-28.png…]()

PR #2572, opened 2020-03-10
Device updates

SwoopX:
- Added support for Tuyatec door/window sensor RH3001 (#2559)
- Allow binding for Tuyatec devices
- Streamline manufacturer name for Tuyatec devices
- Added initial support for Samsung SmartThings Outlet 2018 (IM6001-OTP01, #2576)
- Added initial support for Xiaomi Mija ZB3.0 smart plug ZNCZ04LM (lumi.plug.mmeu01 ,#2583)
- Revert general Xiaomi binding
- Enhancements for Xiaomi special attribute reporting
- Solved float conversion for Xiaomi special attribute reporting. Power and consumption reporting is expected to be working for other devices than lumi.plug.mmeu01 as well (#2583, #2533, #2094, #1692, #1513, #1592, #2302)
- Added initial support for Hive motion sensor MOT003 (#2606)
- Fix for iHorn sensors not being created (#2406)
- Added initial support for Centralite contact sensor 3320-L (#2632)

Issue #2475, opened 2020-02-20
Input needed: Refactor logic setting light state


ebaauw:
I've finally found the time to refactor the setting of the `state` of a `/lights` resource, see https://github.com/dresden-elektronik/deconz-rest-plugin/issues/1111#issuecomment-456464872.  Next to the functional changes, I'm also adding parameter checking.

The logic for handling `on` and `bri` is described in the issue mentioned above.  The logic for handling the other attributes is less clear.  I run into some design decisions for these, @manup, everybody, please provide your input.

1. The Hue API clearly states: `xy` trumps `ct` trumps `hue` and `sat`.  It ignores the over-trumped attributes without issuing an error, but clients can tell because they aren't reported in the response.  Not sure how `effect` is supposed to relate to these.  I want to do the following: for `"none"`, clear the colour loop before handling the colour attributes; for `"colorloop"`, set the colour loop after handling the colour attributes.  And return an error for the colour attributes while colour loop is active (just as for `bri` and colour attributes when the light is off).
2. While not documented, I think the Hue API also acts like: `bri` trumps `bri_inc` and `ct` trumps `ct_inc`.  Note that the deCONZ API doesn't expose `xy_inc`, `hue_inc`, nor `sat_inc`.  While they might be great for rule actions for the magic cube, I haven't seen any use for these;
3. There is no `ResourceItem` for `effect`.  Looking at the code, `effect` is exposed only when `xy` is.  On the other hand, setting the colour loop changes `colormode` to `hs`?
4. I'm not sure I fully understand the code handling `hue` and `sat`.  It would seem the plugin tries to calculate and update the `xy` values from the `hue` and `sat` values for the benefit of `taskToLocalData()`.  Why isn't there similar logic for updating `ct`.  And for updating `xy` and `hue` / `sat` when setting `ct`?  And for updating `ct` and `hue` / `sat` when updating `xy`?  
I think API clients should take notice of `colormode` and only process the corresponding attributes (homebridge-hue does).  
I would really like to defer this setting of other colour attributes to the light, although not all non-Hue lights update `xy`, `ct`, or `hue` / `sat` from other colour modes;
5. I'm missing an `addTaskSetEnhancedHueAndSaturation()` in `zcl_tasks.cpp`.  Is that intentional or just for historic reasons?  Preferably, I would have a single `addTask...()` to set any combination of hue/enhanced hue and sat, depending on the light capabilities;
[...]

PR #2658, opened 2020-04-05
Multiple enhancements and bug fixes.

ebaauw:
_Window covering_ devices, tested on IKEA Fyrtur and Xiaomi `lumi.curtain`, see #2475:
- Implement `state.open`, `state.lift`, and `state.tilt` for _Window Covering_ `/lights` resources;
- Minor bug fixes, light refactoring, and fix indentation in `window_covering.cpp`.

Xioami plug:
- Fix float conversion from #2572.

_IAS WD_ devices (Sirens, smoke detectors, etc), tested on Heiman Siren and Develco Smoke Detector, see #2576:
- Implement `{"alert": "all"}` as a dirty hack to send a network-wide broadcast to activate all sirens;
- Only set `lightNode->rx()`, before sending _Start Warning_ command, when the node is actually zombie, as this incorrectly updates `lastseen`;
- Set `lightNode->rx()` when end-device polls coordinator;
- Fix values sent to _options_ parameter of _Start Warning_ command;
- No more need to whitelist smoke detectors (one less place to whitelist!).

OSRAM SMART+ Motion Sensor (a.k.a. Centralite Motion Sensor-A):
- Don't expose `state.tampered` nor `state.lowbattery`, see #2672.

IKEA SYMFONISK controller, see #1898:
- Don't create group, nor group binding.  Instead create unicast binding to coordinator on pairing, to prevent denial-of-service on Zigbee network due to frequent broadcasts (3-4 per second);
- Expose turns as x002 `buttonevent` value, instead of a pair of x001/x003.  The controller doesn't distinguish start vs stop rotation; it just sends _Move_ and _Stop_ commands while rotating.  The number of commands indicate the duration of the turn, not the angle nor the speed.

Triggering of rules with `ddx`, `in`, and `not in` conditions on `/config/localtime`, see #2148:
- Millisecond resolution for `RConfigLocalTime`;
- Log previous value for `/config/localtime`;
- Pass `previousNow` to `evaluateRule()`;
- Check for condition time between `previousNow` and `now` to deal with missing events during database saving.


window_covering.cpp
```diff
@@ -104,132 +107,171 @@ void DeRestPluginPrivate::handleWindowCoveringClusterIndication(const deCONZ::Ap
...
+                case deCONZ::Zcl16BitUint:
+                case deCONZ::Zcl16BitInt:
+                case deCONZ::Zcl16BitEnum:
+                    quint16 attrVal16;
+                    stream >> attrVal16;
+                    break;
+                default:
+                    // unsupported data type
+                    return;
+            }
```
manup, line 156: Cool, very glad the magic numbers for attribute ids are gone.

rest_lights.cpp
```diff
@@ -1327,30 +1330,61 @@ int DeRestPluginPrivate::setWindowCoveringState(const ApiRequest &req, ApiRespon
...
+        }
+        else if (param == "on" && taskRef.lightNode->item(RStateOn))
+        {
+            paramOk = true;
+            hasCmd = true;
+            if (map[param].type() == QVariant::Bool)
+            {
+                valueOk = true;
+                hasOpen = true;
+                targetOpen = !(map[param].toBool());
```
manup, line 1367: The open attribute is nice, so if I understand correctly, for clients which still use the "on" attribute nothing will break?

rest_lights.cpp
```diff
@@ -1667,28 +1715,34 @@ int DeRestPluginPrivate::setWarningDeviceState(const ApiRequest &req, ApiRespons
...
+        else if (alert == "all")
+        {
+            // FXIME: Dirty hack to send a network-wide broadcast to activate all sirens.
+            task.req.dstAddress().setNwk(deCONZ::BroadcastAll);
+            task.req.setDstAddressMode(deCONZ::ApsNwkAddress);
+            task.req.setTxOptions(0);
+            task.req.setDstEndpoint(0xFF);
+            task.options = 0x17; // Warning mode 1 (burglar), Strobe, Very high sound
+            task.duration = onTime > 0 ? onTime : 1;
+        }
```
manup, line 1745: We have two sirens installed in the office now, this can be fun, totally need to test this out.

manup:
Thanks a lot for the PR, sorry it took a while. I was off grid for a few weeks, now catching up integrating the PRs and code from the team.

rest_lights.cpp
ebaauw: That's the idea.  Both `on` and `open` are reported by GET.  You can still PUT `on` but the API reports `open` instead of `on` in the response, to ease people into using the `open`.  Same for `bri` vs `lift` and `sat` vs `tilt`.

rest_lights.cpp
ebaauw: Make sure to wear protective ear plugs.

What type of sirens do you have?  Do they support groups?

ebaauw:
> Thanks a lot for the PR, sorry it took a while. I was off grid for a few weeks, now catching up integrating the PRs and code from the team.

Welcome back.  Hope you and your loved ones are doing OK in these crazy times.

manup:
Thanks, hope you and your family is well too. Hopefully the global situation gets better as quickly as possible.

rest_lights.cpp
manup: Absolutely :) We have one from Heiman and one from Tuya (branded as wesmartify).

image

rest_lights.cpp
manup: Perfect, really like the more meaningful attribute names here.

rest_lights.cpp
ebaauw: Nice, mains-powered, battery-backed-up and OTAU as well.  Do they publish firmware updates?

You might want to whitelist this siren, so the API plugin creates a ZHAAlarm `/sensors` resource.  I think that's needed for automatic enrolment.  Also, it will expose the level of the backup battery.

After refactoring the `/lights` API, I want to do `/groups`, supporting groups of warning devices, and window covering devices.  Simplest way is bluntly to expose `action.open` on every group; smarter way would be to keep track what type of devices are in the group (what would also be needed e.g. not to expose `action.xy` on a group of dimmable lights).  I might need to introduce another attribute for `alert`, to distinguish between alerting lights and alerting sirens.

Alternatively, use homogenous groups of `type` `WarningGroup` and `WindowCoveringGroup` (next to `LightGroup`).  That would block heterogeneous scenes, though.  I've yet to see a siren that supports scenes, but both the IKEA FYRTUR and `lumi.curtain` support it.

Kane610:
@ebaauw @manup Will these changes be documented somewhere, it is hard to keep up with API changes/additions when I have to sniff out every relevant PR

ebaauw:
@manup does an excellent job listing all changes in the release notes, see https://github.com/dresden-elektronik/deconz-rest-plugin/releases.

What does worry me is that the API documentation is incomplete and outdated.  See http://dresden-elektronik.github.io/deconz-rest-doc/.

Kane610:
The API documentation is good for the most basic parts, but it lacks a lot of detail, I don't really like to guess how stuff works

ebaauw:
Same here.  I still refer to the Hue API documentation a lot, but there are subtle differences.  For those, I tend to look in the code rather than at the API documentation.  They did open-source the documentation, see https://github.com/dresden-elektronik/deconz-rest-doc, but I never got around to looking at that.  So much to do, so little time.

Issue #2732, opened 2020-04-28
New Device Support - RGBgenie ZB-5121


puithove:
I'd like to request support of these nice battery powered ZigBee 3.0 remotes.  I previously had these working well under Home Assistant's ZHA but am working on converting over to deconz.  Under ZHA, they create very reliable zha_event events that can be used to trigger automation.  I'd like to be able to continue to use these in deconz the same way, exposing deconz_event to Home Assistant.

https://rgbgenie.com/?product=rgbgenie-mini-remote-with-single-scene-recall-zigbee

<img width="808" alt="image" src="https://user-images.githubusercontent.com/10265248/80542865-cd813080-897b-11ea-91b5-508bcca22f47.png">

<img width="812" alt="image" src="https://user-images.githubusercontent.com/10265248/80542974-028d8300-897c-11ea-82f6-0b252abaedee.png">

<img width="812" alt="image" src="https://user-images.githubusercontent.com/10265248/80543127-4aaca580-897c-11ea-9962-8ccf5f62a092.png">

<img width="809" alt="image" src="https://user-images.githubusercontent.com/10265248/80543185-64e68380-897c-11ea-8d1a-397c067386fc.png">

SwoopX:
Can you also please provide a screenshot of the node info panel?

puithove:
@SwoopX It looks like that last screenshot is the Node Info panel?  Is there something that I'm missing?  Can you give me an example of what to look for?

puithove:
Here's another one, just in case
<img width="437" alt="image" src="https://user-images.githubusercontent.com/10265248/81512846-d665f080-92f1-11ea-9c7c-1a187944dd0e.png">

SwoopX:
Sry, my bad. Thanks anyway.

puithove:
No worries 👍 let me know if you need anything else.

gmitch64:
Would it be possible to add support for the remote in the issue that I just see linked :)


G

SwoopX:
So, support should be in and available with version .76. You need to try it out and let me know if it works alright.

puithove:
Nice, I will check it out and let you know here once I see my deconz container updated in HA.  Thanks!

puithove:
I added it via Switches (like I did my one Phillips Hue Dimmer and the others of these) in Phoscon, but while deconz registered it, Phoscon didn't.

[...]

Issue #2722, opened 2020-04-26
New Device Support - Sercomm SZ-DWS04 Micro Door Window Sensor


meremortals70:
Please add support for the above Device

![image](https://user-images.githubusercontent.com/52735978/80294131-82172880-87a9-11ea-82d0-962427e9d86c.png)

![image](https://user-images.githubusercontent.com/52735978/80294121-5eec7900-87a9-11ea-9ae7-86ed0b0fe2d4.png)

Thank you

SwoopX:
Please provide all relevant screenshots. You may take #2583 as an example. Thanks!

meremortals70:
I apologise for not providing all the information.  Your info on providing info only mentions the basic cluster.

![image](https://user-images.githubusercontent.com/52735978/80322860-08e00a00-886b-11ea-86d6-b69407844ea3.png)


![image](https://user-images.githubusercontent.com/52735978/80322844-f960c100-886a-11ea-90a9-f97aff485a79.png)

![image](https://user-images.githubusercontent.com/52735978/80322869-1b5a4380-886b-11ea-8647-3943631d27ae.png)

![image](https://user-images.githubusercontent.com/52735978/80322886-3a58d580-886b-11ea-93fe-2e88ad8e51bb.png)

![image](https://user-images.githubusercontent.com/52735978/80322877-2c0ab980-886b-11ea-8559-7f6aef66a31a.png)

![image](https://user-images.githubusercontent.com/52735978/80322900-4e9cd280-886b-11ea-9a9b-b07405a8be34.png)

![image](https://user-images.githubusercontent.com/52735978/80322908-5bb9c180-886b-11ea-88ed-1606b5f38f37.png)

![image](https://user-images.githubusercontent.com/52735978/80322917-6b390a80-886b-11ea-9bcf-16059df942ad.png)

![image](https://user-images.githubusercontent.com/52735978/80322929-755b0900-886b-11ea-83e0-bd86691ef555.png)

![image](https://user-images.githubusercontent.com/52735978/80322942-83108e80-886b-11ea-9efb-7ac5d3c20b20.png)

![image](https://user-images.githubusercontent.com/52735978/80322953-91f74100-886b-11ea-802d-571574725689.png)

SwoopX:
[...]

Issue #2583, opened 2020-03-13
[Device Support Request] Xiaomi Mija ZB3.0 smart plug ZNCZ04LM (lumi.plug.mmeu01)


SwoopX:
![grafik](https://user-images.githubusercontent.com/4005212/76663227-aa8fee00-6580-11ea-83b0-45df88313ff3.png)

![grafik](https://user-images.githubusercontent.com/4005212/76660812-14f16000-657a-11ea-95c8-a022eedd9714.png)

![grafik](https://user-images.githubusercontent.com/4005212/76661109-e2943280-657a-11ea-86b6-c60a6b3e4145.png)

![grafik](https://user-images.githubusercontent.com/4005212/76661161-f8a1f300-657a-11ea-97ff-58f2ba6b0660.png)

![grafik](https://user-images.githubusercontent.com/4005212/76661196-140cfe00-657b-11ea-9785-473bb692aa0d.png)

Endpoint 15:
![grafik](https://user-images.githubusercontent.com/4005212/76663356-fe9ad280-6580-11ea-93dc-21ea1fdc4400.png)

Endpoint 16:
![grafik](https://user-images.githubusercontent.com/4005212/76663399-11150c00-6581-11ea-9cc3-af80bd468eee.png)

SwoopX:
The attributes from cluster **FCC0**:

```
ZigBee Cluster Library Frame, Mfr: Unknown (0x115f), Command: Discover Attributes Response, Seq: 157
    Frame Control Field: Profile-wide (0x1c)
        .... ..00 = Frame Type: Profile-wide (0x0)
        .... .1.. = Manufacturer Specific: True
        .... 1... = Direction: Server to Client
        ...1 .... = Disable Default Response: True
    Manufacturer Code: Unknown (0x115f)
    Sequence Number: 157
    Command: Discover Attributes Response (0x0d)
    Attribute Status Record
        Attribute: 0x00f3
        Data Type: 8-Bit Unsigned Integer (0x20)
    Attribute Status Record
        Attribute: 0x00f5
        Data Type: 32-Bit Unsigned Integer (0x23)
    Attribute Status Record
        Attribute: 0x00f6
        Data Type: 16-Bit Unsigned Integer (0x21)
    Attribute Status Record
        Attribute: 0x00fa
        Data Type: Boolean (0x10)
    Attribute Status Record
        Attribute: 0x00fc
        Data Type: Boolean (0x10)
    Attribute Status Record
[...]

PR #2761, opened 2020-05-12
Refactor IAS zone enrollment and further device integration

SwoopX:
- Added initial support for Sercomm open/close sensor SZ-DWS04 (#2722)
- Prevent lumi.plug.mmeu01 creating wrong sensors in rare cases (#2583)
- Update for LifeControl MCLH-08 environmental sensor
- Refactor IAS zone enrollment
- Added initial support for RGBgenie ZB-5121 (#2732)
- Enhancements for RGBgenie ZB-5001 (#1256)
- Added initial support for Centralite 3323-G contact sensor (#2783)
- Allow binding for Xiaomi Mijia Smart Light Sensor lumi.sen_ill.mgl01 (#2300)

According to Zigbee spec (07-5123-06-zigbee-cluster-library-specification.pdf, 8.2.2.2.3 Implementation Guidelines), writing the IAS CIE address is mandatory to be done first in the enrollment process. Apparently, not much end devices seem to adhere to that, as an enrollment request is fired very early from those devices.
However, some devices (e.g. Sercomm siren) stick to the specs and require the CIE write before, otherwise they do not earn status enrolled. The refactor should be more robust adhering to the procedure and also consider "orphan" sensors (a device with an IAS sensors gets reset, in those cases, the sensor is already available).
Tested with Sercomm siren, Heiman and Develco smoke detector.

### DIFF
--- a/bindings.cpp
+++ b/bindings.cpp
@@ -2019,6 +2019,7 @@ bool DeRestPluginPrivate::checkSensorBindingsForAttributeReporting(Sensor *senso
         sensor->modelId().startsWith(QLatin1String("RH3052")) ||
         // Xiaomi
         sensor->modelId().startsWith(QLatin1String("lumi.plug.maeu01")) ||
+        sensor->modelId().startsWith(QLatin1String("lumi.sen_ill.mgl01")) ||
         // iris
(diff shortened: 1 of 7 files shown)


SwoopX:
You got no idea... 😞  Thanks for keeping an eye though.

Issue #932, opened 2018-11-08
Fan Module (Hampton Bay)


mjruotolo:
I would like to use this module with the gateway.
https://www.homedepot.com/p/Hampton-Bay-Universal-Wink-Enabled-White-Ceiling-Fan-Premier-Remote-Control-99432/206591100
This is a universal zigbee fan controller with fan and light controls.
I have included a bunch of screeen shotes of the nodes as well as the device handler from smartthings.  Not sure if this will help in any way. Please let me know what other info you will need.

Smartthings Device handler. 
https://github.com/dcoffing/KOF-CeilingFan/blob/master/devicetypes/dcoffing/kof-zigbee-fan-controller.src/kof-zigbee-fan-controller.groovy
![fan6](https://user-images.githubusercontent.com/3311453/48177949-5c604000-e2e5-11e8-921b-a6a2c3a7535a.PNG)
![fan7](https://user-images.githubusercontent.com/3311453/48177950-5c604000-e2e5-11e8-932b-168fd7666ef5.PNG)
![hampton-bay-ceiling-fan-remote-wall-controls-99432-64_1000](https://user-images.githubusercontent.com/3311453/48177951-5c604000-e2e5-11e8-904a-90e76feb65fc.jpg)
![hampton-bay-ceiling-fan-remote-wall-controls-99432-c3_1000](https://user-images.githubusercontent.com/3311453/48177952-5c604000-e2e5-11e8-93b8-9e04bb67810a.jpg)
![fan1](https://user-images.githubusercontent.com/3311453/48177953-5c604000-e2e5-11e8-8249-35d24c1f1935.PNG)
![fan2](https://user-images.githubusercontent.com/3311453/48177955-5cf8d680-e2e5-11e8-8e5e-d99313a2883c.PNG)
![fan3](https://user-images.githubusercontent.com/3311453/48177956-5cf8d680-e2e5-11e8-94dd-60be5b202d29.PNG)
![fan4](https://user-images.githubusercontent.com/3311453/48177957-5d916d00-e2e5-11e8-9e23-cfd034c62783.PNG)
![fan5](https://user-images.githubusercontent.com/3311453/48177958-5d916d00-e2e5-11e8-8276-a1aa75913778.PNG)

jjlawren:
I have a couple of these that I’d love to integrate.

manup:
So this is only the remote? Is there also some data available about the controller?

[...]

PR #2842, opened 2020-05-28
Hampton Bay fan module

ebaauw:
Workaround for broken `deCONZ::ZclAttribute`: doesn't support `Zcl8BitEnum` either, see #932.


ebaauw:
And confirmation that it now works!

Issue #2590, opened 2020-03-15
Proposal: deprecate `reachable` attributes


ebaauw:
I propose to deprecate and eventually remove the `state.reachable` attribute from `/lights` resources and the `config.reachable` attribute from `/sensors` resources.  These attributes are utterly useless.  They do **not** indicate whether a device is, in fact, reachable.  They certainly don't indicate whether a device is powered.  They cause a lot of confusion and fruitless discussions.

As indicated [earlier](https://github.com/dresden-elektronik/deconz-rest-plugin/issues/2047#issuecomment-551580888), I think the REST API should expose a `lastseen` attribute instead, indicating the last time a message from the device was received.

**Background**
ZigBee doesn't know the concept of connections; it only deals with messages.  The deCONZ GUI is like a city map: it displays roads, not actual traffic.  The GUI draws lines for neighbour table entries, as reported by the devices.  Just like a city map doesn't show whether a road is actually being used, the GUI doesn't show whether there's any messages using these routes.

Technically, end devices are unreachable almost all of the time.  They turn off their radio, to preserve (battery) power.  They can only be reached through their parent router, who acts like a post office holding [_poste restante_](https://en.wikipedia.org/wiki/Poste_restante) (general delivery) mail.  ZigBee routers, however only keep the message addressed to the end device for like 7 seconds.  Light sleepers, end-devices that wake up once every 5 seconds to check whether their parent has any messages for them, can be reached "normally", albeit with a slight delay.  Deep sleepers, that only poll their parent like once an hour, can only be reached for 10 to 100 milliseconds per hour.

The REST API plugin sets the `reachable` attribute when it receives a message from the corresponding device.  So far, so good.  However, clearing the `reachable` attribute is far more elaborate.  For devices that are polled by the gateway `reachable ` is cleared after not receiving a response to a message for a couple of times in succession.  Depending on the size of your network (devices take turns being polled), this might take a couple of seconds to minutes.  Devices that support attribute reporting aren't polled, so `reachable` cleared after missing a couple of expected reports in a row.  For deep sleepers, you'd only expect a report once per hour, so it takes several hours for `reachable` to be cleared.

MattL0:
Thanks !!

ebaauw:
Did some testing and managed to add `lastseen` (with millisecond resolution, see #1687)  as top-level attribute to `/lights` and `/sensors` resources, based on the `lastRx()` value of the `RestNodeBase` superclass.  

[...]

Issue #1209, opened 2019-02-02
[Request Device Support] Tint remote control by Müller-Licht


Thomas-Vos:
Hello, would it be possible to add support for the Tint remote control? This is a Zigbee remote control made by [Müller-Licht](https://www.mueller-licht.de/produktserien/tint/). I was able to add the remote to deCONZ by pressing the reset button in the battery compartment and starting a sensor search.

![20190202_122347](https://user-images.githubusercontent.com/11077217/52163672-84452100-26e5-11e9-9e33-ba6a42bf591b.jpg)

Clusters of the node:
![screenshot from 2019-02-02 12-10-20](https://user-images.githubusercontent.com/11077217/52163586-4dbad680-26e4-11e9-90df-2bec266ff4e2.png)

Basic cluster:
![screenshot from 2019-02-02 12-12-57](https://user-images.githubusercontent.com/11077217/52163598-62976a00-26e4-11e9-9091-42bb778faebd.png)

Node info panel:
![screenshot from 2019-02-02 12-12-07](https://user-images.githubusercontent.com/11077217/52163595-58756b80-26e4-11e9-9dc1-0e3866a7724b.png)

blaupert:
Yes! Another requests to add this Remote Devices from me!

curryhoden:
Any news here?

Trying to connect my Tint remote with deconz.

But the sesor search in not finding my remote. 

How did you connect it?

Thomas-Vos:
> How did you connect it?

Press the reset button in the battery compartment and start a sensor search. It won't be visible in Phoscon because support for this remote has not been added yet.

realwax:
+1 on the request list for Müller tint support of the remote.  Thank you very much!

Philje123:
Just bought a GU10 and remote so hoping for support in the future.

I noticed Also are selling these off cheap now (€14 for the GU10 RGB + remote, €5 for the GU10 white and €7 for GU10 RGB). That's all they had.

So is the remote usuable as it is? Does it publish anything to the API?

curryhoden:
@Philje123 
Short form: YES

If you are starting a sensor search from the api, you will get the address and the id of your remote control.

But it does NOT appear in Phoscon app.

Philje123:
Thank you.

That will be fine as I mainly use home assistant to control my lights so as long as it the message gets through to there i can set it up to do what i want.

Philje123:
Any update on this?

[...]

PR #2849, opened 2020-05-30
Tint remote control by Müller Light; Refactor setting light state

ebaauw:
- Add support for Tint remote control by Müller Light, see #1209.
- Add support for Tint by Müller Light special scenes, through `state.effect`, see #1209.
- Refactor logic of PUT to light state, take 2, see #2475.
  - Only report an error for violations of the data type (e.g. setting setting `ct` to `true`, or setting `ct` to an integer value < 0 or > 0xFFFF). Coerce the value if of the correct type, but invalid;
  - Return HTTP status 200 on API errors (just like the Hue bridge). This means that invalid keys or values in the body are ignored, but valid ones are still executed;
  - Add a sanity check that `ctmax` > `ctmin`. Change default `ctmax` to 65279 (0xFEFF).
  - For window covering devices, use {"stop": true} instead of {"tilt": "stop"}.
  - Remove `state.alert` for _Window Covering Device_.
- Fully implement `/lights` resource `lastannounced` and `lastseen`, see #2590:
  - Persist values in database;
  - Web socket notifications (limited to 1 per second per device);
  - In UTC (ending with `Z`);
  - Allow rule `dx` condition on `lastannounced` (e.g. to turn off a light on power on, in case it doesn't support _PowerOn On/Off_).
- Refactor updating `ResourceItem` values for `/lights` resources.
- Full web socket notifications for `/lights` resource top-level attributes.  Not all events are generated yet, though.
- Re-enable Hue app, see #2837.
- Re-enbale siren of Bitron Home 902010/24 Smoke Sensor, see #2886.


de_web_plugin.cpp
```diff
@@ -4987,6 +5163,12 @@ void DeRestPluginPrivate::addSensorNode(const deCONZ::Node *node, const SensorFi
             clusterId = VENDOR_CLUSTER_ID;
             sensorNode.addItem(DataTypeUInt16, RStateEventDuration);
         }
+        else if (modelId.startsWith(QLatin1String("ZBT-Remote-ALL-RGBW")))
+        {
+            sensorNode.addItem(DataTypeUInt16, RStateX);
+            sensorNode.addItem(DataTypeUInt16, RStateY);
+            sensorNode.addItem(DataTypeUInt16, RStateTiltAngle);
```
Thomas-Vos, line 5170: Shouldn't this be `RStateAngle` instead of `RStateTiltAngle`? From my testing so far it shows `"tiltangle":null` in the REST API instead of the angle.

Thomas-Vos, line 5170: Shouldn't this be RStateAngle instead of RStateTiltAngle? From my testing so far it shows "tiltangle":null in the REST API instead of the angle.

ebaauw: Indeed, thanks for noticing!

It's correct in `database.cpp`.  If you restart deCONZ, you should see `state.angle` instead.

bortim:
Hello @ebaauw,

when will this branch be part of an official build?
I am using deconz and api with in a docker-container on a raspberry pi.
Because of multiple missing libs I was not able to compile it directly within the container.

Regards

ebaauw:
@manup, could you please have a look at this PR?   Running it on my production for a day and haven't encountered any issues, but it's turned into a bigger one.

bortim:
Hello @ebaauw ,

I have a problem after rebooting, because the Tint remote is forgotten.
I have to add a new sensor and reset the remote by the button over the batteries.

Pressing some buttons on the tint does not result in a reachable = true.
I checked this by curl-ing the API.

Regards
bortim

rest_lights.cpp
```diff
@@ -624,31 +620,36 @@ int DeRestPluginPrivate::setLightState(const ApiRequest &req, ApiResponse &rsp)
...
-                    targetBriInc = briInc;
+                    targetBriInc = briInc < -0xFE ? -0xFE : briInc > 0xFE ? 0xFE : briInc;
                 }
             }
         }
         else if (param == "xy"  && taskRef.lightNode->item(RStateX) && taskRef.lightNode->item(RStateY) &&
                  taskRef.lightNode->modelId() != QLatin1String("FLS-PP"))
         {
-            // @manup: is check for FLS-PP needed, or is this already handled by check for state.xy?
+            // @manup: is check for FLS-PP needed, or is this already handled by check for state.x and state.y?
```
manup, line 634: Hmm good question: The check refers to the very old "FLS-PP" wireless ballast from 2012 which didn't support xy color mode there are less than 200 deployed but still active. When they are configured we'd need to strip the xy ResourceItems (the ballast wrongly claims it supports this color mode).

The newer ones with model id "FLS-PP3" support xy and don't fall into this check.

rest_lights.cpp
```diff
@@ -2445,7 +2501,14 @@ void DeRestPluginPrivate::handleLightEvent(const Event &e)
                     }
                     else if (item->lastSet().isValid() && (gwWebSocketNotifyAll || (item->lastChanged().isValid() && item->lastChanged() >= lightNode->lastStatePush)))
                     {
-                        state[key] = item->toVariant();
+                        if (rid.suffix == RStateEffect)
+                        {
+                            state[key] = RStateEffectValuesMueller[item->toNumber()];
```
manup, line 2506: Is a guard here for Müller Licht needed?

ebaauw:
> I have a problem after rebooting, because the Tint remote is forgotten.

What are you rebooting?  How do you start deCONZ after reboot?  Are you sure it starts with  your self-compiled REST API plugin?

What do you mean by “ forgotten”?  Does it still show in the GUI?  Does the `/sensors` resource still exist?  Can the remote still control lights in the associated groups?  Does the node in the GUI blinkt blue when you press a button?

> I have to add a new sensor and reset the remote by the button over the batteries.

Resetting the remote would only be needed if it has lost the association to the Zigbee network.  That cannot be caused by restarting deCONZ.

> Pressing some buttons on the tint does not result in a reachable = true.

Is that before or after reboot?  Before or after reset and re-pair?  Does `lastseen` change?

de_web_plugin.cpp
```diff
@@ -343,6 +345,8 @@ QString ApiRequest::apikey() const
 DeRestPluginPrivate::DeRestPluginPrivate(QObject *parent) :
     QObject(parent)
 {
+    plugin = this;
```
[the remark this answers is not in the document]
ebaauw: Yeah, using a global variable definitely doesn't deserve a beauty prize.  I tried to add a variable to the constructors at first, but couldn't get that to work.  Got lost in wanting to put that into `RestNodeBase` but needing to pass it to the `LightNode` and `SensorNode` classes.  I'm afraid my C++ is a bit rusty here, and Qt doesn't help.

bindings.cpp
```diff
@@ -2881,7 +2895,7 @@ void DeRestPluginPrivate::checkOldSensorGroups(Sensor *sensor)
                     i->removeDeviceMembership(sensor->id());
                 }
 
-                if (i->state() == Group::StateNormal && !i->hasDeviceMembers())
+                if (i->address() != 0 && i->state() == Group::StateNormal && !i->hasDeviceMembers())
```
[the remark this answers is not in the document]
ebaauw: There's several things going wrong with `/groups/0`.  Events are issues on `/groups/65520`  instead of on `/groups/0`.  A GET of `/groups/65520` returns the same as a GET of `/groups/0`.  The entire code handling this should be reviewed.  I think we might need to decouple the REST resource ID from the group address for all groups, similar to the Hue bridge.

rest_lights.cpp
ebaauw: You make devices of too high quality, now you never get rid of old stuff ;-)

Another argument for capabilities...

rest_lights.cpp
ebaauw: No, but better naming is.  Should probably have used `RStateEffectAllValues` vs `RStateEffectDefaultValues`.  No clue how to expand this to other lights with similar capabilities, but implemented differently.

bortim:
> > I have a problem after rebooting, because the Tint remote is forgotten.
> 
> What are you rebooting? How do you start deCONZ after reboot? Are you sure it starts with your self-compiled REST API plugin?

I am rebooting my "host" - a Raspberry Pi 3 B.
deCONZ is started as a service: systemctl enable deconz // systemctl start deconz
Yes I am, because I compiled the branch with the remote-support and copied it to the deconz folder.

> What do you mean by “ forgotten”? Does it still show in the GUI? Does the `/sensors` resource still exist? Can the remote still control lights in the associated groups? Does the node in the GUI blinkt blue when you press a button?

I am using deconz headless so I can't 
The ressource still exists - curl -X GET /sensors/21 => exists but not reachable (or something like that).
Only if i repair it again, events are created on the websocket for my iobroker.
And the remote-sensor does not have any entry in the Phoscon-Browser-App - is this an issue?

> > I have to add a new sensor and reset the remote by the button over the batteries.
> Resetting the remote would only be needed if it has lost the association to the Zigbee network. That cannot be caused by restarting deCONZ.

Hm okay - strange behavier within my network.

> > Pressing some buttons on the tint does not result in a reachable = true.
> 
> Is that before or after reboot? Before or after reset and re-pair? Does `lastseen` change?

Usually I do the following:
1. Start Raspberry with deCONZ & iobroker
2. Add batteries to the tint (I remove them after playing with iobroker, deconz, a led-stripe and the remote).
3. Reset the tint to get it working again
4. play with iobroker, deconz, a led-stripe and the remote
5. shutdown raspberry
6. begin with 1. on the next weekend

ebaauw:
> And the remote-sensor does not have any entry in the Phoscon-Browser-App - is this an issue?

For Phoscon: it doesn't (yet ?) support the remote.

> I am using deconz headless so I can't

That's unfortunate.  I wouldn't know how to troubleshoot without access to the GUI.

> The ressource still exists - curl -X GET /sensors/21
[{"error":{"address":"/sensors/21","description":"resource, /sensors/21, not available","type":3}}]

This actually suggest that the resource no longer exists.  Is it included if you do a GET of `/sensors`?

> With
sudo systemctl stop deconz and
sudo systemctl start deconz
it is reproducable for me.

Very strange.  How long after pairing the remote do you stop deCONZ?  It would seem that the newly paired sensor resource hasn't been written to the database.  Do you see the same for other devices?  I don't see how this could be related to this PR; best open an separate issue for it.

bortim:
> > And the remote-sensor does not have any entry in the Phoscon-Browser-App - is this an issue?
> 
> For Phoscon: it doesn't (yet ?) support the remote.

Thank you for clearification.

> > The ressource still exists - curl -X GET /sensors/21
> > [{"error":{"address":"/sensors/21","description":"resource, /sensors/21, not available","type":3}}]
> 
> This actually suggest that the resource no longer exists. Is it included if you do a GET of `/sensors`?

It is not included. Thank you for that hint.
But I got a solution for that:
I have installed my ubuntu and deconz new and now it is working.

Thanks a lot!

PR #2898, opened 2020-06-08
Changed Link button not pressed message

Mimiix:
The "link button not pressed" error on API is kinda confusing as it is called "authenticate" in Phoscon. This change makes that easier.


Mimiix:
@manup Is this the right way for me to create a PR? I don't have another option...

ebaauw:
And it’s called “unlock gateway” in the old web app.  The error message is copied from the Hue API.  An API client should catch it and present a user-friendly message.

Mimiix:
I did not know that. Than we probably should edit it in the documentation,

PR #3025, opened 2020-07-09
Quick & dirty support of Window Covering groups

ebaauw:
This PR adds support for `{"open": true}` and `{"open": false}` to a PUT to a group `action`, to generate _Window Covering_ cluster commands _Open_ and _Close_ to the group.  No checks whatsoever, that the group actually contains a _Windows Covering_ device.  Also, `open` is not exposed in the `action` object.

Works brilliantly with my `lumi.curtain` - haven't had a failed attempt yet (where unicast commands fails regularly due to routing issues).  Doesn't seem to work for my FYRTUR - not sure why the open/close controller works just fine.  Both FYRTUR and the controller have the same parent, though: a Hue light.

### DIFF
--- a/rest_groups.cpp
+++ b/rest_groups.cpp
@@ -768,6 +768,7 @@ int DeRestPluginPrivate::setGroupState(const ApiRequest &req, ApiResponse &rsp)
 
     bool hasOn = map.contains("on");
     bool hasOnTime = map.contains("ontime");
+    bool hasOpen = map.contains("open");
     bool hasBri = map.contains("bri");
     bool hasHue = map.contains("hue");
     bool hasSat = map.contains("sat");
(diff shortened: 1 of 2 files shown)


ebaauw:
@manup, as you suggested, the group command is send to broadcast address 0xfffd (all nodes with _Receiver On When Idle_ set), so it won't reach the FYRTUR.  How do I specify a group command using 0xffff?  

I tried to set both:
```
            task.req.dstAddress().setGroup(group->address());
            task.req.dstAddress().setNwk(deCONZ::BroadcastAll);
```
But when using `task.req.setDstAddressMode(deCONZ::ApsGroupAddress);` the group command is sent to 0xfffd, and when using `task.req.setDstAddressMode(deCONZ::ApsNwkAddress);` the command is sent to 0xffff, but the group is missing.

Issue #3079, opened 2020-07-25
API doesn't update colour for Hue Bloom


ebaauw:
## Describe the bug
The colour of the Hue Bloom isn't updated in the API.

## Steps to reproduce the behavior
Set the colour to `{"xy": [0, 0]}`.  The bloom sets itself to its blue point, [0.138, 0.08], but the API continues to report [0, 0], until the _Color Control_ cluster attributes are read manually in the GUI.

## Expected behavior
API should update the `xy` value.

## Additional context
Sniffer log confirms that API plugin isn't polling the colour attributes.  It is polling the _Color Control_ cluster alright, but for attributes 0x0008, 0x4001, 0x400a, 0x400b, 0x400c; even when the light is off.  Looks like the API is stuck trying to initialise ctmin (0x400b) and ctmax (0x400c), but these attributes are unsupported for type Color Light.

See also https://github.com/dresden-elektronik/deconz-rest-plugin/issues/1209, where this issue was discovered.

ebaauw:
I'm pretty sure this used to work in the (distant) past.  When creating `ph probe`, I did test it on the Bloom, and it worked.  But now `ph probe` times out, because the API doesn't update the colour.

ebaauw:
The issue is that for a _Color Light_, the `RConfigColorCapabilities` resource item isn't added, so the response to reading the _Color Capabilities_ isn't stored, causing the poll manager to thing the light hasn't yet been initialised.

ebaauw:
Haven't tested yet with the Tint remote, but `ph probe` now works again.

Mimiix:
@ebaauw Thanks for making me smile this morning Erik :)

thedannymullen:
> The issue is that for a _Color Light_, the `RConfigColorCapabilities` resource item isn't added, so the response to reading the _Color Capabilities_ isn't stored, causing the poll manager to think the light hasn't yet been initialised.

Would this problem exist in other color lights possibly?  My tradfri color lights have developed some issues since I updated.   I am looking through the issues before I start debug.

ebaauw:
Yes, all lights of type _Color light_.

ebaauw:
Note that `RConfigColorCapabilities` was suppressed from the response to a GET of a `/lights` resource, but included in the web socket notification.  It's now exposed as `colorcapabilities` top-level attribute on a GET as well (so the value can be checked with access to the GUI).

klada:
[...]

Issue #2393, opened 2020-01-29
New Device Support Request - eCozy Smart Heating Thermostat


ddppddpp:
I'd like to request the integration of a automatic radiator valve (TRV) by eCozy, the 1TST.
Currently the device is visible via VNC, I'd ultimately like to use it in a hassio/deConz add-on/conbee setup.
<img width="348" alt="Screenshot 2020-01-29 at 21 59 33" src="https://user-images.githubusercontent.com/45116849/73392852-48866c80-42e3-11ea-9212-5599f9ac46d4.png">
<img width="338" alt="Screenshot 2020-01-29 at 21 59 48" src="https://user-images.githubusercontent.com/45116849/73392888-5805b580-42e3-11ea-8cf9-71998295350a.png">
<img width="340" alt="Screenshot 2020-01-29 at 22 00 12" src="https://user-images.githubusercontent.com/45116849/73392908-6227b400-42e3-11ea-8c9b-6ae1a33c393a.png">
<img width="341" alt="Screenshot 2020-01-29 at 22 01 08" src="https://user-images.githubusercontent.com/45116849/73392925-694ec200-42e3-11ea-93be-f5cdc9090f83.png">

Thanks in advance!

SwoopX:
Can you please also add a screenshot of the node info panel?

ddppddpp:
Sure, here it is, more structured:
Node info:
<img width="349" alt="Screenshot 2020-02-07 at 16 12 23" src="https://user-images.githubusercontent.com/45116849/74085342-d6e7b480-4a80-11ea-80b0-4f55fd06dd60.png">
Basic Cluster:
<img width="347" alt="Screenshot 2020-02-07 at 16 17 44" src="https://user-images.githubusercontent.com/45116849/74085376-12827e80-4a81-11ea-81f2-877a95488890.png">
Power Configuration Cluster:
<img width="351" alt="Screenshot 2020-02-07 at 16 17 54" src="https://user-images.githubusercontent.com/45116849/74085382-2201c780-4a81-11ea-9dee-4d3b25e6647b.png">
Identify Cluster:
<img width="345" alt="Screenshot 2020-02-07 at 16 18 04" src="https://user-images.githubusercontent.com/45116849/74085390-2e862000-4a81-11ea-9a25-d2ac697faba2.png">
Time Cluster:
[...]

PR #3087, opened 2020-07-28
eCozy Thermostat and more

ebaauw:
- Additional support for the eCozy Thermostat, see #2393.  Still work in progress:
  - Full GUI support for _Thermostat_ cluster, including schedules and tracking setpoint changes;
  - Support _Time_ cluster as `ZHATime` `/sensors` resource;
  - Sync on-device hardware clock with system clock;
  - Fix hang in API plugin when receiving a schedule with more than one transition;
  - Expose _PI Heating Demand_ as `state.valve`;
  - Base `config.scheduleron` on _Temperature Setpoint Hold_.
- Report `lastseen` with a resolution of minutes, and issue at most one web socket notification per minute per device for `lastseen`;
- Bug fix: the colour attributes for _Color Light_ lights weren't updated, see #3079.


ebaauw:
Hm, I was still working on this one.  The changes are not complete, see #2393.

ebaauw:
Getting a segmentation fault in v2.05.80 when reading an old-version schedule from the database.
```
$ gdb /usr/bin/deCONZ core-deCONZ-sig11-user1000-group1000-pid2656-time1597505417
(...)
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib/arm-linux-gnueabihf/libthread_db.so.1".
Core was generated by `/usr/bin/deCONZ --http-port=80 --dbg-info=2 --dbg-aps=2 --dbg-error=1'.
Program terminated with signal SIGSEGV, Segmentation fault.
#0  0x75fe247c in QString::split(QString const&, QString::SplitBehavior, Qt::CaseSensitivity) const ()
   from /usr/lib/arm-linux-gnueabihf/libQt5Core.so.5
[Current thread is 1 (Thread 0x76fbea00 (LWP 2656))]
(gdb) bt
#0  0x75fe247c in QString::split(QString const&, QString::SplitBehavior, Qt::CaseSensitivity) const ()
   from /usr/lib/arm-linux-gnueabihf/libQt5Core.so.5
#1  0x6bd9d2e8 in DeRestPluginPrivate::deserialiseThermostatTransitions(QString const&, QList<QVariant>*) ()
   from /usr/share/deCONZ/plugins/libde_rest_plugin.so
#2  0x6bd9eb3c in DeRestPluginPrivate::deserialiseThermostatSchedule(QString const&, QMap<QString, QVariant>*) ()
   from /usr/share/deCONZ/plugins/libde_rest_plugin.so
#3  0x6bd5bf40 in DeRestPluginPrivate::sensorToMap(Sensor const*, QMap<QString, QVariant>&, ApiRequest const&) ()
   from /usr/share/deCONZ/plugins/libde_rest_plugin.so
#4  0x6bd5cda4 in DeRestPluginPrivate::getAllSensors(ApiRequest const&, ApiResponse&) ()
   from /usr/share/deCONZ/plugins/libde_rest_plugin.so
#5  0x6bd6d3a4 in DeRestPluginPrivate::handleSensorsApi(ApiRequest const&, ApiResponse&) ()
   from /usr/share/deCONZ/plugins/libde_rest_plugin.so
#6  0x6bcc9dc0 in DeRestPlugin::handleHttpRequest(QHttpRequestHeader const&, QTcpSocket*) ()
   from /usr/share/deCONZ/plugins/libde_rest_plugin.so
#7  0x000bd06c in zmHttpClient::handleHttpRequest() ()
#8  0x7615fbfc in QMetaObject::activate(QObject*, int, int, void**) () from /usr/lib/arm-linux-gnueabihf/libQt5Core.so.5
#9  0x76402964 in ?? () from /usr/lib/arm-linux-gnueabihf/libQt5Network.so.5
Backtrace stopped: previous frame identical to this frame (corrupt stack?)
```

ma-ca:
> Rename `config.scheduleron` to `config.schedule_on` and expose _Temperature Setpoint Hold_ (0x0023) as `config.schedule_on`

@ebaauw The Bitron Thermostat 902010/32 does not have attribute 0x0023 and the `config.scheduleron` was originally mapped to attribute _Thermostat Programming Operation Mode_ 0x0025. This is a breaking change.

I am using HomeKit to turn off and on the Bitron Thermostat scheduler. In HomeKit the thermostat status on or off was mapped to the scheduler being enabled or disabled. That is a very useful feature for me, for example when leaving for holidays in the winter time. 

Is it possible to revert the that change and bring back the mapping in HomeKit between scheduler and thermostat status? Or at least to make it configurable how the thermostat scheduler is mapped into HomeKit?

ebaauw:
Does the Bitron support on-device schedules?  Using the Zigbee standard or some proprietary method?

I only mapped `config.schedule_on` to 0x0023 for the eCozy; for others it’s still mapped to 0x0025.

I think there’s something wrong in the way Homebridge Hue currently uses the _Thermostat_ mode.  In HomeKit, you can set the target mode to _Off_, _Heat_, _Cool_, or _Auto_, indicating whether the thermostat should automatically heat, cool, or both.  The current mode is _Off_, _Heat_, or _Cool_, indicating whether the thermostat is currently heating or cooling.  This has nothing to do with whether the schedule is enabled.  It also has nothing to do whether the Spirit is in automatic mode (where it sets the valve position based on the heatsetpoint) or in manual mode (where you set the value position manually).

Eve shows a separate on/off switch for the schedule, I want to see if I can map that to `config.schedule_on`.  I also want to support setting the on-device from Eve.  Showing the on-device schedule in Eve won’t always be possible, since the on-device schedule is way richer than the Eve schedule.  See https://github.com/ebaauw/homebridge-hue/issues/728.

ma-ca:
> Does the Bitron support on-device schedules? Using the Zigbee standard

Yes, using Zigbee standerd. I did add support in PR #1003 for Thermostat cluster (including get/set schedule) and Bitron Thermostat 902010/32.

> I only mapped `config.schedule_on` to 0x0023 for the eCozy; for others it’s still mapped to 0x0025.

Makes me happy again.

> I think there’s something wrong in the way Homebridge Hue currently uses the _Thermostat_ mode.

That is true. I don't use Eve very often and was not aware about the option _Enabled_  in the thermostat. Anyway, I am also happy to use Eve to enable and disable the scheduler.

ebaauw:
> I did add support in PR #1003 for Thermostat cluster (including get/set schedule) and Bitron Thermostat 902010/32.

Ah, cool.  Did you see in #2393 that `config.schedule` now returns proper JSON and that I changed the way the schedule is set?  Also, I setup periodically reading the schedule for the eCozy, we might want to do the same for the Bitron.  Does the Bitron support attribute reporting for all (relevant) attributes, or do we need to poll it for some of the attributes?

ebaauw:
Another question, @ma-ca: how does the Bitron handle the timing of the schedule?  Does it have an on-device clock?  How do you keep that sync'ed?

ma-ca:
> how does the Bitron handle the timing of the schedule?

The Bitron has a Time cluster that syncs the attributes for UTC time, time zone, dst start, dst end and dst shift.

ma-ca:
> Does the Bitron support attribute reporting for all (relevant) attributes, or do we need to poll it for some of the attributes?

Reporting is enabled and configured every 5 minutes for attribute 0x0000. Polling is already configured every 15 minutes for attributes 0x0012, 0x0025 and 0x0029.

ebaauw:
> The Bitron has a Time cluster that syncs the attributes for UTC time, time zone, dst start, dst end and dst shift.

Is that a client (grey) _Time_ cluster, or a server (blue) one?  Does it find the _Time_ cluster on the coordinator and sync itself automatically?  Or did you set something up?

ma-ca:
A Time server cluster that syncs automatically to the coordinator. I did implement #774 especially for the Bitron.

ebaauw:
The Bitron also supports 0x0030, 0x0031, and 0x0032, so it could also expose `config.lastchange`.

ma-ca:
> There's a bug in the eCozy that it reports _Setpoint Change Time Stamp_ in local time instead of UTC. Does the Bitron do the same?

The Bitron attribute 0x0032 value is the correct UTC time (right after changing heat setpoint the 0x0032 attribute value is close to _Time_ cluster 0x000A attribute 0x0000 value with UTC time)

PR #3127, opened 2020-08-08
Outsource button maps to external JSON file

SwoopX:
All available button maps have been outsourced to an external JSON file named `button_maps.json`, which must reside in the same directory as the deCONZ database `zll.db`. The purpose is to speed up integration of new switches as users can now create it theirselves (given that some appropriate guidance is provided).

The code currently runs flawlessly in my production environment.

Tested and therefore working with:
- Busch Jäger Switches
- EcoDim Switches


manup:
Be aware, when compiling the plugin before the next release the `button_maps.json` needs to be copied (or symlinked) into `~/.local/share/dresden-elektronik/deCONZ` in order to be found.

PR #3207, opened 2020-08-29
Optimize poll manager to consider only avaliable attributes

SwoopX:
When cluster to be polled is found, extend the check to also determine if an attribute is available. If not, they get discarded (mainly applicable for instantaneous demand attribute on simple metering cluster). Also, freshness of values is only checked on available attributes now.


manup:
The PR is included in:

http://deconz.dresden-elektronik.de/raspbian/alpha/deconz-2.05.80-raspbian-jessie-alpha~git.90b522.474ac1_armhf.deb

Issue #2108, opened 2019-11-22
Add Support for Illumra Dual Rocker Switch


mdolnik:
![Product Image](https://illumra.com/wp-content/uploads/2016/07/ZBT-S2AWH-1.png)
[Product Page](https://illumra.com/products/wireless-light-switches/dual-rocker-zigbee-switch/)

I'm aware that this switch uses the PTM 215Z (or PTM 215ZE?) module which has already been somewhat covered in issues [#1579](https://github.com/dresden-elektronik/deconz-rest-plugin/issues/1579) and [#1754](https://github.com/dresden-elektronik/deconz-rest-plugin/issues/1754) but those seem to be for a different type of switch.

I am able to pair this switch in Phoscon by using any of the 4 buttons and then:
1. Hold the button down for 10 seconds and let go.
1. Hold the button down for 1 second and let go.
1. Hold the button down for 10 seconds and let go.
1. Press the button between 1-10 times (to run through the channels) until a connection is made.
1. Press any other button to complete pairing.

(The instructions on [pairing-friends-of-hue-switch](https://phoscon.de/en/support#pairing-friends-of-hue-switch) do not work for me)

The intended way to pair this switch (directly to a zigbee receiver) is to do the steps above for either of the left buttons, then do the same for either of the right buttons [as shown here, but for a single switch](https://www.youtube.com/watch?v=V43Jcw_uUPE)

The issue is only two of the buttons seem to respond (only the top-left and bottom-right) I cannot get the bottom left or top right to trigger any events.

When paired in Phoscon, the switch shows up as a Philips Tap, and I am able to use the switch editor to bind the Philips Tap's "1" and "4" buttons successfully, but the "2" and "3" button bindings do not seem to get triggered.

I have also not been able to get the button holding functionality working, but I don't know where to start with that, I would like to get all 4 buttons working first.

Any suggestions to get this working?

mdolnik:
The switch seems to show up in deConz attached to one of my lights ([sunricher switch/receiver](https://sunricher.en.alibaba.com/product/62359944204-811954196/Zigbee_3_0_2_way_no_neutral_Rotary_dimmer.html)) as `Green Power GP Combo Basic`
![2019-11-21_22h22_10](https://user-images.githubusercontent.com/22206300/69402741-9c565300-0cad-11ea-81a9-e794ed913df2.png)

And there is also this, which I don't know if it's related or not, I don't know what this is...
[...]

PR #3285, opened 2020-09-19
Add support for Illumra Dual Rocker Switch which uses the FOH PTM215ZE module

mdolnik:
This is a pull request for issue [#2108](https://github.com/dresden-elektronik/deconz-rest-plugin/issues/2108)

I am new to C++ so feel free to make any code suggestions.

### DIFF
--- a/de_web_plugin.cpp
+++ b/de_web_plugin.cpp
@@ -1167,30 +1167,55 @@ void DeRestPluginPrivate::gpProcessButtonEvent(const deCONZ::GpDataIndication &i
     quint32 btn = ind.gpdCommandId();
     if (sensor->modelId() == QLatin1String("FOHSWITCH"))
     {
-        const quint32 buttonMap[] = {
-            0x10, S_BUTTON_1,
-            0x11, S_BUTTON_2,
-            0x12, S_BUTTON_4,
-            0x13, S_BUTTON_3,
-            0x14, S_BUTTON_1,
-            0x15, S_BUTTON_2,
-            0x16, S_BUTTON_4,
-            0x17, S_BUTTON_3,
-            0x62, S_BUTTON_6,
-            0x63, S_BUTTON_6,
-            0x64, S_BUTTON_5,
-            0x65, S_BUTTON_5,
-            0x68, S_BUTTON_7,
-            0xe0, S_BUTTON_7,
+        // Map the command to the mapped button and action.
+        // PTM215ZE Friends of Hue switch.
+        quint32 buttonMapPTM215ZE[] = {
+            0x12, S_BUTTON_1, S_BUTTON_ACTION_INITIAL_PRESS,
+            0x13, S_BUTTON_1, S_BUTTON_ACTION_SHORT_RELEASED,
+            0x14, S_BUTTON_2, S_BUTTON_ACTION_INITIAL_PRESS,
+            0x15, S_BUTTON_2, S_BUTTON_ACTION_SHORT_RELEASED,
+            0x18, S_BUTTON_3, S_BUTTON_ACTION_INITIAL_PRESS,
+            0x19, S_BUTTON_3, S_BUTTON_ACTION_SHORT_RELEASED,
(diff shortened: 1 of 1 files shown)


manup:
I'd like to merge this for the 15th October beta. Did you have a chance to test this with the other ZGP switch types?

mdolnik:
> I'd like to merge this for the 15th October beta.

Great, thanks for the quick response!

> Did you have a chance to test this with the other ZGP switch types?

Unfortunately this Illumra switch is the only green-power switch I have, so I am unable to test with any others, but just by judging the code it _should_ avoid changing anything for other models (provided the `Philips` model below the Illumra declaration doesn't share the `ind.payload().size() == 27`)

Also I don't understand the impacts of setting the SW version to a string ID instead of a number, I just used that as a means of differentiating this model in the button mapping section.

I'm open for any code suggestions that I can try out on my end.

manup:
That's fine I think we have all these switches in the company, will schedule some tests.
As far as I'm aware the swversion is only displayed but not uses otherwise, so it should be save.

mdolnik:
Also a minor thing is I set the manufacturer to `PhilipsFoH` when it seems it is actually [EnOcean](https://www.enocean.com/en/products/enocean_modules_24ghz/ptm-215ze/).

I believe I may have either copy/pasted another switch declaration and forgot to change the manufacturer or I may have though Phillips was the manufacturer [from here since it uses the same module](http://www.docs.lighting.philips.com/en_gb/oem/download/xitanium/SCP_PHIL_164296_PMT215ZE_A03.pdf)

manup:
The PhilipsFoH was used to reflect the original Hue REST-API (under the hood all ZGP switches use a EnOcean module).

manup:
I've added a check for the sourceID to prevent confusion with Hue Tap, however I couldn't test this.
Can you please provide the sourceID prefix your switch is using?

mdolnik:
> I've added a check for the sourceID to prevent confusion with Hue Tap, however I couldn't test this.
> Can you please provide the sourceID prefix your switch is using?

I don't know where to find the sourceID, I can't find anything in the deCONZ GUI that mentions anything about a source ID...

Judging by your commit ef15ee1 you are comparing using `ind.gpdSrcId()`. I can't say I know much about working with hexadecimals, but I debugged using:
```
QVariantMap map;
map["ind.gpdSrcId()"] = ind.gpdSrcId();
map["(ind.gpdSrcId() & 0x01700000)"] = (ind.gpdSrcId() & 0x01700000);
map["(ind.gpdSrcId() & 0x01700000) == 0x01700000"] = (ind.gpdSrcId() & 0x01700000) == 0x01700000;
map["0x + QString::number(ind.gpdSrcId(),16).toUpper()"] = "0x" + QString::number(ind.gpdSrcId(),16).toUpper();
DBG_Printf(DBG_INFO, "\n\n\n%s\n\n\n", qPrintable(Json::serialize(map)));
```

And got the following results:
```
{
  "(ind.gpdSrcId() & 0x01700000)":22020096,
  "(ind.gpdSrcId() & 0x01700000) == 0x01700000":false,
  "0x + QString::number(ind.gpdSrcId(),16).toUpper()":"0x15058D2",
  "ind.gpdSrcId()":22042834
}
```

So it seems it is a different ID than the commit is expecting...

Are these debugging values helpful? Or should I try something else?

manup:
Thanks that helps, they seem to use a different vendor prefix than the other Friends of Hue switches.
I think in future we need a more complete approach to detect various switches, perhaps by first testing all buttons in a multi stage process.

mdolnik:
I can confirm the latest master commit 45a769 works now with the Illumra switch.

Works for both commissioning as type "Other" and is able to control lights with no issues 

Thanks for your help.

ICTSMART:
Started testing today with the PTM 215Z (EnOcean) as discussed on https://github.com/dresden-elektronik/deconz-rest-plugin/issues/2780. Unfortunately the switch commands are not passed through the mesh.
When near to the Raspbee, I see the events coming in - when sufficient away with at least 2 nodes in between, messages are lost. Please let me know how I can help here.

PR #3300, opened 2020-09-22
Add support Sonoff SNZB-03 with different ModelIDs

SwoopX:
Including this PR, the following modelIDs are known: "MS01", "MSO1" and "ms01".


ASNNetworks:
How long does it normally take for PR's to be included with the HA addon release?

SwoopX:
It may take more than 5 weeks since HA folks only use stable releases for their addon. If you use docker, than chances are high to have it available real soon after the next beta release, usually provided on the 15th.

PR #3326, opened 2020-09-27
Enable button map assignment to switches via button_maps.json

SwoopX:
The button maps have been extended to now also contain the associated zigbee model IDs, as this allows individual modification in button_maps.json without touching the code.

Internally, variable `buttonMapForModelId` holds the mapping. Model IDs are currently compared by `startsWith()`, which should preferably superseeded by an exact string comparision. Due to the previous string comparison in the code, a handful of switches might experience some issues at the beginning. This can, however, be easily resolved by adding/updating the corresponding model IDs in button_maps.json.


SwoopX:
Closing this one for better merge conflict resolution.

Issue #3123, opened 2020-08-06
Get Elko Thermostat to show temperature for floor sensor


corvy:
This is a feature request to fully get this device supported: 

https://github.com/dresden-elektronik/deconz-rest-plugin/issues/1291

From a developer of Homey I found that the floor sensor is attribute 0x409, but this does not show in REST. Local Temperature is always the air temperature. Would that be possible?

What information would be needed to get this done? Thank you!

Device
Screenshots

olekenneth:
Also missing state/operation. Home Assistant always says it's off.

chairman2s:
@olekenneth I have the same problem. Are you using the api? Mine just appeared in HA but after a while it's gone. Then I must go into VNC and it comes back....

SwoopX:
It might take a bit time to realize that since we have 2 challenges here. One is the extention of the API, the other one that deconz currently misbehaves if there's more than one manufacturer specific attribute with the same ID. The one coming first gets precedence (which shouldn't be).

You only need the attribute to be read, right?

corvy:
Yes that's right. The thermostat is set to warm based on the air or floor sensor. Is we could read this then we could get proper reading of current status. Setting the target works no matter what sensor is selected.

kojax-net:
This device has many vendor specific attributes. A developer of a smartthings app has listed them here: 
https://github.com/nilskaa/Smartthings/blob/master/devicetypes/Elko%20Super%20Thermostat%20-%20vendor%20specific%20attributes.txt

I think these attributes are the most important, and should also be implemented:
0x403 (encoding:30, value: <verified 00=local sensing mode, 01=remote sensing mode, 03=floor max temp mode>
0x409 (encoding:29 value: <verified remote/floor temperature sensor measurement>
0x415 (floating encoding:10, value: <verified heating active/inactive> 00=idle 01=heating

corvy:
I would like to keep this issue open. It is very much wanted still!

gesv:
> I would like to keep this issue open. It is very much wanted still!

I couldn't agree more. Please keep this issue open until the issue is fixed.

t112013:
Please keep this alive. 
this is something we need 👍

Mimiix:
@SwoopX Is this added?

SwoopX:
[...]

Issue #1291, opened 2019-02-22
[Request device support] Elko thermostat


Gaute64Bit:
ELKO Thermostat
http://proff.elko.no/elko-smart-home-super-termostat/esh-plus-super-tr-rf-ph-article2629-1779.html 

Need support for reading local temp and setting Heating (and cooling?) Setpoint.

![image](https://user-images.githubusercontent.com/27805850/53239290-f339e800-369b-11e9-8c70-b9e4d8c4afdf.png)

![image](https://user-images.githubusercontent.com/27805850/53239303-f92fc900-369b-11e9-8c26-6e2a0f999623.png)

![image](https://user-images.githubusercontent.com/27805850/53239312-fd5be680-369b-11e9-998a-41722b5da1e1.png)

![image](https://user-images.githubusercontent.com/27805850/53239417-457b0900-369c-11e9-8c2c-0baadc597112.png)

Contact me if more infor is required.

epiespen:
Yes, a big +1 here. Kinda dissappointed there were no support for this.

avassdal:
+1

thomas70:
+1

ingvaldhaaland:
+1

t112013:
oh yes please. this is very much needed :D

dreuf83:
Making a +1 on this and hope that someone with the knowledge sees it :)

jyrki69:
+1
Would really appreciate support for Elko Thermostat. Let me know if I can help in any way.

epiespen:
@ma-ca Is this what you need in reference to #2009 ?

![image](https://user-images.githubusercontent.com/10612746/68247443-27331e80-001b-11ea-9099-88547787e290.png)

![image](https://user-images.githubusercontent.com/10612746/68247578-69f4f680-001b-11ea-8253-f86407f66fff.png)

t112013:
Plsss someone be kind and fix it :)

eidsaa:
Hi

There are several more vendor specific attributes, such as floor temperature, child lock, heating on/off etc. In the link below there is a list showing vendor specific attributes. 

https://github.com/prj84/com.Elko/blob/master/drivers/ESHSUPERTR/ST_Code/Elko%20Thermostat%20-%20Vendor%20Specific%20attributes.txt

ricmik:
@eidsaa Maybe Elko will be helpful and give you a specification for the vendor specific attributes if you contact them?

[...]

PR #3329, opened 2020-09-27
Improve support for ELKO Super TR thermostat

SwoopX:
Improve support for ELKO Super TR thermostat (#1291, #3123)

Enablement of sensor bindings is not included, but covered by a different PR (#3409).
- Correction of manufacturer specific attributes
- Bindings: heating setpoint, thermostat mode, device on/off, floor temperature, child lock, heating active/inactive
- Introduce `state/floortemperature`, `state/heating` and `config/temperaturemeasurement` to REST API
- `state/floortemperature` provides the floor temperature in addition, analog to the already given temperature (Int16)
- `state/heating` indicates if heating is active/inactive (Bool)
- `config/temperaturemeasurement` sets the mode of operation and allows to set the following string values: **air sensor**, **floor sensor**, **floor protection**
- The manufacturer specific attribute 0x0406 (Device on) is mapped to `state/on`


arnerek:
I use the Elko Super TR in regulator mode, i.e. I create my own thermostat by having a signee thermometer in the bathroom where the heating is applied. In this mode the local heatsetpoint attribute is the load in %, i.e. 500 is mapped to 5%. 

I noticed that the Rest API is not allowing me to set a lower set point than 5% and higher than 30%. I have tracked the problem to this PR: https://github.com/dresden-elektronik/deconz-rest-plugin/pull/3155/. Should this sanity check be dependent on the device? I noticed that the attributes 0x0003 (abs min heat setpoint limit) and (0x0004 (max heat setpoint limit) change depending if I use regulator or thermostat mode, but I don't know if all manufacturers use these attributes. Should I make a new issue since this is not only related to Elko Super TR device support?

SwoopX:
What sanity check? The PR doesn't include anything you're referring to as far as I can tell. Each thermostat handles the min/max values differently and the attributes you mentioned are not responsible for setting the boundaries. So bottom line is: it's a zoo out there.

arnerek:
I was referring to this line:
`if (!ok || heatsetpoint < 500 || heatsetpoint > 3000)`

SwoopX:
This is irrelevant as it applies for on-device schedules only.

arnerek:
Ok. Sorry, I did not trace to the correct section. I am guessing the sanity check of requests are within resource.cpp? 
Still, I am not able to set e.g. heatsetpoint: 5000 (50% load), but I guess my use of this is a real corner-case and not a problem for many users.

SwoopX:
Erm, that's gonna be funny. Setheatpoint is degrees, not percent. In that sense, you' try to set the thermostat to 50° 😂

Maybe I can discuss with the others

oywino:
The great work that you have done, have brought us all a long step further. Fantastic!
But there are still some loose ends; The ELKO Super TR is still "invisible" in the Phoscon UI, and as a consequence of that (i think) it doesn't show up in the available device list in node-red-contrib-deconz OUT node.
The author of the node-red-contrib-deconz Pallette says that this is not something he can control as he only gets his device list from Phoscon. If it isn't in Phoscon, then - sorry.  So his advice is for us to issue a new device request in Phoscon. 
Well, I tried to follow that path, and it eventually just took me back here again.  So I'm running in circles now.
The ELKO Thermostat is listed as a supported device for deCONZ, but how can you make it become visible in Phoscon also?
Obviously, Phoscon does playa role in the equation, because the only way to add the Thermostat to deCONZ is through Phoscon. This lead me to believe that Phoscon is aware of its existence (how else can it be able to add it?), but for some reason it doesn't show up in the sensor list in Phoscon.
Can that be fixed ? Please ?

SwoopX:
>  So his advice is for us to issue a new device request in Phoscon.

Correct. But note that this is the deconz rest api plugin repo, so you need to raise it at the phoscon-beta repo. Technically, the device is ready for use.

oywino:
OK, so I have now files a new Device Request in the phoscon-beta repo here:
[https://github.com/dresden-elektronik/phoscon-app-beta/issues/248](url)

PR #3335, opened 2020-09-30
Update bindings.cpp

aofc:
Added initial support for WL4200 water leak sensor in addition to sibbling WL4200S and description for Sinope thermostats


SwoopX:
This is not gonna work since a couple of other relevant places have been missed.

aofc:
If you know where there is something to add, please tell me, I've searched everywhere and have not found other references for this particular sensor. I'm new to this, but very willing to help.

Issue #3397, opened 2020-10-11
Missing button map for ZGRC-KEY-009


rvanmoorsel:
## Describe the bug

I've got two wall switches that appear to be quite common in Europe. They fit in a normal wall frame and can replace traditional switches. They're marketed under several names, but deCONZ recognizes them as Sunricher ZGRC-KEY-XXX. In a old version of deCONZ I could use the 'deconz_event' service listener in HA, but now they aren't sending the switch presses anymore. 
They are recognized correctly in deCONZ (and obviously HA), but the button map seems off. 
I've posted this issue in another thread (https://github.com/dresden-elektronik/deconz-rest-plugin/issues/1286) and the user SwoopX advised me to file a bug report.

## Steps to reproduce the behavior

  If the problem is reproducable, list the steps here:

  1. I press a button and it isn't registered.
  2. The logfile produces an [INFO] - No button map for... entry.

## Expected behavior

  I'd like to see a button pressed event in HA / deCONZ

## Screenshots

  Not applicable.

## Environment

  deCONZ version: 2.05.84 / 14-9-2020
  Firmware version: 26660700

 - Host system: Raspberry Pi
 - Running method: Home Assistent deCONZ Add-on
 - Firmware version: (26660700)
 - deCONZ version: (2.05.84)
 - Device: ConBee II
 - Do you use an USB extension cable: yes -- only relevant for ConBee I/II

### deCONZ Logs

  There are two rows with each four buttons.
  When I press the 'ON' (Top left) button, this debug log is written:
    [INFO] - No button map for: ZGRC-KEY-009 endpoint: 0x01 cluster: ONOFF (0x0006) command: ON (0x01) payload[0]: 000
  When I press the 'OFF' (Top right) button, this debug log is written:
    [INFO] - No button map for: ZGRC-KEY-009 endpoint: 0x01 cluster: ONOFF (0x0006) command: OFF (0x00) payload[0]: 000
  When I press the second button from the top on the left (Brighter), this debug log is written:
    [INFO] - No button map for: ZGRC-KEY-009 endpoint: 0x01 cluster: LEVEL_CONTROL (0x0008) command: STEP_WITH_ON_OFF (0x06) payload[0]: 000
  When I press the second button from the top on the right (Dim), this debug log is written:
[...]

Issue #3396, opened 2020-10-11
Missing button map for ZGRC-KEY-007


rvanmoorsel:
## Describe the bug

I've got two wall switches that appear to be quite common in Europe. They fit in a normal wall frame and can replace traditional switches. They're marketed under several names, but deCONZ recognizes them as Sunricher ZGRC-KEY-XXX. In a old version of deCONZ I could use the 'deconz_event' service listener in HA, but now they aren't sending the switch presses anymore. 
They are recognized correctly in deCONZ (and obviously HA), but the button map seems off. 
I've posted this issue in another thread (https://github.com/dresden-elektronik/deconz-rest-plugin/issues/1286) and the user SwoopX advised me to file a bug report.

## Steps to reproduce the behavior

  If the problem is reproducable, list the steps here:

  1. I press a button and it isn't registered.
  2. The logfile produces an [INFO] - No button map for... entry.

## Expected behavior

  I'd like to see a button pressed event in HA / deCONZ

## Screenshots

  Not applicable.

## Environment

  deCONZ version: 2.05.84 / 14-9-2020
  Firmware version: 26660700

 - Host system: Raspberry Pi
 - Running method: Home Assistent deCONZ Add-on
 - Firmware version: (26660700)
 - deCONZ version: (2.05.84)
 - Device: ConBee II
 - Do you use an USB extension cable: yes -- only relevant for ConBee I/II

### deCONZ Logs

  When I press the 'ON' (Left) button, this debug log is written:
    [INFO] - No button map for: ZGRC-KEY-007 endpoint: 0x01 cluster: ONOFF (0x0006) command: ON (0x01) payload[0]: 000
  When I press the 'OFF' (right) button, this debug log is written:
    [INFO] - No button map for: ZGRC-KEY-007 endpoint: 0x01 cluster: ONOFF (0x0006) command: OFF (0x00) payload[0]: 000
  When I press the BOTH buttons, this debug log is written:
    [INFO] - No button map for: ZGRC-KEY-007 endpoint: 0x01 cluster: 0x0001 command: 0x0A payload[0]: 021

## Additional context

  It's a two-button switch. I've also got an eight-button version, and I've posted a separate bug report for this one (https://github.com/dresden-elektronik/deconz-rest-plugin/issues/3397).
[...]

Issue #3393, opened 2020-10-11
Bosch  ISW-ZPR1-WP13 pir sensor missing features like tamper and battery percentage


mamrai1:
## Device

- Product name: Bosch  ISW-ZPR1-WP13
- Manufacturer: Bosch
- Model identifier:  ISW-ZPR1-WP13
- Device type : Please remove all unrelated device types. 
  - Sensor

  - Other: 

## Screenshots

![basic](https://user-images.githubusercontent.com/57288344/95672373-f81d1980-0ba8-11eb-92c2-d430613d7a96.png)
![diagnostics1](https://user-images.githubusercontent.com/57288344/95672374-f94e4680-0ba8-11eb-8359-ff9bdb96856a.png)
![diagnostics2](https://user-images.githubusercontent.com/57288344/95672376-f9e6dd00-0ba8-11eb-8101-e328306540eb.png)
![iaszone](https://user-images.githubusercontent.com/57288344/95672379-fb180a00-0ba8-11eb-9b9a-385ba702b9d2.png)
![identify](https://user-images.githubusercontent.com/57288344/95672381-fbb0a080-0ba8-11eb-81a1-f8715a53e015.png)
![nodeinfo](https://user-images.githubusercontent.com/57288344/95672383-fc493700-0ba8-11eb-9382-4a17644f5814.png)
![otau](https://user-images.githubusercontent.com/57288344/95672384-fce1cd80-0ba8-11eb-94c8-409fea87ade0.png)
![phoscon](https://user-images.githubusercontent.com/57288344/95672385-fd7a6400-0ba8-11eb-8ea1-0b5873342bb0.png)
![poll](https://user-images.githubusercontent.com/57288344/95672386-fe12fa80-0ba8-11eb-9f59-90335c1b8282.png)
![power1](https://user-images.githubusercontent.com/57288344/95672387-feab9100-0ba8-11eb-8dfc-833a6c1ddd3c.png)
![power2](https://user-images.githubusercontent.com/57288344/95672389-ff442780-0ba8-11eb-9ab6-c24868ce78fa.png)
![temp](https://user-images.githubusercontent.com/57288344/95672390-ffdcbe00-0ba8-11eb-9be3-c45b5d1b3214.png)

mamrai1:
1) motion sensor is triggered for 1 second and then sleeps for 3 minutes..
2) temperature is reported only when there is a change over 0.5 degrees celcius...( as i can see from home assistant history)
[...]

Issue #3375, opened 2020-10-06
SmartThings Motion Sensor (2018 model) report water no presence 


martinezmp3:
deCONZ REST-API is reporting motion sensor (SmartThings Motion Sensor (2018 model) | IM6001-MTP01) as a water sensor 

## Describe the bug
the info received on the subscribe WS is the fallow:
when motion:
{"e":"changed","id":"19","r":"sensors","state":{"lastupdated":"2020-10-07T22:46:57.691","lowbattery":false,"tampered":false,"water":true},"t":"event","uniqueid":"28:6d:97:00:01:00:55:fd-01-0500"}
end of motion:
{"e":"changed","id":"19","r":"sensors","state":{"lastupdated":"2020-10-07T22:47:13.191","lowbattery":false,"tampered":false,"water":false},"t":"event","uniqueid":"28:6d:97:00:01:00:55:fd-01-0500"}

if sensor information is request "sensor/19"
this is the response: 
 {
  "config": {
    "battery": 94,
    "on": true,
    "pending": [],
    "reachable": true
  },
  "ep": 1,
  "etag": "68ed19dc930eb3eb38c3f0e72288053d",
  "lastseen": "2020-10-07T22:51:29.443",
  "manufacturername": "Samjin",
  "modelid": "motion",
  "name": "Motion Sensor",
  "state": {
    "lastupdated": "2020-10-07T22:47:13.960",
    "lowbattery": false,
    "tampered": false,
    "water": false
  },
  "type": "ZHAWater",
  "uniqueid": "28:6d:97:00:01:00:55:fd-01-0500"
}
## Steps to reproduce the behavior
none it start as soon you add the sensor 

## Expected behavior

state.water should be state.presence same value
## Environment

 - Host system: (Raspberry Pi 4)
 - Running method: (Raspbian imagen download from https://phoscon.de/)
 - Firmware version: (26520700)
 - deCONZ version: (2.5.77)
 - Device: ( RaspBee II)
 - Do you use an USB extension cable: (N/A) -- only relevant for ConBee I/II

SwoopX:
Can you double check in deconz GUI as what the device exposes itself? Should be sufficient to read the attributes of IAS Zone cluster.

Could be that we should hardcode the sensor type.

martinezmp3:
attach screenshot temperature is working the way it should but i attach it to just in case you need it 


[...]

PR #3408, opened 2020-10-13
Enhancements

SwoopX:
- Hardcode Samjin motion sensor as such (#3375)
- Hive motion sensor now correctly exposed as such
- Expose battery value for Bosch motion sensor ISW-ZPR1-WP13 (#3393)
- Add button maps for ZGRC-KEY-007 and ZGRC-KEY-009 (#3396, #3397)


manup:
There are still some modelIds missing, should we wait with merging until these are filled?

mamrai1:
One more question: Why 'tampered' attribute ia not shown in phoscon?

PR #3423, opened 2020-10-15
Add initial support for Owon AC201 Thermostat [02]

SwoopX:
- Add initial support for Owon AC201 Thermostat
- Introduce `fan_control.cpp` to accomodate AC fan control for thermostats
- Introduce `RConfigFanMode` to set AC fan mode
- Introduce `RConfigSwingMode` to set AC louvers position

### DIFF
--- a/fan_control.cpp
+++ b/fan_control.cpp
@@ -76,21 +76,21 @@ void DeRestPluginPrivate::handleFanControlClusterIndication(const deCONZ::ApsDat
                 if (sensor->modelId() == QLatin1String("AC201"))    // Owon
                 {
                     qint8 mode = attr.numericValue().u8;
-                    QString mode_set;
+                    QString modeSet;
 
-                    mode_set = QString("off");
-                    if ( mode == 0x00 ) { mode_set = QString("off"); }
-                    if ( mode == 0x01 ) { mode_set = QString("low"); }
-                    if ( mode == 0x02 ) { mode_set = QString("medium"); }
-                    if ( mode == 0x03 ) { mode_set = QString("high"); }
-                    if ( mode == 0x04 ) { mode_set = QString("on"); }
-                    if ( mode == 0x05 ) { mode_set = QString("auto"); }
-                    if ( mode == 0x06 ) { mode_set = QString("smart"); }
+                    modeSet = QLatin1String("off");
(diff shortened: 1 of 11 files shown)


fan_control.cpp
```diff
@@ -0,0 +1,194 @@
...
+    {
+        task.zclFrame.setFrameControl(task.zclFrame.frameControl() | deCONZ::ZclFCManufacturerSpecific);
+        task.zclFrame.setManufacturerCode(mfrCode);
+    }
+
+    // payload
+    QDataStream stream(&task.zclFrame.payload(), QIODevice::WriteOnly);
+    stream.setByteOrder(QDataStream::LittleEndian);
+
+    stream << (quint16) attrId;
```
manup, line 161: Just a note for the future: We don't have a coding style yet, but always use c++ named casts instead old C style casts.

`static_cast<quint16>(attrId)`

This might seam a bit superfluous at first. But soon the next generation code will get proper types for certain values like `ZclAttributeId_t`, `EndpointId_t`, etc to prevent common mistakes.

The named casts help the compiler to figure out invalid type conversions at compile time to prevent accidental errors.

For example:

```
ZclAttributeId_t attrId(0x0077);

stream << (quint8) attrId;              // programmer mistake won't be detected by compiler
stream << static_cast<quint8>(attrId);  // compile error no valid conversion from ZclAttributeId_t to quint8
stream << static_cast<quint16>(attrId); // ok
stream << attrId;                       // ok, because operator<< for QDataStream
```

So once we have the new types we only need to change the function signature:

    addTaskFanControlReadWriteAttribute(TaskItem &task, uint8_t readOrWriteCmd, uint16_t attrId, uint8_t attrType, uint32_t attrValue, uint16_t mfrCode)

to

    addTaskFanControlReadWriteAttribute(TaskItem &task, uint8_t readOrWriteCmd, ZclAttributeId_t attrId, ZclDatatypeId_t attrType, uint32_t attrValue, ZclManufacturerCode_t mfrCode)

If we already use C++ named casts we get a hole lot of error checking without needing to rewrite all those C style casts, searching for them is also quite hard/impossible.

manup (commented):
No errors just a few comments :)

PR #3433, opened 2020-10-17
Fixes

SwoopX:
- Fix setting temperature mode and child lock for Elko Super TR thermostat
- Prevent polling for further thermostats, thereby relying on reporting for the follwing thermostats: Elko Super TR, Owon AC201, Stelpro Orleans Fan SORB, Centralite pearl 3157100, Bitron 902010/32

### DIFF
--- a/de_web_plugin.cpp
+++ b/de_web_plugin.cpp
@@ -6315,6 +6315,11 @@ void DeRestPluginPrivate::addSensorNode(const deCONZ::Node *node, const SensorFi
             item = sensorNode.addItem(DataTypeUInt8, RConfigPending);
         }
     }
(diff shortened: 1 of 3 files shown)


manup:
The PR seems to activate reporting for a few additional devices, can these be added to the PR description?

ma-ca:
@SwoopX  may I ask to revert the change that prevents polling for Bitron 902010/32 because reporting only works for the local temperature attribute 0x0000 but not for the other attributes. Thanks.

SwoopX:
@ma-ca sure thing, sorry for that inconvenience. Will try to recall that later this evening. In case I shoul forget, just ping me once more.

PR #3435, opened 2020-10-17
Whitelist Bitron 902010/24A to activate siren

SwoopX:


kptkip:
is it now possible to activate the siren also in the Bitron Video 902010/24A smoke detector?

PR #3466, opened 2020-10-21
Support for OWON AC201 including Swing mode control and Attributes-Set

aryelevin:
This PR is for using the OWON AC201.
Most of the work done by @SwoopX and I used to polish it and added the swing functionality.


SwoopX:
We can skip this one, as all changes have been included in my PR ;)

PR #3497, opened 2020-10-24
Handle case when view scene returns no entry status

manup:
In some cases a Add Scene command was never sent, causing a scene not to be created after 3 times trying to read the scene.

Story behind this: My 4x IKEA GU10 color temperature lights corrupted 4 scenes after an OTA update. The scenes were still there but the colorY field was no longer part of the scene, only on, bri and colorX.

To fix the scenes:
1) Execute Remove All Scenes for the group
2) Executed Get Scene Membership for the group

After that the plugin started to recreate the scenes via Add Scene command, but on two lights only 3 out of 4 scenes were created. This commit fixed the remaining ones and all scenes work again.

### DIFF
--- a/de_web_plugin.cpp
+++ b/de_web_plugin.cpp
@@ -13645,6 +13645,28 @@ void DeRestPluginPrivate::handleSceneClusterIndication(TaskItem &task, const deC
                         onOff, bri, x, y, transitionTime);
             }
         }
+        else if (status == 0x8b && zclFrame.payload().size() >= 4) // scene not found
+        {
+            uint16_t groupId;
+            uint8_t sceneId;
+
+            stream >> groupId;
+            stream >> sceneId;
+
+            Group *group = getGroupForId(groupId);
+            Scene *scene = group->getScene(sceneId);
+
+            if (!group || !scene)
+            {
+                return;
+            }
+
+            auto *ls = scene->getLightState(lightNode->id());
+            if (ls)
+            {
+                ls->setNeedRead(false); // move to add scene
+            }
+        }
     }
     else if (zclFrame.commandId() == 0x05 && !(zclFrame.frameControl() & deCONZ::ZclFCDirectionServerToClient)) // Recall scene command
     {


ebaauw:
> The scenes were still there but the colorY field was no longer part of the scene, only on, bri and colorX. 

Isn’t that IKEA’s way of indicating colour temperature?

manup:
Perhaps it was used like that in older firmware, I'm not sure (I made this mistake with our older FLS firmware myself). It would also explain why my scenes were broken after the OTA update.

But the recent IKEA firmware seems to store and report the color temperature in scenes in colorX/colorY values as described in the specification.

image

ebaauw:
What does it return after a _Store Scene_, when the light state was set using _Move to Color Temperature_?

manup:
Ha, that scrambles the damn thing again, colorY is missing.

image

manup:
Both seem to work somehow, but for the scenes where colorY is specified the color mode attribute stays at xy which are also updated but the color temperature isn't.

For the scene which only contains colorX color temperature is updated but not xy and color mode is set to ct on recall.

manup:
Some more testing with the GU10 color temperature lights. It appears it's not possible to use the Add Scene command to set the ct mode. I've tried to only include the colorX field and also append the color temperature behind the other fields, but it makes no difference and the resulting scene always has colorX and colorY.

Another very strange thing: The Store Scene command creates a scene after color temperature was set and this scene can be restore the color temperature fine, but View Scene returns some complete non sense values not related to the color temperature. For example I have two scenes with different ct and both can be recalled fine but the colorX value returned in View Scene is equal.

Currently I think to disable Add Scene usage for IKEA lights and also not to read back ct from View Scene response. The store scene command seems to work but it sucks that a scene can't be verified easily (other than to check the ct value after recall).

MattWestb:
@manup IKEA is adding the 3 "default scenes" then doing firmware on the bulbs.
I think its because not losing the default funktion with the 5 button remotes next / revius scene.
The problem with the default scene and remote is that the remote is locked sending the scene commands to groupe 0xff09 if not configured with touchlink and getting one group from the inisiator (The On/Off and level cluster can being bounded to group and sending their commands to the "new" group). Then the remote dont have any scenes cluster its not possible doing one new bind to it so its always sending scene commands to group 0xff09.

Please make auto add the group 0xff09 then pairing one IKEA remote then the next / previous is working for the "default" and from deCONZ added group for the user.
IKEA_CT19
The from deCONZ is absolute ones = not only colour is set but also light level then the "IKEA magic" is not applied.

After that the remote can control the scene (for all bulbs with the 0xff09) but if having more remotes they is doing  the same (On/Off and Light level is send to bonded group only). 

IKEA_CT23
 The 0xff09 working in HA and from the remote !!!!


And a large thanks for doing great work with our systems !!

MattWestb:
One notis about adding scenes: IKEAs GW is deleting all groups, Adding one group, Reading group for verifying its added, deleting all scenes, sending one IKEA cluster command, Verifying that the scene have the groupe, Adding scene 0-2, Reading the scene 0-2 and most times adding the scenes one more time (perhaps failing to adding them correct).

Have lolking thru around sniffs then adding remotes and bulbs to the GW for finding the "IKEA magic".

MattWestb:
Thanks @Mimiix  :-))

MattWestb:
Some good sniffs from IKEA GW then pairing remote and bulbs to it if like how IKEA is doing the scene things. https://github.com/Koenkk/zigbee2mqtt/issues/4414
The problem is that wireshark is not decoding / parsing the touchlink commands OK but the group is possible reading out from the response from the remote :-)

PS: network key is longer down in the postings.

manup:
Interesting stuff. I did a small test with my five button IKEA remote, it doesn't use 0xff09 for group commands but a different random generated one. Which makes sense, otherwise it would not be possible to have remotes controlling different groups or rooms. But other than that it would be sure interesting to be able to create the IKEA default scenes for IKEA lights which belong to that group. I'll have a look in the sniffer logs :)

MattWestb:
OK its explaining something then i have only used one of my remotes for testing. Is looks that one remote allways is using the same group but perhaps its generated from the MAC address in the device.
The strange thing is that the bulbs is "restoring" the groupe 0xff09 then getting firmware updated.
(Do one forced OTA downgrading and then updating it its working form my in deCONZ and getting the default 0xff09 groups scenes in the bulbs)
Is it possible reading the groupe from the remote or is it only possible sniffing it then the remote is sending scene commands ?
If only then sending then deCONZ can still "sniffing" if and using it.
If getting the same grupe its being more or less one "global scene" that the remotes is controlling.
On good / bad thing is that IKEA have not using the default scene groupe that zigbee 3 standard is forcing (groupe 0x0000).

The group is being settled in the second pair of Get Group Identifiers Request (0x41).
See the https://github.com/zigpy/zha-device-handlers/issues/561#issuecomment-723258295.

By the way it can being better sniffing the groupe from the device then doing one new groupe and binding the on/off and level cluster to it. (scenes is not possible then it dont  have any scene cluster). Then all commands is working on the sniffed group without problems (if not having more remotes with the same groupe).

If you have home office and is being bored implanting touchlink initiation and setting the groupe and then continuing with normal pairing (TL is the recommended commissioning method in ZB3) so its being controlled by the coordinator / TC (many saying its not possible TL in ZB3 but its not true, the problem is getting it working with the TC).

Its one more device that is being in "limbo" then not being touchlinked. The signal repeater is getting the groupe thru TL and is bonding the "relay" funktion to the blinds from the open/remote but its not working then doing classical joining with TC. I think its more or less useless then not having one groupe configured :-((

I have more sniffs done but i dont like spreading my network key around then its in my production system.

manup:
Ok, looking at the sniffer logs and some testing with GU10 it's now working in my setup :)
The group **0xff09** is indeed the magical group.

* Remotes use a random group when sending out commands, lets call it remote group (RG);
* When deCONZ receives a command from an unknown group RG, a new group will be created in the REST-API (not visible in Phoscon App but maybe in the old interface);
* The light needs to be added to the remote group RG in order to receive commands from it;
* However the scenes need to be configured for group **0xff09** not the remote group RG;
* For that I've just used the Store Scene command in Cluster Info Panel to store the current state in scenes 0–5 (it isn't limited to 3 scenes).

Works even when deCONZ/ConBee is not running, quite nice.

MattWestb:
So its pretty much doing as one zigbee 3 light suld doing with one expecting that is the default scene grupe / reserved in the zigbee 3 / ZCL is 0x000. The good thing is if they was using it then its being problems with much systems that is using it for the coordinator (as deCONZ have doing before).
I have the old web app sniffing the groupe (0xff09) one time but not always so i have patching the zigbee.db from deleted to normal so getting it working ( I think i was adding it with add groupe command on the group cluster and deCONZ was sniffing it).
If first adding the groupe 0xff09 to the light and then  binding the on/off and level cluster to it i think all is working very OK IF having only ONE remote in the system.
My GU10 WS have plenty space for scenes :-))))))))

```
ZigBee Cluster Library Frame
    Frame Control Field: Cluster-specific (0x19)
    Sequence Number: 34
    Command: Get Scene Membership Response (0x06)
    Payload
        Scenes Status: Success (0x00)
        Scene Capacity: 61
        Group ID: 0x002a
        Scene Count: 0
```
The second part is how IKEA is getting the scene group to the controlling device.
Thru sniffs all is normal pairing with TL except the "Command: Get Group Identifiers Request (0x41)" that wireshark dont parsing correctly but its possible seeing the added group in the sniffs data.
And one time then adding scenes in the lights one IKEA spesial commands but i think its for making the scenes only reacting on the colour and transaction time and not the on/off and light level stored in the scene (the scenes is added as enhanced scene with all attributes for lights and blinds).
```
ZigBee Cluster Library Frame
    Frame Control Field: Cluster-specific (0x11)
    Sequence Number: 39
    Command: Add Scene (0x00)
    Payload, String: 
        Group ID: 0x002a
        Scene ID: 0x02
        Transition Time: 1 seconds
        Length: 0
        String: 
        Extension field set 1
            Cluster: On/Off (0x0006)
            On/Off: 1
        Extension field set 2
            Cluster: Color Control (0x0300)
            Color X: 0,3818
            Color Y: 0,3797
        Extension field set 3
            Cluster: Level Control (0x0008)
            Level: 254
        Extension field set 4
            Cluster: Unknown (0x0102)
            Current Position Lift Percentage: 0
            Current Position Tilt Percentage: 0
```
The adding scenes thru deCONZ GUI all parameter is setted and also being used the recaled from the remote = absolute scenes.

If you can sending the add group command to the remote you can schusing the group for the scenes but I think its only being possible thru touch link that Rasp/CornBee II not having (My PaspBee one have it but i dont have the knowledge to do that).

One more IKEA specific thing with the remote: If long pressing the toggle button its (soft) resetting all the lights in the group to default so the light being synced in the groupe (as long all lights in the groupe have the same number of scenes stored ;-))  ).

I have doing little other criminal things this afternoon that can helping little later.
Have written one long but gently e-mail to head of technical office at IKEA tradfri division in swedish and kindly asking for implanting fix (implanting scene cluster) or workaround (internal making binding of the scene group to the same as on/off cluster) for the problem with the remote scene then joining thru classical pairing (no touchlink pairing) or implanting adding group to the "internal scene cluster"by "IKEA cluster commands" that can being used in ZB3 systems.
And the same for the signal repeater that is only floating around in one ZB3 system without one is group configured and not working properly as one relay for command being sended to the blinds.

I hope IKEA is listening and doing some "community futur" upgrade of the current and upcoming (E2001/E2002) remotes.

MattWestb:
Pleas Manuel can you adding one function that is adding the groupe 0xff09 to IKEA lights then its being paired then deCONZ is automatically finding the scenes in all IKEA bulbs that have them after firmware update and adding them in the system (also in HA with cracy names).

PR #3498, opened 2020-10-24
Periodically poll SwBuildId for IKEA, OSRAM and Philips

manup:
Due OTA updates the version attribute may change.

### DIFF
--- a/poll_manager.cpp
+++ b/poll_manager.cpp
@@ -463,7 +463,11 @@ void PollManager::pollTimerFired()
             else
             {
                 if (item->toString().isEmpty() ||
+                    lightNode->manufacturerCode() == VENDOR_IKEA ||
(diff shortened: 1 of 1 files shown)


ebaauw:
Wouldn’t it be enough to be reading these on device announcement?  And the descriptors and other _Basic_ cluster attributes.  From what I’ve seen, devices tend to reboot after firmware update.

manup:
Good point, I guess after an update it makes sense to query all static attributes and descriptors again to be sure. For the new Device state machine that's rather easy, after certain events like update finished or device announce. For now I'd like to avoid bringing more complexity in the "old" code. The PR queries on an per hour basis, not optimal but hopefully shouldn't hurt, it also updates SwBuildId for devices which were OTA updated in the past but never queried automatically afterwards (like OSRAM and Philips).

PR #3525, opened 2020-10-28
Fix too broad mac check that prevents from pairing

dn0sar:
```checkMacVendor(sc->address, VENDOR_X)``` is not exclusive, and while it evaluates as true for VENDOR_X, it might evaluate as true for VENDOR_Y as well.

This creates a problem in ```void DeRestPluginPrivate::delayedFastEnddeviceProbe(const deCONZ::NodeEvent *event)``` when deciding whether to skip getting additional attributes (such as modelId and manufacturer).

Since getting additional attributes is skipped, adding a node fails later on, and it is not exposed to the API. Note that this only happens when adding a new node, but not if a node is already paired.

I suspect this might also be relevant for issue #3482 and maybe other issues related to pairing new devices.

Finally, I'm not sure whether ```checkMacVendor``` in general is expected to evaluate as true for multiple vendors for the same mac address. If it is not, then there are probably other bugs in the source code due to this.

### DIFF
--- a/de_web_plugin.cpp
+++ b/de_web_plugin.cpp
@@ -2093,11 +2093,11 @@ void DeRestPluginPrivate::addLightNode(const deCONZ::Node *node)
                     {
                         if (hasServerOnOff)
                         {
-                            if (checkMacVendor(node->address(), VENDOR_JENNIC) &&
+                            if (checkMacAndVendor(node, VENDOR_JENNIC) &&
                                 // prevent false positives like Immax IM-Z3.0-DIM which has only two endpoints (0x01)
                                 // lumi.ctrl_neutral1 and lumi.ctrl_neutral2 have more 5 endpoints
                                 node->simpleDescriptors().size() > 5  &&
-                                node->nodeDescriptor().manufacturerCode() == VENDOR_JENNIC && i->endpoint() != 0x02 && i->endpoint() != 0x03)
+                                i->endpoint() != 0x02 && i->endpoint() != 0x03)
                             {
                                 // TODO better filter for lumi. devices (i->deviceId(), modelid?)
                                 // blacklist switch endpoints for lumi.ctrl_neutral1 and lumi.ctrl_neutral2
@@ -2244,7 +2244,7 @@ void DeRestPluginPrivate::addLightNode(const deCONZ::Node *node)
         QString uid = generateUniqueId(lightNode.address().ext(), lightNode.haEndpoint().endpoint(), 0);
         lightNode.setUniqueId(uid);
 
-        if (checkMacVendor(node->address(), VENDOR_DDEL) && i->deviceId() != DEV_ID_CONFIGURATION_TOOL)
+        if (existDevicesWithVendorCodeForMacPrefix(node->address(), VENDOR_DDEL) && i->deviceId() != DEV_ID_CONFIGURATION_TOOL)
         {
             ResourceItem *item = lightNode.addItem(DataTypeUInt32, RConfigPowerup);
             DBG_Assert(item != 0);
@@ -2299,8 +2299,7 @@ void DeRestPluginPrivate::addLightNode(const deCONZ::Node *node)
             lightNode.setNeedSaveDatabase(true);
         }
 
-        if (checkMacVendor(node->address(), VENDOR_OSRAM) &&
-            (node->nodeDescriptor().manufacturerCode() == VENDOR_OSRAM || node->nodeDescriptor().manufacturerCode() == VENDOR_OSRAM_STACK))
+        if (checkMacAndVendor(node, VENDOR_OSRAM) || checkMacAndVendor(node, VENDOR_OSRAM_STACK))
         {
             if (lightNode.manufacturer() != QLatin1String("OSRAM"))
             {
@@ -2309,7 +2308,7 @@ void DeRestPluginPrivate::addLightNode(const deCONZ::Node *node)
             }
         }
 
-        if (checkMacVendor(node->address(), VENDOR_PHILIPS))
+        if (existDevicesWithVendorCodeForMacPrefix(node->address(), VENDOR_PHILIPS))
         {
             if (lightNode.manufacturer() != QLatin1String("Philips"))
             { // correct vendor name, was atmel, de sometimes
@@ -2339,7 +2338,7 @@ void DeRestPluginPrivate::addLightNode(const deCONZ::Node *node)
         }
 
         //Add missing values for Profalux device
-        if (checkMacVendor(node->address(), VENDOR_PROFALUX))
+        if (existDevicesWithVendorCodeForMacPrefix(node->address(), VENDOR_PROFALUX))
         {
             //Shutter ?
             if (i->deviceId() == DEV_ID_ZLL_COLOR_LIGHT)
@@ -6887,12 +6886,12 @@ void DeRestPluginPrivate::updateSensorNode(const deCONZ::NodeEvent &event)
             case VENDOR_CLUSTER_ID:
             {
                 // ubisys device management (UBISYS_DEVICE_SETUP_CLUSTER_ID)
-                if (event.endpoint() == 0xE8 && checkMacVendor(event.node()->address(), VENDOR_UBISYS))
+                if (event.endpoint() == 0xE8 && existDevicesWithVendorCodeForMacPrefix(event.node()->address(), VENDOR_UBISYS))
                 {
                     break;
                 }
                 // dresden elektronik spectral sensor
-                else if (i->modelId() == QLatin1String("de_spect") && checkMacVendor(event.node()->address(), VENDOR_DDEL))
+                else if (i->modelId() == QLatin1String("de_spect") && existDevicesWithVendorCodeForMacPrefix(event.node()->address(), VENDOR_DDEL))
                 {
                     break;
                 }
@@ -6914,7 +6913,7 @@ void DeRestPluginPrivate::updateSensorNode(const deCONZ::NodeEvent &event)
             // filter endpoint
             if (event.endpoint() != i->fingerPrint().endpoint)
             {
-                if (checkMacVendor(event.node()->address(), VENDOR_JENNIC))
+                if (existDevicesWithVendorCodeForMacPrefix(event.node()->address(), VENDOR_JENNIC))
                 {
                     if (i->modelId().startsWith(QLatin1String("lumi.sensor_86sw")) ||
                         i->modelId().startsWith(QLatin1String("lumi.ctrl_neutral")) ||
@@ -8576,7 +8575,7 @@ void DeRestPluginPrivate::updateSensorNode(const deCONZ::NodeEvent &event)
                         }
                     }
                     else if (event.clusterId() == UBISYS_DEVICE_SETUP_CLUSTER_ID && event.endpoint() == 0xE8 &&
-                            checkMacVendor(event.node()->address(), VENDOR_UBISYS)) // ubisys device management
+                            existDevicesWithVendorCodeForMacPrefix(event.node()->address(), VENDOR_UBISYS)) // ubisys device management
                     {
 //                        bool updated = false;
                         for (;ia != enda; ++ia)
@@ -8794,7 +8793,7 @@ bool DeRestPluginPrivate::isDeviceSupported(const deCONZ::Node *node, const QStr
     while (s->modelId)
     {
         if ((!node->nodeDescriptor().isNull() && node->nodeDescriptor().manufacturerCode() == s->vendorId) ||
-            ((node->address().ext() & macPrefixMask) == s->mac) || checkMacVendor(node->address(), s->vendorId))
+            ((node->address().ext() & macPrefixMask) == s->mac) || existDevicesWithVendorCodeForMacPrefix(node->address(), s->vendorId))
         {
             if (modelId.startsWith(QLatin1String(s->modelId)))
             {
@@ -10895,7 +10894,7 @@ void DeRestPluginPrivate::handleXalClusterIndication(const deCONZ::ApsDataIndica
         return;
     }
 
-    if (!checkMacVendor(lightNode->address(), VENDOR_XAL))
+    if (!existDevicesWithVendorCodeForMacPrefix(lightNode->address(), VENDOR_XAL))
     {
         return;
     }
@@ -11076,14 +11075,14 @@ void DeRestPluginPrivate::handleZclAttributeReportIndication(const deCONZ::ApsDa
         checkReporting = true;
         sendZclDefaultResponse(ind, zclFrame, deCONZ::ZclSuccessStatus);
     }
-    else if (checkMacVendor(ind.srcAddress(), VENDOR_PHILIPS) ||
+    else if (existDevicesWithVendorCodeForMacPrefix(ind.srcAddress(), VENDOR_PHILIPS) ||
             macPrefix == tiMacPrefix ||
-            checkMacVendor(ind.srcAddress(), VENDOR_DDEL) ||
-            checkMacVendor(ind.srcAddress(), VENDOR_IKEA) ||
-            checkMacVendor(ind.srcAddress(), VENDOR_OSRAM_STACK) ||
-            checkMacVendor(ind.srcAddress(), VENDOR_JENNIC) ||
-            checkMacVendor(ind.srcAddress(), VENDOR_SI_LABS) ||
-            checkMacVendor(ind.srcAddress(), VENDOR_CENTRALITE))
+            existDevicesWithVendorCodeForMacPrefix(ind.srcAddress(), VENDOR_DDEL) ||
+            existDevicesWithVendorCodeForMacPrefix(ind.srcAddress(), VENDOR_IKEA) ||
+            existDevicesWithVendorCodeForMacPrefix(ind.srcAddress(), VENDOR_OSRAM_STACK) ||
+            existDevicesWithVendorCodeForMacPrefix(ind.srcAddress(), VENDOR_JENNIC) ||
+            existDevicesWithVendorCodeForMacPrefix(ind.srcAddress(), VENDOR_SI_LABS) ||
+            existDevicesWithVendorCodeForMacPrefix(ind.srcAddress(), VENDOR_CENTRALITE))
     {
         // these sensors tend to mac data poll after report
         checkReporting = true;
@@ -11122,7 +11121,7 @@ void DeRestPluginPrivate::handleZclAttributeReportIndication(const deCONZ::ApsDa
         }
     }
 
-    if (zclFrame.isProfileWideCommand() && checkMacVendor(ind.srcAddress().ext(), VENDOR_XIAOMI) && (ind.clusterId() == BASIC_CLUSTER_ID || ind.clusterId() == 0xfcc0))
+    if (zclFrame.isProfileWideCommand() && existDevicesWithVendorCodeForMacPrefix(ind.srcAddress().ext(), VENDOR_XIAOMI) && (ind.clusterId() == BASIC_CLUSTER_ID || ind.clusterId() == 0xfcc0))
     {
         handleZclAttributeReportIndicationXiaomiSpecial(ind, zclFrame);
     }
@@ -15642,21 +15641,20 @@ void DeRestPluginPrivate::delayedFastEnddeviceProbe(const deCONZ::NodeEvent *eve
 
             bool skip = false;
 
-            if (thermostatClusterEndpoint == 0) // e.g. Eurotronic SPZB0001 thermostat
+            if (thermostatClusterEndpoint > 0) // e.g. Eurotronic SPZB0001 thermostat
             {  }
             else if (iasZoneType > 0) // IAS motion and contact sensors
-            {  }
-            else if (node->nodeDescriptor().manufacturerCode() == VENDOR_IKEA)
-            {  }
-            else if (node->nodeDescriptor().manufacturerCode() == VENDOR_DANFOSS)
             {  }
             else if (modelId.startsWith(QLatin1String("lumi.")))
             {
                 skip = true; // Xiaomi Mija devices won't respond to ZCL read
+                DBG_Printf(DBG_INFO, "[4] Skipping additional attribute read - Model starts with 'lumi.'\n");
             }
-            else if (checkMacVendor(sc->address, VENDOR_JENNIC))
+            else if (checkMacAndVendor(node, VENDOR_JENNIC) ||
+                     checkMacAndVendor(node, VENDOR_ADUROLIGHT))
             {
                 skip = true; // e.g. Trust remote (ZYCT-202)
+                DBG_Printf(DBG_INFO, "[4] Skipping additional attribute read -  Assumed Trust remote (ZYCT-202)\n");
             }
 
             if (skip)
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

**Comment by dn0sar** [comment]
Ps. Also, I'm not sure whether ```checkMacVendor``` in general is expected to evaluate as true for multiple vendors for the same mac address. If it is not, then there are probably other bugs in the source code due to this.

**Comment by SwoopX** [comment]
I guess you spotted a valid point, but for different reasons. `checkMacVendor()` just checks if for a certain vendor, one of the whitelisted MAC address prefixes is given. The logical OR in that function ensures, that it will always evaluate to `true` if such a prefix is given and available. However, as time changes and also vendors silently change their hardware, it cannot be ensured that we know all vendor/prefix combinations for every device. And this seems to be the critical path for me.

Now, if you got a device with Jennic manufacturer code, but it has a different mac address prefix other than mentioned in the code, then `checkMacVendor()` will return false, resulting in the behaviour you described above (attributes are not read). 

Additionally, that might have further impact throughout the code as the function is used a couple of times.

**Comment by dn0sar** [comment]
I see, thanks for the explanation.

This also causes a second problem besides the one you described. A device with manufacturer code != Jennic, but with the same mac prefix will evaluate as true and it will not read attributes when it should actually have. This affects several manufactures (more than 10) as I can see from the current ``checkMacVendor`` implementation.

**Comment by SwoopX** [comment]
> A device with manufacturer code != Jennic, but with the same mac prefix, will evaluate as true and it will not read attributes when it should actually have.

Why should that be? The manufacturer code is checked by a switch statement prior to the evaluation and since VENDOR_JENNIC is explicitly used, I currently don't see the issue for other devices.

**Comment by dn0sar** [comment]
I mean it's just the way a logical _or_ works, I don't know how to say this more clearly 😄 

```
case VENDOR_JENNIC:
            return prefix == jennicMacPrefix ||
                   prefix == silabs2MacPrefix;
```

If this evaluates as true then either ```prefix == jennicMacPrefix``` or ```prefix == silabs2MacPrefix``` is true. Therefore

```
        case VENDOR_IKEA:
            return prefix == ikeaMacPrefix ||
                   prefix == silabsMacPrefix ||
                   prefix == silabs2MacPrefix ||
                   prefix == silabs4MacPrefix ||
                   prefix == energyMiMacPrefix ||
                   prefix == emberMacPrefix;
```
Will also return true. But if ```checkMacVendor(sc->address, VENDOR_JENNIC)``` is called on an IKEA product you would not expect it to return true I guess.

**Comment by dn0sar** [comment]
Assume you have an IKEA product with ```prefix == silabs2MacPrefix``` evaluating to true. Then when you call ```checkMacVendor(sc->address, VENDOR_JENNIC)``` this will also return true.

**Comment by dn0sar** [comment]
Sorry, I realize I wasn't clear enough in the first comment of the PR. Hopefully now it is a bit more clear, but in short what I meant is **not** that ```checkMacVendor(address_from_vendor_x, VENDOR_X) == checkMacVendor(address_from_vendor_x, VENDOR_Y) == true``` but rather that it can happen that ```checkMacVendor(address_vendor_y, VENDOR_X) == true```

**Comment by SwoopX** [comment]
> (address_ikea & macPrefixMask) == silabs2MacPrefix evaluates to true.

There is no such code. The story goes: Does VENDOR_JENNIC have mac X or Y assigned to it? If yes, skip attributes, if no, continue.

In that sense, if you have a device with that vendor/mac combination, you don't get the attributes read although you might want that. With the proposed code, the code skips reading the attributes for **all** Jennic devices then and not just a few chosen ones.

**Comment by dn0sar** [comment]
Ok, I didn't understand from the code that the skip was meant only for a subset of Jennic devices. I've made it more restrictive in the new commit. The example is use

> > (address_ikea & macPrefixMask) == silabs2MacPrefix evaluates to true.
> 
> There is no such code. The story goes: Does VENDOR_JENNIC have mac X or Y assigned to it? If yes, skip attributes, if no, continue.

This was meant to be an example to show the problem, not a verbatim code snippet. Just so it's clear, are you still not convinced that e.g., an IKEA device whose address is such that ```(address_ikea & macPrefixMask) == silabs2MacPrefix``` will evaluate to true both in ```checkMacVendor(address_ikea, VENDOR_IKEA)``` and ```checkMacVendor(address_ikea, VENDOR_JENNIC)```?

**Comment by dn0sar** [comment]
Ps. I had a quick look at the recently open issues and this probably affects also #3529 #3528 #3516 #3474 as XIAOMI and IKEA are two vendors whose device's mac addresses can evaluate to true in ```checkMacVendor(sc->address, VENDOR_JENNIC)```

**Comment by Jaguaza** [comment]
> Ps. I had a quick look at the recently open issues and this probably affects also #3529 #3528 #3516 #3474 as XIAOMI and IKEA are two vendors whose devices' mac addresses can evaluate to true in `checkMacVendor(sc->address, VENDOR_JENNIC)`

- Thanks for your interest. For me it is frustrating that this happens with my xiaomi devices. I hope they find the cause and can correct it soon.

**Comment by SwoopX** [comment]
> Ps. I had a quick look at the recently open issues and this probably affects also #3529 #3528 #3516 #3474 as XIAOMI and IKEA are two vendors whose devices' mac addresses can evaluate to true in checkMacVendor(sc->address, VENDOR_JENNIC)

If there are IKEA devices using VENDOR_JENNIC as mfc and a jennicMacPrefix or silabs2MacPrefix, then this is indeed possible resulting in the attributes not being read.

> Just so it's clear, are you still not convinced that e.g., an IKEA device whose address is such that (address_ikea & macPrefixMask) == silabs2MacPrefix will evaluate to true both in checkMacVendor(address_ikea, VENDOR_IKEA) and checkMacVendor(address_ikea, VENDOR_JENNIC)?

No, I'm afraid not. At least unless you can reasonably explain to me how, in the corresponding code below, the case of the switch statement can automagically can jump to a different manufacturer code other than the one provided as the function parameter.

grafik

**Comment by dn0sar** [comment]
I don't know how what I wrote is still not clear. I am **not** saying that the switch jumps to a different vendor. 

 I'll try to dumb it down a bit.

Assume you have an IKEA device whose address is ```cc:cc:cc:0d:ea:db:4a:14```. This is a valid IKEA device whose mfc code is VENDOR_IKEA, and whose prefix is ```silabs2MacPrefix```. When you call ```checkMacVendor(cc:cc:cc:0d:ea:db:4a:14, VENDOR_JENNIC)``` it returns true, skipping the attribute read for an IKEA device even if it should have not been skipped.

Please tell me it's clear now. If not, do try to pair an IKEA device with prefix ```silabs2MacPrefix``` and see how that goes.

**Comment by Jaguaza** [comment]
Dear deconz staff
I dont know what the problem is, and if you will study my request yo solve It.
I only know i am a customer that believe in you, and i have a dispositive with a hardware i can not use.
I have written to support by mail several times in last 15 days, and i have not answer. I have opened a case and i feel you has not tried yo give me a solution, or at least, try yo see what IS the problem.
I dont know abot programming, but you are able, and you are not giving me any response.

**Comment by manup** [comment]
It's a very tricky part of the code, imho the vendor code/mac check has become fuzzy over time there are just too many combinations (the new Device mechanism will replace this entirely later on).

As workaround and to not mess further with that I would suggest adding the check for VENDOR_IKEA to fix readings for IKEA devices with Jennic mac prefix, it's dirty as well but very specific and only applies to IKEA devices.

For Xiaomi devices this part of the code should not prevent the sensor to be added, since they report their modelId on their own. I've tested a few Xiaomi devices already which were added fine, but we need to get a broader picture here to know exactly what happens.

```
            if (thermostatClusterEndpoint == 0) // e.g. Eurotronic SPZB0001 thermostat
            {  }
            else if (iasZoneType > 0) // IAS motion and contact sensors
            {  }
            else if (node->nodeDescriptor().manufacturerCode() == VENDOR_IKEA)
            {  }
            else if (modelId.startsWith(QLatin1String("lumi.")))
            {
                skip = true; // Xiaomi Mija devices won't respond to ZCL read
            }
            else if (checkMacVendor(sc->address, VENDOR_JENNIC))
            {
                skip = true; // e.g. Trust remote (ZYCT-202)
            }
```

**Comment by dn0sar** [comment]
Hi Manuel, thanks for your reply and for acknowledging the problem. However, I still think your proposed fix would not completely fix the problem. As IKEA devices are not the only ones affected, but any non Jennic device that has a ```jennicMacPrefix``` or a ```silabs2MacPrefix``` will skip the attribute read.

I'll try to explain why again in a different way. The following is the problematic line:

```
else if (checkMacVendor(sc->address, VENDOR_JENNIC)) { skip = true }
```

If you check what ```checkMacVendor``` does, it is clear that only the switch for ```VENDOR_JENNIC``` matters. If we inline and expand the function the if condition above is functionally equivalent to the following (I hope it is easy for you to see why, but do tell me if you are not convinced this is true): 

```
else if ((sc->address & macPrefixMask) == jennicMacPrefix ||
    (sc->address & macPrefixMask) == silabs2MacPrefix ) { skip = true }
```

I hope with the rewritten code above it is easy to see how **all** devices with those prefixes were being skipped the attribute read (well except if ```thermostatClusterEndpoint == 0``` or ``` modelId.startsWith(QLatin1String("lumi.")) ``` which were whitelisted explicitly).

What do you think?

**Comment by Jaguaza** [comment]
> Para los dispositivos Xiaomi, esta parte del código no debería evitar que se agregue el sensor, ya que informan su modelId por su cuenta. Ya probé algunos dispositivos Xiaomi que se agregaron bien, pero necesitamos obtener una imagen más amplia aquí para saber exactamente qué sucede.

And what would be the solution for my case #3528 ???

**Comment by Mimiix** [comment]
@Jaguaza I really do not appreciate hijacking other issues.

**Comment by Jaguaza** [comment]
> @Jaguaza Realmente no aprecio secuestrar otros temas.

Give a solution to my problem, I believe in a serious store and I do not get an answer to my problem. Yeah, put routers. Sorry, I already have them, read my post.

**Comment by Mimiix** [comment]
@Jaguaza Then this still is not the place to reply. This is a pull request, not a issue. I am asking you once again to stick to your issue: https://github.com/dresden-elektronik/deconz-rest-plugin/issues/3528

Please note: We are mostly all volunteers and i do not allow any disrespect to any of us.  Xiaomi's are notorious for odd behaviour.

**Comment by dn0sar** [comment]
Adding a message here in an attempt to bring this back on topic. 

My disagreement with the IKEA hotfix is due to the fact that I have two devices that I could not pair due to this bug. An Ikea remote and a Danfoss thermostat. Maybe I was unlucky that both of these had a ```silabs2MacPrefix``` and they don't send later on the missing attributes (they are both end-nodes if this makes any difference).

So given this, I have some questions:

Q1) Is the attribute read skip meant to be for all devices with ```silabs2MacPrefix``` and ```jennicMacPrefix``` (as opposed to only those with a Jennic manufacturer code)?
Q2) If not, are you ok with the changes proposed in my PR?
Q3) If Q2 is relevant and you are not ok with the proposed fix, what is it that you think should be done instead?

**Comment by manup** [comment]
The check to skip VENDOR_JENNIC was at the time added to skip only Trust remote (ZYCT-202), as we now see that didn't age well. No matter how we change the checks in the current code, it remains fuzzy which isn't good.

So currently we know that for some IKEA mac/vendor combinations (I guess newer devices) and the Danfoss the current check wrongly skip the reading. I'd propose to only pinpoint these two explicitly so that they are not skipped but leaving the old check at the end of the if else chain.

It's not nice to look at, but should fix the two known problems without having unforeseen side effects for other devices which we might not think of here.

```
            if (thermostatClusterEndpoint == 0) // e.g. Eurotronic SPZB0001 thermostat
            {  }
            else if (iasZoneType > 0) // IAS motion and contact sensors
            {  }
            else if (node->nodeDescriptor().manufacturerCode() == VENDOR_IKEA)
            {  }
            else if (node->nodeDescriptor().manufacturerCode() == VENDOR_DANFOSS)
            {  }
            else if (modelId.startsWith(QLatin1String("lumi.")))
            {
                skip = true; // Xiaomi Mija devices won't respond to ZCL read
            }
            else if (checkMacVendor(sc->address, VENDOR_JENNIC))
            {
                skip = true; // e.g. Trust remote (ZYCT-202)
            }
```

**Comment by dn0sar** [comment]
If it's meant just for the ZYCT-202, I still don't understand how the following is not a cleaner solution:

```if (if (node->nodeDescriptor().manufacturerCode() == VENDOR_JENNIC && checkMacVendor(sc->address, VENDOR_JENNIC))```

Do ZYCT-202 have different manufacturer codes? Or do they not even give a manufacturer code?

But in any case, your solution works for my specific devices so that's fine for me. Should I just add a commit here with the version that you proposed in your comment?

**Comment by manup** [comment]
According to the GH issues Trust remote (ZYCT-202) has been seen with 0x1037 (0x00158d0001) and 0x122d (0x001580004) perhaps the latter is a fixed version of the earlier models, I couldn't find anything in the code about it.

https://github.com/dresden-elektronik/deconz-rest-plugin/issues/3117 and https://github.com/dresden-elektronik/deconz-rest-plugin/issues/261

My worries with the Jennic related checks, or to be frank with any silicon vendor mac/silicon vendor check, is that too many products use them. Any conditional branch we add or change which rely on only silicon vendor mac/vendor code (which is used by multiple product manufacturers) is hard to reason about how other devices are affected.

For example the following fixes the IKEA and Danfoss problem:
```
node->nodeDescriptor().manufacturerCode() == VENDOR_JENNIC && checkMacVendor(sc->address, VENDOR_JENNIC))
```

But now the check for the Trust remote which formerly triggered just on the mac address would not trigger any longer for the newer 0x122d AduroLight vendor code of the Trust remote (ZYCT-202), hence fixing two problems but perhaps creating a new one due changed logic.

The VENDOR_IKEA and VENDOR_DANFOSS are more specific in that sense and affect a smaller amount of products. Until the new Device code arrives which fixes this mess, I'm very much in favor to prioritize more specific checks to keep the unforeseen side effects as small as possible. I'm aware it might be a bit hyperbole :)

**Comment by dn0sar** [comment]
Thanks a lot for taking the time to explain it. These are nuances I didn't understand just by looking at the source code, and it makes more sense now.

**Comment by SwoopX** [comment]
@manup 

> But now the check for the Trust remote which formerly triggered just on the mac address would not trigger any longer for the newer 0x122d AduroLight vendor code of the Trust remote (ZYCT-202), hence fixing two problems but perhaps creating a new one due changed logic.

That can easily be fixed from my perspective and as of now, I'd support @dn0sar's PR.

If we add the following to `de_web_plugin_private.h`, **including** the proposed change, everything should go smoothly for the newer device version of the Trust remote

```
case VENDOR_ADUROLIGHT:
	return prefix == silabs2MacPrefix;
```

The point is that the previously given explanations were meant good, but stressed the wrong detail (sorry mate), so I also tipped into that mine a dozen times. The real issue is that the vendor here is hardcoded and not taken from the actual sensor candidate, the variable part is only the MAC prefix. Now, we have 2 prefixes assigned for VENDOR_JENNIC:

```
case VENDOR_JENNIC:
	return prefix == jennicMacPrefix ||
		prefix == silabs2MacPrefix;
```

**ANY** device accidentially having one of those 2 prefixes will not read the modelID, since VENDOR_JENNIC is hardcoded. So with the current code, we kick IKEA, DANFOSS, JENNIC, and LDS. However, the original intention is (presumably) just to skip the attributes if:
1. The sensor candidates manufacturer code is indeed VENDOR_JENNIC **AND**
2. The MAC prefix is jennicMacPrefix

In that sense, another potential fix could be just to remove `prefix == silabs2MacPrefix` for VENDOR_JENNIC in `de_web_plugin_private.h`. Either way, we significantly shrink the number of devices being potentially affected and imho, there's no risk to make things worse.

All that leaves the fact aside that the code is suboptimally suited for excluding particular devices.

**Comment by manup** [comment]
```
(node->nodeDescriptor().manufacturerCode() == VENDOR_JENNIC && checkMacVendor(sc->address, VENDOR_JENNIC)) ||
(node->nodeDescriptor().manufacturerCode() == VENDOR_ADUROLIGHT && checkMacVendor(sc->address, VENDOR_ADUROLIGHT))
```

This should work fine too yes, my approach was to go with minimal risk and only fix the known broken devices, checkMacVendor() scares the hell out of me as it's yields to many false positives. Imo the `checkMacVendor()` function meanwhile only serves a "database" for vendor code / mac prefixes which have been seen in the wild, a better name would be `existDevicesWithVendorCodeForMacPrefix(mac, code)` :D The function doesn't tell the actual vendor anymore it's too fuzzy in many cases.

Digging a bit deeper I've found another device (iHorn sensor) which seems to be affected by (jennicMacPrefix, VENDOR_JENNIC) combo where the last check triggers in the old code, my proposal and also the one above:

https://github.com/dresden-elektronik/deconz-rest-plugin/issues/2406

From the screenshot it appears that the device indeed provides a model identifier, but due the checks it never is queried. `addSensorNode()` contains code based on the simple descriptor like it is done for the Trust remote to workaround that. Damn false positives :/

If we can trust the comment the original check was supposed to just skip the Trust remote, which was working at the time but doesn't anymore, in retrospect this broke at least newer IKEA, Danfoss and iHorn. Perhaps we should add a  debug print whenever the skip = true happens.

>In that sense, another potential fix could be just to remove prefix == silabs2MacPrefix for VENDOR_JENNIC in de_web_plugin_private.h (we currently have 4 occurrences, more might come over time). Either way, we significantly shrink the number of devices being potentially affected and imho, there's no risk to make things worse.

Having silabs2MacPrefix under VENDOR_JENNIC is indeed ugly and smells like a hack. But removing it might potentially break something, I guess it was added sometime for a reason, which would first need to be checked. A quick search yields at least:

```
{ VENDOR_JENNIC, "Plug-230V-ZB3.0", silabs2MacPrefix }, // Immax NEO ZB3.0 smart plug
```

**Comment by SwoopX** [comment]
@manup both devices mentioned came from me :) However, now I probably understand why I had some challenges integrating them, especially the iHorn gave me hell ending up in leaving the simple descriptors fingerprint in the code. I knew it could read the modelID though, but it never did.

Struggled also with the Immax as it took time for the sensors to be created. Added the mac prefix since it's kinda handy in `isDeviceSupported()` check. From my perspective, the silabs2MacPrefix must not be in explicitly assigned to VENDOR_JENNIC.

Still got both devices here if any testing might be required.

> This should work fine too yes, my approach was to go with minimal risk and only fix the known broken devices, checkMacVendor() scares the hell out of me as it's yields to many false positives. Imo the checkMacVendor() function meanwhile only serves a "database" for vendor code / mac prefixes which have been seen in the wild, a better name would be existDevicesWithVendorCodeForMacPrefix(mac, code) :D The function doesn't tell the actual vendor anymore it's too fuzzy in many cases.

Totally agree, it's a "classic". I assume, it made very good sense back in the days but as the code has grown, it now makes more trouble than it helps.

**Comment by dn0sar** [comment]
Having no idea about the set of possible devices' manufacturer and mac combinations (and how they affect the skip), my opinion is not worth much here.

That said, I personally prefer the solution proposed by @SwoopX as it seems the closer one to give less false positives. I think logging when the skip is done makes sense in any case as long as this stays fuzzy.

Renaming the function to ```existDevicesWithVendorCodeForMacPrefix(mac, code)``` sounds actually like a good idea as well, as when using it in the future (or when reading the code) it will be more obvious that there are false positives. Maybe alongside that, a function like the following would also be good:

 ```
isKnownDeviceOfVendor(const deCONZ::Node *node, quint16 vendor) {
    return node->nodeDescriptor().manufacturerCode() == vendor && existDevicesWithVendorCodeForMacPrefix(node->address(), vendor);
}
```
that could be a shortcut to ensure the vendor is properly checked. This seems useful only if both checks are done together a lot of times (but a quick search in the code seems to suggest that).

In any case, I'm happy that at least the issue I raised is understood now (@SwoopX, sorry for stressing the wrong detail).

As soon as you guys agree on the best solution, let me know, and I'm also happy to implement it and push it here if you want.

manup:
Sounds fine by me, happy to go for @SwoopX proposal plus the debug print for the skipping case.
Perhaps also adding a comment to the iHorn specific code in `addSensorNode()` so that it likewise can be removed in future after testing if it still works when this PR is merged and the related part is commented out.

>Struggled also with the Immax as it took time for the sensors to be created. Added the mac prefix since it's kinda handy in isDeviceSupported() check. From my perspective, the silabs2MacPrefix must not be in explicitly assigned to VENDOR_JENNIC.

Yes I think so too, since Jennic and Silaps are different, but before removing this we need to be sure that it doesn't break anything:

If I understand the code correctly following scary code might break and needs to be adapted:

https://github.com/dresden-elektronik/deconz-rest-plugin/blob/master/de_web_plugin.cpp#L2093-L2113

SwoopX:
@manup You're probably not gonna like this one. I played around with the iHorn sensor last night to see if the proposed changes would have any noticable effect. Astonishingly, the device was happy to eventually query `modelId` and `manufacturer`, regardless of how that check is shaped (current code, proposed new code, anything else).

So, I poked a bit around and noticed, the whole if-statement deciding to skip reading the attributes or not starts with a thermostat check:

https://github.com/dresden-elektronik/deconz-rest-plugin/blob/3a0751abaf9eb8014ebfa90f8d51127da3762b22/de_web_plugin.cpp#L15657-L15668

However, this check seems to be responsible for skipping the VENDOR_JENNIC part, as it will be most likely be true (and therefore almost always `skip = false;`). The corresponding variable is declared a bit above (initialized with 0) and will, as indicated, most likely not change

https://github.com/dresden-elektronik/deconz-rest-plugin/blob/3a0751abaf9eb8014ebfa90f8d51127da3762b22/de_web_plugin.cpp#L15490

In the end, this was the reason why the attributes have been queried (which is in that case fine and wanted) although it was not intended by the code. Commenting out the `thermostatClusterEndpoint == 0` check finally resulted in the expected behaviour the attributes **NOT** being queried because of `skip = true;`.

As a side note, the iHorn sensor would still require the code in `addSensorNode()` as it sends garbage as manufacturer.

I guess bottom line is that the proposed change will not be the ultimate solution whereas amending the aforementioned check could introduce more unforseen side effects.

What do you think?

manup:
Ha, good catch! It looks like in thermostat check I used the wrong operator ==, related commit: https://github.com/dresden-elektronik/deconz-rest-plugin/commit/302093d5495e83347a0e909e763ab355ab339ec5

```
 if (thermostatClusterEndpoint == 0) // e.g. Eurotronic SPZB0001 thermostat 
 {  } 
```

The Eurotronic SPZB0001 has VENDOR_JENNIC and the 0x001508d mac prefix and would fall into the skip = true trap since thermostatClusterEndpoint is > 0 but not 0. Perhaps that's also the reason for reported issues pairing this thermostat?

I think the check should be:

```
 if (thermostatClusterEndpoint > 0) // e.g. Eurotronic SPZB0001 thermostat 
 {  } 
```

SwoopX:
Yeah, agree that this could be the reason and on how it should be changed. As I see it, amending the thermostat check alone should already resolve quite some pairing issues.

You lucky guys, I pretty much got all devices mentioned throughout the discussion here for testing (Eurotronic, Danfoss, iHorn, Immax smart plug) here for testing. Will do so the next couple of days with the code as is and with the amendments. Anyway, in contrast to my first impression above, I guess it's safe to say that the situation will definitively only become better than worse...

dn0sar:
I've made the changes as discussed here. See the commit info for details of what I've added in the commit.  Note that I've also renamed the `checkMacVendor` function to what @manup proposed. If you don't like that do let me know and I can revert it back.

In general, do let me know if there is something that I misunderstood and I've added although it was not intended by you two.

SwoopX:
Some interim test results with the code as is:

- Eurotronic spirit (VENDOR_JENNIC, jennicMacPrefix)
**Result**: Does not read attributes, `checkMacVendor()` kicking in
**Expected result:** Read attributes

- Danfoss Ally (VENDOR_DANFOSS, silabs2MacPrefix)
**Result**: Does not read attributes as it couldn't get past reading the active enpoints. So, different story...
**Expected result:** Read attributes

- Immax smart plug (VENDOR_JENNIC, silabs2MacPrefix)
**Result**: Does read attributes
**Expected result:** Read attributes

- iHorn temp./hum. sensor (VENDOR_JENNIC, jennicMacPrefix)
**Result**: Does read attributes, `thermostatClusterEndpoint()` kicking in before
**Expected result:** No attribute read

Will update this comment later after having amended the thermostat check.

SwoopX:
So, bottom line for me after testing is, we should do the following:

- Change `thermostatClusterEndpoint == 0` to `thermostatClusterEndpoint > 0`
- Change `checkMacVendor(sc->address, VENDOR_JENNIC)` to `node->nodeDescriptor().manufacturerCode() == VENDOR_JENNIC && checkMacVendor(sc->address, VENDOR_JENNIC)`
- Add a comparable check for Trust's changed manufacturer code
- Remove `silabs2MacPrefix` for VENDOR_JENNIC

Additionally, if the `checkMacVendor` here is really just for the Trust device, I'd propose to copy the simple descriptor for Trust and iHorn from `addSensorNode()` down here to be as specific as possible, and not to universally exclude the combination VENDOR_JENNIC and jennicMacPrefix for any other devices.

> 
> If I understand the code correctly following scary code might break and needs to be adapted:
> 
> https://github.com/dresden-elektronik/deconz-rest-plugin/blob/master/de_web_plugin.cpp#L2093-L2113

@manup I'm pretty relaxed on that code passage, as it is very specific from my perspective.

de_web_plugin.cpp
```diff
@@ -15642,40 +15641,53 @@ void DeRestPluginPrivate::delayedFastEnddeviceProbe(const deCONZ::NodeEvent *eve
...
             else if (iasZoneType > 0) // IAS motion and contact sensors
             {  }
             else if (modelId.startsWith(QLatin1String("lumi.")))
             {
                 skip = true; // Xiaomi Mija devices won't respond to ZCL read
+                DBG_Printf(DBG_INFO, "[4] Skipping additional attribute read - Model starts with 'lumi.'\n");
             }
-            else if (checkMacVendor(sc->address, VENDOR_JENNIC))
+            else if (checkMacAndVendor(node, VENDOR_JENNIC) ||
+                     checkMacAndVendor(node, VENDOR_ADUROLIGHT))
```
dn0sar, line 15654: I've added also the check for ADUROLIGHT. Let me know if you want to remove this

dn0sar:
All the other changes that you mentioned in your last comment are already there.

manup:
Awesome, very good job on testing :+1: 

I think the VENDOR_JENNIC and VENDOR_ADUROLIGHT checks needs to be more specific to only trigger for the Trust remote ZYCT-202 but not the Trust contact and motion sensors.

```
else if ((checkMacAndVendor(node, VENDOR_JENNIC) ||  checkMacAndVendor(node, VENDOR_ADUROLIGHT))
             && node->simpleDescriptors().size() == 2 && node->simpleDescriptors().front().deviceId() == DEV_ID_ZLL_NON_COLOR_CONTROLLER)
{
    skip = true; // e.g. Trust remote (ZYCT-202)
}
```

PR #3580, opened 2020-11-03
Set ConfigPending flag for Heiman Watersensor

ChrisHae:
This fix the problem that Heiman water sensor could not be paired.
It will still join as "unreachable" until it is activated once.
TODO: remove the unreachable state after pairing


SwoopX:
The code you suggested is already included a bit later and not only Heiman specific

https://github.com/dresden-elektronik/deconz-rest-plugin/blob/3a0751abaf9eb8014ebfa90f8d51127da3762b22/de_web_plugin.cpp#L6421-L6438

manup:
Should be obsolete

PR #3607, opened 2020-11-07
Add initial support for Develco air quality sensor AQSZB-110

SwoopX:


Stram66DE:
Is this change specific to Conbee 2?

PR #3614, opened 2020-11-08
Relax IKEA lights reporting configuration

manup:
The initial approach to use the settings from the IKEA gateway results in a huge amount of rapid reports during state transitions. The lights are sending reports for every tiny change especially for Level and Color  Control clusters.
This commit ensures that there are at least 5 seconds between two reports for an attribute. For color temperature lights reporting of ColorXY attributes is disabled (these need more logic to be handled properly).

### DIFF
--- a/bindings.cpp
+++ b/bindings.cpp
@@ -725,13 +725,13 @@ bool DeRestPluginPrivate::sendConfigureReportingRequest(BindingTask &bt, const s
                 out.push_back(rq);
             }
         }
-        else if (lightNode)
+        else if (lightNode && rq.maxInterval != 0xffff /* disable reporting */)
         {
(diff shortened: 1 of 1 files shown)


MattWestb:
IKEA with the ZLL GW is using very heavy broadcast for all possible and impossible things but because they have limited to 10 (perhaps updated to 20 or 30) light devices / controller device = group. 
They cant building large systems with their system architecture but its not necessarily for them to do.
They is very easy getting the network layer getting blocked because of broadcast storming (its limited to around 1.2 / seconds in the network layer).

Rapid status response its nice but command response and reliability is 200% more important and more then building larger systems.

Keep diving and improving and in the end its being great for sure !!

PR #3615, opened 2020-11-08
Fix refreshing SwBuildId last set time stamps

manup:
The problem was that periodic polling of the attribute was triggered endlessly because the last query time stamp was never updated.

lightNode->setSwBuildId(str) was removed because item->setValue(str) does the same and faster since item is already known.


Mimiix:
It this the reason some sensors stopped working?

manup:
Nope there was no harm here other than increased network traffic.

PR #3626, opened 2020-11-10
Fix updating heatpoint for newer Eurotronic Spirit Zigbee thermostats

joukestoel:
This pull request fixes issue # 1098. 
In short: some newer versions of the Eurotronic Spirit Zigbee thermostats do not accept updating the manufacturer specific attribute 0x4003 when changing the current heating setpoint. These models only accept the standard 0x0012 attribute, Occupied Heating Setpoint. 
For backward compatibility sake a check is added to compare the current software version. If this is lesser than 22190903 the old attribute (0x4003) is written to, otherwise the value is written to 0x0012 (without a specific vendor code).

### DIFF
--- a/rest_sensors.cpp
+++ b/rest_sensors.cpp
@@ -1083,7 +1083,15 @@ int DeRestPluginPrivate::changeSensorConfig(const ApiRequest &req, ApiResponse &
                         hostFlags &= ~0x04; // clear `boost` flag
                         hostFlags |=  0x10; // set `disable off` flag
 
-                        if (addTaskThermostatReadWriteAttribute(task, deCONZ::ZclWriteAttributesId, VENDOR_JENNIC, 0x4003, deCONZ::Zcl16BitInt, heatsetpoint))
+                        // Older models of the Eurotroninc Spirit updated the heat set point via the manufacturer custom attribute 0x4003. 
+                        // For newer models it is not possible to write to this attribute.
+                        // Newer models must use the standard Occupied Heating Setpoint value (0x0012) using a default (or none) manufacturer.
+                        // See GitHub issue #1098
+                        bool success = sensor->swVersion().toInt() < 22190930 ? 
+                            addTaskThermostatReadWriteAttribute(task, deCONZ::ZclWriteAttributesId, VENDOR_JENNIC, 0x4003, deCONZ::Zcl16BitInt, heatsetpoint) :
+                            addTaskThermostatReadWriteAttribute(task, deCONZ::ZclWriteAttributesId, VENDOR_NONE, 0x0012, deCONZ::Zcl16BitInt, heatsetpoint);
+                
+                        if (success)
                         {
                             updated = true;
                         }


Mimiix:
@manup On next pulls please make sure to link the issue then it gets closed

@joukestoel Thanks for the PR and fix! Can you let me know if the merge indeed fixes it? Then i can close the issue itself.

joukestoel:
@Mimiix Yes! Will do.

rest_sensors.cpp
```diff
@@ -1083,7 +1083,15 @@ int DeRestPluginPrivate::changeSensorConfig(const ApiRequest &req, ApiResponse &
                         hostFlags &= ~0x04; // clear `boost` flag
                         hostFlags |=  0x10; // set `disable off` flag
 
-                        if (addTaskThermostatReadWriteAttribute(task, deCONZ::ZclWriteAttributesId, VENDOR_JENNIC, 0x4003, deCONZ::Zcl16BitInt, heatsetpoint))
+                        // Older models of the Eurotroninc Spirit updated the heat set point via the manufacturer custom attribute 0x4003. 
+                        // For newer models it is not possible to write to this attribute.
+                        // Newer models must use the standard Occupied Heating Setpoint value (0x0012) using a default (or none) manufacturer.
+                        // See GitHub issue #1098
+                        bool success = sensor->swVersion().toInt() < 22190930 ?
```
dowhiletrue, line 1090: swVersion might be null or empty if not read correctly. it might change to something non-numeric in the future

teddy-rpi:
Hi guys, thank you very much for your effords in fixing this issue. Unfortunately the update has not yet solved the problem in my setup. Is it possible, that the swVersion you are using is not correct? In my deCONZ software the software Version of the Eurotronic Spirit thermostat (0x0006) is 20191014. This looks like a date to me. If you are using 22190930 this would mean, I would have to wait another 200 years for the problem to be fixed ;). Could you check that? Thanks a lot! teddy

joukestoel:
Hi @teddy-rpi , based on your software version it seems you have an earlier model. According to earlier reports your version should still respond to updates to the 0x4003 attributes (and this is also the attribute that is written to in the current fix implementation). Can you confirm if this works or is the thermostat not updated?

joukestoel:
And indeed, the earlier software versions look like a date but the software versions of the newer models (like the one I own) do not resemble a date format :s I do not know what the underlying idea is with the version numbering

teddy-rpi:
Hi @joukestoel, no unfortunately the temperature is not updated. I can change it via deCONZ but not via Phoscon/home assistant. Also the package of my spirit is equal to the newer version. See [https://forum.iobroker.net/post/355405](url) teddy

teddy-rpi:
sorry, the linked post is in German from the iobroker forum. They are having the same issue and it seems that the older Version of the thermostat (multicolor packaging) seems to work while the newer one (golden "Spirit" on packaging) seems to have the issues that I am experiencing. Is the software version the value that is given under the basic cluster 0x0006 "Date Code"?

teddy-rpi:
@joukestoel Hassio gives me the device info: 
SPZB0001
by Eurotronic
Phoscon-GW
Firmware: 20191014 
So this seems to be the correct attribute (0x0006). Is it possible that the guys at Eurotronic misstakenly published a firmware version with a typo in its name before going back to the original datestamp naming convention (the name of the attribute "Date Code" implies a datestamp)? That would mean that my firmware is younger than yours (just bought the thermostat at amazon this week) and the simple "<" comparison would fail. Question would be - how many firmware versions were there starting with 22? teddy

joukestoel:
@teddy-rpi I have just submitted a new pull request that changes the behaviour of my earlier fix. This new 'fix' simply writes to both attributes (0x4003 and 0x0012) of which one will be accepted while the other one will be rejected.
This is not a proper fix but since it seems that the manufacturer is working on a firmware update I feel this workaround could be temporarily accepted. If my pull request is accepted it means that your problems (with regard to changing the heating setpoint) should be resolved after the next deconz release

teddy-rpi:
Thanks a lot, I will impatiently wait for the next release ;)

Issue #3382, opened 2020-10-08
Add the ability to adjust cooling setpoints of zigbee thermostats


andynbaker:
## Feature request type

Extend the REST-API to allow setting of the cooling setpoint of zigbee thermostats, specifically the Zen-01.
## Description

Full disclosure, I have no idea what I'm doing here. I have a conbee II and run the deCONZ Home Assistant add-on and integration. I also have three Zen-01 thermostats that control traditional combined central heating and cooling HVAC systems, (so both HEAT and COOL modes are supported). I have the thermostats successfully paired, and by [tweaking the deCONZ home assistant integration,](https://github.com/home-assistant/core/issues/41182#issuecomment-703255600) I can successfully change the mode between heating and cooling.

However, it appears the deCONZ REST-API (and therefore the deCONZ HA integration) cannot control the cooling setpoint. In cooling mode, I cannot change the setpoint. But in heating mode I can change the setpoint. As I posted [here](https://github.com/home-assistant/core/issues/41182#issuecomment-705759416) in the HA integration, it appears the REST-API does not currently support adjusting the cooling setpoint?

## Considered alternatives

I have run the ZHA Home Assistant integration instead of deCONZ, using the conbee II, and was able to successfully pair the thermostats and control the cooling setpoints. However, I'd prefer to use deCONZ instead of ZHA and would love to get this feature  carried over.

SwoopX:
I got this already waiting locally to be pushed. I expect this to be available with the beta next week.

Does the zen has a fan control cluster btw?

andynbaker:
Yay! And yes, I think it does have a fan control cluster. At least, using the ZHA integration in Home Assistant I was able to change the fan from auto to on (and it was a really nice feature and enabled me to set up circulation routines when the HVAC system was idle for an extended period of time).

Does this help?

![Screen Shot 2020-10-08 at 5 53 15 PM](https://user-images.githubusercontent.com/53503316/95521459-2fb07980-098f-11eb-8c9d-ca656ff8a88c.png)
help?

SwoopX:
Yeah, that helps. Can you please double check if you got a light resource for steering the fan? If not, try searching for lights and change the fan mode.

andynbaker:
Just checked again. I do not have any extra lights representing the fan. I just removed and repaired one to be sure and no lights popped up.

SwoopX:
[...]

PR #3665, opened 2020-11-15
Improve Zen thermostat support to enable cooling control

SwoopX:
Improve Zen thermostat support to enable cooling control (#3382)

- Rely on attribute reporting
- Expose `RConfigCoolSetpoint`, `RConfigMode` and `RConfigFanMode` via API

### DIFF
--- a/database.cpp
+++ b/database.cpp
@@ -3587,6 +3587,9 @@ static int sqliteLoadAllSensorsCallback(void *user, int ncols, char **colval , c
                 }
                 else if (sensor.modelId() == QLatin1String("Zen-01"))
                 {
+                    sensor.addItem(DataTypeInt16, RConfigCoolSetpoint);
+                    sensor.addItem(DataTypeString, RConfigMode);
+                    sensor.addItem(DataTypeString, RConfigFanMode);
                 }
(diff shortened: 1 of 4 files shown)


andynbaker:
Hey @SwoopX! I'm playing with the latest deconz image now. I was able to follow the [steps ](https://dresden-elektronik.github.io/deconz-rest-doc/endpoints/sensors/) to find my username and start sending commands to the api via http messages. I was able to change the mode back and forth between cooling and heating without issue. I was also able to change the fan mode back and forth between on and auto without issue.

But changing the cooling set points always resulted in an error. I set the setpoints manually on my thermostat to 74 and 75 deg F and then polled the API for the current state and got values of 2350 and 2400, respectively. When I tried to push those values back with the sensors/x/config endpoint, it returned an error each time:

This is the PUT payload: `{"coolsetpoint":"2350"}` This is the end point: `http://[MY_HOST_IP:PORT]/api/[USERNAME]/sensors/[MY_SENSOR_ID]/config`. And this is the response:
`[{"error":{"address":"/sensors/2/config/coolsetpoint","description":"invalid value, 2350, for parameter coolsetpoint","type":7}}]`

PR #3675, opened 2020-11-15
No updates if current version is newer than announced version.

KodeCR:
Current versions of deCONZ are much newer than the announced version and since it was a simple compare not equal it always showed an update was available.
Changed it so only if a newer version is announced an update is shown.

### DIFF
--- a/discovery.cpp
+++ b/discovery.cpp
@@ -361,9 +361,10 @@ void DeRestPluginPrivate::internetDiscoveryExtractVersionInfo(QNetworkReply *rep
             {
                 QStringList gwUpdateVersionList = gwUpdateVersion.split('.');
                 QStringList versionList = version.split('.');
-                if (((gwUpdateVersionList[0].toInt() <  versionList[0].toInt())) ||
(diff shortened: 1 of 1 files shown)


discovery.cpp
```diff
@@ -359,13 +359,21 @@ void DeRestPluginPrivate::internetDiscoveryExtractVersionInfo(QNetworkReply *rep
 
             if (!version.isEmpty())
             {
-                if (gwUpdateVersion != version)
+                QStringList gwUpdateVersionList = gwUpdateVersion.split('.');
+                QStringList versionList = version.split('.');
+                if (((gwUpdateVersionList[0].toInt() <  versionList[0].toInt())) ||
```
manup, line 364: I'd recommend to add a check that both lists contain at least 3 values:

    (gwUpdateVersionList.size() >= 3 && versionList.size() >= 3)

manup (changes requested):

manup:
Hi, good catch, can you please add the check to prevent potential SEGV.

discovery.cpp
KodeCR: Thanks, done

KodeCR:
Thanks, just noticed this now and have added the check.

PR #3680, opened 2020-11-16
Second attempt to make the heating setpoint changeable for all Eurotronic Spirit Zigbee thermostats (issue #1098)

joukestoel:
This is a second attempt to 'fix' the changing of the heating setpoint from within the API (issue #1098). 

This is actually not a real and proper fix but a temporarily workaround. The fix will always send two requests to the thermostat when the user changes the heating setpoint. One write to the 0x4003 attribute and another write to the 0x0012 attribute. Depending on the model it will accept one of these attribute write requests while 'ignoring' the other (actually it will respond back to the client that it did not accept the write attribute request).

The reasons why I think this workaround is accepted is that the manufacturer of the device is working on a new firmware version that should not have this issue (and others). Until then this code can be left in place to enable altering of the heating setpoint


rest_sensors.cpp
```diff
@@ -1089,12 +1089,14 @@ int DeRestPluginPrivate::changeSensorConfig(const ApiRequest &req, ApiResponse &
...
-                        bool success = sensor->swVersion().toInt() < 22190930 ?
-                            addTaskThermostatReadWriteAttribute(task, deCONZ::ZclWriteAttributesId, VENDOR_JENNIC, 0x4003, deCONZ::Zcl16BitInt, heatsetpoint) :
-                            addTaskThermostatReadWriteAttribute(task, deCONZ::ZclWriteAttributesId, VENDOR_NONE, 0x0012, deCONZ::Zcl16BitInt, heatsetpoint);
-
-                        if (success)
+                        // UPD 16-11-2020: Since there is no way to reckognize older and newer models correctly and a new firmware version is on its way this
+                        //                 'fix' is changed to a more robust but ugly implementation by simply sending both codes to the device. One of the commands
+                        //                 will be accepted while the other one will be refused. Let's hope this code can be removed in a future release.
+    
+                        if (addTaskThermostatReadWriteAttribute(task, deCONZ::ZclWriteAttributesId, VENDOR_JENNIC, 0x4003, deCONZ::Zcl16BitInt, heatsetpoint) &&
```
DerOetzi, line 1096: I'm not sure whether a logical AND is correct here? I think the writing to 0x4003 will fail on those new devices with wrong firmware and therefor this statement want get true?

DerOetzi (changes requested):

rest_sensors.cpp
joukestoel: The method `addTaskThermostatReadWriteAttribute` returns `true` whenever a correct request message has been send. It does not depend on the response of the client. So, as far as I know, as long as the request message is syntactically correct the method will return `true`. Since this message is constructed in the code (apart from the request heating setpoint) I think it will always succeed (unless you enter out-of-range heating setpoint values).

joukestoel: In other words: having a logical `or` here would only evaluate the left hand side (since it most likely returns `true`) and ignore the right hand side discarding one of the two requests.

DerOetzi: Thanks @joukestoel for pointing this out for me. I'm really new with the deconz code. And have no running Dev-environment yet, because I mainly programming in Java and Python. 

You can set my comment to resolved.

joukestoel: 👍

PR #3701, opened 2020-11-18
Support checking for software updates.

KodeCR:
When update state indicates 'noupdates', this allows checking for updates by a PUT of 'checkforupdate: true' to swupdate2 (as the hue app does).

### DIFF
--- a/de_web_plugin_private.h
+++ b/de_web_plugin_private.h
@@ -1641,6 +1641,7 @@ public:
     bool gwLANBridgeId;
     QString gwBridgeId;
     QString gwUuid;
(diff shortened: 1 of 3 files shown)


KodeCR:
Keeping this alive

manup:
Hi, I'm holding this one back for now, since I'd like to refactor the discovery handling with some simple state machine first.

The REST request should only mark/notify that a check for update is requested. This will be an event in the state machine version. In the PR the timer function is called directly, leaving it vulnerable to queue endless calls towards the discovery server.

PR #3768, opened 2020-11-28
Improve setup of Aqara Opple switches

manup:
This PR ensures the magic attribute to enable button events is written in the hourly report, and at least once per deCONZ start.
After startup writing the attribute can be enforced by pressing the 'C' button on the back of the switch.
There was also another issue when the switch is paired a simple descriptor request loop could happen, this is fixed in deCONZ core now.

 Issue https://github.com/dresden-elektronik/deconz-rest-plugin/issues/3670

### DIFF
--- a/database.cpp
+++ b/database.cpp
@@ -3770,6 +3770,13 @@ static int sqliteLoadAllSensorsCallback(void *user, int ncols, char **colval , c
                 //item = sensor.addItem(DataTypeInt16, RConfigOffset);
                 //item->setValue(0);
             }
+
+            if (sensor.modelId().endsWith(QLatin1String("86opcn01")))
+            {
+                // Aqara Opple switches need to be configured to send proper button events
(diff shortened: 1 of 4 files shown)


omolko:
Hi I have this issue. I don't understand what to do

manup:
> Hi I have this issue.(my switch is seen on deconz but the button doesn't works) I don't understand what to do
> Conbee II
> version 2.05.88 / 15/10/2020
> Firmware 26660700

With the latest deCONZ version 2.7.0 and the firmware 0x26680700 the switch should be configured properly after pressing the 'C' setup button on the back, can you try this?

omolko:
I have downgrade deconz to 6.5.0 in home assistant and it works fine. But I
don't see the 2.7 version in HA.
thanks again

Mimiix:
@omolko HA needs to update the addon themselves to the latest version.

PR #3784, opened 2020-11-30
Various fixes

SwoopX:
- Re-enable typ check for RConfigCoolSetpoint
- Remove unsupported config parameters for Danfoss/Hive thermostats (`RConfigSchedule` and `RConfigScheduleOn` got created while loading the sensor from DB without the device supporting it)
- Some cleanup for thermostat capabilities (re-order the entries, remove `RConfigMode` duplicate for Zen and prevent ELKO Super TR to expose `RConfigSchedule` and `RConfigScheduleOn`)
- Additions for Xiaomi specific cluster
- Update for Xiaomi special reporting output
- Add Sinope specific clusters/attributes to general.xml (#2104)
- Process Konke motion sensor battery reports (#1607)
- Further additions for Xiaomi special reporting output
- Ensure button map is only loaded once

### DIFF
--- a/de_web_plugin.cpp
+++ b/de_web_plugin.cpp
@@ -6304,15 +6304,6 @@ void DeRestPluginPrivate::addSensorNode(const deCONZ::Node *node, const SensorFi
                 sensorNode.addItem(DataTypeString, RConfigMode);
             }
 
-            if (sensorNode.modelId() == QLatin1String("Super TR"))   // ELKO
-            {
-                sensorNode.addItem(DataTypeString, RConfigTemperatureMeasurement);
-                sensorNode.addItem(DataTypeInt16, RStateFloorTemperature);
-                sensorNode.addItem(DataTypeBool, RStateHeating);
-                sensorNode.addItem(DataTypeBool, RConfigLocked);
-                sensorNode.addItem(DataTypeString, RConfigMode);
-            }
-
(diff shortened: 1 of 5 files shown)


de_web_plugin.cpp
```diff
@@ -761,9 +761,9 @@ DeRestPluginPrivate::DeRestPluginPrivate(QObject *parent) :
     }
 
     const QStringList buttonMapLocations = {
-        deCONZ::getStorageLocation(deCONZ::ApplicationsDataLocation) + QLatin1String("/devices/button_maps.json")
 #ifdef Q_OS_LINUX
-        , "/usr/share/deCONZ/devices/button_maps.json"
+        QLatin1String("/usr/share/deCONZ/devices/button_maps.json"), deCONZ::getStorageLocation(deCONZ::ApplicationsDataLocation) +
+        QLatin1String("/devices/button_maps.json")
```
manup, line 766: I think this change is problematic.

`deCONZ::getStorageLocation(deCONZ::ApplicationsDataLocation) +  QLatin1String("/devices/button_maps.json")`

- This line should always come first so it has a higher priority as the packaged button_maps.json;
- It's not Linux specific, with this change it wouldn't work on Windows anymore.

manup (changes requested):
See review comments on button maps loading.

de_web_plugin.cpp
SwoopX: Change's reverted

PR #3838, opened 2020-12-08
Lidl plugs fill manufacturername, modelid and swversion for all endpoints

manup:
- Fill the data from endpoint 0x01, which is the only valid Basic Cluster, into endpoints 0x02 and 0x03;
- Use Basic Cluster application version attribute for `swversion`;
- Don't endlessly read empty string attributes in poll manager.


DynaMake:
OK With Conbee II firmware 26660700

Issue #3719, opened 2020-11-20
Evaluation of rules broken in v2.05.88


ebaauw:
## Describe the bug

The evaluation of rules in v2.05.88 is broken.  It still works in v2.04.84.  Consequently my toggle buttons no longer work.

## Steps to reproduce the behavior

Create a CLIP sensor and a pair of rules to toggle a light state when pressing a button:
```
$ ph get /rules/172
{
  "actions": [
    {
      "address": "/sensors/281/state",
      "body": {
        "status": 0
      },
      "method": "PUT"
    }
  ],
  "conditions": [
    {
      "address": "/sensors/285/state/buttonevent",
      "operator": "eq",
      "value": "1002"
    },
    {
      "address": "/sensors/285/state/lastupdated",
      "operator": "dx"
    },
    {
      "address": "/sensors/281/state/status",
      "operator": "gt",
      "value": "0"
    }
  ],
  "created": "2019-12-08T18:46:29",
  "etag": "774523b5d284c23c40cadb85ec36d772",
  "lasttriggered": "2020-11-20T19:12:43",
  "name": "Balcony Switch On/Off Press (1/2)",
  "owner": "3635D7B19D",
  "periodic": 0,
  "status": "enabled",
  "timestriggered": 1
}
$ ph get /rules/173
{
  "actions": [
    {
      "address": "/sensors/281/state",
      "body": {
        "status": 1
      },
      "method": "PUT"
    }
  ],
  "conditions": [
    {
      "address": "/sensors/285/state/buttonevent",
      "operator": "eq",
      "value": "1002"
    },
    {
      "address": "/sensors/285/state/lastupdated",
      "operator": "dx"
    },
    {
      "address": "/sensors/281/state/status",
      "operator": "lt",
      "value": "1"
    }
  ],
  "created": "2019-12-08T18:46:29",
  "etag": "3561bd37e71b6d57d31099ac5a4c9200",
  "lasttriggered": "2020-11-20T19:12:41",
  "name": "Balcony Switch On/Off Press (2/2)",
  "owner": "3635D7B19D",
  "periodic": 0,
  "status": "enabled",
  "timestriggered": 1
}
```
In v2.05.88 both rules trigger when pressing the button; in v2.05.84 only one of them.

## Expected behavior

Only one of the rules should tigger, as in v2.05.84 or on the Hue bridge.

## Screenshots

n/a

## Environment

 - Host system: Raspberry Pi
[...]

PR #3918, opened 2020-12-14
Fix ambiguous loading of button maps

manup:
- The missing break, did load all entries which started with a modelid. So lumi.sensor_switch.aq2 was overwritten by lumi.sensor_switch;
- Prevent single press button event to be emitted twice for lumi.sensor_switch.aq2;
- Don't use `QString::startsWith()` always require exact modelId;
- Put manufactuer name of Tuya switch in button map and use it as modelId;
- Add helper `isTuyaManufacturerName()` to detect Tuya devices.

Issue: https://github.com/dresden-elektronik/deconz-rest-plugin/issues/3719

### DIFF
--- a/button_maps.json
+++ b/button_maps.json
@@ -854,7 +854,7 @@
         "Tuya3gangMap": {
             "vendor": "Tuya",
             "doc": "3-gang remote",
-            "modelids": [""],
+            "modelids": ["_TZ3000_bi6lpsew", "_TZ3400_keyjhapk", "_TYZB02_key8kk7r", "_TZ3400_keyjqthh", "_TZ3400_key8kk7r", "_TZ3000_vp6clf9d", "_TYZB02_keyjqthh"],
             "map": [
                 [1, "0x01", "ONOFF", "0xfd", "0", "S_BUTTON_1", "S_BUTTON_ACTION_SHORT_RELEASED", "B1 short"],
                 [1, "0x01", "ONOFF", "0xfd", "1", "S_BUTTON_1", "S_BUTTON_ACTION_DOUBLE_PRESS", "B1 double"],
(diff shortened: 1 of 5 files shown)


SwoopX:
You might want to add the Tuya manufacturer name regex here as well without any action

https://github.com/dresden-elektronik/deconz-rest-plugin/blob/1ecda0c57f4c49d0cd8828f2b139e85facd3b69b/de_web_plugin.cpp#L6502

to prevent any sensor assuming `Heiman` based on the MFC.

manup:
> You might want to add the `isTuyaManufacturerName()` here as well without any action
> 
> https://github.com/dresden-elektronik/deconz-rest-plugin/blob/1ecda0c57f4c49d0cd8828f2b139e85facd3b69b/de_web_plugin.cpp#L6502
> 
> to prevent any sensor assuming `Heiman` based on the MFC.

Good point, yes for Tuya we absolutely need to query the actual manufacturer code.

SwoopX:
Well, we have exactly that assignment a bit further below in the if-else jungle. Apparently, VENDOR_EMBER has already been silenced

'''
else if ( //node->nodeDescriptor().manufacturerCode() == VENDOR_EMBER ||
             node->nodeDescriptor().manufacturerCode() == VENDOR_HEIMAN)
    {
        sensorNode.setManufacturer("Heiman");
    }
'''

Espacially that passage gave me a hard time back in the days changing the manufacturer out of the blue. From my perspective, the function should be applied for consistency and to retain the original manufacturer name undoubtly.

manup:
I'd like to remove the manual manufacturer name setup for all vendor codes which are shared across different manufacturers.
And then if it is empty it needs to be queried from the device (which should already happen automatically).

PR #3953, opened 2020-12-17
Merry Christmas #4.

ebaauw:
See #3716: revert `bri_inc` support and trying to keep _Current Level_ up to date with actual brightness.

### DIFF
--- a/de_web_plugin.cpp
+++ b/de_web_plugin.cpp
@@ -3103,6 +3103,10 @@ LightNode *DeRestPluginPrivate::updateLightNode(const deCONZ::NodeEvent &event)
             }
             else if (ic->id() == LEVEL_CLUSTER_ID && (event.clusterId() == LEVEL_CLUSTER_ID))
             {
+                if (isXmasLightStrip(lightNode))
+                {
+                    continue;
+                }
                 std::vector<deCONZ::ZclAttribute>::const_iterator ia = ic->attributes().begin();
                 std::vector<deCONZ::ZclAttribute>::const_iterator enda = ic->attributes().end();
                 for (;ia != enda; ++ia)
--- a/xmas.cpp
+++ b/xmas.cpp
@@ -295,8 +295,6 @@ int DeRestPluginPrivate::setXmasLightStripState(const ApiRequest &req, ApiRespon
     bool targetOn = false;
     bool hasBri = false;
     quint8 targetBri = 0;
-    bool hasBriInc = false;
-    qint16 targetBriInc = 0;
     bool hasHue = false;
     quint16 targetHue = 0;
     bool hasSat = false;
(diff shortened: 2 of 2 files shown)


sonar98:
@ebaauw it's the time of the year again so I'm using the led string again.
Now I noticed that when using the "color" (physical) button on the device it is possible to set it to a multi color mode.
It looks like the carnival mode of the effect but without the twinkle/movement.

image

Is it possible to set this with the api as well?

LaCruz75:
I bought today this nice LED xmas-string.
:-)

Those effects are avaible:"steady","snow","rainbow","snake","twinkle","firework","horizontal_flag","waves","updown","vintage","fading","collide","strobe","sparkles","carnaval","glow" and I can set them via REST-API:

curl -H 'Content-Type: application/json' -X PUT -d '{"effect":"sparkles"}' http://172.30.33.2:40850/api/33E7B7A7B7/lights/14/state

with success

or

curl -H 'Content-Type: application/json' -X PUT -d '{"effect":"snow","speed":100,"colors":[{"r":164,"g":168,"b":50},{"r":168,"g":50,"b":153},{"r":102,"g":50,"b":168},{"r":50,"g":155,"b":168},{"r":50,"g":168,"b":72},{"r":168,"g":88,"b":50}]}' http://172.30.33.2:40850/api/33E7B7A7B7/lights/14/state

with error an success

TribunMaximus:
Hi, also looking for a way to the different colors of all the effects:

`service: deconz.configure
data:
  entity: light.lichterkette_2
  field: /state
  data:
    'on': true
    effect: fading
    effectSpeed: 50
    hue: 30000
    bri: 254
    sat: 254
  bridgeid: 00212E07CDED`

works fine. But changing effect to carnaval comes to an error. I think this is because the effect needs different colors.
Any idea how to set all the different colors for the carnaval effect?

Thanks for your help.

Br
JEns

MarjanRothman:
> Hi, also looking for a way to the different colors of all the effects:
> 
> > service: deconz.configure
> > data:
> > entity: light.lichterkette_2
> > field: /state
> > data:
> > 'on': true
> > effect: fading
> > effectSpeed: 50
> > hue: 30000
> > bri: 254
> > sat: 254
> > bridgeid: 00212E07CDED`
> 
> works fine. But changing effect to carnaval comes to an error. I think this is because the effect needs different colors. Any idea how to set all the different colors for the carnaval effect?
> 
> Thanks for your help.
> 
> Br JEns

@TribunMaximus In case you are still looking for the answer. Use `effectColours` to set different colours, i.e.
```
effect: carnival
effectSpeed: 10
effectColours: [[255,0,0],[255,255,255],[255,197,25],[255,255,255]]
```

Issue #3744, opened 2020-11-23
groups/0 (all devices) not working


popy2k14:
## Describe the bug

Yesterday i have updated to 2.05.88 from 02.05.84 and now my groups/0 doesnt work at all.
I cant set it with phoscon "all off switch" or over rest api.
There is a 404 error with HUEEssentials.

## Steps to reproduce the behavior

Switch the group 0 as stated above.

## Expected behavior

All lights should be controlable with group 0.

## Environment

02.05.84 (downgraded again)
26660700

 - Host system: Raspberry Pi
 - Running method: Raspbian
 - Firmware version: 26660700
 - deCONZ version: 2.05.84 and 02.05.88
 - Device: ConBee II
 - Do you use an USB extension cable: yes
 - Is there any other USB or serial devices connected to the host system? If so: Which? yes SSD

## Additional context

When ill do a get on group "0" manuall with the browser the answer is:

GET: http://192.168.0.9:8080/api/XXXXXXXXXX/groups/0
Answer: [{"error":{"address":"/groups/0","description":"resource, /groups/0, not available","type":3}}]

With other groups all is working.

ANy hints how i can get my group/0 working again?

MattWestb:
Try with 0xFFF0 = 65520 that is the "new" groupe Al

popy2k14:
Thank you for your help.
Sadly, not working on 2.05.84 after downgrade from 02.05.88.
result is:

`[{"error":{"address":"/groups/65520","description":"resource, /groups/65520, not available","type":3}}]`

Any hints?
Do i have to upgrade to 02.05.88 again to get it working?

PS.: why should the groups/0 be changed?

MattWestb:
Then taking on look in zigbee.db in the table group wot number "All" is having but not the one that is marked "deleted"

MattWestb:
About group 0 is in Zigbee 3 reserved for global scene commands and not as one trashcan for HA devices.

popy2k14:
ok, can you please guide me in the right direction regarding zigbee.db.
Dont want to mess something up.

I just updated deconz yesterday to 02.05.88 and now my groups/0 is gone :-(

MattWestb:
Sorry I was having wrong with the name but on my Pi its installed in
/home/pi/.local/share/dresden-elektronik/deCONZ/zll.db
[...]

PR #3966, opened 2020-12-18
Only add lights to the "All" group.

KodeCR:
This PR intents to only add actual lights to the "All" group/group 0.
This hopefully addresses #3744 where the Fyrtur blinds are both a "light" and a "sensor", but the "sensor" is not part of the "All" group on the REST side, causing the group to be removed from the device again when syncing (and with it the complete group from the REST API).
The Fyrtur actually being a window covering device it should now no longer be added at all.

### DIFF
--- a/de_web_plugin.cpp
+++ b/de_web_plugin.cpp
@@ -2526,13 +2526,43 @@ void DeRestPluginPrivate::addLightNode(const deCONZ::Node *node)
         }
 
         // add light node to default group
-        GroupInfo *groupInfo = getGroupInfo(&lightNode, gwGroup0);
-        if (!groupInfo)
+        switch (i->deviceId())
         {
-            groupInfo = createGroupInfo(&lightNode, gwGroup0);
-            lightNode.setNeedSaveDatabase(true);
-            groupInfo->actions &= ~GroupInfo::ActionRemoveFromGroup; // sanity
-            groupInfo->actions |= GroupInfo::ActionAddToGroup;
+            // HA lighting devices
+            case DEV_ID_HA_ONOFF_LIGHT:
+            case DEV_ID_HA_DIMMABLE_LIGHT:
+            case DEV_ID_HA_COLOR_DIMMABLE_LIGHT:
+            case DEV_ID_HA_ONOFF_LIGHT_SWITCH:
+            // ZLL lighting devices
+            case DEV_ID_ZLL_ONOFF_LIGHT:
+            case DEV_ID_ZLL_ONOFF_PLUGIN_UNIT:
+            //case DEV_ID_ZLL_DIMMABLE_LIGHT: // same as DEV_ID_HA_ONOFF_LIGHT
+            case DEV_ID_ZLL_DIMMABLE_PLUGIN_UNIT:
+            case DEV_ID_ZLL_COLOR_LIGHT:
+            case DEV_ID_ZLL_EXTENDED_COLOR_LIGHT:
+            case DEV_ID_ZLL_COLOR_TEMPERATURE_LIGHT:
+            // ZigBee 3.0 lighting devices
+            case DEV_ID_Z30_ONOFF_PLUGIN_UNIT:
+            case DEV_ID_Z30_DIMMABLE_PLUGIN_UNIT:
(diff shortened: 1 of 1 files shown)


popy2k14:
Pump

@manup any news when this gets merged into next beta?

It's running here without any issue since weeks. 

Thx

manup:
The switch case misses some devices like DEV_ID_MAINS_POWER_OUTLET and DEV_ID_SMART_PLUG.

However, we have discussed the group0 problems in the last developer meeting, the plan is to remove the special group handling all together since it creates more trouble than it solves.

The `/groups/0` API endpoint will remain to not break the API  but it will use normal  "Broadcast to all Routers" addressing instead of the special *group0* &rarr; 0xFFF0 like it used to be in the past.

Devices won't be added to *group0* anymore and existing group membership will be removed from devices — this is the tough part.

popy2k14:
Thx for clearing things up.
Don't asking for ETA but is there any time frame for this?
Because of the issue with the current group handling, I can't update. Currently using KodeCR's binary from above.

KodeCR:
> The switch case misses some devices like DEV_ID_MAINS_POWER_OUTLET and DEV_ID_SMART_PLUG.

I don't know all the details but I thought one of the reasons the Hue bridge doesn't use a broadcast is that only lights are added to the all group?

>However, we have discussed the group0 problems in the last developer meeting, the plan is to remove the special group handling all together since it creates more trouble than it solves.

You/@ebaauw? wanted to mimic the way the Hue bridge implements this (#1005). I'm fine if you want to change it to a broadcast but why not merge this until that's done so that @popy2k14's problem is solved in the meantime?

ebaauw:
The Hue bridge designates a Zigbee group for special usage.  It exposes this group as `/groups/0`.  Note that the resource ID is in no way related to the Zigbee group number.  The bridge adds this group to each endpoint with a server _Groups_ cluster that it exposes as a `/lights` resource, irrespective of the _Device Type_ on that endpoint.  The API confusingly shows this like it has added all `/lights` resources to the `/groups/0` resource.

The Hue API has a separate `/scenes` endpoint.  All scenes use the group ID of the designated group, exposed as `/groups/0`.  This way you can recall the same scene from different groups, changing on the lights in that group.  Technically, it sends a _Recall Scene_ for the scene with the `/groups/0` group ID to the group specified in the API call.

Imho `/groups/0` is not about broadcast vs groupcast, but about (eventually) offering a compatible `/scenes` API endpoint.  I think:
- Only device endpoints with a server _Groups _ cluster should be exposed as `/lights` endpoints;
- The group corresponding to `/groups/0` should be added to all these endpoints (or: all `/lights` resources should be added to `/groups/0`);
- `state.on` and `action.on` should be used only for the server _On/Off_ cluster.  Other devices (like _Window Covering_) should use different attributes (`state.lift`, `action.lift`) that result in proper commands being sent to groups, thus enabling a group of e.g. _Window Covering_ or _Warning Device_ devices.
- Whenever we expose non-light endpoints under a different API endpoint (`/plugs`, `/windowCoverings`, `/warningDevices`), these resources should be in `/groups/0`.

Not sure how to deal with device endpoints that don’t carry _Groups_, like the second endpoint on some Xiaomi wired in-wall switches, or the _IAS WD_ endpoint on most smoke detectors.  I think its confusing to expose these as `/lights`, as they cannot be added to a group, but its also confusing to expose these differently

SwoopX:
I guess that's the 2nd step to follow. However we should first ensure all sensors/switches get removed from group 0xFFFD as this is "dangerous" and really no joy if your whole place gets shut down in the middle of the night.

KodeCR:
Thanks @ebaauw, that's clear and very helpful. I would like to support the `/scenes` Hue API properly of course (#2384). And I'm happy to work on getting the code to this point, but considering the number of PRs of mine that are still open I would first need buy-in from @manup before spending more time on this.

With @SwoopX's that's three votes for merging this in the meantime, any issue with that?

manup:
I've added PR https://github.com/dresden-elektronik/deconz-rest-plugin/pull/4608 to hopefully solve the underlying issue. This hopefully solves also some of the "ghost events" which some devices send to group `0xfff0`.

PR #4034, opened 2020-12-26
Introduce getSensorNodeForAddressEndpointAndCluster()

SwoopX:
This function provides the ability to chose a sensor more precisely than both functions `getSensorNodeForAddressAndEndpoint()`. The first implementation (which doesn't use the sensor type as argument) returns the first sensor found, which is problematic if more than one "sensor-worthy" cluster is available on one endpoint. The second implementation (which usea the sensor type as argument) is more accurate, but can only handle one sensor type (there's multiple sensor types based on IAS zone cluster).

The proposed function would eliminate both weaknesses, however, it relies on the correctness of a sensor's fingerprint. In this regard, some minor updates on the corresponding code are required.

As small caveat remains: As of now, basic cluster and power configuration clusters are not exclusively added to the sensor fingerprints. In those cases, the function will just return the first sensor found.

### DIFF
--- a/de_web_plugin.cpp
+++ b/de_web_plugin.cpp
@@ -9466,6 +9466,31 @@ Sensor *DeRestPluginPrivate::getSensorNodeForAddress(const deCONZ::Address &addr
     return 0;
 }
 
+/*! Returns the Sensor for its given \p Address and \p Endpoint and \p cluster or 0 if not found.
+ */
+Sensor *DeRestPluginPrivate::getSensorNodeForAddressEndpointAndCluster(const deCONZ::Address &addr, quint8 ep, quint16 cluster)
+{
+    for (Sensor &sensor: sensors)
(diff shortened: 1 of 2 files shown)


SwoopX:
We should leave this one out for the time being. Implications are a bit larger than anticipated...

manup:
Interesting, for the `PC_GetPollControlEndpoint(Node*)` I've used the actual Simple Descriptor to check for the cluster, also since the finger print doesn't have this cluster listet.

>We should leave this one out for the time being. Implications are a bit larger than anticipated...

Should this PR get `backlog` label then?

SwoopX:
Happy to do that. I noticed it misbehaves, since it currently doesn't include the client clusters.

Also, there might be further updates required for it to work. And for those to propagate, one needs to turn on sensor search again...

Issue #3926, opened 2020-12-15
edp re:dy plug


vpedromarques:
## Device

- Product name: edp re:dy plug.
- Manufacturer: EDP-WITHUS.
- Model identifier: ZM-SmartPlug-1.0.0.
- Device type : Please remove all unrelated device types. 
  - Other: Smart Plug with Energy Metering
![Screenshot 2020-12-15 at 18 17 02](https://user-images.githubusercontent.com/10771662/102256416-4a5d1c80-3f03-11eb-8ef9-41e673e3205c.png)
![Screenshot 2020-12-15 at 18 17 40](https://user-images.githubusercontent.com/10771662/102256419-4c26e000-3f03-11eb-9c11-016094ad38d3.png)
![Screenshot 2020-12-15 at 18 24 03](https://user-images.githubusercontent.com/10771662/102256420-4cbf7680-3f03-11eb-873d-597d4a63b62d.png)

Mimiix:
Missing screenshots.

vpedromarques:
![Screenshot 2020-12-15 at 23 35 36](https://user-images.githubusercontent.com/10771662/102285482-4b577380-3f2e-11eb-9951-5d812e8346b9.png)
![Screenshot 2020-12-15 at 23 35 13](https://user-images.githubusercontent.com/10771662/102285487-4c88a080-3f2e-11eb-93e8-470fdf0322fe.png)
![Screenshot 2020-12-15 at 23 34 57](https://user-images.githubusercontent.com/10771662/102285489-4d213700-3f2e-11eb-9e64-939a348e56a7.png)
![Screenshot 2020-12-15 at 23 34 44](https://user-images.githubusercontent.com/10771662/102285490-4d213700-3f2e-11eb-9a9d-0d1d397472a5.png)
![Screenshot 2020-12-15 at 23 34 36](https://user-images.githubusercontent.com/10771662/102285491-4db9cd80-3f2e-11eb-868a-bd45eea9f371.png)
![Screenshot 2020-12-15 at 23 33 56](https://user-images.githubusercontent.com/10771662/102285492-4e526400-3f2e-11eb-8360-712ddcbd4fe6.png)
![Screenshot 2020-12-15 at 23 32 57](https://user-images.githubusercontent.com/10771662/102285493-4e526400-3f2e-11eb-8dcd-a4068116e70f.png)
[...]

PR #4041, opened 2020-12-28
Various enhancements and fixes

SwoopX:
- Extend sensor fingerprints
- Improve sensor selection
- Initialize bool values for Thermostats with "false"
- Correct typo in modelID for edp re:dy smart plug (#3926)
- Enhance support for Stelpro baseboard thermostat STZB402+ (#4049)
- Add complete EcoDim and Namron ModelIDs to button maps (#4076)
- Add complete iCasa ModelID to button_maps.json (ICZB-RM11S) (#4125)
- Enhance debug output on valid button event
- Correct model ID in button maps for LZL4BWHL01 Remote (#4151)
- Show info on button presses in debug log for Philips switches
- Increase time resolution and add date for etag to decrease hash collisions (#4052)
- Streamline way of setting etags


de_web_plugin.cpp
```diff
@@ -1816,9 +1816,9 @@ void DeRestPluginPrivate::updateEtag(QString &etag)
 {
     QTime time = QTime::currentTime();
 #if QT_VERSION < 0x050000
-    etag = QString(QCryptographicHash::hash(time.toString().toAscii(), QCryptographicHash::Md5).toHex());
+    etag = QString(QCryptographicHash::hash(time.toString("hh:mm:ss.zzz").toAscii(), QCryptographicHash::Md5).toHex());
```
hbkoenig, line 1819: why do you use *only* time for etag hash, why not ad date too ?
and why not add measurement values and/or other stuff too ?
the md5 hash is much longer than the raw time string (12 vs. 32 chars), I don't get it what here a md5 hash can be good for ?!?

diagnostics.cpp
```diff
@@ -7,11 +7,11 @@
  */
 void DeRestPluginPrivate::handleDiagnosticsClusterIndication(const deCONZ::ApsDataIndication &ind, deCONZ::ZclFrame &zclFrame)
 {
-    Sensor *sensor = getSensorNodeForAddressAndEndpoint(ind.srcAddress(), ind.srcEndpoint());
+    Sensor *sensor = getSensorNodeForAddressAndEndpoint(ind.srcAddress(), ind.srcEndpoint(), QLatin1String("ZHAThermostat"));
 
     if (!sensor)
     {
-        DBG_Printf(DBG_INFO, "No sensor found for 0x%016llX, endpoint: 0x%08X\n", ind.srcAddress().ext(), ind.srcEndpoint());
+        DBG_Printf(DBG_INFO, "No thermostat sensor found for 0x%016llX, endpoint: 0x%08X\n", ind.srcAddress().ext(), ind.srcEndpoint());
```
manup, line 14: `handleDiagnosticsClusterIndication()` isn't this to be supposed to be general, why thermostat?

`endpoint: 0x%08X` shouldn't this be `endpoint: 0x%02X`?

air_quality.cpp
```diff
@@ -140,8 +140,7 @@ void DeRestPluginPrivate::handleAirQualityClusterIndication(const deCONZ::ApsDat
 
         if (configUpdated || stateUpdated)
         {
-            updateEtag(sensor->etag);
-            updateEtag(gwConfigEtag);
```
manup, line 144: Note the `gwConfigEtag` needs to be updated as well since querying the full config state is affected by this.

manup (commented):
Added a few comments.

air_quality.cpp
SwoopX: Using `updateSensorEtag(&*sensor);` instead does take care of that, doesn't it?

diagnostics.cpp
SwoopX: For the endpoint: Right, corrected that in other files as well

For the generic handling: Happy to remove the "thermostat" from the message. Regarding the generic usage, currently only 2 thermostats require this cluster. However, we allow further processing only if a sensor is available. And to pick the right one, I currently have to pin the sensor type some lines above.

Issue #4038, opened 2020-12-27
Lidl Melinera Smart LED string lights usage


Preto:
## Describe the question or issue you are having
This was awesome! Had a lot of Christmas fun thanks to you guys and Lidl :-)

I used a lot of different effects and used RGB colors in them, all flawlessly. I am only unable to use the _colour mode_ correctly. I'm on the latest version and removed and repaired the lights as stated in the support issue with the following steps:

 - Removed device in GUI view, then pressed delete in Phoscon
 - Put Phoscon in search light mode
 - Repaired by pressing F button > 5 seconds
 - Got recognized as Fairy Lights (model), only with a new ID.

**Reproducible steps**
1. Setting the _white mode_ according to https://github.com/dresden-elektronik/deconz-rest-plugin/issues/3716#issuecomment-735467996:
```bash
curl -d '{ "bri":254, "sat":0 }' -X PUT "http://192.168.1.1:8080/api/KEY/lights/1/state"
```
&emsp;&emsp;The API returns a success for `bri` and a separate success for the `sat`. The REST-API reflects all this too.
&emsp;&emsp;The lights however, show shortly the _white mode_ but quickly reverts to _colour mode_. In my case it flashes
&emsp;&emsp;white and returns to the color green.

2. Issuing two commands in quick succession does keep the _white mode_:
```bash
curl -d '{ "bri":254, "sat":0 }' -X PUT "http://192.168.1.1:8080/api/KEY/lights/1/state" ; curl -d '{ "effect":"none" }' -X PUT "http://192.168.1.1:8080/api/KEY/lights/1/state"
```
&emsp;&emsp;The API returns a success for `bri`, a separate success for `sat` and the second command returns also a success
&emsp;&emsp;on `effect`. The lights look the way they should and the REST-API reflect this too.

3. So lets try to combine those commands:
```bash
curl -d '{ "bri":254, "sat":0 , "effect":"none" }' -X PUT "http://192.168.1.1:8080/api/KEY/lights/1/state"
```
&emsp;&emsp;The API returns **only** a success on `effect` and ignoring the other two. Is this expected behavior?
&emsp;&emsp;Should I fire separate commands at the API?

4. Lets change the color to the mentioned yellow color (see https://github.com/dresden-elektronik/deconz-rest-plugin/issues/3716#issuecomment-736123450):
```bash
[...]

Issue #4022, opened 2020-12-25
`state.effect` not initialised properly


ebaauw:
## Describe the bug

After creation of a new `/lights` resource, `state.effect` is `null` instead of `"none"`.

## Steps to reproduce the behavior

```
$ ph get /lights/12
{
  "colorcapabilities": 0,
  "ctmax": 65279,
  "ctmin": 0,
  "etag": "bb0b792390b8e37858d2304b7cb2859e",
  "hascolor": true,
  "lastannounced": null,
  "lastseen": "2020-12-21T15:55Z",
  "manufacturername": "LIDL Livarno Lux",
  "modelid": "HG06106C",
  "name": "Extended color light 12",
  "state": {
    "alert": "none",
    "bri": 255,
    "colormode": "ct",
    "ct": 153,
    "effect": null,
    "hue": 0,
    "on": false,
    "reachable": false,
    "sat": 254,
    "xy": [
      0.7,
      0.3
    ]
  },
  "swversion": "0x00000050",
  "type": "Extended color light",
  "uniqueid": "84:2e:14:ff:fe:22:29:1a-01"
}
```
## Expected behavior

`state.effect` should be initialised to `"none"`.
## Environment

 - Host system: All
 - Running method: All
 - Firmware version: All
 - deCONZ version: 2.9.0, but probably introduced in v2.6.0, when changing `RStateEffect` back to `DataTypeString`
 - Device: All

## deCONZ Logs

n/a

## Additional context

n/a

Issue #4021, opened 2020-12-25
Cannot set light state for light when `state.reachable` is false


ebaauw:
## Describe the bug

The API returns api error 3, "resource, /lights/xx, not available" when trying to set a light state of a light that's unreachable (for which `state.reachable` is false).

## Steps to reproduce the behavior

```
$ ph get /lights/12
{
  "colorcapabilities": 0,
  "ctmax": 65279,
  "ctmin": 0,
  "etag": "bb0b792390b8e37858d2304b7cb2859e",
  "hascolor": true,
  "lastannounced": null,
  "lastseen": "2020-12-21T15:55Z",
  "manufacturername": "LIDL Livarno Lux",
  "modelid": "HG06106C",
  "name": "Extended color light 12",
  "state": {
    "alert": "none",
    "bri": 255,
    "colormode": "ct",
    "ct": 153,
    "effect": null,
    "hue": 0,
    "on": false,
    "reachable": false,
    "sat": 254,
    "xy": [
      0.7,
      0.3
    ]
  },
  "swversion": "0x00000050",
  "type": "Extended color light",
  "uniqueid": "84:2e:14:ff:fe:22:29:1a-01"
}
$ ph put -v /lights/12/state '{"on": true}'
ph put: warning: api error 3: resource, /lights/12, not available
[
  {
    "error": {
      "address": "/lights/12",
      "description": "resource, /lights/12, not available",
      "type": 3
    }
  }
]
```

## Expected behavior

This behaviour doesn't seem correct to me:
- The resource is available, so the error message is highly confusing.  At least it should read "not reachable".
- I think this should be a warning, not an error; i.e. the API plugin should still send the Zigbee commands to change the state, even though the light is marked unreachable.
## Environment

 - Host system: All
 - Running method: All
 - Firmware version: All
 - deCONZ version: 2.9.0, but probably all.
 - Device: all

## deCONZ Logs

n/a

## Additional context

n/a

ebaauw:
I think is't best to remove the check for reachability altogether.  The Hue bridge (`swversion` 1941132080, API version 1.41.0) accepts changing the state without any warning when the light is unreachable.

@manup, do you agree?

```
$ ph -H hue-3 get /lights/27
{
  "state": {
    "on": false,
    "bri": 77,
    "hue": 6144,
    "sat": 254,
    "effect": "none",
    "xy": [
      0.561,
      0.4042
    ],
    "ct": 499,
[...]

PR #4048, opened 2020-12-28
Minor bug fixes

ebaauw:
- Initialise `state.effect` to `"none"`, see #4022;
- ~~Skip test whether light is reachable on updating the light state, see #4021;~~
- Xmas lights strip: allow setting `effect to "none"` in combination with `bri`, `hue`, and/or `sat`, see #4038.


manup:
>Skip test whether light is reachable on updating the light state, see #4021;

Tricky, see my comment in the issue.

ebaauw:
> > Skip test whether light is reachable on updating the light state, see #4021;

> Tricky, see my comment in the issue.

OK reverted that commit, pending discussion.

Issue #3927, opened 2020-12-15
Danfoss Ally setting Heatsetpoint accepted but device does not change valve


easybeat:
## Describe the bug
 Setting the heatsetpoint on the device does not "trigger" the valve to open or close.

## Steps to reproduce the behavior
 I change the heatsetpoint to temperature 27 and check on device and it shows that the heatsetpoint has changed on the display but nothing happens. Local temperature on device is much lower. If I turn on the device and change heatsetpoint on the device and wait for 2-3 seconds the valve changes and you can hear the valve opening.
 
The problem is happening always when I change the heatsetpoint to a value which should change the valve position.

## Expected behavior

The valve should open or close by some percentage after the heatsetpoint was changed.

## Screenshots

This is the valve value when I change the heatsetpoint in ioBroker
![image](https://user-images.githubusercontent.com/37217887/102282707-4e079800-3f31-11eb-8d23-740762aeafbd.png)

This is the valve value when I change the heatsetpoint by 0.5 degrees on the device
![image](https://user-images.githubusercontent.com/37217887/102282824-8018fa00-3f31-11eb-9031-fb4565f702b2.png)

## Environment

![image](https://user-images.githubusercontent.com/37217887/102282916-b0f92f00-3f31-11eb-99da-6ce08eac6796.png)

 - Host system: NUC
 - Running method: Ubuntu Virtual Machine in Proxmox
 - Firmware version: (26680700)
 - deCONZ version: (2.07.00)
 - Device: ConBee II 
 - Do you use an USB extension cable: no
 - Is there any other USB or serial devices connected to the host system? If so: Which? yes. Aeotec Zwave Stick

easybeat:
my test was with a full value. I did set to 26 and then turned it manually down by 0.5 to 25.5

SwoopX:
Although you might not want to hear that, there's nothing we can do about it since we can't control how the device reacts on the valve by setting the heatsetpoint. This is more a case for Danfoss itself

But my TRV also doesn't jump to fully open if I set the highest temp, it takes a while...

easybeat:
@SwoopX After your input I did a test this morning and you are right. It opens but very slowly:
![image](https://user-images.githubusercontent.com/37217887/102456462-78e11180-4041-11eb-9f87-bbe375af7afd.png)

Like that it will take about 2 hours until the valve is open 100%!

I expect there are maybe different modes?

And yes this is not a bug from deconz.

Thanks for help.

[...]

Issue #3699, opened 2020-11-18
heating plan for Danfoss Ally fails


Einstein2150:
## Describe the bug
Danfoss Ally:
Setting a heating Plan is not possible. It is rejected by the API.

## Steps to reproduce the behavior
Setting/unsetting a heating plan but the setting/unsetting fails

## Expected behavior
Heating plan is set / unset

## Screenshots
![Bildschirmfoto 2020-11-18 um 07 19 31](https://user-images.githubusercontent.com/22019133/99492856-96ea2200-296e-11eb-9adb-e023d11dcd9f.png)
![Bildschirmfoto 2020-11-18 um 07 19 12](https://user-images.githubusercontent.com/22019133/99492868-99e51280-296e-11eb-856f-e19f97c9211e.png)

## Environment
RaspBee I
2.05.88 / 15.10.2020
Firmware
26390500
Raspberry Pi 4 - Raspbian
 

## deCONZ Logs
none

## Additional context
Set:
[18.11.2020, 07:03:47] [Hue] Temperaturregler Büro: 00 begin
[18.11.2020, 07:03:47] [Hue] Temperaturregler Büro: 13 schudule_on true
[18.11.2020, 07:03:47] [Hue] Temperaturregler Büro: F4 temp: 21.5°C, 15.0°C, 21.5°C
[18.11.2020, 07:03:47] [Hue] Temperaturregler Büro: FC time: 2020-11-18T07:03
[18.11.2020, 07:03:47] [Hue] Temperaturregler Büro: Mon 05:00 17:00
[18.11.2020, 07:03:47] [Hue] Temperaturregler Büro: Tue 05:00 17:00
[18.11.2020, 07:03:47] [Hue] Temperaturregler Büro: Wed 05:00 17:00
[18.11.2020, 07:03:47] [Hue] Temperaturregler Büro: Thu 05:00 17:00
[18.11.2020, 07:03:47] [Hue] Temperaturregler Büro: Fri 05:00 17:00
[18.11.2020, 07:03:47] [Hue] Temperaturregler Büro: Sat 05:00 17:00
[18.11.2020, 07:03:47] [Hue] Temperaturregler Büro: Sun 05:00 17:00
[18.11.2020, 07:03:47] [Hue] Temperaturregler Büro: Free 05:00 17:00
[18.11.2020, 07:03:47] [Hue] Temperaturregler Büro: 06 end
[...]

PR #4086, opened 2021-01-03
added additional danfoss ally attributes

eglia:
I added some additional attributes for the Danfoss Ally TRV based on
[Danfoss Ally Radiator Thermostat 1.08 Zigbee Cluster Specifications 16122020.pdf](https://raw.githubusercontent.com/SimpleUserHA/DanfosseTRV/main/Danfoss%20Ally%20Radiator%20Thermostat%201.08%20Zigbee%20Cluster%20Specifications%2016122020.pdf) (found [here](https://community.home-assistant.io/t/danfoss-ally-thermostat-firmware-update-ota-v1-08-open-window-external-temp-zha-deconz/261951))

PR #4122, opened 2021-01-07
Enhancements for Danfoss Ally

SwoopX:
- Updates on manufacturer specific cluster
- Change the command to set the heatpoint. so the device instantly reacts by opening/closing the valve (#3927)
- Enabled on-device schedules (as supported with firmware version 1.08) (#3699)


SwoopX:
Supercedes #4086

Issue #4130, opened 2021-01-07
Enhance Battery Reporting Time (specifically for Samsung Smartthings Devices)


jameshirka:
When using Deconz, Smartthings devices report battery too frequently causing drain on the battery and therefore reduced battery life.

The problem seems to be that if the configured reporting is more than 1.5 times the original value that reporting is reset back to default values.

This is to reopen issue:
https://github.com/dresden-elektronik/deconz-rest-plugin/issues/2908

It seems the issue has been tracked down in comment:
https://github.com/dresden-elektronik/deconz-rest-plugin/issues/2908#issuecomment-642698345

Issue #2908, opened 2020-06-09
Samsung SmartThings Multipurpose sensor (2018) reporting and battery life


NickWaterton:
I am running the latest FW (conbee II) and deconz server (2.5.77).

My SmartThings Multipurpose sensors (2018) are working, but I have noticed some odd things.

* The temperature, orientation/vibration and open/close status is reported every 5 minutes.
* Battery status is reported every 45 minutes (although this varies a lot)
* Battery status is reported 3 times (once for each of the sensors).
* battery state is dropping by about 1% per day.

At this rate, the sensor will be out of battery in about 3 months (I was hoping for longer battery life). I am using the sensors as door open/close sensors.

I don't mind the contact state being reported every 5 minutes (its actually useful to me), but is there a way to change the reporting interval of the other sensors? This may not be an issue if it's not the cause of the battery drain.

I have the websocket set to `websocketnotifyall` so I can see all the device updates (whether they change or not).

Why does the battery state get reported for each of the sensors? it's the same battery for all three.

Is there a way to extend the battery life? say by increasing the reporting interval? or disabling some of the sensors (temperature or orientation/vibration)?

I have changed the orientation threshold to 90, to prevent very frequent reports from the orientation sensor (due to door vibrations).

This is my high level log (with nothing happening to the door - ie idle state):
```
2020-06-09 08:18:21,346[     FrontDoor.DoorSensors]DEBUG|       sensor_update: Item frontDoorDoorContact_ST (ST) received update CLOSED
2020-06-09 08:23:24,854[     FrontDoor.DoorSensors]DEBUG|       sensor_update: Item frontDoorDoorContact_ST (ST) received update CLOSED
2020-06-09 08:28:28,366[     FrontDoor.DoorSensors]DEBUG|       sensor_update: Item frontDoorDoorContact_ST (ST) received update CLOSED
2020-06-09 08:33:31,862[     FrontDoor.DoorSensors]DEBUG|       sensor_update: Item frontDoorDoorContact_ST (ST) received update CLOSED
2020-06-09 08:38:35,373[     FrontDoor.DoorSensors]DEBUG|       sensor_update: Item frontDoorDoorContact_ST (ST) received update CLOSED
2020-06-09 08:43:38,879[     FrontDoor.DoorSensors]DEBUG|       sensor_update: Item frontDoorDoorContact_ST (ST) received update CLOSED
[...]

PR #4163, opened 2021-01-11
Fix Poll Control cluster configuration

manup:
This should also fix draining battery of Samsung SmartThings devices, which do MAC Data Polling every 5–10 seconds if not configured otherwise.

In general devices which contain an Poll Control cluster will be configured as well.

Issues: https://github.com/dresden-elektronik/deconz-rest-plugin/issues/2908 https://github.com/dresden-elektronik/deconz-rest-plugin/issues/4130

### DIFF
--- a/de_web_plugin.cpp
+++ b/de_web_plugin.cpp
@@ -4729,6 +4729,8 @@ void DeRestPluginPrivate::addSensorNode(const deCONZ::Node *node, const deCONZ::
         std::vector<Sensor>::iterator i = sensors.begin();
         std::vector<Sensor>::iterator end = sensors.end();
 
+        bool pollControlInitialized = false;
+
         for (; i != end; ++i)
         {
             if (i->address().ext() == node->address().ext())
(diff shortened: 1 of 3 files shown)


ebaauw:
Tried to change the _Long Poll Interval_ of my Samjin `multi` it to `14400` in the GUI, but get an _INVALID_VALUE_.  Tried `0x3840` as well, but no joy.  deCONZ v2.9.0.

ebaauw:
I'm able to set the _Short Poll Interval_.  After setting the _Check-In Interval_ to `28800`, I could set the _Long Poll Interval_.  It was `0` before (see screenshot in the issue); according to the ZCL spec, that special value should be except from the check _Check-In_ >= _Long Poll_ >= _Short Poll_.  Samjin firmware bug?  Could be an issue in the change as well?

manup:
I've have only tested with the Samjin button, I assumed they work the same.

With the PR the *Long Poll Interval* is set to 3600 and *Check-in Interval* to 14400. 
Did a small test in the GUI by setting 0x960 (2400) as *Poll Control Interval* works with success.

I'll see to get the Samjin multi and motion sensors, should have them in the office, for some tests.

PR #4196, opened 2021-01-15
Allow throttling of `lastseen` updates

SwoopX:
Currently, the period for updating `lastseen` is fixed to 60 secs and if the actual timestamp is older, the value is updated, emitting a websocket event for `/light` resources. By introducing `lightlastseeninterval` for the REST API gateway configuration, this value is now configurable between 1 and 655535 seconds, keeping the default of 60 seconds.

### DIFF
--- a/rest_configuration.cpp
+++ b/rest_configuration.cpp
@@ -2173,6 +2173,29 @@ int DeRestPluginPrivate::modifyConfig(const ApiRequest &req, ApiResponse &rsp)
         rspItem["success"] = rspItemState;
         rsp.list.append(rspItem);
     }
+    if (map.contains("lightlastseeninterval")) // optional
+    {
+        int lightLastSeen = map["lightlastseeninterval"].toInt(&ok);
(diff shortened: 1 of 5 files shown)


rest_configuration.cpp
```diff
@@ -2173,6 +2173,29 @@ int DeRestPluginPrivate::modifyConfig(const ApiRequest &req, ApiResponse &rsp)
         rspItem["success"] = rspItemState;
         rsp.list.append(rspItem);
     }
+    if (map.contains("lightlastseeninterval")) // optional
+    {
+        int lightLastSeen = map["lightlastseeninterval"].toInt(&ok);
+        if (!ok || lightLastSeen = 0 || lightLastSeen > 65535)
```
manup, line 2179: Shouldn't this also guard again negative values?

`if (!ok || lightLastSeen <= 0 || lightLastSeen > 65535)`

SwoopX: Ah, yes. Probably better.

PR #4209, opened 2021-01-17
Add initial support for Heiman remote controller RC-EF-3.0

SwoopX:
See #3828


NicolasBoulanger:
thanks for this release :) i can add device through app, but i cannot see it in devices section,
i have no event in home assistant, how can i get it  ?
thanks

SwoopX:
Phoscon does not support the device, so not surprising that you don't see anything.

How does the device look like in the REST API? Got any logs which contain the button presses?

Issue #3383, opened 2020-10-08
Adurosmart ERIA Smart Wireless Dimming Switch Remote 81825


stevenwfoley:
## Device
- Product name: ERIA Smart Wireless Dimming Switch Remote
- Manufacturer: ADUROLIGHT
- Model identifier: Adurolight_NCC
- Device type: Switch
https://adurosmart.com/product/eria-smart-wireless-dimming-switch-remote/

## Screenshots
![image](https://user-images.githubusercontent.com/37962790/95523612-131b3e00-099d-11eb-9bd4-c2f0e7e0e15f.png)
![image](https://user-images.githubusercontent.com/37962790/95524593-ca18b900-099f-11eb-82ab-05685daac2a8.png)

mdolnik:
FYI there is an closed / stale issue for this remote (#1730) which never got resolved.

I'm willing to make changes to the source code and test with my Eria remote if someone with more expertise can chime in and suggest what would need to be changed.

Mimiix:
Would like to continue in #1730 ! :)

Issue #1730, opened 2019-08-04
Eria Adurosmart Wireless Dimming Switch - Request Support


jbkuma:
![image](https://user-images.githubusercontent.com/20398640/62426237-12f42400-b6b0-11e9-96b4-f3f720c7d8cb.png)
![image](https://user-images.githubusercontent.com/20398640/62426240-21dad680-b6b0-11e9-8e92-397a25331817.png)

The Eria wireless dimmer has 4 buttons similar to the Hue wireless dimmer, but it's smaller and fits in a decora style wall plate.

I've tried various methods of pairing it, but I'm not getting any thing from the API  when press the buttons.

Device
Screenshots

skluthe:
I would very much like this. How were you able to pair it? Mine wont connect even in paring mode.

jbkuma:
> I would very much like this. How were you able to pair it? Mine wont connect even in paring mode.

There are two ways to pair it: press and hold the top button, press and hold the bottom one.  I can't remember which way I got it to pair.  I either used generic or Ikea as the profile.

skluthe:
Wasn't able to get it to pair with Deconz but it paired with home assistant
no problem (holding the bottom button).

mdolnik:
> I would very much like this. How were you able to pair it? Mine wont connect even in paring mode.

I was able to get this paired in deCONZ via phoscon, but it seems very hit and miss. I am fairly certain I chose `Other` in the `Add new switch` window and holding either the down or up button for 5 or more seconds.

This doesn't always work, I would say 1 out of 20 tries this will work for seemingly random reasons, and most times I would try to "factory reset" the remote by holding both the up and down buttons at the same time until the green LED flashes on the remote.

Once it is "paired":

- It does not get successfully added to Phoscon, it only shows up in deCONZ. Phoscon doesn't register that anything was ever found.
- Most of the remote's values in deCONZ matches the screenshots above, but I also have:
    - `Model Identifier` : Adurolight_NCC
    - `Date Code` : `20180402`
- Every button press will flash the green LED on the remote. When it's not paired, the LED does not light up on every button press.
- When deCONZ is in debugging mode I get the following output on any button press:

```
05:37:48:426 MAC Poll 0x02 0x4B71
05:37:48:426 verify 0x00158d000198d590 is child node after 809128 s
05:37:48:436 verify neighbor status: APP_SUCCESS (0x00)
05:37:48:436 MAC Poll 0x02 0x4B71
[...]

PR #4211, opened 2021-01-17
Add initial support for Eria Adurosmart Wireless Dimming Switch

SwoopX:
See #3383 and #1730

### DIFF
--- a/bindings.cpp
+++ b/bindings.cpp
@@ -2799,6 +2799,8 @@ bool DeRestPluginPrivate::checkSensorBindingsForAttributeReporting(Sensor *senso
         sensor->modelId() == QLatin1String("eTRV0100") ||
         // LIDL
         sensor->modelId() == QLatin1String("HG06323")
+        // Eira
+        sensor->modelId() == QLatin1String("Adurolight_NCC")
         )
     {
         deviceSupported = true;
@@ -3540,7 +3542,8 @@ bool DeRestPluginPrivate::checkSensorBindingsForClientClusters(Sensor *sensor)
     // Linkind 1 key Remote Control / ZS23000178
     // SR-ZG2835 Zigbee Rotary Switch
     else if (sensor->modelId().startsWith(QLatin1String("ZBT-DIMSwitch")) ||
-             sensor->modelId().startsWith(QLatin1String("ZG2835")))
+             sensor->modelId().startsWith(QLatin1String("ZG2835")) ||
+             sensor->modelId().startsWith(QLatin1String("Adurolight_NCC")))
     {
         clusters.push_back(ONOFF_CLUSTER_ID);
         clusters.push_back(LEVEL_CLUSTER_ID);
@@ -3670,6 +3673,7 @@ void DeRestPluginPrivate::checkSensorGroup(Sensor *sensor)
         sensor->modelId().startsWith(QLatin1String("TRADFRI wireless dimmer")) ||
         // sensor->modelId().startsWith(QLatin1String("SYMFONISK")) ||
         sensor->modelId().startsWith(QLatin1String("902010/23")) || // bitron remote
+        sensor->modelId().startsWith(QLatin1String("Adurolight_NCC")) || // Eria Adurosmart Wireless Dimming Switch
         sensor->modelId().startsWith(QLatin1String("WB01")) || // Sonoff SNZB-01
         sensor->modelId().startsWith(QLatin1String("WB-01")) || // Sonoff SNZB-01
         sensor->modelId().startsWith(QLatin1String("Bell")) || // Sage doorbell sensor
--- a/bindings.cpp
+++ b/bindings.cpp
@@ -2798,8 +2798,8 @@ bool DeRestPluginPrivate::checkSensorBindingsForAttributeReporting(Sensor *senso
         // Danfoss
         sensor->modelId() == QLatin1String("eTRV0100") ||
         // LIDL
-        sensor->modelId() == QLatin1String("HG06323")
-        // Eira
+        sensor->modelId() == QLatin1String("HG06323") ||
+        // Eria
         sensor->modelId() == QLatin1String("Adurolight_NCC")
         )
     {
--- a/de_web_plugin.cpp
+++ b/de_web_plugin.cpp
@@ -327,6 +327,7 @@ static const SupportedDevice supportedDevices[] = {
     { VENDOR_KONKE, "3AFE220103020000", konkeMacPrefix }, // Konke Kit Pro-BS Temp Humidity Sensor ver ???
     { VENDOR_EMBER, "3AFE130104020015", konkeMacPrefix }, // Konke Kit Pro-Door Entry Sensor
     { VENDOR_NONE, "RICI01", tiMacPrefix}, // LifeControl smart plug
+    { VENDOR_JENNIC, "Adurolight_NCC", jennicMacPrefix}, // Eria Adurosmart Wireless Dimming Switch
     { VENDOR_JENNIC, "VOC_Sensor", jennicMacPrefix}, // LifeControl Enviroment sensor
     { VENDOR_JENNIC, "SN10ZW", jennicMacPrefix }, // ORVIBO motion sensor
     { VENDOR_OSRAM_STACK, "SF20", heimanMacPrefix }, // ORVIBO SF20 smoke sensor


mdolnik:
I ran into a compiler error with this branch/commit:

```
bindings.cpp: In member function ‘bool DeRestPluginPrivate::checkSensorBindingsForAttributeReporting(Sensor*)’:
bindings.cpp:2803:9: error: expected ‘)’ before ‘sensor’
         sensor->modelId() == QLatin1String("Adurolight_NCC")
         ^~~~~~
```

This is fixed by adding and OR operator to the line above after `HG06323`:

eg
```
// LIDL
sensor->modelId() == QLatin1String("HG06323") ||
// Eira
sensor->modelId() == QLatin1String("Adurolight_NCC")
```

mdolnik:
After successfully compiling I am still having the same issues as [my previous comment in issue ](https://github.com/dresden-elektronik/deconz-rest-plugin/issues/1730#issuecomment-706621272) where:

- I can get it connected to deCONZ but not Phoscon
- Every button press will flash the green LED on the remote. When it's not paired, the LED does not light up on every button press.
- When deCONZ is in debugging mode I get the following output on any button press where `0x5517` is the ID of the Eria switch:

```
22:08:54:556 CTRL skip polling while permit join is set
22:08:54:563 CTRL skip polling while permit join is set
22:08:54:574 APS-DATA.indication from child 0x5517
22:08:54:673 CTRL skip polling while permit join is set
22:08:54:682 CTRL skip polling while permit join is set
```

With the code that was recently added for this switch, is there any specific brand/model that should be selected in Phoscon? I have tried:

- Other
- Ikea TRÅDFRI remote control
- Phillips Hue dimmer switch
- Phillips Tap

Not sure if it would help, but for either of these brands/models, when I am in the pairing mode in Phoscon, I will get the something similar to the following output in deCONZ debugging:

```
22:26:53:681 APS-DATA.indication from child 0x5517
22:26:53:682 void deCONZ::zmNode::setFetched(deCONZ::RequestId, bool) fetched item: 7, node: 0x5517
22:26:53:682 FP indication 0x0000 / 0x8005 (0x00158D000198D590 / 0x5517)
22:26:53:682                       ...     (0x00158D000198D590 / 0x5517)
22:26:53:682 ZDP indication search sensors 0x00158D000198D590 (0x5517) cluster 0x8005
22:26:53:683 ZDP indication search sensors 0x00158D000198D590 (0x5517) clear timeout on cluster 0x8005
22:26:53:732 CTRL skip polling while permit join is set
```

I am very unfamiliar with this code so I don't have any idea what most of it does.

Is there anything specific I should try? Including modifying any certain parts of the code?

SwoopX:
Thanks for checking. I meanwhile corrected the small error in the code.

How does the switch look like in deconz GUI after you've paired it? Has a corresponding sensor resource been created in the REST API? If the device got paired correctly and everything worked out as expected, you should see some `[INFO] - XYZ` messages in the logs upon each button press. Right now, those should indicate that there's no button map found (this is the part where you need to provide me some input to create it).

You'll not see the device in Phoscon unless explicitly supported there, which must be requested in the Phoscon beta repository.

mdolnik:
Well I somehow managed to bork my Ubuntu installation by installing a wrong driver so I currently cannot use any mouse or keyboard, and I never set up a means to remote into it. This means I'll have quite the trouble being of any help right now until I un-fudge my OS.

At this point I'm a little hesitant to try messing around with this on my Raspberry Pi in case I mess up my setup, so I'll try my best from memory...

> How does the switch look like in deconz GUI after you've paired it?

A couple things I remember:

- I was able to "pair" it without Phoscon listening and it would show up in deCONZ and:
    - Only one circle appears on the right side of the deCONZ node
    - The Eria remote lights up green on each button press, but deCONZ does not seem to register any button presses.
- When I "pair" it with Phoscon listening:
    - Phoscon does not register that anything was found, but deCONZ does.
    - Both circles appear on the right side of the deCONZ node.
    - The Eria remote lights up green on each button press, and the circle on the left side deCONZ will flash green on every button press.

> Has a corresponding sensor resource been created in the REST API?

How would I be able to tell is one was created? Regardless I would be going off memory at this point, but I can see if anything rings a bell.

> If the device got paired correctly and everything worked out as expected, you should see some [INFO] - XYZ messages in the logs upon each button press.

Does this answer my previous question? Is this referring to the REST API?

I don't remember anything in the CLI output specifically saying `[INFO]` but I do recall some logs with square brackets, but I can't recall any patterns that occurred on EVERY button press except for what I mentioned above. I could be wrong though.

Would these `[INFO]` logs have included the ID (in my case `0x8005`)? I was only looking through the logs for anything in/around that ID.

> You'll not see the device in Phoscon unless explicitly supported there, which must be requested in the Phoscon beta repository.

What would be the steps for this? Alternatively would it be easier to classify it as a `Phillips Hue dimmer switch` since the buttons are the same and it looks similar?

Sorry I'm not much help at the moment, I'll try to see if I can get my Ubuntu working this weekend, and if not, I'll try debugging on my Pi.

mdolnik:
So I managed to get my Ubuntu up and running again and did a bit of investigating...

> Has a corresponding sensor resource been created in the REST API?

I think I can now confidently that no sensor has been created.

While debugging with print statements I noticed whenever any buttons are pressed, the code ends up getting to  `line 1021 in DeRestPluginPrivate::apsdeDataIndication() in de_web_plugin.cpp` and the line `sensorNode = getSensorNodeForAddress(ind.srcAddress());` seems to not return a `sensorNode` as the following if statement is never entered.

I am also not getting the `[INFO] - No button map for...` message nor any `[INFO] - ` messages. 

What would be involved in getting the sensor resource / sensor node set up?

SwoopX:
@mdolnik Sorry, I guess this one slipped through. The reason is that the manufacturer code and mac prefix are identical to the Xiaomi's and therefore, neither model ID nor manufacturer name are querried. The solution for this should be starting sensor search in Phoscon and then reading the basic cluster in deconz GUI while the device is awake (just press a button on the device and press the read button 1-2 seconds afterwards).

mdolnik:
@SwoopX 

Juts an FYI the merged branch above does not resolve this issue, which I'm sure is already known since nothing was changed since 65acc33 on Jan 19th, but I just wanted to confirm this is still an issue.

I found the culprit as to why the sensorNode is not being created for this switch:

https://github.com/dresden-elektronik/deconz-rest-plugin/blob/08e1181e8e841669b089c4770d6c9600f70f25fb/de_web_plugin.cpp#L5641-L5643

This code will ignore all nodes with `VENDOR_JENNIC`

One work-around that works for me is to change the logic in an if statement above:

https://github.com/dresden-elektronik/deconz-rest-plugin/blob/08e1181e8e841669b089c4770d6c9600f70f25fb/de_web_plugin.cpp#L5626-L5629

To:
```
if (modelId == QLatin1String("ZYCT-202") ||
    modelId == QLatin1String
    // Eria Adurosmart Wireless Dimming Switch.
    modelId == QLatin1String("Adurolight_NCC"))
{
    fpSwitch.outClusters.push_back(ci->id());
}
```

Which essentially whitelists this specific model (which there are no other existing ) before it is backlisted for its manufacturer code.

After this change, I am now getting the following `[INFO] - No button map for` outputs:
- On Button:
    - `[INFO] - No button map for: Adurolight_NCC, unicast to: 0x0000, endpoint: 0x01, cluster: 0xFCCC, command: 0x00, payload: 000000, zclSeq: 174`
- Up Button:
    - `[INFO] - No button map for: Adurolight_NCC, unicast to: 0x0000, endpoint: 0x01, cluster: 0xFCCC, command: 0x00, payload: 000100, zclSeq: 177`
- Down Button:
    - `[INFO] - No button map for: Adurolight_NCC, unicast to: 0x0000, endpoint: 0x01, cluster: 0xFCCC, command: 0x00, payload: 000200, zclSeq: 180`
- Off Button:
    - `[INFO] - No button map for: Adurolight_NCC, unicast to: 0x0000, endpoint: 0x01, cluster: 0xFCCC, command: 0x00, payload: 000300, zclSeq: 183`

But I'm not sure how to proceed or what to make of cluster `0xFCCC` which is labeled `unknown` in the deCONZ GUI.

mdolnik:
Also regarding the manufacturer codes, is there a reason that this branch used `VENDOR_JENNIC` and did not add a new `VENDOR_ADUROLIGHT`?

The constant definitions refer to the wireshark github page:
https://github.com/dresden-elektronik/deconz-rest-plugin/blob/08e1181e8e841669b089c4770d6c9600f70f25fb/de_web_plugin_private.h#L304-L308

And on that page, it has a definition for `VENDOR_ADUROLIGHT`
https://github.com/dresden-elektronik/deconz-rest-plugin/blob/08e1181e8e841669b089c4770d6c9600f70f25fb/de_web_plugin_private.h#L369

mdolnik:
Refer to the new Pull Request #4495

Issue #4189, opened 2021-01-13
GamaBit Ltd. Window/Door Sensor


Kristian8606:
## Device

- Product name: Contact Switch.
- Manufacturer: GamaBit Ltd.
- Model identifier: GMB-HAS-DW-B01.
- Device type : Please remove all unrelated device types. 
- Sensor
 

## Screenshots
![Screenshot 2021-01-14 at 0 49 59](https://user-images.githubusercontent.com/34716572/104525721-799da100-5609-11eb-93e0-2b6a0f37b825.png)
![Screenshot 2021-01-14 at 0 50 19](https://user-images.githubusercontent.com/34716572/104525725-7bfffb00-5609-11eb-858a-1b0d96e1d320.png)
![Screenshot 2021-01-14 at 0 51 45](https://user-images.githubusercontent.com/34716572/104525728-7c989180-5609-11eb-95f0-a9bee318dc08.png)

SwoopX:
That would be indeed good if you could test it upfront.

Kristian8606:
Not yet visible in Phoscon

SwoopX:
That's because deconz is not Phoscon. It's the REST API what matters here.

For visibility in Phoscon, you'd need to request it in their repository.

Kristian8606:
I tried `GET API request ip: 80/api/67CE474F7D/sensors` but I don't see it, i see all the other sensors i have.

SwoopX:
Have you re-paired it?

Kristian8606:
Yes I tried twice, the different I see is that there is now a line in the deconz GUI connected to two bulbs as well as a connection to the conbee

SwoopX:
How does the node look like in deconz GUI.

Kristian8606:
I tried a third time. I do the search through Phoscon.
![Screenshot 2021-01-18 at 13 38 44](https://user-images.githubusercontent.com/34716572/104910860-a5ff4780-5992-11eb-8d68-3de8a222289d.png)

Kristian8606:
How could I find it in a database to make sure it's there?

Kristian8606:
GMB-HAS-WD-B01 I noticed a discrepancy in the Model identifier
![Screenshot 2021-01-18 at 13 46 55](https://user-images.githubusercontent.com/34716572/104911557-b6fc8880-5993-11eb-9609-a01b2e3e54a4.png)

Kristian8606:
[...]

PR #4215, opened 2021-01-17
Add initial support for GamaBit GMB-HAS-WD-B01

SwoopX:
See #4189


manup:
Needs sync with master, otherwise ready to merge.

PR #4217, opened 2021-01-18
Fix possible segfault by wrong printf argument type

sebres:
Proposed fix avoid possible segfault by wrong type of argument supplied to DBG_Printf, expected int (got wide-int?),
see https://github.com/dresden-elektronik/deconz-rest-plugin/issues/3611#issuecomment-761917726 for details

I'm pretty sure it is not the single place where it is used in same way, but have no time at the moment to check whole codebase.
FWIW, it is doubtful (at least not advisable) to use printf without check of argument types, let alone the arguments declared as auto-variable supplied without cast, etc.

### DIFF
--- a/de_web_plugin.cpp
+++ b/de_web_plugin.cpp
@@ -4641,7 +4641,7 @@ void DeRestPluginPrivate::checkSensorButtonEvent(Sensor *sensor, const deCONZ::A
 
                             if (dt > 0 && dt < 500)
                             {
-                                DBG_Printf(DBG_INFO, "[INFO] - Button %u %s, discard too fast event (dt = %d) %s\n", buttonMap.button, qPrintable(cmd), (int)dt, qPrintable(sensor->modelId()));
+                                DBG_Printf(DBG_INFO, "[INFO] - Button %u %s, discard too fast event (dt = %d) %s\n", buttonMap.button, qPrintable(cmd), static_cast<int>(dt), qPrintable(sensor->modelId()));
                                 break;
                             }
                         }
--- a/de_web_plugin.cpp
+++ b/de_web_plugin.cpp
@@ -4641,7 +4641,7 @@ void DeRestPluginPrivate::checkSensorButtonEvent(Sensor *sensor, const deCONZ::A
 
                             if (dt > 0 && dt < 500)
                             {
-                                DBG_Printf(DBG_INFO, "[INFO] - Button %u %s, discard too fast event (dt = %d) %s\n", buttonMap.button, qPrintable(cmd), dt, qPrintable(sensor->modelId()));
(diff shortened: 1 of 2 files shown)


de_web_plugin.cpp
```diff
@@ -4641,7 +4641,7 @@ void DeRestPluginPrivate::checkSensorButtonEvent(Sensor *sensor, const deCONZ::A
 
                             if (dt > 0 && dt < 500)
                             {
-                                DBG_Printf(DBG_INFO, "[INFO] - Button %u %s, discard too fast event (dt = %d) %s\n", buttonMap.button, qPrintable(cmd), dt, qPrintable(sensor->modelId()));
+                                DBG_Printf(DBG_INFO, "[INFO] - Button %u %s, discard too fast event (dt = %d) %s\n", buttonMap.button, qPrintable(cmd), (int)dt, qPrintable(sensor->modelId()));
```
manup, line 4644: Good catch.

Please always use C++11 casts instead old C casts, as in future all old C style casts in the code base will be replaced by the new ones for various reasons. For example:

```
DBG_Printf(DBG_INFO, "[INFO] - Button %u %s, discard too fast event (dt = %d) %s\n",
    buttonMap.button, qPrintable(cmd), static_cast<int>(dt), qPrintable(sensor->modelId()));
```

However for this PR I'd suggest the correct printf modifier `%lld` and avoid casting at all.

```
DBG_Printf(DBG_INFO, "[INFO] - Button %u %s, discard too fast event (dt = %lld) %s\n",
    buttonMap.button, qPrintable(cmd), dt, qPrintable(sensor->modelId()));
```

manup (changes requested):
Added a few notes, can you please have a look.

de_web_plugin.cpp
sebres: The main reasons against classic C casts are almost to neglect here since it is not an object but built-in type int (so the checking of compatibility does no matter) and a search and spot ability argument does not cut any ice here either (due to printf formatting rules, so an intention to change it implies check of argument type and vice versa).

But OK, I'll change it to the static-cast.

As for suggestion to change to `%lld`, it is wrong in the same way as `%d` was previously, because of an `auto` declaration and so depending on result of function (but platform or whatever else can cause prototype change, exactly as it happening by switch somewhere between 2.7 and 2.9). So without to cast or changing of variable type declaration it is not good idea (at least unless DBG_Printf is unable to generate an error/warning by compilation for the wrong type, similar to `printf` at al). Let alone such a large value is totally unneeded here (since it is difference equals about maximal thousands ms, and even a switch of granularity to µs would yet match an integer range entirely).

manup (approved):

Issue #3308, opened 2020-09-24
Sengled Smart Light Switch E1E-G7F


stevenwfoley:
## Device
- Product name: Sengled Smart Light Switch
- Manufacturer: sengled
- Model identifier: E1E-G7F
- Device type : Switch

## Screenshots
![image](https://user-images.githubusercontent.com/37962790/94089018-fcdb8280-fddf-11ea-9046-0ace058375a7.png)
![image](https://user-images.githubusercontent.com/37962790/94089082-298f9a00-fde0-11ea-81ba-391e50b6bd30.png)
![image](https://user-images.githubusercontent.com/37962790/94089136-50e66700-fde0-11ea-9fbb-236c04f49c5d.png)
![image](https://user-images.githubusercontent.com/37962790/94089239-a6227880-fde0-11ea-8301-7b9371caf7e2.png)

SwoopX:
@stevenwfoley We would still need the button events to create a fitting button map. See here for reference
#3397

stevenwfoley:
Is this the data you're looking for?
Top (ON) button - No button map for: E1E-G7F endpoint: 0x01 cluster: 0xFC10 command: 0x00 payload[0]: 001
Second from top (Dim Up) button - No button map for: E1E-G7F endpoint: 0x01 cluster: 0xFC10 command: 0x00 payload[0]: 002
Third from top (Dim Down) buton - No button map for: E1E-G7F endpoint: 0x01 cluster: 0xFC10 command: 0x00 payload[0]: 003
Bottom (OFF) button - No button map for: E1E-G7F endpoint: 0x01 cluster: 0xFC10 command: 0x00 payload[0]: 004

SwoopX:
Hm, interesting. I would have rather expected to see comparable messages from on/off and level control cluster instead of FC10. On/off and level control are typically used to control regular lights and also should allow to used zigbee groups. Maybe the FCC10 commands are used to communicate differently with Sengled lights...
Could you please double check?

stevenwfoley:
I don't think the switch supports the zigbee group cluster as its greyed out in the deconz gui. I reran the debug and received the same results. Is there something else you want to test or provide?

SwoopX:
It just feels a bit strange that this switch seems to behave significantly differently than all others, that's what I'm trying to understand here. Could also be that this is something device specific, which can be configured via the unknown FC clusters. However, a map based on that is easily doable...

[...]

PR #4264, opened 2021-01-23
Enhancements for Sengled E1E-G7F

SwoopX:
- Add Sengled E1E-G7F button map (#3308)


manup:
I'd propose to rename this PR to "Add support for Sengled E1E-G7F"
So stays a nice device support PR :)

PR #4275, opened 2021-01-25
Add support for Sengled E1E-G7F

MoshiBin:
See #3308

### DIFF
--- a/button_maps.json
+++ b/button_maps.json
@@ -51,7 +51,8 @@
         "WINDOW_COVERING": 258,
         "COLOR_CONTROL": 768,
         "IAS_ZONE": 1280,
-        "IAS_ACE": 1281
+        "IAS_ACE": 1281,
+        "SENGLED": 64528
 	},
 	"commands": {
         "DOOR_LOCK": {
@@ -112,6 +113,9 @@
         },
         "MULTISTATE_INPUT": {
             "ATTRIBUTE_REPORT": 10
+        },
+        "SENGLED": {
+            "CLICK": 0
         }
 	},
 	"maps": {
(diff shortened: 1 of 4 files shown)


SwoopX:
See #4264 and #4235 from a couple of days ago 😄

MoshiBin:
Yeah, I was definitely over-eager 😅 
At least I got to poke around the code and understand a little more about zigbee and deCONZ

MoshiBin:
By the way, the above PRs do not fix `de_web_plugin.cpp` - the Sengled switch sends the button number as the payload. Without [this part](https://github.com/dresden-elektronik/deconz-rest-plugin/pull/4275/files#diff-d0aaee66bc2f60fdd243c0e5b86aebca01fa3eb36c7ec39578d5beeca47078abR4628-R4634) all buttons were being identified as the top button.

SwoopX:
Ah, yeah, you're right. Took me a little while to get what you mean. Since it's a new cluster to consider, the already available code doesnt cater for that. Good catch!

kyletpugh:
Was this fully merged? Has the pressing any button results in only the top button event been resolved? The switch works but only generates the event for the first button single press no matter what you press.

Mimiix:
@kyletpugh Depending on your version, it is.
Check Swoops PR :)

kyletpugh:
Which version has fully working behavior? I have version 2.09.03. I am using the beta channel, looks like this change is there but does not produce the correct results.

SwoopX:
Please open a dedicated issue for this and provide logs with info level 2. The latest stable and beta contain the required code.

kyletpugh:
The above comments say that all buttons are recognized as the first single press because something else needed changed. The PR is closed with unmerged changes. I am confirming that every button results in being mapped to first button single press.

PR #4282, opened 2021-01-27
Add support for Lidl/Silvercrest door bell (TS0211)

jurriaan:
Adds support for the new door bell device, which shows up as a IAS Zone 'Contact Switch' device.

After these changes deconz emits the correct events, but it doesn't show up in Phoscon, is that expected behavior?

### DIFF
--- a/bindings.cpp
+++ b/bindings.cpp
@@ -2622,6 +2622,7 @@ bool DeRestPluginPrivate::checkSensorBindingsForAttributeReporting(Sensor *senso
         // lidl / SilverCrest
         sensor->modelId()== QLatin1String("TY0203") ||  // Door sensor
         sensor->modelId() == QLatin1String("TY0202") || // Motion Sensor
+        sensor->modelId() == QLatin1String("TS0211") || // Door bell
         // Konke
         sensor->modelId() == QLatin1String("3AFE140103020000") ||
         sensor->modelId() == QLatin1String("3AFE130104020015") ||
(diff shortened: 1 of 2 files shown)


Jasperste (approved):

Mimiix:
Linked the issue on PR.

jurriaan:
@manup Is there anything I can do to get this merged?

I don't really like that we need to allowlist all the separate mac prefixes, can we improve that somehow?

Mimiix:
@jurriaan Feel free to compile yourself  :) Having the whitelist enables Dresden-Elektronik to keep control and have some quality assurance so that every device works. Otherwise, we would be getting flooded with "Device doesn't work" requests. So i do understand your frustration, but at this point they are in no position to simplyfy  it afaik.

manup:
There are quite some device requests which didn't made it into v2.9.2 since releases after x.y.0 are meant to prepare for stable release. I think this one here is safe to include in the next v2.10.0 beta release where new devices get merged.

jurriaan:
Thanks for the response, I'm running deconz with this change for almost two weeks already, it's working perfectly. 

Was just asking this in case you want me to improve something before this can be merged :)

manup:
Can you please sync the PR with latest master for merging.

In a followup PR it would be good to replace the model id (TS0211) with a unique product id, see https://github.com/dresden-elektronik/deconz-rest-plugin/pull/4377 for an example how it works for other Tuya devices.

jurriaan:
Rebased on master, will do a followup to fix that :)

jurriaan:
@manup I just looked into it and was a bit confused, I noticed that in de_web_plugin.cpp there is a lidlDevice list as well. 

See https://github.com/dresden-elektronik/deconz-rest-plugin/blob/master/de_web_plugin.cpp#L463-L479

How does this work exactly. Lidl devices are Tuya devices (mostly?) should those tables be integrated into one? They seem to serve a similar purpose.

MattL0:
Is that only a button? https://zigbee.blakadder.com/Lidl_HG06668.html

Not sure I got the right one here.

jurriaan:
Yes, that's the one

knutbrecht:
ist there any solution for that?
I got the exact same behaviour - device is not recognized as a butten. It is recognized as a sensor (why?) but not listed in Phoscon.
It is listed in deconz objects (iobroker in my case) but no correct behavior when buuton is pushed. There is someting happening but the "button" object is not changing.

Has anyone succeeded integrating the doorbell correctly?

mohdshaqdih:
Hello 
Do you have edge drive for Lidl doorbell 
Thanks

Issue #4287, opened 2021-01-28
Hue Dimmer Switch (2021 model)


gvssr:
## Device

- Product name: Philips Hue Dimmer switch
- Manufacturer: Signify Netherlands B.V.
- Model identifier: RWL022
- Device type :
  - Switch

## Screenshots
![Basic cluster](https://user-images.githubusercontent.com/61377476/106145373-c2ab3480-6175-11eb-8160-a75d34ee5add.png)
![Node info](https://user-images.githubusercontent.com/61377476/106145387-c6d75200-6175-11eb-8252-36974fa2a3ea.png)
![Power cluster](https://user-images.githubusercontent.com/61377476/106145394-c9d24280-6175-11eb-92d9-22be6ef8f395.png)

ebaauw:
I'll have a look this weekend.  Should be fairly similar to the Hue smart button, with the two endpoints of the older model combined into one endpoint.

ebaauw:
Ha, looking at the code, the ZHASwitch resource is created, and then removed.  There's some archaic code to fix the original Hue dimmer switch being exposed wrongly (in a version from 2017 or older), forcing the endpoint to 0x02.
https://github.com/dresden-elektronik/deconz-rest-plugin/blob/f46e3321715f755a057c19992e51ce9cf6f01e35/de_web_plugin.cpp#L5649-L5657

@manup I propose to remove this block altogether, would you agree?  Alternatively, I could add a check for `RWL022`.

ebaauw:
Couldn't resist and got me one.  And a good thing I did, too, because it doesn't work.  On pairing, two identical resources are created, both for non-existing endpoint 0x02:
```
{
  "14": {
    "config": {
      "battery": null,
      "group": "18",
      "on": true,
      "reachable": true
    },
    "ep": 2,
    "etag": "9c2ef3035c5520e72ee1c8751a98a058",
    "lastseen": "2021-02-05T18:03Z",
    "manufacturername": "Philips",
    "mode": 1,
    "modelid": "RWL022",
    "name": "RWL022 14",
    "state": {
      "buttonevent": null,
      "eventduration": null,
      "lastupdated": "none"
    },
    "swversion": "2.42.0_h1F5E860",
    "type": "ZHASwitch",
    "uniqueid": "00:17:88:01:09:27:59:e1-02-fc00"
  },
  "15": {
    "config": {
      "battery": null,
      "group": "19",
      "on": true,
      "reachable": true
[...]

Issue #4248, opened 2021-01-21
Lidl Livarno Lux Leuchten


Schnuecks:
## Device

- Product name: LIVARNO LUX® Stimmungsleuchte
- Link: https://www.lidl.de/de/livarno-lux-stimmungsleuchte-mit-akku-und-lichtfarbensteuerung-rgb-led-zigbee-smart-home/p365242
- Manufacturer: _TZ3000_9cpuaca6
- Model identifier: TS0505A
- Device type : Please remove all unrelated device types. 
  - Light

- Product name: LIVARNO LUX® Lichtleiste
- Link: https://www.lidl.de/de/livarno-lux-lichtleiste-mit-lichtfarbensteuerung/p365241
- Manufacturer: _TZ3000_gek6snaj
- Model identifier: TS0505A
- Device type : Please remove all unrelated device types. 
  - Light

## Screenshots

  Required screenshots:
Node Info
![nodeinfo](https://user-images.githubusercontent.com/27301329/105388696-42398080-5c17-11eb-9750-f0ff1157d8b7.PNG)

Node List
![nodelist](https://user-images.githubusercontent.com/27301329/105388721-4a91bb80-5c17-11eb-8f10-1fdae6416654.PNG)

Coordinator ClusterInfo
![coordinator_clusterinfo](https://user-images.githubusercontent.com/27301329/105388750-52e9f680-5c17-11eb-905e-da23a71461f6.PNG)

Lightbar ClusterInfo
![lidl_livarno_lux_lightbar_clusterinfo](https://user-images.githubusercontent.com/27301329/105388777-5a110480-5c17-11eb-894f-524ad7b83821.PNG)

Lightbox1 ClusterInfo
![lidl_livarno_lux_lightbox1_clusterinfo](https://user-images.githubusercontent.com/27301329/105388813-64330300-5c17-11eb-91a8-bef54b1a4c98.PNG)

Lightbox2 ClusterInfo
![lidl_livarno_lux_lightbox2_clusterinfo](https://user-images.githubusercontent.com/27301329/105389509-3601f300-5c18-11eb-9064-8c8679d7aac2.PNG)

ebaauw:
Looks like they're already exposed by the REST API Plugin?  If not, please provide all screenshots, including, for each of the lights, the node with endpoints and clusters, and the _Node Info_ panel.

[...]

PR #4302, opened 2021-01-29
Hue dimmer switch (2021), Livarno Lux lights

ebaauw:
- Add support for Hue dimmer switch (2021 model, RWL022), see #4287;
- Add _Manufacturer Name_/_Model Identifier_ translations for LIDL Livarno Lux light bar (Lichtleiste) and mood light (Stimmingsleuchte), see #4248;
- Add _Manufacturer Name_/_Model Identifier_ translations for LIDL Xmas lightstrip.

### DIFF
--- a/de_web_plugin.cpp
+++ b/de_web_plugin.cpp
@@ -5655,7 +5655,10 @@ void DeRestPluginPrivate::addSensorNode(const deCONZ::Node *node, const deCONZ::
                 {
                     sensor = nullptr;
                 }
-                fpSwitch.endpoint = 2;
+                if (modelId != QLatin1String("RWL022")) // new model with one endpoint
+                {
+                    fpSwitch.endpoint = 2;
+                }
             }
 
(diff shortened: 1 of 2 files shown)


Thomas-Vos:
Thanks again Erik for delaying the model id for the LIDL Xmas lightstrip! I added the LIDL model IDs in Hue Essentials a few months ago, so it is alright to add this one now.

ebaauw:
@manup, I double-checked that both new model and the old model pair OK and work with this PR.  I think it's good to go.

ebaauw:
OK, this was self-inflicted.  Now both old and new models pair OK, getting only one resource.

manup:
> OK, this was self-inflicted. Now both old and new models pair OK, getting only one resource.

Happens, I didn't know the new model was already available.
Will be part of v2.9.2. :)

ebaauw:
Still an issue: when restored from database, endpoint 2 is used (as for the original model), causing the `uniqueid` to change to `-02-fc00` and button events not to be registered.

PR #4319, opened 2021-02-01
Optional exposure of coordinator as light resource, default false

SwoopX:


manup:
We should discuss this in the next developer meeting and also take the opinions from REST API clients into account.

Personally I'd like to deprecate and remove the coordinator as light all together, since it causes more confusion as any benefits (at least which I know of). Imho we shouldn't have extra code to support coordinator as light and also clients shouldn't need that, and as in case of the Phoscon App extra code to hide it. All that is relevant to the coordinator should be kept in `/configuration`.

SwoopX:
Sure, let's discuss this. This was meant as a compromise which could make everybody kinda happy. I recall exposing it as light cause quite some irritation so this should be the walk back to the time when it wasn't while preserving the flexibility to put it whatever you want.

ebaauw:
Please explain to me why we do want to expose repeaters through the REST API, but not the coordinator?  And what about an additional RaspBee / Conbee configured as router?

Zehir:
On my side the main issue with it is that he share the same `uniqueid` between the `light` and `sensor`.

ebaauw:
`uniqueid` is only unique for resources corresponding to Zigbee devices.  The builtin Daylight sensor isn't.

Zehir:
> `uniqueid` is only unique for resources corresponding to Zigbee devices. The builtin Daylight sensor isn't. Neither are groups, nor CLIP sensors.

yes but that id should be unique, it's in the name. Or rename it to `notalwaysuniqueid` ;)

ebaauw:
Or it should only have been exposed on resources related to Zigbee devices.

Mimiix:
From a community perspective it never was made perfectly clear why we wanted this. Nobody seems to use it or is familiar with the reason to have it exposed.

That being said, I rather see it gone.

manup:
Imho we should figure out how we can cleanup `/lights` in future. The repeater addition is my fault and a hack which I now regret :)

Coordinator (or router) and repeater devices neither fit into `/lights` or `/sensors` — but are already exposed via the experimental `/devices` endpoint. For the repeater the main reason to expose them via `/lights` was to make them visible for the API client.

The following  is the output with current `master` branch
```
GET devices/00:0d:6f:ff:fe:9f:7f:52

{
    "modelid": "TRADFRI signal repeater",
    "sub": [
        {
            "lastannounced": "2021-04-20T10:42:04Z",
            "lastseen": "2021-04-24T14:01Z",
            "name": "Range extender",
            "state": {
                "alert": "none",
                "reachable": false
            },
            "type": "Range extender",
            "uniqueid": "00:0d:6f:ff:fe:9f:7f:52-01"
        }
    ],
    "swversion": "2.2.005",
    "uniqueid": "00:0d:6f:ff:fe:9f:7f:52"
}
```

Here is the same request in the `device_descriptions` current branch, the output is a bit more rich due the discussion about timestamps earlier in Discord.

```
GET devices/00:0d:6f:ff:fe:9f:7f:52
{
    "lastannounced": "2021-04-20T10:42:04Z",
    "lastseen": "2021-04-24T14:01Z",
    "modelid": "TRADFRI signal repeater",
    "name": "Range extender",
    "subdevices": [
        {
            "id": "64",
            "state": {
                "alert": {
                    "lastupdated": "2021-04-24T14:01:00Z",
                    "value": null
                },
                "reachable": {
                    "lastupdated": "2021-04-24T14:01:00Z",
                    "value": false
                }
            },
            "type": "Range extender",
            "uniqueid": "00:0d:6f:ff:fe:9f:7f:52-01"
        }
    ],
    "swversion": "2.2.005",
    "uniqueid": "00:0d:6f:ff:fe:9f:7f:52"
}
```

So if we can figure out and decide on a generic layout for the `/device` endpoint we could move at least simpler devices to that endpoint. Would that be an approach?

Zehir:
At the end does all devices will be in `/device` ?

ebaauw:
Sounds like a plan: deprecate and plan to remove all `/lights` resources for endpoints that aren’t “lightoids”, leaving only lights, plugs, and wired switches (cf. the Hue bridge).  Expose other devices (window covering, sirens, repeaters, fans, vents, locks, ...) only under `/devices`.  Would need to expand the `/groups` endpoint to allow for subdevices in groups.  Would also solve the weirdness with the Hampton Bay, where the fan and light are on the same endpoint (but different clusters).

`lastupdated` doesn’t make sense on `alert`: it is write-only.

manup:
*Lightoids* is a very fine description :smile:  This looks like a good direction, in combination with self describing resources this can be really powerful. I think how `alert` or `ResourceItems` in general represent them self in the REST API can also be accomplished in a generic manner since items are described via JSON files. We can add attributes to specify wherever they should show `lastupdated` or are visible at all which is represented by the `public` attribute:

[items/state_alert_item.json](https://github.com/manup/deconz-rest-plugin/blob/device_descriptions/devices/generic/items/state_alert_item.json)
```
{
	"id": "state/alert",
	"datatype": "String",
	"access": "RW",
	"public": true,
	"description": "The currently active alert effect.",
	"values": [
		["\"none\"", "light is not performing an alert"],
		["\"select\"", "light is blinking a short time"],
		["\"lselect\"", "light is blinking a longer time"]
	]
}
```

Note in this particular case `alert` was write-able and lets the LED of the IKEA repeater blink.

ebaauw:
Not sure I like the syntax of `values`.  You already know ip`alert` takes a String value (why capilatised?), so the double quotes seem superfluous.

The descriptions are wrong: `alert` does *not* report the current blinking state, it simply echoes the last value set.  Hence my designation write-only.  It should be in `action` rather than in `state`.

manup:
I think we just found a bug, I tried `PUT /lights/5 {"alert": "lselect"}` this does indeed bring the light to flash a few times (hence my interpretation it's write-able, albeit this is the wrong term here). But `GET /lights/5` always returns `{"alert": "none"}` during the blinking as well as afterwards :thinking: 

https://dresden-elektronik.github.io/deconz-rest-doc/endpoints/lights/#set-light-state seems a bit misleading in that regard.

A separation into `action` would be really helpful, I guess that's a bigger topic to figure out a clean interface for what is `action`, `config` and `state`.

>Not sure I like the syntax of values. You already know ipalert takes a String value (why capilatised?), so the double quotes seem superfluous.

The `datatype` is there so that an item description is fully self describing, It would be nice in future to to get completely dynamic types as well which don't require a former definition in C++, that's what the ZCLDB does. I think it might be a good idea to have separate JSON descriptions for data types, in the ZCLDB I found that very useful.

The capitalization is currently there because it's taken directly for the auto generated documentation (hack).
Same for the wonky values syntax, the double quotes are a temporary hack to show quotes in the docs as well.

Example result: https://dresden-elektronik.github.io/deconz-rest-doc/devices/ikea/gu10_ws_400lm_light/#state-attributes

This part is currently a bit hacky and needs to be streamlined, for example:

```json
{
	"id": "state/alert",
	"datatype": "string",
	"access": "rw",
	"public": true,
	"description": "The currently active alert effect.",
	"values": [
		{ "value": "none", "description": "light is not performing an alert" },
		{ "value": "select", "description": "light is blinking a short time" },
		{ "value": "lselect", "description": "light is blinking a longer time" }
	]
}
```

ebaauw:
> I guess that's a bigger topic to figure out a clean interface for what is `action`, `config` and `state`.

As a first step, we could accept a PUT on `state`, but not report it on GET, cf. `bri_inc`.

> for example

❤️

Issue #3258, opened 2020-09-13
Elko ESH 316 Endevender RF - device support request


Knodd:
This is the battery operated wireless remote for the 316GLED RF requested here: https://github.com/dresden-elektronik/deconz-rest-plugin/issues/2497

Phoscon will not find it but it shows up in deCONZ after pairing. I can bind it to a light to turn on/off and dim but cannot bind to groups.

## Device

- Product name: Elko ESH 316 Endevender RF
- Manufacturer: ELKO
- Model identifier: ElkoDimmerRemoteZHA
- Device type : Please remove all unrelated device types. 
  - Remote
  - Switch

## Screenshots
Basic cluster
![image](https://user-images.githubusercontent.com/5361815/93026959-1e7f8180-f60a-11ea-9750-d81f80e0bfdc.png)

Node info
![image](https://user-images.githubusercontent.com/5361815/93027092-09572280-f60b-11ea-8410-c4d99ac5f248.png)

WTell:
I would be happy to help with testing with my ESH 316 remotes. In my experience, binding is hit-or-miss, and some remotes refuse to show up in deCONZ at all.

SwoopX:
I guess I can give it a look within the next days. It would really help if anybody could compile the plugin then for testing.

WTell:
@SwoopX I would love to help in any way I can, as long as I get a little bit of guidance on the way!

The product in question here is described by the manufacturer on https://proff.elko.no/316-endevender/category2990.html .

The battery operated remote mentioned above is meant to work in conjunction with an in-wall mounted dimmer that features the same design. The dimmer is called _Elko ESH 316GLED dimmer RF_, and is described on https://proff.elko.no/smart-316gled/category2968.html .

I have with success been able to add the in-wall dimmers to my ZigBee network, and they show up as Lights in deCONZ and Phoscon. They are have model _ElkoDimmerZHA_, and vendor either _Heiman_ **or** _ELKO_ depending on what hardware version it is and when it was manufactured. It seems reasonable to add support for the _ElkoDimmerZHA_ along with the _ElkoDimmerRemoteZHA_ that @Knodd mentioned above. Should I add the requested screenshots for the _ElkoDimmerZHA_ in this thread, or should I start a separate thread?

WTell:
[...]

PR #4396, opened 2021-02-11
Add button map for Elko ESH 316 Endevender RF

Knodd:
Added button map for Elko ESH 316 Endevender RF


Knodd:
To include fixes found here: https://github.com/dresden-elektronik/deconz-rest-plugin/issues/3258

PR #4408, opened 2021-02-12
Add API config "externalsensortemp" for Danfoss Ally thermostat

hanskroner:
Allows for setting an externally-measured temperature value via the REST
API. Useful for those of us who have radiators under desks or covered by
radiator covers.

### DIFF
--- a/database.cpp
+++ b/database.cpp
@@ -3669,6 +3669,7 @@ static int sqliteLoadAllSensorsCallback(void *user, int ncols, char **colval , c
                     // Supported with Danfoss firmware version 1.08
                     sensor.addItem(DataTypeBool, RConfigScheduleOn)->setValue(false);
                     sensor.addItem(DataTypeString, RConfigSchedule);
+                    sensor.addItem(DataTypeInt16, RConfigExternalTemperatureSensor);
                 }
                 else if (sensor.modelId() == QLatin1String("AC201")) // OWON AC201 Thermostat
                 {
--- a/de_web_plugin.cpp
+++ b/de_web_plugin.cpp
@@ -6606,6 +6606,7 @@ void DeRestPluginPrivate::addSensorNode(const deCONZ::Node *node, const SensorFi
                 // Supported with Danfoss firmware version 1.08
                 sensorNode.addItem(DataTypeBool, RConfigScheduleOn)->setValue(false);
                 sensorNode.addItem(DataTypeString, RConfigSchedule);
+                sensorNode.addItem(DataTypeInt16, RConfigExternalTemperatureSensor);
             }
             else if (modelId == QLatin1String("AC201")) // OWON AC201 Thermostat
             {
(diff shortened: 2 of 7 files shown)


luka6000:
@hanskroner this is great! Would you mind adding a few other stuff for Ally?
On same cluster
0x0025 Thermostat programming operation mode
0x4003 External Open Window Detected
I think those 2 are also very usefull.
And maybe 
0x0001 KeypadLockout
on cluster (0x0204) Thermostat UI Configuration
That would make this trv a great option with deconz

SwoopX:
@luka6000 What you request is already there, even since day 1 if I recall correctly

SwoopX (commented):
Please have a look at the minor remarks.

hanskroner:
@SwoopX, thanks for having a look at this! I've implemented the suggestions and updated the PR.

hanskroner:
> 0x0025 Thermostat programming operation mode

This looks to be already implemented

> 0x4003 External Open Window Detected

The REST API already exposes the current state of the "Window Open detection" feature but doesn't yet allow hinting to the TRV that a window is open, which is what `0x4003` seems to be for. Implementing support for this should be straightforward, just need to be careful to not mess up the Eurotronic Spirits, who use `0x4003` for some sort of setpoint reporting.

It might make sense to also add support for `0x4051` which is used to enable/disable the "Window Open detection" feature. I'll try to get both together in a separate PR.

hanskroner:
I had missed a couple of renames. The branch still built properly, but didn't work 😞 I've made additional changes, and it's both building and working properly now.

Mimiix:
Great work @hanskroner Thanks :)!

tkintscher:
Thanks so much for this patch!

As I understand it, this requires the 1.08 firmware.
I can confirm that the firmware can be upgraded to 1.08 using the OTA tool in deCONZ, with the firmware file from the manufacturer (https://www.danfoss.com/en/products/smart-heating/dhs/smart-heating/danfoss-ally/danfoss-ally-support/#tab-software, download the zip, then use the ...ota file inside).

It took ~8 hours for me to upgrade, so I let it run overnight and this morning it was done. :-)

simoneras:
hello, I have entered the temperature value from an external sensor, but to enable the detection of the external ambient probe what should I do? should I just put the value in / config / externalsensortemp?

thank you

GoogleTranslate

hanskroner:
> hello, I have entered the temperature value from an external sensor, but to enable the detection of the external ambient probe what should I do? should I just put the value in / config / externalsensortemp?
> 
> thank you

I didn't have to do anything other than putting a value in `/config/externalsensortemp`. It seems that as long as the value in there is different from -8000, the TRVs take the measurement into account. The Danfoss documentation suggests not writing a value more often than every 30 minutes, and at least once every 3 hours.

At least for me, the TRVs don't seem to go exclusively after the value reported by the external sensor, but try to get to some middle ground between the externally reported temperature and the internally measured one.

simoneras:
ok. thank you

PR #4426, opened 2021-02-15
Add support for Develco External Meter Interface ZHEMI101

SwoopX:
Tested and working. 

Introduce `RConfigPulseConfiguration` and `RConfigInterfaceMode` to API


simple_metering.cpp
```diff
@@ -0,0 +1,213 @@
...
+        }
+        else if (attrType == deCONZ::Zcl16BitInt)
+        {
+            stream << static_cast<qint16>(attrValue);
+        }
+        else
+        {
+            return false;
+        }
+    }
```
manup, line 203: I think the following should work for the given data types:

```cpp

stream << attrId;

if (readOrWriteCmd == deCONZ::ZclWriteAttributesId)
{
    deCONZ::ZclAttribute attr(attrId, attrType, "", deCONZ::ZclWrite, true);
    attr.setValue(QVariant(attrValue));

    if (!attr.writeToStream(stream))
    {
         return false;
    }
}
```

manup (commented):
added a few comments

simple_metering.cpp
SwoopX: Ah, yes. Way more elegant. However, the code must be slightly amended. Tested and working locally.

simple_metering.cpp
```diff
@@ -0,0 +1,213 @@
...
+                    quint8 mode = 0;
+                    
+                    if (interfaceMode == 0x0000) { mode = 1; }
+                    else if (interfaceMode == 0x0001) { mode = 2; }
+                    else if (interfaceMode == 0x0002) { mode = 3; }
+                    else if (interfaceMode == 0x0100) { mode = 4; }
+                    else if (interfaceMode == 0x0101) { mode = 5; }
+                    else if (interfaceMode == 0x0102) { mode = 6; }
+                    else if (interfaceMode == 0x0103) { mode = 7; }
+                    else if (interfaceMode == 0x0104) { mode = 8; }
```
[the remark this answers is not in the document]
SwoopX: A Magician doesn't reveal his tricks... 😛 

Now commented.

rest_sensors.cpp
```diff
@@ -1825,6 +1825,86 @@ int DeRestPluginPrivate::changeSensorConfig(const ApiRequest &req, ApiResponse &
...
+                            else if (mode == 3) { interfaceMode = 0x0002; }
+                            else if (mode == 4) { interfaceMode = 0x0100; }
+                            else if (mode == 5) { interfaceMode = 0x0101; }
+                            else if (mode == 6) { interfaceMode = 0x0102; }
+                            else if (mode == 7) { interfaceMode = 0x0103; }
+                            else if (mode == 8) { interfaceMode = 0x0104; }
+                            else
+                            {
+                                rspItemState[QString("error unknown interface mode for %1").arg(sensor->modelId())] = map[pi.key()];
+                            }
```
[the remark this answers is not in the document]
SwoopX: Modes are commented.

We should have a quick discussion on how the error reporting shall take place (also in the future). This is all over the place.
Either we add `rspItem["success"] = rspItemState;` where it is missing to display the error, or we directly respond with an error message. However, not too sure in how far that might "interfere" with Hue API behaviour...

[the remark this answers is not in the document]
manup: As far as I'm aware there is only one proper way to report errors, which is the array of **error objects** in the defined format:

https://dresden-elektronik.github.io/deconz-rest-doc/misc/errors/

For example:

```
GET /api/<apikey>/lights/does_not_exists
```

```json
[
    {
        "error": {
            "address": "/lights/does_not_exist",
            "description": "resource, /lights/does_not_exist, not available",
            "type": 3
        }
    }
]
```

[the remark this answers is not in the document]
SwoopX: Amended accordingly

rest_sensors.cpp
```diff
@@ -1825,6 +1825,89 @@ int DeRestPluginPrivate::changeSensorConfig(const ApiRequest &req, ApiResponse &
...
+            {
+                if (rid.suffix == RConfigPulseConfiguration)
+                {
+                    if (map[pi.key()].type() == QVariant::Double)
+                    {
+                        if (sensor->modelId() == QLatin1String("ZHEMI101"))
+                        {
+                            quint16 pulseConfiguration = map[pi.key()].toUInt(&ok);
+
+                            if (addTaskSimpleMeteringReadWriteAttribute(task, deCONZ::ZclWriteAttributesId, 0x0300, deCONZ::Zcl16BitUint, pulseConfiguration, VENDOR_DEVELCO))
```
manup, line 1840: Extra points if `ok` is also used :)

```cpp
if (ok && addTask...)
```

simple_metering.cpp
```diff
@@ -0,0 +1,200 @@
...
+    }
+
+    // payload
+    QDataStream stream(&task.zclFrame.payload(), QIODevice::WriteOnly);
+    stream.setByteOrder(QDataStream::LittleEndian);
+
+    if (readOrWriteCmd == deCONZ::ZclWriteAttributesId)
+    {
+        stream << static_cast<quint16>(attrId);
+        stream << static_cast<quint8>(attrType);
```
manup, line 181: Doesn't need casting as parameters already have the right type.

simple_metering.cpp
```diff
@@ -0,0 +1,200 @@
...
+/*! Write Attribute on thermostat cluster.
+ *  Iterate over every day and get schedule for each day.
+   \param task - the task item
+   \param attrId
+   \param attrType
+   \param attrValue
+   \return true - on success
+           false - on error
+ */
+bool DeRestPluginPrivate::addTaskSimpleMeteringReadWriteAttribute(TaskItem &task, uint8_t readOrWriteCmd, uint16_t attrId, uint8_t attrType, uint32_t attrValue, uint16_t mfrCode)
```
manup, line 148: Nothing wrong, but I'd suggest to stick to quint8, 16, ... instead uint8_t types for consistency.

manup (commented):
Small comments

rest_sensors.cpp
SwoopX: New highscore then ;)

Even more points for correcting it throughout the file...

simple_metering.cpp
SwoopX: kk

simple_metering.cpp
SwoopX: Got the idea. However, if you're fine with it, I'd prefer to leave it for now as it is. Since we got this all over the code, I'd rather got with a new PR to address this consistently. What do you think?

rest_sensors.cpp
```diff
@@ -2111,6 +2111,89 @@ int DeRestPluginPrivate::changeSensorConfig(const ApiRequest &req, ApiResponse &
...
+                            }
+                            else
+                            {
+                                rsp.list.append(errorToMap(ERR_ACTION_ERROR, QString("/sensors/%1/config/%2").arg(id).arg(pi.key()).toHtmlEscaped(),
+                                                           QString("Could not set attribute")));
+                                rsp.httpStatus = HttpStatusBadRequest;
+                                return REQ_READY_SEND;
+                            }
+                        }
+                    }
```
manup, line 2138: `if (sensor->modelId() == QLatin1String("ZHEMI101")) ...`

When this is called for sensors which have a `RConfigPulseConfiguration` but aren't "ZHEMI101" there seems to be a missing error else branch?

`quint16 pulseConfiguration = map[pi.key()].toUInt(&ok);`

Implicit cast, maybe not realistic but a client could call this with values > UINT16_MAX which would be silently truncated.

```
const uint pulseConfiguration = map[pi.key()].toUInt(&ok);

if (ok && pulseConfiguration <=  UINT16_MAX && addTaskSimpleMeteringReadWriteAttribute(task, deCONZ::ZclWriteAttributesId, 0x0300, deCONZ::Zcl16BitUint, pulseConfiguration, VENDOR_DEVELCO)
{
}
```

rest_sensors.cpp
```diff
@@ -2111,6 +2111,89 @@ int DeRestPluginPrivate::changeSensorConfig(const ApiRequest &req, ApiResponse &
...
+                        return REQ_READY_SEND;
+                    }
+                }
+                else if (rid.suffix == RConfigInterfaceMode)
+                {
+                    if (map[pi.key()].type() == QVariant::Double)
+                    {
+                        if (sensor->modelId() == QLatin1String("ZHEMI101"))
+                        {
+                            quint8 mode = map[pi.key()].toUInt(&ok);
```
manup, line 2153: The checks later on are solid, to prevent implicit cast I'd go for `const uint` or `const auto`.

rest_sensors.cpp
```diff
@@ -2111,6 +2111,89 @@ int DeRestPluginPrivate::changeSensorConfig(const ApiRequest &req, ApiResponse &
...
+                            quint16 interfaceMode = 0;
+                            
+                            if (ok && mode == 1) { interfaceMode = 0x0000; }          // Pulse Counting on an Electricity Meter – Unit KWh
+                            else if (ok && mode == 2) { interfaceMode = 0x0001; }     // Pulse Counting on a Gas Meter – Unit m3
+                            else if (ok && mode == 3) { interfaceMode = 0x0002; }     // Pulse Counting on a Water Meter – Unit m3
+                            else if (ok && mode == 4) { interfaceMode = 0x0100; }     // Kamstrup KMP Protocol
+                            else if (ok && mode == 5) { interfaceMode = 0x0101; }     // Not Supported - Linky Protocol
+                            else if (ok && mode == 6) { interfaceMode = 0x0102; }     // DLMS-COSEM - IEC62056-21 mod A
+                            else if (ok && mode == 7) { interfaceMode = 0x0103; }     // P1 Dutch Standard – DSMR 2.3 Version
+                            else if (ok && mode == 8) { interfaceMode = 0x0104; }     // P1 Dutch Standard – DSMR 4.0 Version
```
manup, line 2163: For the `mode` and `interfaceMode` values I'd suggest defines instead hard coded values, since they are used later in this file again and this might be extended in future?

rest_sensors.cpp
```diff
@@ -2111,6 +2111,89 @@ int DeRestPluginPrivate::changeSensorConfig(const ApiRequest &req, ApiResponse &
...
+                            else
+                            {
+                                rsp.list.append(errorToMap(ERR_INVALID_VALUE, QString("/sensors/%1/config/%2").arg(id).arg(pi.key()).toHtmlEscaped(),
+                                                           QString("invalid value, %1, for parameter %2").arg(map[pi.key()].toString()).arg(pi.key()).toHtmlEscaped()));
```
manup, line 2167: The `errorToMap()` function already sanitizes the arguments 2 and 3 internally with `.toHtmlEscaped()` it can be removed here and in other places in the file.

simple_metering.cpp
manup: Yes fine by me :)

manup (commented):
Had a closer look, no errors but might be worth considering these changes.

rest_sensors.cpp
SwoopX: Right, that's a remnant from the previous PR aimed at (mostly) unifying the error handling.

rest_sensors.cpp
SwoopX: There's indeed a 2nd device to follow. I'd probably even go a step further and later see if a device specific map would make sense, as it is used every now and then for other devices/attributes. That could aid in shrinking the code and make it less error prone, I suppose

rest_sensors.cpp
SwoopX: ok

rest_sensors.cpp
SwoopX: > When this is called for sensors which have a RConfigPulseConfiguration but aren't "ZHEMI101" there seems to be a missing error else branch?

Well, this is true, however, that is elsewhere in the function also not the case. I assume it should throw a `ERR_PARAMETER_NOT_AVAILABLE`?

I'm currently really thinking this function could use a decent refactor to take out complexity and repetitions while making it more robust and shrinking size...

> Implicit cast, maybe not realistic but a client could call this with values > UINT16_MAX which would be silently truncated.

Totally agree

manup: >Well, this is true, however, that is elsewhere in the function also not the case. 

I guess we have unfortunately many places with missing error handling.

>I assume it should throw a ERR_PARAMETER_NOT_AVAILABLE?

Sounds fine to me.

>I'm currently really thinking this function could use a decent refactor to take out complexity and repetitions while making it more robust and shrinking size...

All in :D I'd really like to split this up into multiple smaller functions each only doing once thing/topic so we get a chance to get closer to automatic testing.

manup (approved):
Looking good :+1:

tombadog99:
Does this support include support for current power usage (kW) as well as the energy meter reading (kWh). I've just integrated this into home assistant via the second plugin but currently only has the energy meter reading and not the current power usage.

PR #4428, opened 2021-02-15
Add initial support for Xiaomi Aqara T1 open/close sensor MCCGQ12LM

SwoopX:


oleyyh:
Hello @SwoopX,
Have you managed to Enroll your MCCGQ12LM?
I'm doing a side project where i'm trying, and after writing the CIE address to the CIE Address attribute, i get the ZoneEnrollRequest which i reply with a ZoneEnrollReponse. Everything seems to look correct on wireshark:
Request:
image
Response:
image

But if i read the ZoneState attribute afterwards, it is still marked as Value: 0x00 [NotEnrolled].
The same process has been tested on a Samsung Multisensor, and works perfectly.

SwoopX:
@oleyyh It never worked, since I don't have any :D

Are you sure you're looking at the right packets? Your request is from Xiaomi, but your response is Develco... Also note that barely any vendor implements that correctly in that sense, that a request/response is sufficient for the enrollment to be successful (without the IAS CIE write).

oleyyh:
@SwoopX The reason the response is from Develco, is because my ZigBee coordinator is made by Develco. 

I've tried quite a few different methods to enroll my device, with and without the IAS CIE write. 
Would you be interested in me sending a couple MCCGQ12LM to you, where you could in return explain the necessary steps required, if you were to succeed?

Please email me if interested, at: olif@softcontrol.dk
Bear in mind i won't be available after 15:45 (CEST) today, until after easter (5. April).

Issue #4099, opened 2021-01-04
Migrating from RaspBee (I or II) and ConBee II, Backup/Restore ConBee II


lukicsl:
## Describe the bug
I am running deCONZ inside docker. 
I started with using RaspBee I and buildup a network of 80+ nodes.
I later bought a ConBee II hoping to migrate the installation to ConBee II.
I failed multiple times and put the idea byside.
2 months ago I ordered RaspBee II and the migration was as: Shutdown Pi, replace the RB I for RB II, boot up, start the container, and it was done. Works as expected.

Yesterday the RB II stopped working as was not visible as a device under /dev.
This incident brought back the idea of migrating the installation on the CB II, having issues to do so. Even a simple backup CB II and restore CB II does not work.

## Steps to reproduce the behavior

I first installed back the RB I into the Pi and it picked up the installation without any issue: 

![image](https://user-images.githubusercontent.com/23736716/103539466-6ca9e080-4e98-11eb-99fb-f0e1f2a05689.png)

I made a fresh backup.
I shut down and removed RB I and install CB II. Booted up, started container, and loaded the backup, as a result, all nodes were visible but not connected to the coordinator:

![image](https://user-images.githubusercontent.com/23736716/103539597-a8dd4100-4e98-11eb-87bf-52fb93fd0733.png)

I have the feeling that the backup restore process does not work on CB II.
I have made a test on a fresh installation with CB II: I added 2 lights from scratch. The installation worked as expected. After back-restore the lights do not connect anymore to the coordinator:

![image](https://user-images.githubusercontent.com/23736716/103540161-93b4e200-4e99-11eb-8aee-4cf8fa25a743.png)

## Expected behavior

  Backup made fro RB I or RB II should be ok to be imported for CB I

## Screenshots

  in the description above

## Environment
 - Host system: (Raspberry Pi )
 - Running method: Marthoc Docker container
 - Firmware version: 264A0700
 - deCONZ version: 2.09.00 / 23.12.2020
 - Device: ConBee II / RaspBee I / RaspBee II
 - Do you use an USB extension cable: no

## deCONZ Logs

No logs
-->

## Additional context

no additional context

lukicsl:
I have downloaded the sd-card image:

![image](https://user-images.githubusercontent.com/23736716/103566741-de971f80-4ec2-11eb-8681-f1b41f728b34.png)

And added my 2 test lights and created a test group. everything worked as expected.

[...]

Issue #3500, opened 2020-10-24
Segmentation fault during upgrade from RaspBee I to RaspBee II


fubar-coder:
## Describe the bug

I get a `Segmentation fault (core dumped)` when I try to upgrade from RaspBee I to RaspBee II following the instructions in issue #1441.

## Steps to reproduce the behavior

1. Stop the Raspberry PI 4B 8GB
2. Open the case
3. Replace the RaspBee I with a RaspBee II
4. Close the case
5. Start the Raspberry PI
6. Follow the instructions in issue #1441
7. Start the network recovery procedure as described in the release notes for deconz V2.05.59

## Expected behavior

No segmentation fault/error.

## Environment

 - Host system: Raspberry Pi 4B 8GB
 - OS: Raspbian arm64
 - Running method: Marthoc Docker container
 - Firmware version: 26660700
 - deCONZ version: 2.05.84 / 14.9.2020
 - Device: RaspBee II

## deCONZ Logs

[deconz-crash-log.txt](https://github.com/dresden-elektronik/deconz-rest-plugin/files/5433664/deconz-crash-log.txt)

## Additional context

EDIT 1: Hopefully this issue will go through. Previous issue was: #3499.
EDIT 2: The same problem exists with a ConBee II

SwoopX:
I feel this is closely related to #3133

fubar-coder:
I tried all three approaches: Restore a backup, load a recent config and editing the database. The error was always the same.

SwoopX:
Well, as you haven't provided any further information as I've requested, don't wonder. I would have had a potential solution available taking only 5 mins... 🤷‍♂️

fubar-coder:
@SwoopX Ok, and what would've been that solution? Would be interesting for all other persons having the same problem.

Mimiix:
> There is a log and I gave all the information as required by the bug report template.

@SwoopX asked for :

Can you describe in more detail what happened directly after device switch? Apparently, deconz ran for a while. Have you tried to restore a Phoscon backup or just to load a recent config via the hidden page?

fubar-coder:
@Mimiix What else can I tell you other than telling you that it gave a segmentation fault? And the other information is just pointless. A segmentation fault is always an application bug and the log should be enough. Except when the log is either incomplete or useless.

Mimiix:
[...]

PR #4458, opened 2021-02-21
Hotfix to prevent segfault while importing backups

SwoopX:
See #3133 for details.

Tested and working in my dev environment

Further references: #3500, #4099, https://github.com/dresden-elektronik/phoscon-app-beta/issues/339

### DIFF
--- a/de_web_plugin.cpp
+++ b/de_web_plugin.cpp
@@ -19384,15 +19384,7 @@ void DeRestPluginPrivate::genericDisconnectNetwork()
 
     apsCtrl->setNetworkState(deCONZ::NotInNetwork);
     
-    if (!reconnectTimer)
-    {
-        reconnectTimer = new QTimer(this);
-        reconnectTimer->setSingleShot(true);
-        connect(reconnectTimer, SIGNAL(timeout()),
-                this, SLOT(reconnectTimerFired()));
-    }
-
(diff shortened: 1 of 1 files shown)


de_web_plugin.cpp
```diff
@@ -19383,6 +19383,14 @@ void DeRestPluginPrivate::genericDisconnectNetwork()
...
 
     apsCtrl->setNetworkState(deCONZ::NotInNetwork);
+    
+    if (!reconnectTimer)
+    {
+        reconnectTimer = new QTimer(this);
+        reconnectTimer->setSingleShot(true);
+        connect(reconnectTimer, SIGNAL(timeout()),
+                this, SLOT(reconnectTimerFired()));
+    }
```
manup, line 19393: Good catch but I think the following is more appropriate since there are more state variables involved, the method will also initialize the timer if not already created:

```cpp
// reconnectTimer->start(DISCONNECT_CHECK_DELAY);

startReconnectNetwork(RECONNECT_CHECK_DELAY);
```

manup (changes requested):

de_web_plugin.cpp
SwoopX: Was tempted to use that, but wasn't too sure since I don't know the parts of the code too well. Hence, a more conservative initial approach ;)

PR #4461, opened 2021-02-22
Add API config "externalwindowopen" for Danfoss Ally thermostat

hanskroner:
Allows for manually signaling to the thermostat that windows in the room where it's located have been opened. Useful for helping the built-in logic enter "window open" mode when the thermostat is not situated directly under a window.

### DIFF
--- a/database.cpp
+++ b/database.cpp
@@ -3672,6 +3672,7 @@ static int sqliteLoadAllSensorsCallback(void *user, int ncols, char **colval , c
                     sensor.addItem(DataTypeBool, RConfigScheduleOn)->setValue(false);
                     sensor.addItem(DataTypeString, RConfigSchedule);
                     sensor.addItem(DataTypeInt16, RConfigExternalTemperatureSensor);
+                    sensor.addItem(DataTypeBool, RConfigExternalWindowOpen)->setValue(false);
                 }
                 else if (sensor.modelId() == QLatin1String("AC201")) // OWON AC201 Thermostat
                 {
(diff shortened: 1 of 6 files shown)


luka6000:
@hanskroner, this doesn't work. It sure set the attribute, but a while later TRV resets it back to false. Can you check? Should I make bug report on this?

hanskroner:
> @hanskroner, this doesn't work. It sure sets the attribute to true, but a while later TRV resets it back to false. Can you check? Should I make bug report on this?
@luka6000, You'll probably want to make a bug report for it, with a lot more details than this 😉 Note that there's also a minimum firmware version required on the TRV. Anecdotally, I sanded and oiled the floors in one of the rooms in our house, and I've had the windows open for around 2 weeks now to get it properly aired out. The TRV in that room has been in "external window open" the whole time (verified with Home Assistant's history). Before filling the bug, make sure you're not accidentally toggling it back.

luka6000:
> The TRV in that room has been in "external window open" the whole time (verified with Home Assistant's history).

What do you mean? Where can you see this attribute in HA?

hanskroner:
@luka6000, please do file the issue. I just made a few quick toggles on my setup to make sure it was working, and I'm noticing that any calls via the REST API are turning the config setting to "false". Lucky for me, looks like I had activated the mode through the deCONZ UI 😕

luka6000:
> Lucky for me, looks like I had activated the mode through the deCONZ UI 😕

Yes, I can confirm setting it with GUI works fine. Ok, I will send a bug report. Thanks.

luka6000:
#4882

Issue #4403, opened 2021-02-11
Support for Orvibo Contact Sensor -  Model no SM11


ashar38:
Hello team,

Going through the old issues I noticed there was a thread open to add support for Orvibo door/window sensors, but looks like that thread went cold. Hence, I am starting a new issue hoping that support could be added for this device in one of the future firmwares.
I have tried to add as much info as possible from what I see through Deconz. But please let me know if I can add some more.
## Device

- Product name: LivingWise / Orvibo Contact Sensor
- Manufacturer: ORVIBO
- Model identifier: e70f96b3773a4c9283c6862dbafb6a99
- Device type : Sensor

## Screenshots

  Required screenshots:
Node Info.
![image](https://user-images.githubusercontent.com/34940560/107681172-ba222600-6c6c-11eb-80e8-24e20106974b.png)

Clusters:
 Basic:
![image](https://user-images.githubusercontent.com/34940560/107681427-08372980-6c6d-11eb-95b7-705e3bbf7e79.png)

Identity:
![image](https://user-images.githubusercontent.com/34940560/107681495-1be29000-6c6d-11eb-8ea3-3a5cb9ec7991.png)

IAS Zone:
![image](https://user-images.githubusercontent.com/34940560/107681574-33217d80-6c6d-11eb-8d79-fc6bf6c7c874.png)

IAZ Zone: cont'd
![image](https://user-images.githubusercontent.com/34940560/107681605-416f9980-6c6d-11eb-8253-b390c3706d12.png)

Power Config:
![image](https://user-images.githubusercontent.com/34940560/107681718-6b28c080-6c6d-11eb-82ac-663b1114fb98.png)

SwoopX:
When support has been added, you need to enter sensor search in Phoscon and pair the device. However, you also need to keep an eye on deconz GUI for the device to show up, as the manufacturer name and model ID are not read automatically. When it appears in deconz GUI, you need to read the basic cluster's attributes while sensor search is still running (which is the case even when Phoscon already issued the greenish success message) for the sensor resource to be created.

ashar38:
> 
> 
> When support has been added, you need to enter sensor search in Phoscon and pair the device. However, you also need to keep an eye on deconz GUI for the device to show up, as the manufacturer name and model ID are not read automatically. When it appears in deconz GUI, you need to read the basic cluster's attributes while sensor search is still running (which is the case even when Phoscon already issued the greenish success message) for the sensor resource to be created.

[...]

PR #4465, opened 2021-02-22
Add initial support for ORVIBO open/close sensor SM11

SwoopX:
See #4403

When support has been added, you should enter sensor search in Phoscon and pair the device. However, you also need to keep an eye on deconz GUI for the device to show up, as the manufacturer name and model ID might not be read automatically. When it appears in deconz GUI, you need to read the basic cluster's attributes while sensor search is still running (which is the case even when Phoscon already issued the greenish success message) for the sensor resource to be created.


manup:
Can you please align the PR to current master due merge conflicts?

SwoopX:
@manup Done!

PR #4492, opened 2021-02-28
Reporting changes for Develco / cluster naming

SwoopX:


manup:
Nice, it's always good to get rid of magic numbers :)

PR #4493, opened 2021-02-28
Harmonize sensor error handling in API

SwoopX:
Ensure a proper error message is returned for API calls.


rest_sensors.cpp
```diff
@@ -984,7 +984,8 @@ int DeRestPluginPrivate::changeSensorConfig(const ApiRequest &req, ApiResponse &
                 }
                 else
                 {
-                    rsp.list.append(errorToMap(ERR_INVALID_VALUE, QString("/sensors/%1/%2").arg(id).arg(rid.suffix), QString("could not set attribute")));
+                    rsp.list.append(errorToMap(ERR_ACTION_ERROR, QString("/sensors/%1/config/%2").arg(id).arg(pi.key()).toHtmlEscaped(),
+                                               QString("Could not set attribute")));
```
manup, line 988: I'd suggest to move the call to `.toHtmlEscaped()` into the `errorToMap()` function for the second and third parameter. So it gets less noisy and also is used always a cross all calls.

manup (commented):
Small suggestion

rest_sensors.cpp
SwoopX: Done

PR #4579, opened 2021-03-13
Introduce getSensorNodeForAddressEndpointAndCluster()

SwoopX:
This function provides the ability to choose a sensor more precisely than both functions `getSensorNodeForAddressAndEndpoint()`. The first implementation (which doesn't use the sensor type as argument) returns the first sensor found, which is problematic if more than one "sensor-worthy" cluster is available on one endpoint. The second implementation (which uses the sensor type as argument) is more accurate, but can only handle one sensor type (e.g. there's multiple sensor types based on IAS zone cluster).

The proposed function would eliminate both weaknesses, however, it relies on the correctness of a sensor's fingerprint. In this regard, some minor updates on the corresponding code might be required.

Presumably, the following code passage could be removed as a consequence as well:

https://github.com/dresden-elektronik/deconz-rest-plugin/blob/db673839f8cf4ec867bb9471cf79477299c5a794/de_web_plugin.cpp#L1040-L1051

**Example on improved sensor selection:**
```
21:04:23:953 [TEST] - 0x000D6FFFFE400CE4 Sensor selected: 00:0d:6f:ff:fe:40:0c:e4-01-0500, LastRx: Mi. März 10 21:04:23 2021...
21:04:23:953 [TEST] - 0x000D6FFFFE400CE4 APSDE data indication details: EP: 0x01, Cluster: 0x0400...
21:04:23:953 [TEST] - 0x000D6FFFFE400CE4 Correct sensor: 00:0d:6f:ff:fe:40:0c:e4-01-0400, LastRx: Mi. März 10 21:04:23 2021...
```

As small caveat remains: As of now, basic cluster and power configuration clusters (which occur in each sensor of multi-sensor devices) are not exclusively added to the sensor fingerprints. In those cases, the function will just return the first sensor found. It also cannot handle cases, where just one sensor has been exposed for multiple endpoints (primarily/exclusively switches).

Supercedes #4034

### DIFF
--- a/de_web_plugin.cpp
+++ b/de_web_plugin.cpp
@@ -1035,7 +1035,7 @@ void DeRestPluginPrivate::apsdeDataIndication(const deCONZ::ApsDataIndication &i
             !(zclFrame.frameControl() & deCONZ::ZclFCDirectionServerToClient) ||
             (zclFrame.isProfileWideCommand() && zclFrame.commandId() == deCONZ::ZclReportAttributesId))
         {
-            Sensor *sensorNode = getSensorNodeForAddressAndEndpoint(ind.srcAddress(), ind.srcEndpoint());
+            Sensor *sensorNode = getSensorNodeForAddressEndpointAndCluster(ind.srcAddress(), ind.srcEndpoint(), ind.clusterId());
 
             if (sensorNode && ind.clusterId() == IAS_ZONE_CLUSTER_ID && sensorNode->type() != QLatin1String("ZHASwitch"))
             {
@@ -9779,6 +9779,23 @@ Sensor *DeRestPluginPrivate::getSensorNodeForAddressAndEndpoint(const deCONZ::Ad
 
     return 0;
 
+}
+
+/*! Returns the first Sensor for its given \p Address and \p Endpoint and \p Cluster or 0 if not found.
+ */
+Sensor *DeRestPluginPrivate::getSensorNodeForAddressEndpointAndCluster(const deCONZ::Address &addr, quint8 ep, quint16 cluster)
+{
+    for (Sensor &sensor: sensors)
+    {
+        if (sensor.deletedState() != Sensor::StateNormal || !sensor.node())                                     { continue; }
+        if (!((sensor.address().hasExt() && addr.hasExt() && sensor.address().ext() == addr.ext()) ||
+            (sensor.address().hasNwk() && addr.hasNwk() && sensor.address().nwk() == addr.nwk())))              { continue; }
+        if (!((sensor.fingerPrint().hasInCluster(cluster) || sensor.fingerPrint().hasOutCluster(cluster))))     { continue; }
+        if (sensor.fingerPrint().endpoint != ep)                                                                { continue; }
+        else { return &sensor; }
+    }
+
+    return nullptr;
 }
 
 /*! Returns the first Sensor which matches a fingerprint.
--- a/de_web_plugin.cpp
+++ b/de_web_plugin.cpp
@@ -9815,12 +9815,12 @@ Sensor *DeRestPluginPrivate::getSensorNodeForAddressEndpointAndCluster(const deC
     for (Sensor &sensor: sensors)
     {
         if (sensor.deletedState() != Sensor::StateNormal || !sensor.node())                             { continue; }
+        if (sensor.fingerPrint().endpoint != ep)                                                        { continue; }
         if (!isSameAddress(sensor.address(), addr))                                                     { continue; }
         if (sensor.fingerPrint().hasInCluster(cluster) || sensor.fingerPrint().hasOutCluster(cluster))  { }
         else                                                                                            { continue; }
 
-        if (sensor.fingerPrint().endpoint != ep)    { continue; }
-        else                                        { return &sensor; }
+        return &sensor;
     }
     return nullptr;
 }
--- a/de_web_plugin.cpp
+++ b/de_web_plugin.cpp
@@ -9781,20 +9781,47 @@ Sensor *DeRestPluginPrivate::getSensorNodeForAddressAndEndpoint(const deCONZ::Ad
 
 }
 
-/*! Returns the first Sensor for its given \p Address and \p Endpoint and \p Cluster or 0 if not found.
+/*! Compare addresses where either NWK or MAC address might not be known.
+    \returns true if both adresses have same MAC address (strong guaranty).
+    \returns true if at least one of the addresses doesn't have MAC but the NWK addresses are equal.
+    \returns false otherwise.
+*/
+bool isSameAddress(const deCONZ::Address &a, const deCONZ::Address &b)
+{
+    if (a.hasExt() && b.hasExt())
+    {
+        // nested if statement, so the NWK check won't be made if both MAC addresses are known
+        if (a.ext() != b.ext()) 
+        {
+             return false;
+        }
+    }
+    else  if (a.hasNwk() && b.hasNwk())
+    {
+       if (a.nwk() != b.nwk())
+       {
+            return false;
+       }
+    }
+    else { return false; }
+
+    return true;
+}
+
+/*! Returns the first Sensor for its given \p Address and \p Endpoint and \p Cluster or nullptr if not found.
  */
 Sensor *DeRestPluginPrivate::getSensorNodeForAddressEndpointAndCluster(const deCONZ::Address &addr, quint8 ep, quint16 cluster)
 {
     for (Sensor &sensor: sensors)
     {
-        if (sensor.deletedState() != Sensor::StateNormal || !sensor.node())                                     { continue; }
-        if (!((sensor.address().hasExt() && addr.hasExt() && sensor.address().ext() == addr.ext()) ||
-            (sensor.address().hasNwk() && addr.hasNwk() && sensor.address().nwk() == addr.nwk())))              { continue; }
-        if (!((sensor.fingerPrint().hasInCluster(cluster) || sensor.fingerPrint().hasOutCluster(cluster))))     { continue; }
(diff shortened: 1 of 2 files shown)


manup:
Thanks, in general I like this addition since more precise control is always preferable.

>The proposed function would eliminate both weaknesses, however, it relies on the correctness of a sensor's fingerprint. In this regard, some minor updates on the corresponding code might be required.

I'm also concerned that there might be sensor finger prints out there that currently not fit the requirement. Therefore I'd propose to merge the PR in the next 2.11.0 beta release, to figure out missing clusters in finger prints in the next beta cycle.

SwoopX:
@manup No problem at all. I'd rather expect the missing clusters in the fingerprint is more a theoretical issue based on the proposed code, since I made the code more conversative than I initially planned. For me, it simply seems not possible to craft a function handling all the specifics, hacks and zigbee misbehaviors here, but, as indicated, this is a good improvement.

There's e.g. the Tradfri 5 button remote which is sending from the scenes cluster, but doesn't have it in its cluster list. Also one resource for multiple endpoints is problematic. That's the reason while I left all the following sensor selection code untouched, as this is a bit too delicate without having the devices at hand for actual testing.

de_web_plugin.cpp
```diff
@@ -9781,6 +9781,23 @@ Sensor *DeRestPluginPrivate::getSensorNodeForAddressAndEndpoint(const deCONZ::Ad
...
+Sensor *DeRestPluginPrivate::getSensorNodeForAddressEndpointAndCluster(const deCONZ::Address &addr, quint8 ep, quint16 cluster)
+{
+    for (Sensor &sensor: sensors)
+    {
+        if (sensor.deletedState() != Sensor::StateNormal || !sensor.node())                                     { continue; }
+        if (!((sensor.address().hasExt() && addr.hasExt() && sensor.address().ext() == addr.ext()) ||
+            (sensor.address().hasNwk() && addr.hasNwk() && sensor.address().nwk() == addr.nwk())))              { continue; }
+        if (!((sensor.fingerPrint().hasInCluster(cluster) || sensor.fingerPrint().hasOutCluster(cluster))))     { continue; }
+        if (sensor.fingerPrint().endpoint != ep)                                                                { continue; }
+        else { return &sensor; }
```
manup, line 9795: I'm a big fan of guard expressions but the more complex ones are a bit hard to reason about :)

*Live code review* 😄 

Here is what I did to understand what the code does since the address check gives me a hard time:

```cpp
if (sensor.deletedState() != Sensor::StateNormal || !sensor.node())                                     { continue; }
if (!
      (
         (sensor.address().hasExt() && addr.hasExt() && sensor.address().ext() == addr.ext()) ||
         // FIXME When both ext() are present but different the nwk() check could still success (albeit unlikely)
         (sensor.address().hasNwk() && addr.hasNwk() && sensor.address().nwk() == addr.nwk())
      )
   )
{ continue; }

if (!
       (
          (sensor.fingerPrint().hasInCluster(cluster) || sensor.fingerPrint().hasOutCluster(cluster))
       )
   )
{ continue; }

if (sensor.fingerPrint().endpoint != ep)
{ continue; }

else { return &sensor; }
```

The difficulty for me when reading code in general is to keep lots of logic and parantheses in my head to follow what is actually happening 😄 

Another attempt, since the address check can be reused in other parts of the code, we can made it a extra function `isSameAddress(a, b)` which is a bit stricter and testable. In `getSensorNodeForAddressEndpointAndCluster()` it becomes one thing less to think about.

```cpp
/*! Compare addresses where either NWK or MAC address might not be known.
    \returns true if both adresses have same MAC address (strong guaranty).
    \returns true if at least one of the addresses doesn't have MAC but the NWK addresses are equal.
    \returns false otherwise.
*/
bool isSameAddress(const deCONZ::Address &a, const deCONZ::Address &b)
{
    if (a.hasExt() && b.hasExt())
    {
        // nested if statement, so the NWK check won't be made if both MAC addresses are known
        if (a.ext() != b.ext()) 
        {
             return false;
        }
    }
    else  if (a.hasNwk() && b.hasNwk())
    {
       if (a.nwk() != b.nwk())
       {
            return false;
       }
    }
    else {  return false; }

    return true;
}

/*! Returns the first Sensor for its given \p Address and \p Endpoint and \p Cluster or nullptr if not found.
 */
Sensor *DeRestPluginPrivate::getSensorNodeForAddressEndpointAndCluster(const deCONZ::Address &addr, quint8 ep, quint16 cluster)
{
    for (Sensor &sensor: sensors)
    {
        if (sensor.deletedState() != Sensor::StateNormal || !sensor.node())  { continue; }
        if (!isSameAddress(sensor.address(), addr))    {  continue; }
        if (sensor.fingerPrint().hasInCluster(cluster) || sensor.fingerPrint().hasOutCluster(cluster)) { }
        else { continue; }

        if (sensor.fingerPrint().endpoint != ep) { continue; }
        else { return &sensor; }
    }
    return nullptr;
}
```

manup (commented):
The code is correct. At first it looked so simple but I had some questions while looking closer to understand how it works, see comments, what do you think?

de_web_plugin.cpp
SwoopX: @manup at times, you really give me a hard time to find the right balance in what to keep together and what to split :stuck_out_tongue_winking_eye: Readability and to not have to twist your thoughts was the initial reason why the code we discussed via PM was longer, let aside the positive/negative checks.

Regardless, I'm all in here but it also raises the question, why we don't take this one step further and consider dissecting all the related `getSensorNodeForXYZ()` we have to create something universal. Maybe by using an additional argument to jump out or so. The other functions are quite similar in code usage, but yet different. What do you think?

What still disturbs me is that there's still a chance you'd need to have more than one sensor (e.g. for Multisensors having one endpoint and updating battery).

manup: Ya I know :) lately I'm becoming more aware that splitting up functions helps a lot to reason about them — at least if they really just do what the function name implies. Digging deeper into unit testing it's neat to be able to test small functions as well as larger ones which use these.

>Regardless, I'm all in here but it also raises the question, why we don't take this one step further and consider dissecting all the related getSensorNodeForXYZ() we have to create something universal. Maybe by using an additional argument to jump out or so. The other functions are quite similar in code usage, but yet different. What do you think?

I'm all in, personally I'd like them to be free standing functions which take the `sensors`container as parameter.
For example the above could be:

```cpp
Sensor *getSensorNodeForAddressEndpointAndCluster(std::vector<Sensor> &sensors,
                                                  const deCONZ::Address &addr,                                                  
                                                  quint8 ep, quint16 cluster)
```

Mainly to decouple this from the plugin and being able to write some tests.

I'm not sure if it is possible for the current code base to create a universial `getSensor(whatGoesInHere?)` function since depending on the context the needed and available "matching arguments" are different. But it would be great if we can figure the contexts out and reduce the getSensorXYZ() functions to the most precise ones.

>What still disturbs me is that there's still a chance you'd need to have more than one sensor (e.g. for Multisensors having one endpoint and updating battery).

Yes tricky with the current code base without lots of checks, the Device code doesn't have this problem anymore as commands are just directly forwarded to where they belong without extra code.

SwoopX:
So, I guess we should take that evolutionary then and do the optimization step by step. I'm happy to take your suggestions in how to split it. However, I'd appreciate a talk where we can leverage synergies from this one. Simply goes faster ;)

manup:
Sounds good, going step by step also works for me best and hopefully prevents bugs due to too large changes at once.

SwoopX:
@manup Done

de_web_plugin.cpp
```diff
@@ -9781,6 +9781,50 @@ Sensor *DeRestPluginPrivate::getSensorNodeForAddressAndEndpoint(const deCONZ::Ad
...
+{
+    for (Sensor &sensor: sensors)
+    {
+        if (sensor.deletedState() != Sensor::StateNormal || !sensor.node())                             { continue; }
+        if (!isSameAddress(sensor.address(), addr))                                                     { continue; }
+        if (sensor.fingerPrint().hasInCluster(cluster) || sensor.fingerPrint().hasOutCluster(cluster))  { }
+        else                                                                                            { continue; }
+
+        if (sensor.fingerPrint().endpoint != ep)    { continue; }
+        else                                        { return &sensor; }
```
manup, line 9823: Cool I can understand it now much quicker :+1: 

The function is called quite often, to tickle a bit more performance out, I'd suggest the following reordering:

```
if (sensor.deletedState() != Sensor::StateNormal || !sensor.node())                             { continue; }
if (sensor.fingerPrint().endpoint != ep)                                                        { continue; }
if (!isSameAddress(sensor.address(), addr))                                                     { continue; }
if (sensor.fingerPrint().hasInCluster(cluster) || sensor.fingerPrint().hasOutCluster(cluster))  { }
else                                                                                            { continue; }

return &sensor; 
```

Here the checks are ordered by "cost", the first two checks are likely in the same cache line and very fast, the remaining ones reach out to memory access since address and clusters are located somewhere else on the heap :)

manup (commented):
I like it, added one more performance hint what do you think?

de_web_plugin.cpp
SwoopX: Well, I guess you know better than I do which order might be faster 😉 However, I reacall there was a reason behind it putting at least the EP check as last one. I've to admit that might have lost relevance as the initial idea was also to be able to cover switches exposing only one resource and the code changed based on our various discussions around it.

For that reason, I'd like to put some thoughts around this again to be sure.

manup: I think the position of the EP check is independent, if the check fails the sensor will always be skipped :)

SwoopX: Yeah. But regardless, there's mostly a reason why I put stuff in there 😆 

Found the very first version, there it made some sense... Anyway, as this has all changed, the order indeed doesn't matter.

PR #4582, opened 2021-03-13
Various bugfixes

SwoopX:
- Add missing handler for IAS ACE data indications
- Prevent loading of erroneous button map entries

### DIFF
--- a/read_files.cpp
+++ b/read_files.cpp
@@ -265,7 +265,7 @@ QMap<QString, std::vector<Sensor::ButtonMap>> loadButtonMapsJson(const QJsonDocu
         {
             QString buttonMapName = i.key();    // Individual button map name
             //DBG_Printf(DBG_INFO, "[INFO] - Button map name: %s\n", qPrintable(buttonMapName));
-            quint8 mapItem = 0;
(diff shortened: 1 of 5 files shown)


read_files.cpp
```diff
@@ -280,6 +280,7 @@ QMap<QString, std::vector<Sensor::ButtonMap>> loadButtonMapsJson(const QJsonDocu
                     for (auto i = buttonMapArr.constBegin(); i != buttonMapArr.constEnd(); ++i)       // Loop through button map items
                     {
                         QJsonValue val = *i;
+                        mapItem++;
```
manup, line 283: `mapItem` is incremented again at the end of the else branch.

Shouldn't it be enough to just put this into the for loop definition:

`for (auto i = buttonMapArr.constBegin(); i != buttonMapArr.constEnd(); ++i, mapItem++)`

I'd also propose to make the type `int mapItem = 0;`

manup (changes requested):
Small question regarding mapItem variable.

read_files.cpp
SwoopX: Done

PR #4585, opened 2021-03-14
Added support for an other Revision of Orvibo Motion Sensor SN10ZW

ninharp:
Added support for an other unknown Revision of Orvibo Motion Sensor SN10ZW
I added new device in supported devices because it differs in MAC prefix and also in the vendor id, i added also the different model id and the slightly different manufacturer name.

Note that this sensor sends its feedback to endpoint 0x0A and not like normal to endpoint 0x01


Mimiix:
@ninharp Thanks for this contribution! Is there a git issue i can link?

manup (changes requested):
Something is off with the indentation, is this already using all 4 spaces per indentation level?

manup:
Any news here in regard to requested changes?

RangeMethod2 (approved):

manup:
Closing due inactivity, for further process this device should be integrated via DDF.

Issue #4589, opened 2021-03-16
No switch event x003 after long press - Aqara/Opple 


klaasnicolaas:
## Describe the question or issue you are having

I've noticed that I don't get an x003 event from my 6-way opple switch after a long press of more than 5 seconds. I can see this would have been fixed in the past, but could there be a new bug now? Or possibly this has already been resolved in the current beta?

#3412

## Screenshots

![image](https://user-images.githubusercontent.com/20448157/111312538-aef15a00-865f-11eb-906d-fd52cbc33870.png)

## Environment

 - Host system: Intel NUC
 - Running method: Home Assistent deCONZ Add-on
 - Firmware version: 26660700
 - deCONZ version: 2.09.03
 - Device: ConBee II
 - Do you use an USB extension cable: no

Mimiix:
To add:

@SAOPP tested this for me on 2.10.2 and he got this response:
![image](https://user-images.githubusercontent.com/43333120/111312940-16a7a500-8660-11eb-8e28-9ad4ff9ea206.png)


I'll test this later today when i update my setup to the latest beta.

SAOPP:
I did not have this problem on the version you specified, now my gateways work on versions 2.10.00, 2.10.01 and 2.10.02, and it is on the latest version (specified by me) that the opple switch works, and it works fine, I just paired it yesterday.

SAOPP:
'til yesterday opple worked for me on conbee2 with firmware `264A0700`, today it works on conbee2 with firmware `26580700`.

klaasnicolaas:
But do you have this problem if you are running version `2.9.3`? Unfortunately, I cannot easily run a beta version myself to see if it has been solved for me 😢 Unfortunately, I cannot easily run a beta version myself to see if it has been solved for me. But I think the next stable release is coming soon 🤔

SAOPP:
> But do you have the problem if you are run version `2.9.3`?

Nope.

SirChickenLittle:
Hi, I have this problem on all my OPPLE switches connected via Deconz. I have OPPLE switch 1-gang WXCJKG11LM, OPPLE switch 2-gang WXCJKG12LM and OPPLE switch 3-gang WXCJKG13LM. All of them give me a x003 command back if I hold less then around 5s, but if I hold longer I only get a x001. Have you tried holding longer than 5s and got a x003 command back? I am also on 2.09.03.

SAOPP:
[...]

Issue #4572, opened 2021-03-12
Device Request IKEA STYRBAR


aje89:
Hey,
i´ve got one of the brand new IKEA Styrbar switches.
Thing its not supported by deconz yet. 

## Device

- Product name: STYRBAR
- Manufacturer: IKEA of Sweden
- Model identifier: Remote Control N2
- Device type : Switch

## Screenshots

###Cluster Overview
![image](https://user-images.githubusercontent.com/29950259/111140523-b55dd400-8582-11eb-95f1-408cc685bcb4.png)

###Node Info
![image](https://user-images.githubusercontent.com/29950259/110990128-3050a000-8373-11eb-9072-1de6ec96686e.png)

### Basic
![image](https://user-images.githubusercontent.com/29950259/110990781-15326000-8374-11eb-9cb7-e10b44f50b74.png)

### Identify
![image](https://user-images.githubusercontent.com/29950259/110989170-c84d8a00-8371-11eb-8dc6-7d779cf1dea5.png)

### PowerConfiguration
![image](https://user-images.githubusercontent.com/29950259/110989239-e0bda480-8371-11eb-8dda-fe6f36dcd772.png)

### Poll Control
![image](https://user-images.githubusercontent.com/29950259/110989282-f206b100-8371-11eb-8719-caec440e673b.png)

### ZLL commissioning
![image](https://user-images.githubusercontent.com/29950259/110989367-16628d80-8372-11eb-8230-f654c134418a.png)

perseoGI:
Me too, I have already buy this Styrbar switch and I'm not getting deCONZ to work with it

Mimiix:
Can you also add a screen of the node / clusters ?

aje89:
> Can you also add a screen of the node / clusters ?

Added to the Overview

ebaauw:
Looks like a new version of the five-button remote, but with the on/off button integrated in dimup and dimdown, as with the two-button remote.  We’ll need the Zigbee commands that the styrbar sends on various buttonactions.

Happy to give this a go, if I can get my hands on one.

bemardev:
Got this new switch today.
I'm trying to support it but I'm seeing some strange behaviors.
@ebaauw you seem to know well the 5-button remote and have provided some fix for symfonisk, so if you have some time and want to enlighten and help me adding this, you can contact me on discord ;)

ebaauw:
[...]

PR #4627, opened 2021-03-22
IKEA STYRBAR and Minor changes

ebaauw:
- Add support for IKEA STYRBAR, see #4572.
  - Whitelist device in `de_web_plugin.cpp`;
  - Add commands to `button_map.json`;
  - Hack in `checkSensorButtonEvent()` to ignore additional commands on holding left or right button (see issue);
  - Streamline code for SHORTCUT and SYMFONISK, in line with STYRBAR:
    - Only bind _OnOff_ client cluster to group;
    - Discard unicast from client cluster to prevent double button events;
    - Fix bugs in verifying bindings/reporting;
    - Remove unneeded code for binding SYMFONISK to coordinator, as it sends commands to coordinator already by default.
  - **TODO**: Check older IKEA end devices for double button events (haven't seen any reports, so don't expect any;
  - **TODO**: Check whether old wireless dimmer (with dial) sends commands to coordinator by default as well;
  - **TODO**: Check whether old IKEA end devices can do with one client binding as well.
- Add support for Aqara Opple in controller mode, see #4589.  No support for switching modes, though; you'll need to do this manually through the GUI (as describes in the issue)... every time deCONZ restarts;
  - Add commands to `button_map.json`;
  - Add support for _Step Color Temperature_ command in `checkSensorButtonEvent()`;
- Add to `general.xml` the IKEA-specific scene commands as sent by the TRADFRI remote control on the left and right buttons, so they can be sent from the GUI.  Still cannot get them to work, though...
- Add new button mapping for TRADFRI wireless dimmer, under `mode` 1, see [Wiki](https://github.com/dresden-elektronik/deconz-rest-plugin/wiki/IKEA-controllers#ictc-g-1).
- Disable check for "too fast event", as this interferes with the correct function of several IKEA and Hue devices.

### DIFF
--- a/bindings.cpp
+++ b/bindings.cpp
@@ -2693,6 +2693,7 @@ bool DeRestPluginPrivate::checkSensorBindingsForAttributeReporting(Sensor *senso
         sensor->modelId().startsWith(QLatin1String("S2")) ||
         // IKEA
         sensor->modelId().startsWith(QLatin1String("TRADFRI")) ||
+        sensor->modelId().startsWith(QLatin1String("Remote Control N2")) || // STYRBAR
         sensor->modelId().startsWith(QLatin1String("FYRTUR")) ||
         sensor->modelId().startsWith(QLatin1String("KADRILJ")) ||
         sensor->modelId().startsWith(QLatin1String("SYMFONISK")) ||
@@ -3483,9 +3484,11 @@ bool DeRestPluginPrivate::checkSensorBindingsForClientClusters(Sensor *sensor)
     }
     // IKEA Trådfri on/off switch
     // IKEA Trådfri shortcut button
+    // IKEA STYRBAR
     // Sonoff SNZB-01
     else if (sensor->modelId().startsWith(QLatin1String("TRADFRI on/off switch")) ||
              sensor->modelId().startsWith(QLatin1String("TRADFRI SHORTCUT Button")) ||
+             sensor->modelId().startsWith(QLatin1String("Remote Control N2")) ||
              sensor->modelId().startsWith(QLatin1String("WB01")) ||
              sensor->modelId().startsWith(QLatin1String("WB-01")))
     {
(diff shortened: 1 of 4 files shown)


manup:
Ah damn I pressed "Ready for review" thinking I was starting review but now PR status changed.
So far PR looks fine to me.

manup (approved):
Looks fine so far

ebaauw:
I think this is good to go for now.  Still sanity-checking the older IKEA controllers (see also new [Wiki](https://github.com/dresden-elektronik/deconz-rest-plugin/wiki/IKEA-controllers page).

@manup, @SwoopX please note that I commented out the check for "too fast event": it's perfectly legal for IKEA and Hue devices to send more events within 0.5s.  The check causes mismatches in x000/x001/x003 sequences. I think we need to whitelist the devices for which this check is needed, if at all.  The double events for IKEA are all due to the device sending the command twice, once by groupcast and once by unicast; as a remedy, I simply ignore the unicast.

manup:
> @manup, @SwoopX please note that I commented out the check for "too fast event": it's perfectly legal for IKEA and Hue devices to send more events within 0.5s. The check causes mismatches in x000/x001/x003 sequences. I think we need to whitelist the devices for which this check is needed, if at all. The double events for IKEA are all due to the device sending the command twice, once by groupcast and once by unicast; as a remedy, I simply ignore the unicast.

This was introduced in v2.5.33 in 2018 https://github.com/dresden-elektronik/deconz-rest-plugin/commit/0f89afebda4afa9a13c05ce72d039fa2945d2564 due problems with Busch Jaeger switches https://github.com/dresden-elektronik/deconz-rest-plugin/commit/0ed3d1745fd1fe99e459b7aaadca6f2883f750c3 So it might be that these are the only ones (or perhaps it isn't needed at all anymore).

ebaauw:
> >  I commented out the check for "too fast event"

> This was introduced in v2.5.33 in 2018 0f89afe due problems with Busch Jaeger switches 0ed3d17 So it might be that these are the only ones (or perhaps it isn't needed at all anymore).

If it's no longer needed, let's please remove it.  If still needed can we whitelist Busch Jaeger on manufacturer code or manufacturer name, or do we need each model identifier?  Alternatively, I could blacklist manufacturer codes for Philips and IKEA.  Not sure if there's other devices that have come to rely on this check.

manup:
>If it's no longer needed, let's please remove it. If still needed, can we whitelist Busch Jaeger on manufacturer code or manufacturer name, or do we need each model identifier? Alternatively, I could blacklist manufacturer codes for Philips and IKEA. Not sure if there's other devices that have come to rely on this check.

That's a good question I can't test the BJ switch currently. Would like to remove this code as well, but I guess the safest bet is to disable it for IKEA and Philips since these are the ones we can be certain?

ebaauw:
> the safest bet is to disable it for IKEA and Philips since these are the ones we can be certain?

Done.

manup (approved):

PR #4635, opened 2021-03-24
Refactor getFreeSensorId(); fix reuse of deleted sensor ids

manup:
- Make it a free standing function, since there is no Database class yet. This is a forecast to remove database code dependency of the plugin, which will ultimately result in a proper testable Database class;
- Remove sqliteGetAllSensorIdsCallback() dependency from plugin class;
- Make `sensorIds` a local variable since it's only used here;
- Modernize code and and reduce unperformant code paths;
- Ensure new sensor ids only grow, don't reuse sensor ids which where formerly deleted.

Code has been tested in debugger for various cases.

The PR also fixes that formerly deleted sensor ids from the database were reused under certain conditions. Which was the whole point to actually access the database in this function, but it never worked. I'm not sure if this is too useful other than to prevent clients which still somehow access old ids getting confused by new completely different sensors with the same id. If that isn't needed the code can be simplified to not use the database at all.

### DIFF
--- a/database.cpp
+++ b/database.cpp
@@ -23,6 +23,8 @@ static const char *pragmaPageCount = "PRAGMA page_count";
 static const char *pragmaPageSize = "PRAGMA page_size";
 static const char *pragmaFreeListCount = "PRAGMA freelist_count";
 
+static sqlite3 *db = nullptr; // TODO should be member of Database class
+
 struct DB_Callback {
   DeRestPluginPrivate *d = nullptr;
   LightNode *lightNode = nullptr;
@@ -950,6 +952,11 @@ void DeRestPluginPrivate::pushZclValueDb(quint64 extAddress, quint8 endpoint, qu
     dbQueryQueue.push_back(sql);
 }
 
+bool DeRestPluginPrivate::dbIsOpen() const
+{
+    return db != nullptr;
+}
+
 /*! Opens/creates sqlite database.
  */
 void DeRestPluginPrivate::openDb()
@@ -967,7 +974,7 @@ void DeRestPluginPrivate::openDb()
     if (rc != SQLITE_OK) {
         // failed
         DBG_Printf(DBG_ERROR, "Can't open database: %s\n", sqlite3_errmsg(db));
-        db = 0;
+        db = nullptr;
         return;
     }
 
@@ -5408,7 +5415,7 @@ void DeRestPluginPrivate::closeDb()
         int ret = sqlite3_close(db);
         if (ret == SQLITE_OK)
         {
-            db = 0;
+            db = nullptr;
 #ifdef Q_OS_LINUX
             QElapsedTimer measTimer;
             measTimer.restart();
--- a/database.cpp
+++ b/database.cpp
@@ -4283,7 +4283,7 @@ static int sqliteGetAllSensorIdsCallback(void *user, int ncols, char **colval , 
     {
         if (colval[i] && (colval[i][0] != '\0'))
         {
-            if (strcmp(colname[i], "id") == 0)
+            if (strcmp(colname[i], "sid") == 0)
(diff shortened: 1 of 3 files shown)


ebaauw (approved):
I see no issues with the changed code.

database.cpp
```diff
@@ -4371,38 +4378,34 @@ static int sqliteLoadAllGatewaysCallback(void *user, int ncols, char **colval ,
...
+    for (const Rule &r : plugin->rules)
     {
         for (const RuleCondition &c : r.conditions())
         {
-            if (c.address().startsWith(QLatin1String("/sensors/")))
+            if (c.resource() == RSensors)
             {
-                uint id = c.id().toUInt(&ok);
+                bool ok;
+                const int id = c.id().toInt(&ok);
```
SwoopX, line 4408: You might wanna use `sid` instead of `id` here and in the following as well for consistency

SwoopX (commented):
Just a minor one ;)

database.cpp
manup: check :)

manup:
> Last time I checked (probably several years ago), the Hue bridge only recycles resource IDs, after it's gone through them all. Of course, that was when their IDs only ran from 1 to 64; it think they changed that. Anyways, I see no fundamental issue in recycling a resource ID, but it might be prudent to wait some time before doing so. We might also want to review the `add` and `delete` web socket notifications, making sure API clients are informed of deleted and new resources.
> 
> Some related open issues:
> 
>* Rules should be disabled when a resource in a condition or action is deleted.  I think the Hue bridge uses `status` `invalid` for that (or something other than `disabled`).  That way they won't (automatically) re-use a new resource with the same ID.  When re-enablng the rule (setting status to `enabled`), the API should check the conditions and actions as if the rule were newly created;

The `evaluateRule()` method seems to disable the rule for the running session, but not store this to the database.
I think back then I did this so that for example a switch can be deleted and rejoined in the Phoscon App with the configured switch rules being still intact once the device is paired again. We use the `uniqueid` in the condition address to not depend on "unstable" sensor ids. However this works only again after a deCONZ restart, a `invalid` state or something more specific could be really handy here.

https://github.com/dresden-elektronik/deconz-rest-plugin/blob/master/rest_rules.cpp#L1243-L1255

> 
>* When deleted, resources should be removed from the `links` in `/resourcelinks` resources;

Interesting topic, implicit deletion might be tricky if a `resourcelink` acts a immutable part in a greater logic — depends on what they are used for. The "hack" I mentioned above wouldn't work with this for the switch editor use case. A middle ground might be to only remove simple ids but not uniqueids? On the other hand it is only a undocumented hack and not that dramatic if it doesn't work anymore.

>* The logic for the `{"reset": false}` body should be default (when deleting resources linked to Zigbee devices).  The other resources for the same device should be deleted as well.  Might be challenging without a proper `Device` class linking all the resources.

In the experimental `Device` class code I'm doing this linking currently on the fly and match MAC address of a `deCONZ::Node` and all related Resources such as `LightNode` and `Sensor`. `Device` itself also inherits from `Resource` and has the MAC address as the `uniqueid` without endpoint and cluster id parts.

Adding sub devices: https://github.com/manup/deconz-rest-plugin/blob/device_descriptions/device.cpp#L785-L801

Returning a vector of them: https://github.com/manup/deconz-rest-plugin/blob/device_descriptions/device.cpp#L785-L801


Not currently implemented but `DELETE /device/<mac address>` is planned.

Once events are generated when a Resource such as `LightNode`, `Sensor` or `Group` is deleted, other modules like `/resourcelink` , `/rules` etc. could act on it and do their cleanup.

Perhaps we should create dedicated issues for each of these?

SwoopX (approved):

Issue #4648, opened 2021-03-26
Mi Light Detection Sensor


goegsig:
## Device

- Product name: Mi Light Detection Sensor
- Manufacturer: XIAOMI
- Model identifier: lumi.sen_ill.mgl01
- Device type : 
  - Sensor
![IMG_7734](https://user-images.githubusercontent.com/7282538/112638283-6005aa80-8e3f-11eb-9dd6-10fd50b62ccb.jpg)
![IMG_7733](https://user-images.githubusercontent.com/7282538/112638288-61cf6e00-8e3f-11eb-9c76-68ce07227ab6.jpg)

### Basic
![billede](https://user-images.githubusercontent.com/7282538/112637706-b9b9a500-8e3e-11eb-9452-719d819c2b33.png)

### Identify
![billede](https://user-images.githubusercontent.com/7282538/112637756-c6d69400-8e3e-11eb-9aa6-21887b81b1ee.png)

### Power Configuration
![billede](https://user-images.githubusercontent.com/7282538/112638702-d1ddf400-8e3f-11eb-96fb-92bacd440d43.png)

### Other clusters that are not mentioned above
![billede](https://user-images.githubusercontent.com/7282538/112637798-d2c25600-8e3e-11eb-96ab-91e358cd2e31.png)
![billede](https://user-images.githubusercontent.com/7282538/112637839-dce45480-8e3e-11eb-9189-a4f22cfb325d.png)

Same type of issue. 
https://github.com/dresden-elektronik/deconz-rest-plugin/issues/2300
https://github.com/dresden-elektronik/deconz-rest-plugin/issues/2380

SwoopX:
Hi,

would you be able to compile and test an updated plugin version?

goegsig:
At the moment iam using Home Assistent.  But if there is some guide i can follow i'am willing to try setting at virtual linux op and trying to compile and verify that it works.

Finne75:
Three devices of this type did arrive today. What is a bit curious is that I see this device in the deCONZ GUI with measurements updated (after I did manual binding as recommended in #2380), but it is not visible in the REST API. I did a restart of the gateway, but no success. That means that I also cannot use the device in home assistant currently :/..

Finne75:
@SwoopX success! Do you need any more details?

![image](https://user-images.githubusercontent.com/76888443/113411089-373c6280-93b5-11eb-8a57-ca02f60cf50d.png)

[...]

PR #4649, opened 2021-03-26
Add support for Mi version of light detector (lumi.sen_ill.mgl01)

SwoopX:
See #4648

Tested and confirmed to be working.

### DIFF
--- a/de_web_plugin.cpp
+++ b/de_web_plugin.cpp
@@ -77,6 +77,7 @@ const quint64 macPrefixMask       = 0xffffff0000000000ULL;
 const quint64 legrandMacPrefix    = 0x0004740000000000ULL;
 const quint64 dishMacPrefix       = 0x0008890000000000ULL;
 const quint64 ikeaMacPrefix       = 0x000b570000000000ULL;
+const quint64 danalockMacPrefix   = 0x000b570000000000ULL; // note: same as ikeaMacPrefix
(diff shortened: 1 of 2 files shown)


de_web_plugin.cpp
```diff
@@ -77,6 +77,7 @@ const quint64 macPrefixMask       = 0xffffff0000000000ULL;
 const quint64 legrandMacPrefix    = 0x0004740000000000ULL;
 const quint64 dishMacPrefix       = 0x0008890000000000ULL;
 const quint64 ikeaMacPrefix       = 0x000b570000000000ULL;
+const quint64 danalockMacPrefix   = 0x000b570000000000ULL; // note: same as ikeaMacPrefix
```
manup, line 80: Should we just keep one and rename it to `silabs1MacPrefix`?

https://maclookup.app/search/result?mac=000b570

manup (commented):
Question for the mac prefix, we have the huge opportunity to remove *one line* from the code base :smile:

de_web_plugin.cpp
SwoopX: You overlooked the other duplicate entry ;)

I'd prefer to leave it here as is and have a seperate PR for that as we need to propagate the changes through all files.

manup: Sounds good

PR #4652, opened 2021-03-27
Change Lidl/Silvercrest Smart Wireless Door Bell from ZHASwitch to ZHAAlarm resource

SwoopX:
This device is technically an alarm sensor and should be treated as such. By doing that, all assiciated IAS attributes will be exposed as supported by the device. Any triggering in automation systems must then be done on the `alarm` resource item.

For this take to take effect in environments where the device is already in use, a deCONZ restart would be required.

### DIFF
--- a/database.cpp
+++ b/database.cpp
@@ -3831,6 +3831,14 @@ static int sqliteLoadAllSensorsCallback(void *user, int ncols, char **colval , c
             item = sensor.addItem(DataTypeString, RConfigAlert);
             item->setValue(R_ALERT_DEFAULT);
         }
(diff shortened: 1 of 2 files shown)


MartinAirN (approved):

ebaauw:
> Any triggering in automation systems must then be done on the `alert` resource item.

I think that should be (read-only)`state.alarm`?  `state.alert` is available only on `/lights` resources.  Like `config.alert` for `/sensors` resources, it's write-only and used to send an _Identify_ command (or similar) to make the device blink.

SwoopX:
Yes, indeed `state.alarm`. Don't ask me why I always mix that up in writing...

manup:
Generally ok with the change to `ZHAAlarm`. Can't we update existing formerly paired devices while loading them from the database?

Since this might be a breaking change for clients, we should announce it.

PR #4653, opened 2021-03-27
Expose RConfigInterfaceMode for Develco EMI Norwegian HAN

SwoopX:
See #2127

### DIFF
--- a/rest_sensors.cpp
+++ b/rest_sensors.cpp
@@ -2175,41 +2175,30 @@ int DeRestPluginPrivate::changeSensorConfig(const ApiRequest &req, ApiResponse &
                         {
                             const uint mode = map[pi.key()].toUInt(&ok);
                             quint16 interfaceMode = 0;
-                            quint8 availableModes = 0;
                             
-                            if(sensor->modelId() == QLatin1String("ZHEMI101"))
+                            if (!ok)
+                            { }
+                            if (sensor->modelId() == QLatin1String("ZHEMI101"))
                             {
-                                availableModes = 8;
-                                
-                                if      (ok && mode == 1) { interfaceMode = PULSE_COUNTING_ELECTRICITY; }
-                                else if (ok && mode == 2) { interfaceMode = PULSE_COUNTING_GAS; }
-                                else if (ok && mode == 3) { interfaceMode = PULSE_COUNTING_WATER; }
-                                else if (ok && mode == 4) { interfaceMode = KAMSTRUP_KMP; }
-                                else if (ok && mode == 5) { interfaceMode = LINKY; }
-                                else if (ok && mode == 6) { interfaceMode = DLMS_COSEM; }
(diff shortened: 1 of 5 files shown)


rest_sensors.cpp
```diff
@@ -2171,28 +2171,45 @@ int DeRestPluginPrivate::changeSensorConfig(const ApiRequest &req, ApiResponse &
...
-                            if (ok && mode == 1)      { interfaceMode = PULSE_COUNTING_ELECTRICITY; }
-                            else if (ok && mode == 2) { interfaceMode = PULSE_COUNTING_GAS; }
-                            else if (ok && mode == 3) { interfaceMode = PULSE_COUNTING_WATER; }
-                            else if (ok && mode == 4) { interfaceMode = KAMSTRUP_KMP; }
-                            else if (ok && mode == 5) { interfaceMode = LINKY; }
-                            else if (ok && mode == 6) { interfaceMode = DLMS_COSEM; }
-                            else if (ok && mode == 7) { interfaceMode = DSMR_23; }
-                            else if (ok && mode == 8) { interfaceMode = DSMR_40; }
-                            else
+                            if(sensor->modelId() == QLatin1String("ZHEMI101"))
```
manup, line 2180: Change request: Spaces after `if` ;-)

The `ok`check could be done first so it doesn't need to be repeated for each value.

```
if (!ok)
{ }
else if (sensor->modelId() == QLatin1String("ZHEMI101"))
{
}
```

rest_sensors.cpp
```diff
@@ -2171,28 +2171,45 @@ int DeRestPluginPrivate::changeSensorConfig(const ApiRequest &req, ApiResponse &
...
+                            if(interfaceMode == 0)
                             {
                                 rsp.list.append(errorToMap(ERR_INVALID_VALUE, QString("/sensors/%1/config/%2").arg(id).arg(pi.key()),
                                                            QString("invalid value, %1, for parameter %2").arg(map[pi.key()].toString()).arg(pi.key())));
                                 rsp.httpStatus = HttpStatusBadRequest;
                                 return REQ_READY_SEND;
                             }
 
-                            if (mode > 0 && mode < 9)
+                            if (availableModes && mode > 0 && mode <= availableModes)
```
manup, line 2212: If I understand the whole code section correctly..

Why not ditch the `availableModes` completely to simplify the code? 

We can only reach at this point when `interfaceMode != 0` due the error checking above, so even an invalid mode from the client is already handled. The `if (availableModes && mode > 0 && mode <= availableModes)` check isn't needed, it's like a duplicated check for the same thing?

manup (commented):
Interesting, added some comments :)

rest_sensors.cpp
SwoopX: Sure

rest_sensors.cpp
SwoopX: Done

SwoopX:
@manup Done

PR #4660, opened 2021-03-27
Allow Xiaomi Aqara Single Switch Module T1 to create sensor resources

SwoopX:
Depends on #4649


manup:
Have the device here now and will make some tests with the PR.

SwoopX:
The recent discoveries could probably explain why the behavior feels a bit unusual here. Gonna be interesting...

manup:
I'll try to join it with the two different Node Descriptors for the coordinator. Wouldn't be surprised to see differences like with the Aqara Cube.

PR #4736, opened 2021-04-08
Support delete device and its REST resources

manup:
The new `deleteDevice(extAddress)` function has following side effects:

- Delete all related Sensor and LightNodes;
- Delete device related entries from database;
- Send ZDP Mgmt_Leave_req to device to remove it from network (also sleeping end-devices).

Coming with v2.11.0 this can be invoked from the GUI via <key>Delete</key> key or node context menu.

image

### DIFF
--- a/database.cpp
+++ b/database.cpp
@@ -5556,6 +5556,42 @@ void DeRestPluginPrivate::getZigbeeConfigDb(QVariantList &out)
     closeDb();
 }
 
+/*! Deletes a device from the database.
+
+    Due the foreign keys this affects the tables:
+    - device
+    - device_descriptors
+    - device_gui
+    - source_routes
+    - source_route_hops
+ */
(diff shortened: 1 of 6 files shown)


ebaauw (changes requested):
Nice!  Still missing:
- Invalidate rules referring to a device resource from `conditions` (or `actions`?);
- Remove device resources from resourcelinks `links`;
- Not sure about this: invalidate schedules referring to a device resource in `command`.

manup:
> Nice! Still missing:
> 
> * Invalidate rules referring to a device resource from `conditions` (or `actions`?);
> * Remove device resources from resourcelinks `links`;
> * Not sure about this: invalidate schedules referring to a device resource in `command`.

Good point, I'd like to address these in separate PR(s) with some more discussion about the details beforehand in an own issue. I think we need to get a better picture how currently clients use rules and resourcelinks to avoid surprises.

PR #4765, opened 2021-04-13
Streamline attribute reporting config and align polling

SwoopX:
This PR is intended to streamline the behavior in `checkLightBindingsForAttributeReporting()` and `checkSensorBindingsForAttributeReporting()` to check if the corresponding bindings exist and require attribute reporting.

For both functions, a grace time of `(val.maxInterval * 3 / 2)` is now given (the resources must have the `maxInterval` already set internally), so that basically 2 reports are allowed to have gone MIA before configuring the reporting again. Otherwise, a default value will be assumed (60 * 6 secs for lights, 60 * 15 secs for sensors (previously 60 * 45)).

All this also has been better aligned with the poll manager, which takes the same code as `checkLightBindingsForAttributeReporting()`. This should ensure light nodes get a proper fall back to polling, in case no reports come in after `(val.maxInterval * 3 / 2)`.

Some XAL related code has been removed as is seemed to be duplicate code, but required a slight amendment of the maxInterval value (now 1200 instead of 3600) to reflect the initial polling conditions.

All in all, this should prevent some unnecessary requests to set up reporting configration and prevent unnecessary polling of light nodes based on a larger maxInterval of the previous allowed 6 minutes (e.g. light nodes maxInterval is configured to 10 mins, but with the old code, it always gets polled after 6 mins. New code would allow for up to 2 missed maxInterval 10 mins reports before falling back to polling).

### DIFF
--- a/poll_manager.cpp
+++ b/poll_manager.cpp
@@ -472,7 +472,6 @@ void PollManager::pollTimerFired()
 
     size_t fresh = 0;
     const int reportWaitTime = 360;
-    const int reportWaitTimeXAL = 60 * 30;
 
     // check that cluster exists on endpoint
     if (clusterId != 0xffff)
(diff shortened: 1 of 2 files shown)


poll_manager.cpp
```diff
@@ -509,20 +508,18 @@ void PollManager::pollTimerFired()
                                 }
 
                                 NodeValue &val = restNode->getZclValue(clusterId, attrId);
+                                quint16 maxInterval = val.maxInterval > 0 && val.maxInterval < 65535 ? (val.maxInterval * 3 / 2) : reportWaitTime;
 
+                                // This lines mean IKEA is never polled (or just once)?
                                 if (lightNode && lightNode->manufacturerCode() == VENDOR_IKEA && val.timestamp.isValid())
                                 {
                                     fresh++; // rely on reporting for ikea lights
                                 }
```
manup, line 517: Hmm strange I could swear that for IKEA lights polling was used until reporting kicks in.
I think it should be safe to remove this IKEA specific handling and just use the general approach.

ebaauw: If memory serves, you did something special for IKEA, since the bulb firmware would hang when polled too often.  Not sure if this has been fixed in recent firmware versions.

ebaauw (commented):
Seems like a good idea, but I don't think I can oversee the impact.  The polling and re-issuing the bindings/reporting configuration have always been like magic to me.

poll_manager.cpp
SwoopX: I pretty much had the similar discussion with Manuel on the IKEA stuff, but for reporting. There, it is explicitly prevented after the first report dripped in. Not sure if the mentioned side effects also occur on polling.

manup: It's just a gut feeling but the latest IKEA firmware versions appear more stable, haven't experienced issues in the last two months.

SwoopX: So, gone they are...

manup:
Merged to check how it behaves in my network, I have 4x IKEA GU10 CW and 2x IKEA E27 dimable lights.

PR #4779, opened 2021-04-16
Fix ResourceItem setValue(QString) for DataTypeTime

manup:
When not using the `setValue(QVariant)` or `setValue(qint64)` the internal `m_num` wasn't updated.

This PR properly sets the `m_num` value. Allowed input formats are "yyyy-MM-ddTHH:mm:ss.zzz" and "yyyy-MM-ddTHH:mm:ss", the time is assumed to be in UTC.

Note that regardless of the input format `toString()` and `toVariant().toString()` always return UTC format "yyyy-MM-ddTHH:mm:ss.zzz".

The issue was detected in a work in progress unit test https://github.com/manup/deconz-rest-plugin/blob/device_descriptions/tests/101-resourceitem-dt-time.cpp

### DIFF
--- a/resource.cpp
+++ b/resource.cpp
@@ -651,6 +651,11 @@ bool ResourceItem::setValue(const QString &val)
 {
     if (m_str)
     {
+        if (m_rid->type == DataTypeTime)
+        {
+            return setValue(QVariant(val));
+        }
+
         m_lastSet = QDateTime::currentDateTime();
         m_flags |= FlagNeedPushSet;
         if (*m_str != val)
@@ -736,10 +741,14 @@ bool ResourceItem::setValue(const QVariant &val)
     {
         if (val.type() == QVariant::String)
         {
-            QDateTime dt = QDateTime::fromString(val.toString(), QLatin1String("yyyy-MM-ddTHH:mm:ss"));
+            const auto str = val.toString();
+            auto fmt = str.contains('.') ? QLatin1String("yyyy-MM-ddTHH:mm:ss.zzz")
+                                         : QLatin1String("yyyy-MM-ddTHH:mm:ss");
+            auto dt = QDateTime::fromString(str, fmt);
 
             if (dt.isValid())
             {
+                dt.setTimeSpec(Qt::UTC);
                 m_lastSet = now;
                 m_numPrev = m_num;
                 m_flags |= FlagNeedPushSet;


ebaauw (commented):
Did you check where we actually do a `setValue()` on a DataTime resource attribute with a string value?  Are you sure it isn't for attributes in local time, (like `config.localtime`).

I really think we should be using `Z` to indicate UTC.

manup:
> Did you check where we actually do a `setValue()` on a DateTime resource attribute with a string value? Are you sure it isn't for attributes in local time, (like `config.localtime`).

Yes, I've checked for usage of this, also searching for `setValue(QVariant(`, from what I see `setValue(QString)` and `setValue(QVariant(QString))` on `DataTypeTime` are never used in the current code base. The `config.localtime` is set via `setValue(QDateTime)` and internally stored as UTC but converted back to local time when used in rules and schedules.

Prior to this PR, during testing  I was surprised too see the following (my timezone UTC offset is 2 hours):

```
item.setValue(QVariant("2021-04-16T18:20:20"));
item.toString() == "2021-04-16T16:20:20";
```

Here the input value was transformed from local time to UTC time, the `toString()` function always returns an UTC timestamp. My reasoning is that other than `config.localtime` all `ResourceItem` based timestamps are indeed UTC.

* For `setValue(qint64)` the input is UTC milliseconds since Epoch and stored directly in `m_num`;
* For `setValue(QVariant(QDateTime))` the input can be either local time or UTC but is always stored as UTC milliseconds since Epoch in `m_num`.

> I really think we should be using `Z` to indicate UTC.

I like that since it becomes clear that the inputs and outputs are either UTC or local time and clients don't have to guess what timestamps represent which is only written in the docs. For internal handling and storage in database this is no problem, but I'm concerned to change the REST API timestamps yet without breaking clients. Perhaps we should plan for this to be changed in a few months in version x.y.z and announce it so that clients can test it and are prepared?

ebaauw:
> My reasoning is that other than `config.localtime` all `ResourceItem` based timestamps are indeed UTC.

There's also a `localtime` in schedules, but that doesn't include the date part.  That is very relevant, because you want the schedule to adjust to daylight saving time.

> I'm concerned to change the REST API timestamps yet without breaking clients.  Perhaps we should plan for this to be changed in a few months in version x.y.z and announce it so that clients can test it and are prepared?

Yes, please.  The documentation should refer to ISO-8601, which clearly support the time zone part, or `Z` for UTC.  Also, the microseconds, and even the seconds are optional.  See also the discussions around `lastseen`.

manup:
> > My reasoning is that other than `config.localtime` all `ResourceItem` based timestamps are indeed UTC.
> 
> There's also a `localtime` in schedules, but that doesn't include the date part. That is very relevant, because you want the schedule to adjust to daylight saving time.

I think we should be safe here since schedules have their own time handling unrelated to `ResourceItem`?

> > I'm concerned to change the REST API timestamps yet without breaking clients.  Perhaps we should plan for this to be changed in a few months in version x.y.z and announce it so that clients can test it and are prepared?
> 
> Yes, please. The documentation should refer to ISO-8601, which clearly supports the time zone part, or `Z` for UTC. Also, the microseconds, and even the seconds are optional. See also the discussions around `lastseen`.

ISO-8601 is already mentioned in the docs [/sensors](https://dresden-elektronik.github.io/deconz-rest-doc/endpoints/sensors/), but when the 'Z' support becomes active we should add a section which formats can be expected from the REST API with examples. There are some fun formats within the ISO-8601 range which we likely won't use :) I'm not sure how good or bad time parsers are in various languages to interpret a given ISO-8601 string.

... would actually be a nice test for Qt.

SwoopX (commented):
I'd also like to see all timestamps streamlined to UTC sometime, and also like adding the "Z", as this eliminates any room for interpretation.

> Perhaps we should plan for this to be changed in a few months in version x.y.z and announce it so that clients can test it and are prepared?

I would have tackled this pretty much the same and maybe, we should use that approach for other items as well (e.g. bri for the window covering) to finally get rid of certain unpopular parts.
As for any timestamp transition in the API, we also might want to consider using a new parameter under `/config`. That would oc introduce a bit more code, but the changes could be done right away while leaving the default setting in line with the current state. By that, we could switch as we please...

As a side note for `setValue(const QVariant &val)`, I was always wondering why this does not reflect the allowed/defined API data types. If we, in future, want to expose floats as well, this also needs to be extended (as I recall from the PoC I made back in the days on this one).

PR #4808, opened 2021-04-24
Electrical measurement: Improve handler performance

manup:
The call `sensor->modelId()` iterates over all ResourceItems and dereferences the QString pointer.
Since the value is used quite often it can be cached to prevent looping every time the `modelId` is used.

Related PR: https://github.com/dresden-elektronik/deconz-rest-plugin/pull/4802


SwoopX (approved):
Ah, could have seen that myself... So we should do it for everything else yet to come. I'll add that to the pending PR.

manup:
Yep I have noticed that also only yesterday for the first time. We have lots of places currently where we to this and can kill some CPU/memory cycles.

PR #4813, opened 2021-04-25
Move Xiaomi special reporting related code to dedicated file

SwoopX:
Move any code related to Xiaomi special reporting to a dedicated file. This shall ensure all relevant code is at one place for better overview and to reduce the size of `de_web_plugin.cpp`. Some code optimizations have also be included.


manup (approved):
I think this is ready to merge

PR #4818, opened 2021-04-27
Fix IKEA KADRILJ and FYRTUR roller blind reporting

manup:
The reporting could stop when the device rejoins the network for various reasons like parent loss or transmission errors.

This PR:
- Marks KADRIL and FYRTUR roller blinds as light sleeper with the new non public `ResourceItem` "attr/sleeper" (borrowed from Device code);
- Support polling light sleeper *Window Covering* devices `state.lift` while reporting isn't setup;
- Fix window covering update by ZCL report / read attributes value (was always set to by report);
- Fix reconfiguration of window covering bindings and ZCL reporting.

Issue: https://github.com/dresden-elektronik/deconz-rest-plugin/issues/1121

### DIFF
--- a/de_web_plugin.cpp
+++ b/de_web_plugin.cpp
@@ -2833,6 +2833,7 @@ void DeRestPluginPrivate::setLightNodeStaticCapabilities(LightNode *lightNode)
     }
 
     ResourceItem *item = nullptr;
+    const QString modelId = lightNode->modelId();
 
     if (lightNode->manufacturerCode() == VENDOR_LEDVANCE &&
             (lightNode->modelId() == QLatin1String("BR30 RGBW") ||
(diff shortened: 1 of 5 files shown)


ebaauw (approved):
I only have the FYRTUR, but I understand the FYRTUR and KADRILJ are the same Zigbee device, with the same firmware, except for the _Model Identifier_ and the _Product Code_.  

Same logic might be opportune for the Aqara B1.

manup:
Cool, thanks I've extended the PR to handle both.

zigomatichub:
Hello, I don't know if this is the right place to post. 
About the FYRTUR, in deconz application, on the cluster info in container Window Covering of the node,  the value Current Position Lift Percentage is not updated after executing a exec command. 
Is it possible to expose the reported position value in home assistant as an entity and having then refreshed correctly ?

PR #4850, opened 2021-05-08
Fix for missing Tradfri remote button events

SwoopX:
This PR fixes the missing button events upon < and > press. Most likely, a handful of further devices is affected by this as well.

The root cause is that for the newer 2.X firmware, the device doesn't natively expose the _scene_ cluster (only used on APS layer), which is hence not considered in the sensor fingerprint. Further devices where a cluster is not hidden, but not included in the sensor fingerprint, could be affected as well. As a consequence, the newly inttroduced `getSensorNodeForAddressEndpointAndCluster()` cannot select the correct sensor. The subsequent code has a more or less appropriate fallback to select the fitting sensor, but gets discarded.

### DIFF
--- a/de_web_plugin.cpp
+++ b/de_web_plugin.cpp
@@ -1126,7 +1126,13 @@ void DeRestPluginPrivate::apsdeDataIndication(const deCONZ::ApsDataIndication &i
                     }
                     else
                     {
-                        sensorNode = 0; // not supported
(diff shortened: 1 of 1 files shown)


SwoopX:
Sure thing. I consider addint the clusters more than a documentation aid for the DDF and a minimal performance tweak compared to the now active fallback. In future, we luckily don't need the fingerprints anyway anymore.

PR #4858, opened 2021-05-10
Prevent automatic client cluster binding to "all" group

SwoopX:


SwoopX:
Hm... Then what about extending the function with an additional, optional parameter just used while binding? Minimum impact/effort.

PR #4881, opened 2021-05-16
Make Regulation SetPoint Offset usable as offset for Danfoss Ally

SwoopX:
This PR allows setting the manufacturer specific attribute 0x404B (Regulation SetPoint Offset) via the offset resource item exposed through REST API. It has an allowed range of +-2.5° C. Using it will basically make the thermostat over-/underheat accoding to the set value, thereby automatically adjusting the set heat setpoint.

Includes some further minor enhancements.

### DIFF
--- a/rest_sensors.cpp
+++ b/rest_sensors.cpp
@@ -1218,8 +1218,8 @@ int DeRestPluginPrivate::changeSensorConfig(const ApiRequest &req, ApiResponse &
                     }
                     else if (ok && (sensor->modelId() == QLatin1String("eTRV0100") || sensor->modelId() == QLatin1String("TRV001")))
                     {
-                        if      (offset <= -25) { offset = -25; }
-                        else if (offset >= 25)  { offset = 25; }
+                        if      (offset < -25) { offset = -25; }
+                        else if (offset > 25)  { offset = 25; }
                         
                         if (addTaskThermostatReadWriteAttribute(task, deCONZ::ZclWriteAttributesId, VENDOR_DANFOSS, 0x404B, deCONZ::Zcl8BitInt, offset))
                         {
@@ -1235,8 +1235,8 @@ int DeRestPluginPrivate::changeSensorConfig(const ApiRequest &req, ApiResponse &
                     }
                     else if (ok)
                     {
-                        if      (offset <= -25) { offset = -25; }
-                        else if (offset >= 25)  { offset = 25; }
+                        if      (offset < -25) { offset = -25; }
+                        else if (offset > 25)  { offset = 25; }
                         
                         if (addTaskThermostatReadWriteAttribute(task, deCONZ::ZclWriteAttributesId, 0, 0x0010, deCONZ::Zcl8BitInt, offset))
                         {
--- a/rest_sensors.cpp
+++ b/rest_sensors.cpp
@@ -1216,22 +1216,42 @@ int DeRestPluginPrivate::changeSensorConfig(const ApiRequest &req, ApiResponse &
                             return REQ_READY_SEND;
                         }
                     }
-                    else
+                    else if (ok && (sensor->modelId() == QLatin1String("eTRV0100") || sensor->modelId() == QLatin1String("TRV001")))
                     {
-                        if (ok && addTaskThermostatReadWriteAttribute(task, deCONZ::ZclWriteAttributesId, 0, 0x0010, deCONZ::Zcl8BitInt, offset))
+                        if      (offset <= -25) { offset = -25; }
+                        else if (offset >= 25)  { offset = 25; }
+                        
+                        if (addTaskThermostatReadWriteAttribute(task, deCONZ::ZclWriteAttributesId, VENDOR_DANFOSS, 0x404B, deCONZ::Zcl8BitInt, offset))
                         {
-                            rspItemState[QString("set %1").arg(rid.suffix)] = offset;
-                            rspItem["success"] = rspItemState;
+                            updated = true;
                         }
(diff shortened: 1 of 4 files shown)


rest_sensors.cpp
```diff
@@ -1216,22 +1216,42 @@ int DeRestPluginPrivate::changeSensorConfig(const ApiRequest &req, ApiResponse &
                             return REQ_READY_SEND;
                         }
                     }
-                    else
+                    else if (ok && (sensor->modelId() == QLatin1String("eTRV0100") || sensor->modelId() == QLatin1String("TRV001")))
                     {
-                        if (ok && addTaskThermostatReadWriteAttribute(task, deCONZ::ZclWriteAttributesId, 0, 0x0010, deCONZ::Zcl8BitInt, offset))
+                        if      (offset <= -25) { offset = -25; }
+                        else if (offset >= 25)  { offset = 25; }
```
manup, line 1222:
```
if      (offset < -25) { offset = -25; }
else if (offset > 25)  { offset = 25; }
```
:)

rest_sensors.cpp
```diff
@@ -1216,22 +1216,42 @@ int DeRestPluginPrivate::changeSensorConfig(const ApiRequest &req, ApiResponse &
...
+                            rsp.list.append(errorToMap(ERR_ACTION_ERROR, QString("/sensors/%1/config/%2").arg(id).arg(pi.key()),
+                                                       QString("Could not set attribute")));
+                            rsp.httpStatus = HttpStatusBadRequest;
+                            return REQ_READY_SEND;
+                        }
+                    }
+                    else if (ok)
+                    {
+                        if      (offset <= -25) { offset = -25; }
+                        else if (offset >= 25)  { offset = 25; }
```
manup, line 1239:
```
if      (offset < -25) { offset = -25; }
else if (offset > 25)  { offset = 25; }
```

rest_sensors.cpp
```diff
@@ -1216,22 +1216,42 @@ int DeRestPluginPrivate::changeSensorConfig(const ApiRequest &req, ApiResponse &
...
+                        
+                        if (addTaskThermostatReadWriteAttribute(task, deCONZ::ZclWriteAttributesId, 0, 0x0010, deCONZ::Zcl8BitInt, offset))
```
manup, line 1241: Would be nice to add constants for the attribute ids  like 0x0010. There are quite a bunch in the surrounding switch statement, could be part of a later PR.

thermostat.cpp
```diff
@@ -517,15 +518,15 @@ void DeRestPluginPrivate::handleThermostatClusterIndication(const deCONZ::ApsDat
...
-                    if ( mode == 0x09 ) { mode_set = QString("sleep"); }
+                    mode_set = QLatin1String("off");
+                    if ( mode == 0x01 ) { mode_set = QLatin1String("auto"); }
+                    else if ( mode == 0x03 ) { mode_set = QLatin1String("cool"); }
+                    else if ( mode == 0x04 ) { mode_set = QLatin1String("heat"); }
+                    else if ( mode == 0x05 ) { mode_set = QLatin1String("emergency heating"); }
+                    else if ( mode == 0x06 ) { mode_set = QLatin1String("precooling"); }
+                    else if ( mode == 0x07 ) { mode_set = QLatin1String("fan only"); }
+                    else if ( mode == 0x08 ) { mode_set = QLatin1String("dry"); }
+                    else if ( mode == 0x09 ) { mode_set = QLatin1String("sleep"); }
```
manup, line 529: Interesting, might be a good candidate for the fancy get mapped value function (doesn't have to be for this PR).

To streamline code style I'd prefer to remove the spaces in the `if` condition and use cameCase instead snake_case.

```
QString modeSet;

if      (mode == 0x01) { modeSet = QLatin1String("auto"); }
else if (mode == 0x03) { modeSet = QLatin1String("cool"); }
else if (mode == 0x04) { modeSet = QLatin1String("heat"); }
else if (mode == 0x05) { modeSet = QLatin1String("emergency heating"); }
else if (mode == 0x06) { modeSet = QLatin1String("precooling"); }
else if (mode == 0x07) { modeSet = QLatin1String("fan only"); }
else if (mode == 0x08) { modeSet = QLatin1String("dry"); }
else if (mode == 0x09) { modeSet = QLatin1String("sleep"); }
else                   { modeSet = QLatin1String("off"); }
```

Note, only the else branch set's the fallback "off" value, no need to set it every time just to overwrite it :)
This raises the question, is every not checked value really "off"?

manup (commented):
:+1:  for using more QLatin1String :)

PR looks fine function wise, added some small style comments.

rest_sensors.cpp
SwoopX: ok, I let that one count ;)

rest_sensors.cpp
SwoopX: Same here

rest_sensors.cpp
SwoopX: Yeah, got that on the list for the refactor.

thermostat.cpp
SwoopX: I had exactly the same thought about the mapping function. Regarding the `off` value, I'd say let's keep that for the moment, shouldn't cause any harm. With the new function, that topic's gone anyway.

Regarding the placement of constants and variables, we could make some decisions on how and where to place them best when we talk next time.

Happy to include the further code beauty.

manup: Perfect, thanks :)

PR #4889, opened 2021-05-17
Initial support for Sinope Smart in-wall outlet – Zigbee (SP2610ZB)

ericvandamme:
Tested and working with Sinope Smart in-wall outlet – Zigbee SP2610ZB (https://www.sinopetech.com/en/product/smart-in-wall-outlet-zigbee/). Should address issue #4880 as well.

Issue #4880, opened 2021-05-15
SINOPE SP2610ZB Wall Smart-Plug


joelwener:
## Device
![PXL_20210515_144209398](https://user-images.githubusercontent.com/84113418/118371726-eb680600-b57b-11eb-8855-68b41a841334.jpg)

- Product name: The device name as shown on the product or package.
- Manufacturer: Sinope.
- Model identifier: SP2610ZB
- Device type : Please remove all unrelated device types. 
  - Other: Smart Plug

### Identity
![PXL_20210515_144242716](https://user-images.githubusercontent.com/84113418/118371876-86f97680-b57c-11eb-87b8-d004e17a7d87.jpg)

### Basic
![PXL_20210515_144234772](https://user-images.githubusercontent.com/84113418/118371884-911b7500-b57c-11eb-8a63-a149a867dddb.jpg)

### On/Off
![PXL_20210515_144251003](https://user-images.githubusercontent.com/84113418/118371895-a98b8f80-b57c-11eb-8b21-8684cff3d7be.jpg)

### Simple Metering
![PXL_20210515_144259924](https://user-images.githubusercontent.com/84113418/118371906-c0ca7d00-b57c-11eb-86b7-5636d3a70ce1.jpg)

### Electrical Measurement
![PXL_20210515_144319021](https://user-images.githubusercontent.com/84113418/118371925-d6d83d80-b57c-11eb-9828-1196b2d751eb.jpg)

### Sinope Specific
![PXL_20210515_144338075](https://user-images.githubusercontent.com/84113418/118371950-f96a5680-b57c-11eb-85c5-ef4136442223.jpg)

Issue #4871, opened 2021-05-13
SINOPE sp2600ZB Smart Plug 


joelwener:
## Device

- Product name: Sinope smart plug
- Manufacturer: Sinope
- Model identifier: SP2600ZB
- Device type :
  - Other: Smart Plug
![16209922419165701589213238983235](https://user-images.githubusercontent.com/84113418/118265480-3e13c600-b487-11eb-926e-511fdba63c49.jpg)

![16209924844925179654626174079037](https://user-images.githubusercontent.com/84113418/118265878-cdb97480-b487-11eb-9927-1926202612c2.jpg)

### Basic
![16209922653184901393567160361820](https://user-images.githubusercontent.com/84113418/118265554-584da400-b487-11eb-8c5f-11782d0b46bf.jpg)

### Identify
![16209923126711544619292872252917](https://user-images.githubusercontent.com/84113418/118265606-68fe1a00-b487-11eb-980f-fed9f7fe832e.jpg)

### Alarms
NA
### Device Temperature
NA
### Groups
NA
### Scenes
NA
### On/Off
![16209923484767793678412355579444](https://user-images.githubusercontent.com/84113418/118265659-803d0780-b487-11eb-8056-4f07d0acb773.jpg)

### Level Control
NA
### Color Control
NA
### Simple Metering
![16209923848238655894612709316820](https://user-images.githubusercontent.com/84113418/118265707-934fd780-b487-11eb-9228-1332b4d7a75a.jpg)

### Diagnostics
NA
### Other clusters that are not mentioned above
![16209924133047843580763105881120](https://user-images.githubusercontent.com/84113418/118265768-a662a780-b487-11eb-854e-ca4a3bcc4c15.jpg)

![16209924494283130900143423690824](https://user-images.githubusercontent.com/84113418/118265809-b9757780-b487-11eb-8cb7-e4dbd1a7ad85.jpg)

joelwener:
Missing on/off
![PXL_20210513_030444862](https://user-images.githubusercontent.com/84113418/118071836-927b5080-b376-11eb-9373-29b7dbe1382a.jpg)

joelwener:
Smart plug / controller may qualify as sensor as well.

joelwener:
To be clear, I would like the Phoscon I/f to recognize the following cluster points (as either sensors or readable points of the smart plug/controller)

Current Temperature

Electrical Measurement Cluster
RMS Current
RMS Voltage
[...]

PR #4943, opened 2021-05-29
Addition of Sinope SP2600ZB, SP2610ZB, RM3250ZB Smart Plugs, Smart Controller

joelwener:
added the devices to be recognized, data properly converted.

### DIFF
--- a/bindings.cpp
+++ b/bindings.cpp
@@ -1902,7 +1902,10 @@ bool DeRestPluginPrivate::sendConfigureReportingRequest(BindingTask &bt)
             modelId == QLatin1String("SKHMP30-I1") ||              // GS smart plug
             modelId.startsWith(QLatin1String("E13-")) ||           // Sengled PAR38 Bulbs
             modelId.startsWith(QLatin1String("Z01-A19")) ||        // Sengled smart led
-            modelId == QLatin1String("Connected socket outlet"))   // Niko smart socket
+            modelId == QLatin1String("Connected socket outlet") ||   // Niko smart socket
+	    modelId == QLatin1String("SP2600ZB") ||                  // Sinope Smart Plug
+	    modelId == QLatin1String("SP2610ZB") ||                  // Sinope Smart Wall Plug
+	    modelId == QLatin1String("RM3250ZB"))                    // Sinope Smart Controller
         {
             rq.reportableChange48bit = 10; // 0.001 kWh (1 Wh)
         }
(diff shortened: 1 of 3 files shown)


joelwener:
These changes complete device addition requests: #4889 #4880 #4871

SwoopX (changes requested):
Thanks for raising this PR. We appreciate that you are eager to get the devices supported, but for the SP2610ZB, this will end up in a bit of a mess due to the Salami tactics. As this and the other associated PR do contain common and individual changes, this will most likely result in duplicate entries and merge conflicts. In that sense, please either fully include the device with all the required changes or spare it out.

As for the content related changes, it looks already pretty good. Please correct indentation throughout the amendments of `bindings.cpp` (it's spaces, not tabs). What's then still missing is the explicit exclusion of all 3 devices in the 2 files mentioned here https://github.com/dresden-elektronik/deconz-rest-plugin/pull/4889#pullrequestreview-664786246 so they do not get an unnecessary `power` resource item in the ZHAConsumption sensor.

joelwener:
Hi,

I didn't include instantaneous power demand, as it is not included in the
devices.  I did included the kWh consumption (current Summation) which
includes the ZhAConsumption sensor.  Not sure why you are indicating to
exclude it.

All devices as far as I am concerned are included completely, I don't
believe I have left anything out.

I will fix the spacing in bindings.cpp

manup:
I can't tell wherever the `RStatePower` item needs to be excluded @SwoopX what's your view here? I remember we had devices which reported bogus values for some attributes, is this the reason?

SwoopX (changes requested):
@manup The reason is quite simple: what is not available, shouldn't be exposed.

By default, `RStateConsumption` and `RStatePower` get created for `ZHAConsumption` sensors and none of the 3 devices supports power there. I can already anticipate the future issues that power is always 0 (we had some for other devices), so I'll stick to my previous comment. This is, I'm afraid, no matter of believe 😉 

Apart from that, everything else looks fine.

manup:
> By default, `RStateConsumption` and `RStatePower` get created for `ZHAConsumption` sensors and none of the 3 devices supports power there. I can already anticipate the future issues that power is always 0 (we had some for other devices), so I'll stick to my previous comment. This is, I'm afraid, no matter of believe wink

Perhaps, we should consider changing the defaults for `ZHAConsumption` and only add the items explicitly when the device actually supports them.

SwoopX:
Yeah, I would appreciate that although I haven't dared yet since I don't know if our supported devices documentation is good enough for that 🤔

BabaIsYou:
Ma y be completing with a DDF like  PR #5962 for SP2610ZB ould unblock this one until  final state  about useless  Power reported ?

BabaIsYou:
Perhaps adding  DDF file  like in PR  #5962 will  permit to release this PR until final state  on this useless  reported power  ?

PR #4947, opened 2021-05-30
Fix: Consider ZCL frame control direction for sending default response

SwoopX:


manup:
Why the extra parameter, does it differ from the direction value in `zclFrame::frameControl()`?

SwoopX:
No sure if I get what you mean. However, discussing this, you can probably just revert the direction without the parameter...

manup:
Perfect, I think this should do the trick.

PR #4950, opened 2021-05-30
Expose `RConfigDelay` instead of `RConfigDuration` for Develco motion sensors

SwoopX:
Develco/frient motion sensors MOSZB-1XX currently have `RConfigDuration` exposed to set presence from `true` to `false`. However, this is not necessary, as the device sends an attribute report with occupancy `false` when no motion is detected within the timeframe specified in `PIR Occupied To Unoccupied Delay` (0x0010). Based on that, `RConfigDelay` is the more accurate choice to set the desired timeframe when the sensor should report unoccupied state.

As the same mechanism like for the Hue motion sensors will be used, `RConfigPending` will also be exposed for the occupancy sensor endpoints. The presence sensor based on the IAS Zone cluster will keep `RConfigDuration`.

### DIFF
--- a/database.cpp
+++ b/database.cpp
@@ -3276,7 +3276,8 @@ static int sqliteLoadAllSensorsCallback(void *user, int ncols, char **colval , c
             {
                 clusterId = clusterId ? clusterId : OCCUPANCY_SENSING_CLUSTER_ID;
                 if (sensor.modelId().startsWith(QLatin1String("FLS")) ||
-                    sensor.modelId().startsWith(QLatin1String("SML00")))
+                    sensor.modelId().startsWith(QLatin1String("SML00")) ||
+                    sensor.modelId().startsWith(QLatin1String("MOSZB-1")))
                 {
                     // TODO write and recover min/max to db
                     deCONZ::NumericUnion dummy;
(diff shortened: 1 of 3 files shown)


manup:
For the Hue motion sensor `SML0..` there is extra initialisation done when loading from database, is this needed for the `MOSZB-1XX` as well?

SwoopX:
Where specifically? I checked all occurrences for the `SML00` and did the necessary changes. The device works with the code as far as I can tell.

manup:
In the following lines, I'm not sure if it is important to do this 0 initialization for `RConfigDelay` since the value is set to the value from the database later on anyway.

https://github.com/dresden-elektronik/deconz-rest-plugin/blob/2a8e5aa4163ebf462c3ff095c20e08a5eb20493c/database.cpp#L3308-L3316

SwoopX:
> 
> 
> In the following lines, I'm not sure if it is important to do this 0 initialization for `RConfigDelay` since the value is set to the value from the database later on anyway.

Well, the 0 is not required as it either gets set by the DB or read by the plugin. It's more meant to prevent a fugly `null` to occur for a brief time iirc ()especially since `RConfigDuration gets discarded`? Regardless, I'm happy to remove the 0 initialization if you like.

manup:
>Regardless, I'm happy to remove the 0 initialization if you like.

The 0 init doesn't hurt for now, lets keep it an cleanup later via Device code :)

SwoopX:
Sure 😄

PR #4951, opened 2021-05-30
Fix: Emit button event 1002 again for TRADFRI remote control after mode switch

SwoopX:
When pushing the on/off button of the switch approx. 5 seconds, the device changes the operational mode so that the arrow buttons do not send any commands on hold/long press. Under certain conditions (apparently when the arrow keys haven't been used before), this can lead to button event 1002 not being emitted, although the corresponding command is send.

As the respective code was intended for the STYRBAR remote, it is now restricted to that device.

### DIFF
--- a/de_web_plugin.cpp
+++ b/de_web_plugin.cpp
@@ -4608,7 +4608,7 @@ void DeRestPluginPrivate::checkSensorButtonEvent(Sensor *sensor, const deCONZ::A
                         }
                     }
                 }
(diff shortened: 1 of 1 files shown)


Nindaleth:
Could you please fix a typo related to this PR in the [changelogs](https://phoscon.de/en/changelog/)? Both English and German changelogs list this fix as #5951 instead of #4951 which gives 404 instead of the proper thing. Thanks!

SwoopX:
Pinged @manup

manup:
Thanks for the hint, link is corrected now.

PR #4961, opened 2021-06-01
Partially harmonize light/sensor selection logic

SwoopX:
With introduction of `isSameAddress()`, a consistent address comparison became availble. Currently, quite a few functions to select alight/sensor have a similar code implemented which varies in some details.

With this PR, the address comparision gets streamlined for the functions `getLightNodeForAddress()`, `getSensorNodeForAddress()`, `getSensorNodeForAddressAndEndpoint()` and `getSensorNodeForAddressAndEndpoint()` while sparing a few lines of code

### DIFF
--- a/de_web_plugin.cpp
+++ b/de_web_plugin.cpp
@@ -9858,7 +9858,7 @@ Sensor *DeRestPluginPrivate::getSensorNodeForAddress(const deCONZ::Address &addr
 {
     for (Sensor &sensor: sensors)
     {
-        if (sensor.deletedState() != Sensor::StateNormal || !sensor.node()) { continue; }
+        if (sensor.deletedState() != Sensor::StateNormal)                   { continue; }
         if (!isSameAddress(sensor.address(), addr))                         { continue; }
 
         return &sensor;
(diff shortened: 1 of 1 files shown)


de_web_plugin.cpp
```diff
@@ -9888,47 +9856,14 @@ Sensor *DeRestPluginPrivate::getSensorNodeForAddress(quint64 extAddr)
...
+    for (Sensor &sensor: sensors)
     {
-        for (; i != end; ++i)
-        {
-            if (i->address().nwk() == addr.nwk() && i->deletedState() != Sensor::StateDeleted)
-            {
-                return &(*i);
-            }
-        }
+        if (sensor.deletedState() != Sensor::StateNormal || !sensor.node()) { continue; }
```
manup, line 9861: The `!sensor.node()` check is good intention but this would break at least ZGP devices and Sensors where the `deCONZ::Node` hasn't been assigned yet. Albeit I'd prefer to have this check here the current code base is more structured to have the node() check outside this function.

manup:
Just checked the ZGP code, in the button handler the `getSensorNodeForAddress()` is called with `sensor.node() == nullptr` in

https://github.com/dresden-elektronik/deconz-rest-plugin/blob/4eed6b1a5e020c5d1da06250f46a8ade0a4e880e/de_web_plugin.cpp#L1442-L1445

If the `sensor.node()` check is removed in that function the PR should be fine with the small tweak mentioned in review comments.

manup:
Tested with Friends of Hue switch, all good :+1:

PR #5025, opened 2021-06-16
Add initial support for Linkind Door/Window Sensor ZS110050078

SwoopX:


manup:
Such a simple PR and still git screams :/

SwoopX:
Per device PRs are nice, but if you choose the wrok position to place the code, PRs interfere with each other :/

PR #5080, opened 2021-06-30
Add custom Node Descriptor response handling

manup:
With next firmware version and deCONZ v2.12.2-beta, Node Descriptor requests are forwarded to the application.

The handler allows to customize the response for specific devices.


Smanar:
If  I m right with this code it will possible to simulate a fake R22 stack ?
Screen_Shot_2021-03-17_at_19 54 06

manup:
>If I m right with this code it will possible to simulate a fake R22 stack?

Kind of, we can at least present to the device a Node Descriptor claiming R22. Note that the current firmware runs on R21 so devices which see R22 might expect different behavior.

PR #5089, opened 2021-07-04
Streamline event handling.

ebaauw:
Streamline the implementation of event handling for `/lights` and `/groups` vs `/sensors`.  Properly implement `lastseen` and `lastannounced` for `/sensors`.

### Changes
- `Sensor`:
  - Create `SetValue()` to update `ResourceItem` and issue event (cf. `LightNode`);
  - Add `RAttrLastAnnounced` and `RAttrLastSeen` on sensor creation;
  - Overwrite `rx()` to update `lastseen`
  - Collect `attr` attributes in single websocket notification.
- `database.cpp`: remove `config/reachable`, `lastseen`, and `lastannounced` from CLIP sensors (and Daylight sensor).
- `Resource`: lose `lastStatePush` and `lastAttrPush`;
- `rest_lights.cpp` and `rest_groups.cpp`: use `ResourceItem::needPushChange()` instead of `Resource::lastStatePush`.

### TODO
- Persist and restore `lastseen` and `lastannounced` for `/sensors`;
- Collect `attr` attributes into single websocket notification for `/groups`;
- Streamline database tables (separate columns for `state` vs `config` for sensors; single nested `items` columns for lights);
- Streamline use of `ResourceItem`s for top-level attributes (like `manufacturer` and `modelid`.

### FOLLOW UP:
- Use `Sensor::SetValue()` instead of `ResourceItem::SetValue()`.


SwoopX:
I'm afraid the current state code gives me a SIGSEGV

grafik

zdp/zdp_handlers.cpp
```diff
@@ -160,6 +160,7 @@ void DeRestPluginPrivate::handleDeviceAnnceIndication(const deCONZ::ApsDataIndic
         if (si->address().ext() == ext)
         {
             si->rx();
+            si->setValue(RAttrLastAnnounced, i->lastRx().toUTC());
```
SwoopX, line 163: Overlooked that little sweety twice... Needs to be `si->lastRx().toUTC()`

ebaauw: Thanks, still haven't seen a _Device Announcement_ for any `/sensors` resource myself.

manup:
For the `Sensor::setValue()` I'd like to decouple this from the plugin so that `Sensor` (and `LightNode`) don't depend on it.

I have an idea for this but need to make some scribbles and tests, I think it's possible to create a small function which works for `Sensor` and `LightNode` and still have it testable (hence no plugin dependency).

ebaauw:
I moved the `setValue(const char *suffix, ... val, bool forceUpdate)` methods to `Resource`, made `Resource` an abstract class by declaring `virtual void didSetValue(ResourceItem *i) = 0`; and implemented `didSetValue()` in the `Resource` subclasses: `LightNode`, `Sensor`, `Group`, and `ApiConfig`.

We need an additional constructor `Event::Event(const char *resource, const char *what, ResourceItem *item)` for `/config` events, that don't have an `id`.  But I don't think we do rules nor websocket notifications on `/config` attributes yet.

ebaauw:
Added `lastseen` and `lastannounced` columns to the `sensors` table to persist the values across restarts.

I think the PR is good to go.

rest_lights.cpp
```diff
@@ -2975,30 +2975,32 @@ void DeRestPluginPrivate::handleLightEvent(const Event &e)
...
                 }
             }
 
             if (ix && ix->lastSet().isValid() && iy && iy->lastSet().isValid())
             {
                 if (gwWebSocketNotifyAll ||
-                    (ix->lastChanged().isValid() && ix->lastChanged() >= lightNode->lastStatePush) ||
-                    (iy->lastChanged().isValid() && iy->lastChanged() >= lightNode->lastStatePush))
+                    (ix->lastChanged().isValid() && ix->needPushChange()) ||
+                    (iy->lastChanged().isValid() && iy->needPushChange()))
```
manup, line 2990: I think we don't need the `.lastChanged().isValid()` anymore. To my understanding when `.needPushChange()` is set, it implies that the timestamp is valid.

rest_lights.cpp
```diff
@@ -3056,17 +3058,17 @@ void DeRestPluginPrivate::handleLightEvent(const Event &e)
                     continue;
                 }
 
-                if (item->lastSet().isValid() && (gwWebSocketNotifyAll || (item->lastChanged().isValid() && item->lastChanged() >= lightNode->lastAttrPush)))
+                if (item->lastSet().isValid() && (gwWebSocketNotifyAll || (item->lastChanged().isValid() && item->needPushChange())))
```
manup, line 3061: As mentioned above, I'd ditch the timestamp checks, here and in further places as well.

```c++
if (item->needPushSet() && (gwWebSocketNotifyAll || item->needPushChange()))
{
}
```

rest_sensors.cpp
```diff
@@ -341,6 +341,10 @@ int DeRestPluginPrivate::createSensor(const ApiRequest &req, ApiResponse &rsp)
         sensor.setSwVersion(map[QLatin1String("swversion")].toString());
         sensor.setType(type);
 
+        sensor.removeItem(RConfigReachable);
```
manup, line 344: Interesting, I'm not sure if `RConfigReachable` is used in some CLIP sensors. It is allowed by the POST REST API call.

rest_sensors.cpp
```diff
@@ -2657,6 +2658,7 @@ void DeRestPluginPrivate::handleSensorEvent(const Event &e)
             if (!(item->needPushSet() || item->needPushChange()))
             {
                 DBG_Printf(DBG_INFO_L2, "discard sensor config push for %s (already pushed)\n", e.what());
+                webSocketServer->flush(); // force transmit send buffer
```
manup, line 2661: I get the intention, but imho it isn't the responsibility of this branch to flush the socket, since obviously there isn't anything to do here. I'd propose to leave the flushing to the places where there's actually something to flush.

manup:
Quick thought on database decoupling for `setValue()`, e.g.:

```c++
plugin->saveDatabaseItems |= DB_SENSORS;
plugin->queSaveDb(DB_SENSORS, DB_SHORT_SAVE_DELAY);
```

We can move this to `DeRestPluginPrivate::handleEvent(const Event &e)` in the plugin (see https://github.com/dresden-elektronik/deconz-rest-plugin/pull/5127). More precisely, to specific handlers like `handleSensorEvent()`. With the logic: If data is pushed to the Websocket, due an event, schedule the database store.

This would result in fewer calls to the database code with the same result.

rest_lights.cpp
ebaauw: OK, happy to move it, but this is older code, that this PR hasn't touched.

rest_sensors.cpp
ebaauw: Again, happy to remove it, but this was copied from handling the `state` and `config` attributes.  Same logic also in lights.

rest_sensors.cpp
ebaauw: Double-checked: the Hue bridge shows `config/reachable` for CLIP sensors.  I'll ditch this line.

rest_sensors.cpp
ebaauw: Ditched 6 calls to `flush()` in three files.

rest_lights.cpp
ebaauw: I don't think we need to check for `needPushSet()` here?

rest_lights.cpp
ebaauw: Removed calls to `item->lastSet().isValid()` and `item->lastChanged().isValid()`.

ebaauw:
Made all changes and use `enqueueEvent()` global function. However the dependency on `plugin` remains, because of `gwLightLastSeenInterval`.

SwoopX:
Does that make sense here?
https://github.com/dresden-elektronik/deconz-rest-plugin/blob/27b9d20df4be70574e88cdb89749453237f9384a/resource.cpp#L701-L708

ebaauw:
Apparently it does, to allow null values.  Has been there for 3 years, , see https://github.com/dresden-elektronik/deconz-rest-plugin/commit/cbf571dd9ffd1d20bb6261d041f9bd857b7b577b.

SwoopX:
Hm... It doesn't get a push, not sure if it should anyway...

Other than that, I had the version from yesterday running in my dev for quite a while, seems to work fine.

rest_sensors.cpp
```diff
@@ -2763,19 +2768,44 @@ void DeRestPluginPrivate::handleSensorEvent(const Event &e)
...
-            const char *key = item->descriptor().suffix + 5;
-            map[key] = item->toVariant();
+            QVariantMap attr;
+
+            for (int i = 0; i < sensor->itemCount(); i++)
+            {
+                item = sensor->itemForIndex(i);
+                const ResourceItemDescriptor &rid = item->descriptor();
+
+                if (strncmp(rid.suffix, "attr/", 5) == 0) {
```
manup, line 2791: Curly brace on the next line :)

rest_sensors.cpp
```diff
@@ -2763,19 +2768,44 @@ void DeRestPluginPrivate::handleSensorEvent(const Event &e)
...
+                    const char *key = item->descriptor().suffix + 5;
```
manup, line 2792: In a future "performance" PR we should replace this, and similar places, by:

`const QLatin1String key(item->descriptor().suffix + 5)`

Background: The call to `attr[key]` creates a new string via `QString(const char*)`, which does a `fromUtf8()` conversion. Since all keys are ASCII, the faster `QLatin1String` constructor can be used.

rest_sensors.cpp
```diff
@@ -2763,19 +2768,44 @@ void DeRestPluginPrivate::handleSensorEvent(const Event &e)
...
+                    if (gwWebSocketNotifyAll || item->needPushChange())
+                    {
+                        attr[key] = item->toVariant();
+                        item->clearNeedPush();
+                    }
+                }
+            }
 
-            webSocketServer->broadcastTextMessage(Json::serialize(map));
+            if (!attr.isEmpty()) {
```
manup, line 2802: Curly brace on the next line.

manup (commented):
Added a few comments.

Overall I found no breaking issues. I like that the PR brings some real performance improvements with the decoupling. By using the push/set flag logic the code got much more readable imho :)

I'll put it in my test networks this week, and consider it for merging in [v2.12.4-beta](https://github.com/dresden-elektronik/deconz-rest-plugin/milestone/17).

rest_sensors.cpp
```diff
@@ -2404,6 +2406,8 @@ bool DeRestPluginPrivate::sensorToMap(const Sensor *sensor, QVariantMap &map, co
                 state[key] = item->toVariant();
             }
         }
+        else if (item->descriptor().suffix == RAttrLastAnnounced) { map["lastannounced"] = item->toString(); }
+        else if (item->descriptor().suffix == RAttrLastSeen) { map["lastseen"] = item->toString(); }
```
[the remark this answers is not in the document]
ebaauw: Check

rest_groups.cpp
```diff
@@ -3449,10 +3449,9 @@ void DeRestPluginPrivate::handleGroupEvent(const Event &e)
         ResourceItem *item = group->item(e.what());
         if (item)
         {
-            if (group->lastStatePush.isValid() && item->lastSet() < group->lastStatePush)
+            if (!(item->needPushSet() || item->needPushChange()))
             {
                 DBG_Printf(DBG_INFO_L2, "discard group state push for %s: %s (already pushed)\n", qPrintable(e.id()), e.what());
```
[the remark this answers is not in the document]
ebaauw: Check

rest_sensors.cpp
```diff
@@ -2558,7 +2558,6 @@ void DeRestPluginPrivate::handleSensorEvent(const Event &e)
             if (!(item->needPushSet() || item->needPushChange()))
             {
                 DBG_Printf(DBG_INFO_L2, "discard sensor state push for %s: %s (already pushed)\n", qPrintable(e.id()), e.what());
```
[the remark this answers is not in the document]
ebaauw: Check, 3x.

rest_lights.cpp
```diff
@@ -2940,10 +2940,9 @@ void DeRestPluginPrivate::handleLightEvent(const Event &e)
         ResourceItem *item = lightNode->item(e.what());
         if (item)
         {
-            if (lightNode->lastStatePush.isValid() && item->lastSet() < lightNode->lastStatePush)
+            if (!(item->needPushSet() || item->needPushChange()))
             {
                 DBG_Printf(DBG_INFO_L2, "discard light state push for %s: %s (already pushed)\n", qPrintable(e.id()), e.what());
```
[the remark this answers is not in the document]
ebaauw: Check 2x.

rest_sensors.cpp
ebaauw: Check.  Fixed other occurrences as well.

rest_sensors.cpp
ebaauw: So not today...

rest_sensors.cpp
ebaauw: Check.  Multiple occurrences in this file as well.

manup (approved):
Seems to work here and events are properly sent :+1: 

A few points in the PR description todo list can be marked as resolved now.

PR #5090, opened 2021-07-04
Use `isSameAddress()` for address comparison

SwoopX:
This PR is intended to make use of `isSameAddress()` more often when address comparisons are required and thereby save some lines of code.

There's further places where `ext()` and its existance is checked exclusively, but I'm not sure if it's appreciated to change those as well (would probably result in a slight processing overhead).

### DIFF
--- a/de_web_plugin.cpp
+++ b/de_web_plugin.cpp
@@ -1308,8 +1308,7 @@ void DeRestPluginPrivate::apsdeDataConfirm(const deCONZ::ApsDataConfirm &conf)
 
         if (conf.dstAddressMode() == deCONZ::ApsNwkAddress &&
             task.req.dstAddressMode() == deCONZ::ApsNwkAddress &&
-            conf.dstAddress().hasNwk() && task.req.dstAddress().hasNwk() &&
-            conf.dstAddress().nwk() != task.req.dstAddress().nwk())
+            !isSameAddress(conf.dstAddress(), task.req.dstAddress()))
         {
             DBG_Printf(DBG_INFO, "warn APSDE-DATA.confirm: 0x%02X nwk mismatch\n", conf.id());
             //continue;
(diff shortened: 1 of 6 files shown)


ebaauw:
Why not implement an `==` operator on address?

manup:
The `==` operator is already implemented but is very strict and requires every field to be equal, which makes sense to compare two instances for example in a vector. The `isSameAddress()` function prefers the MAC address if available and falls back to NWK address which is useful for our purposes.

```cpp
bool Address::operator==(const Address &other) const
{
    return hasNwk() == other.hasNwk() &&
           hasExt() == other.hasExt() &&
           hasGroup() == other.hasGroup() &&
           nwk() == other.nwk() &&
           ext() == other.ext() &&
           group() == other.group();
}
```

poll_manager.cpp
```diff
@@ -163,16 +163,9 @@ void PollManager::apsdeDataConfirm(const deCONZ::ApsDataConfirm &conf)
...
+    if (!isSameAddress(dstAddr, conf.dstAddress()))
     {
-
-    }
-
-    else if (dstAddr.hasNwk() && conf.dstAddress().hasNwk()
-        && dstAddr.nwk() != conf.dstAddress().nwk())
-    {
-
+        return;
```
manup, line 168: This is an interesting spot, I think the reason why I didn't add a `return;` back in the day was that the `confirm.id()` already uniquely matches the confirm to the request. In theory we can ditch the address check here altogether, but it doesn't hurt and in worst case would run in a small timeout and polling keeps going.

de_web_plugin.cpp
```diff
@@ -1308,8 +1308,7 @@ void DeRestPluginPrivate::apsdeDataConfirm(const deCONZ::ApsDataConfirm &conf)
 
         if (conf.dstAddressMode() == deCONZ::ApsNwkAddress &&
             task.req.dstAddressMode() == deCONZ::ApsNwkAddress &&
-            conf.dstAddress().hasNwk() && task.req.dstAddress().hasNwk() &&
-            conf.dstAddress().nwk() != task.req.dstAddress().nwk())
+            !isSameAddress(conf.dstAddress(), task.req.dstAddress()))
```
manup, line 1311: Nice! This is an improvement, tasks which only had the MAC address set weren't handled before.

manup (approved):
Beside simpler code the PR also brings some improvements and fixes cases which weren't handled before :+1:

SwoopX:
Like I said, there might be more places when we consider usage of `.ext()`, e.g. in `connectivity.cpp`

manup:
Doesn't compile,  `#include "utils/utils.h"` needs to be added in respective files.
Note the include should *not* be added to the `de_web_plugin_private.h` monster ;-)

manup (approved):

PR #5127, opened 2021-07-15
Add EventEmitter class to decouple enqueueEvent() from plugin

manup:
For now with a global `enqueueEvent()` free function, to not change hundret's lines of code. In future we can make this cleaner and get rid of the globals.

One timer call handles multiple events, if the total processing time is below 10 ms. This is done because one timer activation might take +-10 ms until called by Qt (borrowed from Device code).

The event queue uses `std::vector` instead `std::queue`, which is several times faster.

### DIFF
--- a/database.cpp
+++ b/database.cpp
@@ -4141,8 +4141,7 @@ static int sqliteLoadAllSensorsCallback(void *user, int ncols, char **colval , c
                         item = sensor.item(RConfigGroup);
                         if (item && !item->toString().isEmpty() && item->toString() != QLatin1String("0"))
                         {
-                            Event e(RSensors, REventValidGroup, sensor.id());
-                            d->enqueueEvent(e);
+                            enqueueEvent(Event(RSensors, REventValidGroup, sensor.id()));
                         }
                     }
                 }
--- a/de_web.pro
+++ b/de_web.pro
@@ -116,6 +116,7 @@ HEADERS  = bindings.h \
            de_web_plugin_private.h \
            de_web_widget.h \
            event.h \
+           event_emitter.h \
            fan_control.h \
            gateway.h \
            gateway_scanner.h \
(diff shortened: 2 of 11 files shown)


ebaauw (approved):
Looks good to me.

Crazy idea: pass the parameters to the `Event` constructor to `enqueueEvent()` instead, and create the `Event` there.  Doesn’t save any line of code, but does save seven characters on many lines, and might reduce the exposure of `Event`.

ebaauw:
I moved the `setValue(const char *suffix, ... val, bool forceUpdate)` methods to `Resource`, made `Resource` an abstract class by declaring `virtual void didSetValue(ResourceItem *i) = 0;` and implemented `didSetValue()` in the `Resource`'s subclasses: `LightNode`, `Sensor`, `Group`, and `ApiConfig`.

The current implementation of `LightNode::didSetValue(ResourceItem *i)` is:
```
void LightNode::didSetValue(ResourceItem *i)
{
    plugin->enqueueEvent(Event(RLights, i->descriptor().suffix, id(), i));
    plugin->updateLightEtag(this);
    setNeedSaveDatabase(true);
    plugin->saveDatabaseItems |= DB_LIGHTS;
    plugin->queSaveDb(DB_LIGHTS, DB_SHORT_SAVE_DELAY);
}
```
`Sensor` has a similar implementation; `Group` and `ApiConfig` don't call `setNeedSataDataBase()` (which is defined on `RestNodeBase`).

So to eliminate the dependency of the `Resource` subclasses on `DeRestPluginPrivate`, we'd need to move the etag and database logic to `enqueueEvent()` (which would then have the dependency on `plugin`), or create global functions for these as well.

To move the logic to `enqueueEvent()`, we'd need to know what resource type was changed by calling `Event::resource()` and probably doing `switch` on the result.  Ugly and badly maintainable (if we ever need another subclass of `Resource`).  I'd favour having global functions `update...Etag()`, and `queSaveDb()`.  I think `queSaveDb()` could set `savaDataBaseItems`.  I might be cleaner to create `LightNode::updateEtag()` instead of `updateLightEtag()`.

PR #5141, opened 2021-07-20
Add initial support for Xiaomi Aqara RTCGQ13LM high precision motion sensor

SwoopX:
See #5141.

The PR now explicitly handles indications from cluster 0xFCC0, but currently only for the Xiaomi special reporting and the sensitivity attribute for this device. It should also allow to configure the motion sensitivity with the same mechanism as the Hue motion sensors.


xiaomi.cpp
```diff
@@ -1,6 +1,110 @@
...
+                    item = sensor->item(RConfigSensitivity);
+
+                    if (item && item->toNumber() != sensitivity)
+                    {
+                        item->setValue(sensitivity);
+                        enqueueEvent(Event(RSensors, RConfigSensitivity, sensor->id(), item));
+                        updateSensorEtag(&*sensor);
+                        sensor->setNeedSaveDatabase(true);
+                        queSaveDb(DB_SENSORS, DB_SHORT_SAVE_DELAY);
+                    }
```
manup, line 95: Perhaps this part can be simplified a bit, now that https://github.com/dresden-elektronik/deconz-rest-plugin/pull/5089 is merged?

SwoopX: Sure, done.

manup (approved):

vinhthaipham:
i can be able to add this sensor to deconz phocon app (add sensor) but it was only shown in deconz GUI.
nothing shown in phocon app

Mimiix:
Which is normal @vinhthaipham. Rest api != phoscon.

vinhthaipham:
Thanks!! I got it

PR #5161, opened 2021-07-25
Do not use `durationDue` for Philips and Develco motion sensors

SwoopX:
Philips SML00X and Develco MOSZB-1XX motion sensors get `RConfigDelay` exposed which represents the occupied to unoccupied delay setting of the device. If no further motion is detected, they report occupied state `false` amongst regular attribute reporting intervals.

As the `durationDue` mechanism is used for those sensors as well, certain unwanted behavior can occur in rare cases, leading to a swift change in presence (`true` to `false` to `true`) rather than a constant `true`. This e.g. can happen, if the attribute report to prevent the automatic change from presence `true` to `false` is processed a second too late.

This PR omits usage of the `durationDue` feature for both motion sensors and additionally adds a 3 seconds grace time for late attribute reports from other motion sensors using the occupancy sensing cluster.

### DIFF
--- a/occupancy_sensing.cpp
+++ b/occupancy_sensing.cpp
@@ -86,22 +86,28 @@ void DeRestPluginPrivate::handleOccupancySensingClusterIndication(const deCONZ::
                 enqueueEvent(Event(RSensors, RStatePresence, sensor->id(), item));
                 stateUpdated = true;
 
-                const NodeValue &val = sensor->getZclValue(OCCUPANCY_SENSING_CLUSTER_ID, OCCUPIED_STATE);
-
-                // prepare to automatically set presence to false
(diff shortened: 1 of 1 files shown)


manup:
Looks good to me, the only (likely academic) caveat is that a presence `true` won't be reset to `false` if the SML00X goes offline or stops sending reports, for ...reasons. But I think that's ok since the Hue motion sensor is pretty robust in that sense.

SwoopX:
Yes, absolutely true. However, that mechanism catches going back to false. You cannot cover the other direction (go to true if there's issues whatsoever), so you die that one for sure 😉

manup (approved):

PR #5173, opened 2021-08-02
Enable APS ACKs for ZCL Default Response

manup:
For end devices it can happen that the APS request is enqueued in the firmware but never sent, in which case the APS confirm ends with a timout status code (0xF0).

It looks like a timing issue, where the plugin doesn't push the APS request fast enough, and the end device MAC data polls too fast for a response. Or when the end device's initial request required an APS ACK from the coordinator.

This can be seen for Samjin Motion sensor, where not a single ZCL Default Response made it to the device. I suspect a few other devices could be affected as well.

The fix is simple, with APS ACK enabled the firmware will just retry to send the response if it doesn't work on the first attempt.
This is more heavy, but given that the ZCL Default Response is kind of important in order to keep devices alive, worth it.

image

The sniffer screenshot shows the applied fix, after the marked packed.

### DIFF
--- a/aps_controller_wrapper.cpp
+++ b/aps_controller_wrapper.cpp
@@ -31,7 +31,7 @@ static bool ZCL_SendDefaultResponse(deCONZ::ApsController *apsCtrl, const deCONZ
     apsReq.setProfileId(ind.profileId());
     apsReq.setRadius(0);
     apsReq.setClusterId(ind.clusterId());
-    //apsReq.setTxOptions(deCONZ::ApsTxAcknowledgedTransmission);
+    apsReq.setTxOptions(deCONZ::ApsTxAcknowledgedTransmission);
 
     deCONZ::ZclFrame outZclFrame;
     outZclFrame.setSequenceNumber(zclFrame.sequenceNumber());


Smanar:
And what does this cause for exemple on the Samjin Motion sensor ?

SwoopX:
Hm, I guess that shouldn't have a too heavy impact anyway? If I'm not mistaken, this is only applicable for the direct communication path between a device and the coordinator and not for an indirect one (so having some router(s) in between). The latter should be intransparent for the coordinator, I assume.

I'd consider that as mutual assurance of reception and increase of transfer robustness.

manup:
Deep diving into the issue, it seems to me more problematic than thought, the above mentioned APS ACK works only if the device MAC data polls lazy enough, preferably multiple times, the later the better (so the APS response is queued in the firmware).

In case of Samjin, the sensor polls very quickly so even with APS ACK for ZCL Default Response it's a hit and miss.

I found one gem in the 802.15.4 MAC specification: The **Frame Pending** flag on MAC level which is usually set to false for Zigbee devices. In a test firmware I set this to true for outgoing APS frames:

image

So we get one more MAC data poll from devices. Nice in general.

But doesn't solve the Samjin problem. Not sure yet how to solve this, I'm thinking of handling the ZCL Default Response entirely in the firmware, which should be as quick as it can get, without any deCONZ or plugin in the way.

image

This screenshot shows a case where the last MAC data poll was made (because of the Frame Pending flag in the ACK frame).
But at this point in time the ZCL Default Response from the plugin wasn't queued in the firmware .... end never reached the device.


------------------------


On the other hand the device should still be awake for some time, it looks like the Zigbee stack doesn't take this into account and sends out responses only *after* the next MAC data poll happens. Not sure if this can be fixed easily, quite complex part of the stack.

manup:
Got some good new and some good news:

I refactored the firmware serial communication and the same in the deCONZ core, which is now up to 8 times faster and more robust.

The ZCL Default Response handling is now implemented directly in the firmware (for non OTA requests).
The responses are send very quickly, I think this might solve reliability issues with some devices. Basically a ZCL Default Response is always sent, when requested, even if deCONZ (or Zigbee2mqtt, ZHA, etc.) isn't running.

The case for SmartThings is also solved by this, since the ZCL Default Response is queued directly and available on the next MAC data poll.

image

Smanar:
>now up to 8 times faster

Lol, it s not an improvement but a miracle, or the initial code realy need it ^^

SwoopX:
That sounds promising after all 😄 Is the PR at hand then still required? I've added the change to my test network some days ago but tbh I haven't noted any real effect so far. Presumably, that's due to the variety of options you have out there. The majority of devices I have on my test network do apparently not require a default response (admittedly, I currently do not have to many in action).

Anyway, it's always good to hear some bugs got caught down the road. And with the default reponse now being part of the firmware not requiring the software to run, I'd say that's a huge benefit.

manup:
The new firmware only handles the ZCL Default Reponse for some commands, e.g. IAS Zone Status Notifications and ZCL profile wide commands like ZCL attribute reports. Having the heavy betty plugin ZCL Default Responder in place is still needed as it handles the rest.

To not break anything the firmware APS confirms any ZCL Default Reponse from deCONZ (or any other host application) if it was already handled by the firmware.

PR #5174, opened 2021-08-03
Fix: Correct erroneous bool value flip

SwoopX:
Affects door locks and eCozy thermostat


alexpomsft:
Thanks @SwoopX ! Looking forward to trying this out :)

PR #5210, opened 2021-08-15
Add sensitivity and selftest for Xiaomi smoke and gas sensor

SwoopX:
See #5210.

This PR adds `RConfigSensitivity` and `RConfigSelfTest` to the Honeywell smoke and gas sensors.

The selftest will be a hit and miss for the smoke sensor, as it doesn't give any kind of response or updates any attribute value (battery powered). As it also mac polls just every 15 secs (instead of the usual 7.5, a quite remarkable deviation), delivery of the command cannot be ensured. The gas sensor reacts instantly (mains powered). For the value to return to `false` after issuing a selftest, a 3 seconds timer is used.

Sensitivity can be handled better. For the smoke sensor, the command is retried every 10 secs until the corresponding attribute holding the sensitivity value is successfully read. The same scale for the allowed values and the meaning is applied as for the Hue motion sensor (0 - low, 1 - medium, 2 - high).

Additionally, the tamper state is removed, as both devices do not support it anyway. They have also been excluded from the IAS enrollment process, as they do not comply with the zigbee specs in this regard, so the devices do not get hammered with unnecessary requests (will be marked as enrolled).

### DIFF
--- a/database.cpp
+++ b/database.cpp
@@ -4002,6 +4002,13 @@ static int sqliteLoadAllSensorsCallback(void *user, int ncols, char **colval , c
             {
                 sensor.removeItem(RConfigBattery);
             }
+            
+            if (sensor.modelId() == QLatin1String("lumi.sensor_smoke") ||
+                sensor.modelId() == QLatin1String("lumi.sensor_natgas"))
+            {
+                sensor.addItem(DataTypeBool, RConfigSelfTest)->setValue(false);
+                item = sensor.addItem(DataTypeUInt8, RConfigSensitivity);
+            }
         }
         else if (sensor.modelId().startsWith(QLatin1String("tagv4"))) // SmartThings Arrival sensor
         {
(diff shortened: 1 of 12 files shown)


Prior99:
Awesome. Will this also enable the siren?

SwoopX:
Afaik, there isn't any capability to do this, so no.

7leman7:
Sorry for the noob question, but is there any way to get this working prior to merge into official deconz release (e.g. by manually copying certain files)?

SwoopX:
Sure, you can compile the corresponding branch anytime and exchange the original plugin with the compiled version.

jip67:
I’m not shure, but I think it will be available in the next beta release. Probably in September. I’m also a noob so maybe someone can confirm this?
Regards, 

Remi

7leman7:
> Sure, you can compile the corresponding branch anytime and exchange the original plugin with the compiled version.

Thanks for the quick response, compiling etc. is way beyond my understanding level currently, but I guess I'll have to educate myself.

As I'm running deConz as HA OS add-on I guess I won't be able to put my hands on beta releases unfortunately...

RealKanashii:
> Just wanted to bump this - I guess this functionality should be already merged in deCONZ 2.12.6, but how exactly do I access sensitivity configuration and trigger self test?

Doesn't seem merged in the master ...

SwoopX:
Replaced by #6245

PR #5233, opened 2021-08-21
Fix: Warning for Develco siren SIRZB-110 via REST API

SwoopX:
Develco concatenates the subfields differently. Instead of mode (4 bits), strobe (2 bits) and level (2 bits), it must be level (2 bits), strobe (2 bits) and mode (4 bits).

So:
"Standard": b00010011 -> 0x13 -> 19
Develco: b11000001 -> 0xC1 -> 193

### DIFF
--- a/rest_lights.cpp
+++ b/rest_lights.cpp
@@ -2216,7 +2216,7 @@ int DeRestPluginPrivate::setWarningDeviceState(const ApiRequest &req, ApiRespons
             }
             else if (taskRef.lightNode->modelId() == QLatin1String("SIRZB-110"))    // Doesn't support strobe
             {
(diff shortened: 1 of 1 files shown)


ebaauw:
Are you sure it does for all firmware versions?

SwoopX:
> 
> 
> Are you sure it does for all firmware versions?

Joker, of course not 😃 Apart from the fact that only a handful of people know how to get the actual firmware version directly from the device (if it's not printed on the packaging) and Develco does not exactly what I'd call throwing firmware files at people, you can never be sure. However, I got confirmation that this is working and the bit sets are handled as described above.

The rather philosophical question which approach is the correct and intended one, however, will remain unanswered.

PR #5246, opened 2021-08-25
Corrections on sensor fingerprints

SwoopX:
All sensors now get the basic cluster added (previously, only switches got it), air quality sensors receive the power configuration cluster and all devices get the OTAU and time client cluster, if available. Besides the corrections, there should also be a minor performance increase as unnecessary sensor iterations are thereby prevented.

### DIFF
--- a/de_web_plugin.cpp
+++ b/de_web_plugin.cpp
@@ -5275,16 +5275,9 @@ void DeRestPluginPrivate::addSensorNode(const deCONZ::Node *node, const deCONZ::
                         }
                     }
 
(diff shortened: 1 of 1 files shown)


manup:
I haven't checked the full code yet, but would it be possible add the Basic cluster just in the `addSensorNode()` to the fingerprint?

SwoopX:
Hm, sure. But is that sufficient so that already existing sensors get the fingerprint update? During my tests, it seemingly wasn't the case.

manup (approved):

PR #5247, opened 2021-08-25
Enable websocket server to listen on IP specified via `--http-listen` command line parameter

SwoopX:
The websocket server can now be bound to the IP address specified via command line, similar to the regular webserver. If the parameter is not provided or the IP is invalid/not found, it will default back to listen on any IP.

The example below has the webserver running on port 8080 and the websocket server on port 8888. This change is already running in my production environment 😉 

grafik

### DIFF
--- a/websocket_server.cpp
+++ b/websocket_server.cpp
@@ -10,6 +10,8 @@
 
 #ifdef USE_WEBSOCKETS
 
+#include "de_web_plugin.h"
+#include "de_web_plugin_private.h"
 #include "deconz/dbg_trace.h"
 #include "websocket_server.h"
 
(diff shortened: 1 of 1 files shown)


websocket_server.cpp
```diff
@@ -10,6 +10,8 @@
 
 #ifdef USE_WEBSOCKETS
 
+#include "de_web_plugin.h"
+#include "de_web_plugin_private.h"
```
manup, line 14: Are these included for `deCONZ::appArgumentString()`?

`#include "deconz/util.h"`  should do the trick :)

SwoopX: Indeed. Seems to work 👍

websocket_server.cpp
```diff
@@ -27,21 +28,31 @@ WebSocketServer::WebSocketServer(QObject *parent, quint16 port) :
         ports[0] = port;
     }
 
-    while (!srv->listen(QHostAddress::AnyIPv4, ports[p]))
+    QHostAddress address;
+    if (deCONZ::appArgumentString("--http-listen", 0).isEmpty()) // Tie websocket server to specified IP, if any
```
manup, line 32: Almost there, I've noticed the second default argument should be a string instead 0.

So in this case:

`deCONZ::appArgumentString("--http-listen", QString())`

manup (approved):

PR #5255, opened 2021-08-27
Enable alarm systems in rules; return alarm systems in full config request

manup:
- With this PR alarm systems can be used as a trigger in rules via `state/armstate`.
- Further alarm systems can be armed and disarmed by rules, e.g. arm at a certain time.
- All alarm systems are returned in the `GET /api/<apikey>` full config response.

Example to turn a light on when the alarm system gets `armed_away`:

```json
{
    "actions": [
        {
            "address": "/lights/15/state",
            "body": {
                "bri": 255,
                "on": true
            },
            "method": "PUT"
        }
    ],
    "conditions": [
        {
            "address": "/alarmsystems/1/state/armstate",
            "operator": "dx"
        },
        {
            "address": "/alarmsystems/1/state/armstate",
            "operator": "eq",
            "value": "armed_away"
        }
    ],
    "name": "Armed away"
}
```

### DIFF
--- a/database.cpp
+++ b/database.cpp
@@ -6009,7 +6009,7 @@ bool DB_LoadSecret(DB_Secret &secret)
         return false;
     }
 
-    return true;
+    return !secret.secret.empty();
 }
 
 static bool initSecretsTable()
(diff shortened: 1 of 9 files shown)


leonardpitzu:
The alarm system succesfully triggers the siren (or light). I have one question though - is there a `not in` conditional operator? As in when them the `alarmsystem` is not in the state `in_alarm` turn of the siren.

PR #5276, opened 2021-09-01
Don't reset Sonoff SNZB-03 presence

erikced:
Attempt to determine whether an IAS Zone motion sensor sends restore reports and only queue presence reset if it does not.

### DIFF
--- a/ias_zone.cpp
+++ b/ias_zone.cpp
@@ -133,6 +133,38 @@ static void IAS_EnsureValidState(ResourceItem *itemIasState)
     }
 }
 
+/*! Configure presence restoration timer */
+static void IAS_QueueRestorePresence(Sensor *const sensor, const ResourceItem &presence)
+{
+    const NodeValue &val = sensor->getZclValue(IAS_ZONE_CLUSTER_ID, IAS_ZONE_CLUSTER_ATTR_ZONE_STATUS_ID);
+    const ResourceItem *const duration = sensor->item(RConfigDuration);
+    if (val.maxInterval > 0)
+    {
+        sensor->durationDue = presence.lastSet().addSecs(val.maxInterval);
+    }
+    else if (duration && duration->toNumber() > 0)
+    {
+        sensor->durationDue = presence.lastSet().addSecs(duration->toNumber());
+    }
+}
+
+/*! Check whether a sensor sends Zone Status Change when an alarm is reset */
+static bool IAS_SensorSendsRestoreReports(const Sensor &sensor, const quint16 zoneStatus)
+{
+    if (zoneStatus & STATUS_RESTORE_REP)
+    {
+        return true;
+    }
+    const std::array<const QLatin1String, 5> supportedSensors = {
+        QLatin1String("TY0202"),
+        QLatin1String("MS01"),
+        QLatin1String("MSO1"),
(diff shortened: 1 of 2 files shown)


SwoopX:
So the device does support restore reports? You can check the IAS zone cluster in deconz GUI.

I'm not too convinced about the proposed solution yet although I've to admit, I was into the same direction fixing some misbehavior of the occupancy based motion sensors. I feel it would be cleaner extracting the restore report bit from a zone status notification in `processIasZoneStatus()` of `ias_zone.cpp` as this is the place where it all starts with `durationDue` 🤔

erikced:
> So the device does support restore reports? You can check the IAS zone cluster in deconz GUI.

Well, the Restore Reports attribute is not set but it _does_ appear to send them anyway

When motion is detected:
```
09:58:51:630 [INFO] - No button map for: MS01, unicast to: 0x0000, endpoint: 0x01, cluster: IAS_ZONE (0x0500), command: STATUS_CHANGE (0x00), payload: 010000640000, zclSeq: 214
```
When motion is no longer detected:
```
10:02:24:412 [INFO] - No button map for: MS01, unicast to: 0x0000, endpoint: 0x01, cluster: IAS_ZONE (0x0500), command: STATUS_CHANGE (0x00), payload: 000000640000, zclSeq: 215

```
i.e. Alarm 1 is set as long there is motion and zone status change commands are sent when its value changes.

> I'm not too convinced about the proposed solution yet although I've to admit, I was into the same direction fixing some misbehavior of the occupancy based motion sensors. I feel it would be cleaner extracting the restore report bit from a zone status notification in `processIasZoneStatus()` of `ias_zone.cpp` as this is the place where it all starts with `durationDue` 🤔

Agreed, it would be better to avoid setting `durationDue` in the first place for sensors that send zone status change to reset. Even if, as is the case for this sensor, it doesn't properly announce that it does.

SwoopX:
Nice job. I just tested your code with a sensor which exposes restore reports bit and it ignored the duration setting nicely. So 👍

erikced:
> Nice job. I just tested your code with a sensor which exposes restore reports bit and it ignored the duration setting nicely. So 👍

Great! Thanks for testing.

manup:
In near future we should add this as `ResourceItem` or some kind of "capability flags" so that the code only needs to check the flag and act accordingly.

SwoopX:
Absolutely. With the DDFs, that should be a piece of cake but ease some pain for others in the meantime.

manup (approved):

trevorbenyack:
Hi, I was wondering if the result of this pull request would be visible through Home Assistant...? I'm on deCONZ V2.13.04 but the behavior of the sensor remains the same when exposes to Home Assistant (Also via Phoscon).

erikced:
> Hi, I was wondering if the result of this pull request would be visible through Home Assistant...? I'm on deCONZ V2.13.04 but the behavior of the sensor remains the same when exposes to Home Assistant (Also via Phoscon).

Sorry for the delay, I was still on my old patched version and had to adapt my setup to the new docker image to be able to test. The result should absolutely be visible through Home Assistant, at least it is for me with with 2.13.04. In general the sensor should report no motion after a one minute delay/hysteresis.
What behavior are you seeing?

PR #5279, opened 2021-09-03
Add support for POPP Smart Thermostat

SwoopX:
See #5278.

### DIFF
--- a/bindings.cpp
+++ b/bindings.cpp
@@ -1361,7 +1361,8 @@ bool DeRestPluginPrivate::sendConfigureReportingRequest(BindingTask &bt)
             return sendConfigureReportingRequest(bt, {rq, rq2, rq3, rq4, rq5});
         }
         else if (modelId == QLatin1String("eTRV0100") || // Danfoss Ally
-                 modelId == QLatin1String("TRV001"))     // Hive TRV
+                 modelId == QLatin1String("TRV001") ||   // Hive TRV
+                 modelId == QLatin1String("eT093WRO"))   // POPP smart thermostat
         {
             rq.dataType = deCONZ::Zcl16BitInt;
             rq.attributeId = 0x0000;       // local temperature
@@ -1591,7 +1592,8 @@ bool DeRestPluginPrivate::sendConfigureReportingRequest(BindingTask &bt)
     else if (bt.binding.clusterId == THERMOSTAT_UI_CONFIGURATION_CLUSTER_ID)
     {
         if (modelId == QLatin1String("eTRV0100") || // Danfoss Ally
-            modelId == QLatin1String("TRV001"))     // Hive TRV
+            modelId == QLatin1String("TRV001") ||   // Hive TRV
+            modelId == QLatin1String("eT093WRO"))   // POPP smart thermostat
         {
             rq.dataType = deCONZ::Zcl8BitEnum;
             rq.attributeId = 0x0001;       // Keypad Lockout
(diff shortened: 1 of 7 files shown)


J-N-K:
Is there something blocking this PR? I have several requests for support from my application‘s users.

Mimiix:
@J-N-K  Manups time.

Pe-MaKer:
Is there any chance of getting this merged soon?
I'm sitting on eight of these for my entire house and can't control them.
If I would know how to put the definition for this Popp derivative of the Danfoss Ally in a DDF, then it would be fine for me as well.

Pe-MaKer:
@manup  @Mimiix 
sorry for poking you. Is there any chance of getting this merged soon?
I'm sitting on eight of these thermostats for my entire house and can't control them. The winter is coming. 🥶
I tried to compile it from @SwoopX branch by myself but it's already 500 commits behind the master which brings other troubles.

If I would know how to put the definition for this Popp derivative of the Danfoss Ally in a DDF, then it would be fine for me as well.

Without your support I'm screwed and the WAF of our new home is at 0.

Pe-MaKer:
@manup @Mimiix 
I don't want to be a pain, but it would ruin our Advent season if I don't finally master the thermostats. 
My wife will ensure there's a fire🔥 otherwise. 

Can you please take care that this commit by @SwoopX will make it in the next beta?
@J-N-K did you managed to get a solution by yourself?

Mimiix:
@Pe-MaKer there's more PR s that are waiting longer. So it isn't fair to prioritize above...

Mimiix:
I'll ask Manuel, but no promises.

SwoopX:
@Pe-MaKer You can program the thermostats meanwhile via deconz GUI.

What you need to know is that you only have 6 transitions per day at your disposal. The time which needs to be entered is minutes since 00:00, het set point must be multiplied by 100. The last temperature set on a day will be taken over to the next day until the next change is programmed.
The example below has two periods with comfort temperature set, ending up with 4 transitions and applies for Mon - Fri with the following times:

05:50 -> 21°
09:00 -> 17°
17:00 -> 22°
21:00 -> 17°

grafik

When the schedule is programmed by pressing `exec`, you need to activate the schedules on the thermostat. This can be done by setting attribute 0x0025 as per below.

grafik

You can override the scheduled temp anytime manually on the device and this will remain until the next transition.

My recommendation would be to overshoot the aimed comfort temperature by 2° in the schedule for the beginning, so set 24° if you want to have 22°.

Pe-MaKer:
Thank you @SwoopX  for the detailed information.
I have made my own and current fork of the master, and incorporated all the changes from your commit there.
https://github.com/Pe-MaKer/deconz-rest-plugin/tree/POPP

Unfortunately, the time sync does not seem to work, as you can see from the screenshots.
If the time is given as Unix Epoch, the time of the termostat is about 35 years 
off.

This makes the use of time schedules a challenge ;)

2021-11-30 18_34_15-phoscon - VNC® Viewer for Google Chrome™
er-images.githubusercontent.com/44093812/144099527-a162555c-ee06-4dcb-8303-fe803eee3d6f.png)

SwoopX:
That more feels like the time sensor has not been created. Can you enter sensor search in Phoscon again and read the attributes of the basic cluster? After that, the sync should directly kick in and I know from own experience that it can be annoying like hell.

manup (approved):

PR #5325, opened 2021-09-26
Add Develco specific attributes to general.xml

SwoopX:
Add some manufacturer specific attributes for Develco IO module.


manup:
Develco has more mf specific stuff than Xiaomi :D

SwoopX:
> 
> 
> Develco has more mf specific stuff than Xiaomi :D

It just feels like that, slightly more than 10 attribs or so... But it's ultimately documented compared to Xiaomi 😂

PR #5392, opened 2021-10-18
Xiaomi multi endpoint fix

aryelevin:
A fix to avoid wrong state of multi-endpoint switches (Tested with Aqara S1 Scene Panel)


SwoopX:
Missed that one.

Tbh, I have my doubts this will be workable for all devices. As to my experience, the Xiaomi special reporting does always come from one endpoint, and all data contained might need to be distrubuted accross all resources a device presents.

As an example, a device has a resource on EP 0x02 and 0x15 and the special reporting comes from EP 0x01, then I'd assume the reporting will not be processed at all due to the missing EP match 🤔

aryelevin:
This fix affects other reports (Such as On/Off, I got wrong lights updated via socket while their state didn't change really).

SwoopX:
I hear you, but I believe this should be shapes differently. Let me give you an example.

grafik

That special report always comes from EP 1, but contains data data which also must be applied to EP 15 and 16. Applying your change, that data wouldn't reach EP 15 and 16 anymore, so it would fix your case but interfere with anything else and that's what I'm concerned about.

Issue #5433, opened 2021-10-31
Eurotronic Spirit ZigBee (SPZB0001) temperature not updated on device since deCONZ 2.12.6


double1968:
## Describe the bug
Temperature set in Home Assistant is never being sent to the device. Local changes are sent back to Home Assistant. As deCONZ sits in between and there were no firmware updates, we believe this issue has been introduced with v2.12.6 (where v2.11.5  not have this issue - full disclosure: I can guarantee the issue didn´t exist with addon v6.8.0/deCONZ v2.10.4).
## Steps to reproduce the behavior
Set new temperature in Home Assistant
Wait forever for the device to update
No temperature update on the valve 

## Expected behavior
Set new temperature in Home Assistant
Valve target temperature is being updated within max. 20 second

## Screenshots
![1](https://user-images.githubusercontent.com/48420012/139591381-b2730070-10f2-4d61-9c69-591f16457a67.PNG)
![2](https://user-images.githubusercontent.com/48420012/139591391-51fe0108-99c2-4822-a478-a5a6e6dfe3cb.PNG)

Reading on the net it seems that the problem is due to an incorrect writing of the value in the ID 0x0012 / 0x4003.
I remember after the DECONZ update if I changed the values via Phoscon the temperatures were transmitted to the valve

## Environment

 - Host system: Raspberry Pi
 - Running method:  Home Assistent deCONZ Add-on 
 - Firmware version: 26720700
 - deCONZ version: 2.12.6
 - Device: ConBee II
 - Do you use an USB extension cable: yes -- only relevant for ConBee I/II
 - Is there any other USB or serial devices connected to the host system? Yes : Zwave Dongle

## Additional context

See issue raised within HA repo (which is not the right place I think) here https://github.com/home-assistant/addons/issues/2244

See another user @bcutter with similar setup and same issue (closed for non-compliant) here https://github.com/dresden-elektronik/deconz-rest-plugin/issues/5422

Mimiix:
Asked manup to check :)

manup:
There are multiple firmware versions for the Spirit out there. Somehow the newer ones are "tricky".
Mine has Datecode 20181205 shown in the Basic Cluster.

Does writing the 0x4003 attribute work for you when done directly in deCONz Cluster Info Panel?
(double click on the attribute which open)

![image](https://user-images.githubusercontent.com/383386/139658081-cb61b08a-8c27-4bd4-864d-7009527353c6.png)

[...]

PR #5440, opened 2021-11-03
Fix Eurotronic Spirit heatsetpoint change for newer models

SwoopX:
See here for further details https://github.com/dresden-elektronik/deconz-rest-plugin/issues/5433#issuecomment-957240588

This PR ensures there're now 2 consecutive requests being send with different task and request IDs (confirmed to be send by sniffing). As it is still unknown which information of the device needs to be considered to undoubtly distinguish between the truly required request, this sort of hack should remain.

This PR also amends the evaluation of requests being on air. The limits are set to 2 for unicast and 6 for groupcast, previously effectively allowed were 1 and 5.

### DIFF
--- a/de_web_plugin.cpp
+++ b/de_web_plugin.cpp
@@ -12668,7 +12668,7 @@ void DeRestPluginPrivate::processTasks()
             {
                 onAir++;
 
-                if (onAir >= maxOnAir)
+                if (onAir > maxOnAir)
                 {
                     ok = false;
                     break;
@@ -12678,7 +12678,7 @@ void DeRestPluginPrivate::processTasks()
             {
                 onAir++;
                 int dt = idleTotalCounter - j->sendTime;
-                if (dt < 5 || onAir >= maxOnAir)
+                if (dt < 5 || onAir > maxOnAir)
                 {
                     if (dt > 120)
                     {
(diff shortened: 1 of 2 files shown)


Mimiix:
@manup Can this be in next updates as wel?

de_web_plugin.cpp
```diff
@@ -12668,7 +12668,7 @@ void DeRestPluginPrivate::processTasks()
             {
                 onAir++;
 
-                if (onAir >= maxOnAir)
```
manup, line 12671: I'm concerned with these, are they really needed for the issue that the PR fixes?

Pushing too many commands in the lower level APS queue may cause more congestion and waiting times.
Since the device code doesn't use the tasks anymore (later superseded by StateChange mechanic) I'd prefer to keep the current limits here until we have a proper tracking system.

SwoopX: Well, as described, then the limits are not correctly set or better to say it is irritating 😄 

Regardless, that is not really required for the fix, so I reverted it.

manup: Thanks :) I've tried to raise the limit in the past, causing weird issues in some networks.

manup (approved):

bcutter:
> @manup Can this be in next updates as wel?

What is a rough estimation on when the next stable containing this fix will be available?

It usually takes ages until they arrive @ https://github.com/home-assistant/addons/tree/master/deconz unfortunately.

PR #5465, opened 2021-11-07
Enforce skipping legacy code while in strict mode for numerous clusters

SwoopX:
In order to better test created DDFs and the associated new code processing, using strict mode will disable legacy data processing for numerous clusters.


manup:
Merging works but compiling the PR brings n error

SwoopX:
Fixed

manup (approved):

Issue #5590, opened 2021-12-20
Half brightness once Sengled LED is linked to deCONZ


steve30avs:
## Describe the bug

When a new Sengled E12-N1E color LED is added through Phoscon, the peak brightness drops to roughly half of what it was.

## Steps to reproduce the behavior

1. Plug in new Sengled E12-N1E or reset it to factory defaults (10 times on/off rapidly)
2. Search for new lights on the Phoscon website
3. Once the new bulb is added, watch its brightness
4. It should be roughly half of what it just was and cannot be raised higher through the brightness slider

## Expected behavior

Dragging the brightness slider to 100% should produce the whole 1300lm of brightness

## Screenshots

Photo of the connected bulb (left) vs an unconnected bulb (right):
![IMG_20211215_143327](https://user-images.githubusercontent.com/10735245/146721894-27817305-1d8e-4091-a127-084a32c8e066.jpg)

## Environment

 - Host system: Raspberry Pi
 - Running method: Raspbian
 - Firmware version: (26660700)
 - deCONZ version: (2.13.03)
 - Device: ConBee II
 - Do you use an USB extension cable: no
 - Is there any other USB or serial devices connected to the host system? Yes, USB mouse

## deCONZ Logs

INFO_L2 log:
11:39:37:993 Websocket 192.168.1.230:42322 send message: {"e":"changed","id":"19","r":"lights","state":{"alert":null,"bri":254,"colormode":"xy","effect":"none","hue":0,"on":true,"reachable":true,"sat":0,"xy":[0.402,0.3768]},"t":"event","uniqueid":"b0:ce:18:14:03:53:83:94-01"} (ret = 219)

## Additional context

All other connected bulbs (Ikea, Philips, and other Sengleds, appear to reach full brightness fine)

steve30avs:
Update: It looks like the lights reach full brightness when on 'Color Temperature' mode through Phoscon, but cannot reach full brightness when selecting a color.

Smanar:
Ha yes, I already see this issue on extented light.
It s because the device can use different color mode, XY or Hue/Sat/Bri.
It will be better if the "colormode" was "hs".

Are you able to try to change the color yourself using the API, using H/S/B instead of XY ? https://dresden-elektronik.github.io/deconz-rest-doc/getting_started/

[...]

Issue #4833, opened 2021-05-02
Color temperature issue with Singled Zigbee Lights


kkilchrist:
## Describe the bug

Singled RGB + Color temp bulbs no longer work properly. RGB colors can be selected just fine, but I am no longer able to activate color temperatures. This previously worked. 

Vendor: singled
Model: E12-N1E
Version: 0x00000015
Link: https://www.amazon.com/Sengled-E12-N1E-Smart-Light-White/dp/B07J6JK2BD

## Expected behavior

I should be able to control color temperature of these lights, in addition to RGB.

## Screenshots

Attached some screenshots.

## Environment

 - Host system: (Raspberry Pi / PC / NAS)
Raspberry Pi

 - Running method: (Raspbian / Ubuntu / Home Assistent deCONZ Add-on / Marthoc Docker container / Windows / Virtual Machine): 
Home Assistant DeCONZ add-on 6.8.0

 - Firmware version: (26xxyy00)
26680700

 - deCONZ version: (2.xx.yy)
2.10.04 

 - Device: (ConBee I / ConBee II / RaspBee I / RaspBee II)
ConBee II

 - Do you use an USB extension cable: (yes / no) -- only relevant for ConBee I/II
Yes

 - Is there any other USB or serial devices connected to the host system? If so: Which?
No

## Additional context

Attached photos of PhosCON info for bulb, as well as HA state for both a single bulb and a group of these bulbs.
![HA Single bulb](https://user-images.githubusercontent.com/11999003/116820177-71ba2c00-ab28-11eb-8879-e24eed298b96.png)
![Singled](https://user-images.githubusercontent.com/11999003/116820178-7252c280-ab28-11eb-8a17-4058d3723bca.png)
![HA group bulb](https://user-images.githubusercontent.com/11999003/116820179-7252c280-ab28-11eb-84e7-bf0aef14b313.png)

kkilchrist:
I should mention: if I manually call the light.turn_on service, e.g., 
```
service: light.turn_on
data:
  kelvin: 3100
target:
  area_id: living_room

```
The lights properly update to warm white

akusokuzan:
I am also having this problem! Is this issue being worked on or should I just return them? Also manually setting them doesn't work instead they turn a light pink looking color. The only way to reset them I've found is to delete them and re-add them.

kkilchrist:
The issue has actually gotten worse for me, as I can no longer manually
[...]

PR #5655, opened 2022-01-10
Added DDF for Sengled E12-N1E

taby1:
Previously color temperature support for this device was broken; added state/ct in DDF to allow this property to be properly picked up by api clients. Has been tested.

I believe this addresses #4833 and possibly #5590


Mimiix:
Can you please put the status on bronze?

taby1:
Done

PR #5681, opened 2022-01-18
added DDF for thirdreality 3RMS16BZ motion sensor

taby1:
Does not appear to read battery percentage correctly

### DIFF
--- a/devices/thirdreality/3RMS16BZ_motion_sensor.json
+++ b/devices/thirdreality/3RMS16BZ_motion_sensor.json
@@ -1,6 +1,10 @@
 {
   "schema": "devcap1.schema.json",
-  "manufacturername": "THIRDREALITY",
+  "manufacturername": [
+    "THIRDREALITY",
+    "Third Reality",
+    " Inc"
+   ],
   "modelid": "3RMS16BZ",
   "product": "Wireless Motion Sensor",
   "sleeper": true,
--- a/devices/thirdreality/3RMS16BZ_motion_sensor.json
+++ b/devices/thirdreality/3RMS16BZ_motion_sensor.json
@@ -1,0 +1,92 @@
+{
+  "schema": "devcap1.schema.json",
+  "manufacturername": "THIRDREALITY",
+  "modelid": "3RMS16BZ",
+  "product": "Wireless Motion Sensor",
+  "sleeper": true,
+  "status": "Bronze",
+  "path": "/devices/third_reality.json",
+  "subdevices": [
+    {
+      "type": "$TYPE_PRESENCE_SENSOR",
+      "restapi": "/sensors",
+      "uuid": [
+        "$address.ext",
+        "0x01",
+        "0x0006"
+      ],
+      "fingerprint": {
+        "profile": "0x0104",
+        "device": "0x0850",
+        "endpoint": "0x01",
+        "in": [
+          "0x0000",
+          "0x0001",
+          "0x1000"
+        ],
+        "out": [
+          "0x0006",
+          "0x0008"
+        ]
+      },
+      "items": [
+        {
+          "name": "attr/id"
+        },
+        {
+          "name": "attr/lastannounced"
+        },
+        {
+          "name": "attr/lastseen"
+        },
+        {
+          "name": "attr/manufacturername"
+        },
+        {
+          "name": "attr/modelid"
+        },
+        {
+          "name": "attr/name"
+        },
+        {
+          "name": "attr/swversion"
+        },
+        {
+          "name": "attr/type"
+        },
+        {
+          "name": "attr/uniqueid"
+        },
+        {
+          "name": "config/alert",
+          "description": "The currently active alert."
+        },
+        {
+          "name": "config/battery",
+          "parse": {
+            "at": "0x0021",
+            "cl": "0x0001",
+            "cppsrc": "power_configuration.cpp:192",
+            "ep": 1,
+            "eval": "Item.val = Attr.val"
+          }
+        },
+        {
+          "name": "config/duration"
+        },
+        {
+          "name": "config/on"
+        },
+        {
+          "name": "config/reachable"
+        },
+        {
+          "name": "state/lastupdated"
+        },
+        {
+          "name": "state/presence"
+        }
+      ]
+    }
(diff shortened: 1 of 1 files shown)


Smanar:
Hello, the DDF solve the issue or have the issue ?

taby1:
Previously the sensor didn’t work at all; now it works, but the battery reads 0%.

majorsl:
Ooooo, I did that device request #5337 - if there is a fix for the battery, I'll test it too.  Thanks for doing this DDF!

Smanar:
```
        {
          "name": "config/battery",
          "parse": {
            "at": "0x0021",
            "cl": "0x0001",
            "cppsrc": "power_configuration.cpp:192",
            "ep": 1,
            "eval": "Item.val = Attr.val"
          }
        },
```

Have you made the DDF with the editor or editing the text file ?
I think something more classic can work, and perhaps need a bind to this cluster/attribute ?

If you are using editor, can you show the result when seleting the config/battery ?

taby1:
I'm using the DDF editor; here's the battery power cluster information:
image

and here it is as shown in the DDF editor:
image
image

It is worth noting that sometimes, when the battery percentage field is manually read (after double-clicking it, etc.), it does read out correctly. So I guess the issue is that it is inconsistent and unreliable, rather than being completely broken.

Also, I'm not sure how much this matters, but I did notice that in the issue ticket above the manufacturer is "Third Reality, Inc", which is what the device enumerates itself as:
image
While the company [appears to brand itself ](https://smile.amazon.com/gp/product/B08RRRWK6B/ref=ppx_yo_dt_b_asin_title_o00_s00?ie=UTF8&psc=1)as THIRDREALITY (which is what this DDF uses).

Smanar:
>It is worth noting that sometimes, when the battery percentage field is manually read (after double-clicking it, etc.), it does read out correctly. So I guess the issue is that it is inconsistent and unreliable, rather than being completely broken.

Ok, so nice, it's normal, if it work when you read the value it mean the setting are good, just miss the bind/report for the device report them naturally periodically

You are missing something like that in the json

```
        {
            "bind": "unicast",
            "src.ep": 1,
            "cl": "0x0001",
            "report": [ {"at": "0x0021", "dt": "0x20", "min": 7200, "max": 7200, "change": "0x00" } ]
        }
```

With the editor https://github.com/dresden-elektronik/deconz-rest-plugin/wiki/DDF-cheat-sheet#bindings-panel


Also, I'm not sure how much this matters, but I did notice that in the issue ticket above the manufacturer is "Third Reality, Inc", which is what the device enumerates itself as:

Smanar:
BTW there is an issue for the manufacture   "Third Reality, Inc" we have this result on the ddf
```
  "manufacturername": [
    "Third Reality",
    " Inc"
  ],
```

Sp not sure it will work

manup:
According to the DDF this one seems to use THIRDREALITY as manufacturer name.
When the battery binding is added I think it's good to merge.

Smanar:
But on the capture here we can see "Third Reality, Inc" as manufacture name 

image

So IDK if 
`
"manufacturername": "THIRDREALITY",`

Can work ? And not possible to add "Third Reality, Inc" because of comma issue.

manup:
Hmm good question I thought the THIRDREALITY was in the Basic Cluster for this DDF?
If it is "Third Reality, Inc" the DDF won't be evaluated.

Smanar:
>Hmm good question I thought the THIRDREALITY was in the Basic Cluster for this DDF?

I can be sure, this "device" have at least 3 clones with thoses manufacture name
- Third Reality. Inc
- Third Reality
- a third brand I haven't found again but totaly different.

If the code compare only the starting word, can work with "Third Reality" and for first both, else not possible.

Smanar:
Ok so according to https://github.com/dresden-elektronik/deconz-rest-plugin/issues/5682#issuecomment-1025066191
It seem it can work even with a "bad DDF"

```
  "manufacturername": [
    "Third Reality",
    " Inc"
  ],
  "modelid": "3RWS18BZ",
```

taby1:
yeah not really sure what to say here; it works now, and doesn't with the DDF. Battery level is updating with the previous commit, but (as in #5682) appears to be divided by 2 for some reason. I will update the manufacturername to be an array as has been suggested.

Smanar:
He ...
It's a "break" ^^
The correct synthax is
```
  "manufacturername": [
    "M1",
    "M2"
  ],
  "modelid": [
    "I1",
    "I2"
  ],
```

M1/I1 is one device and M2/I2 is another one, but the problem is the M1 is "Third Reality, Inc", and the DDF don't escape the comma, so we have this result.
From the code it can't work, because we need the exact manufacture name, so only "Third Reality" need not work, but it seem it work ...

Smanar:
BTW, it's not possible using direclty a text editor ? To correct the manufacture name ?

majorsl:
> 

Shouldn't this limitation be fixed in deconz? ",inc" ",ltd" ",llc" and others would be common in company names that could be run into in the future.

Smanar:
>Shouldn't this limitation be fixed in deconz? ",inc" ",ltd" ",llc" and others would be common in company names that could be run into in the future.

Sure, but this part is coded on the deconz core, so only manup can do something, he already know the problem.
Ofc my proposition is just a workaround waiting for the corrective.

manup:
Small update on this, I really like to get this in and tune the DDF handling to support "foo, bar" as one string within the DDF editor.

Not sure yet about the best approach, ideally this becomes multiple input fields but this is quite cumbersome. Perhaps as in-between solution a different separator which hasn't been used yet in manufacturer names and model ids can be used, for example `;` or `|`.

Smanar:
It's not possible to "escape" the "," ?
Just with looking with " character ?

Smanar:
For information have made a PR for the manufacture name with the comma https://github.com/dresden-elektronik/deconz-rest-plugin/pull/5941 (it's not possible using editor, but possible using text editor)
So this PR can use just the classic manufacture name.

BabaIsYou:
Could this PR be closed as #5941 as been merged now ? This one will generated a duplicate in another folder.

Mimiix:
Yes.

PR #5707, opened 2022-01-24
DDF add ubisys S1

manup:
The ubisys S1 wasn't working for a few releases due some other code changes.
The DDF replaces most of the C++ implementation.

Note for the ubisys S1-R a separate DDF needs to be created, except if it is equal... needs to be tested.


ebaauw:
If memory serves, the S1-R has two inputs for switches, where the S1 has only one.

manup:
Ah makes sense, I was wondering why the `button_maps.json` did list them in one entry. I have split these into two so that the `buttonevent` introspection API returns correct number of buttons and events.

PR #5720, opened 2022-01-27
Add DDF for Develco/frient WISZB-120 open close sensor

SwoopX:


Mimiix:
@SwoopX these are only ddfs right? No c++ code involved?

manup (approved):

PR #5728, opened 2022-01-30
Add DDF : Third Reality 3RWS18BZ Water Leak Sensor

devWaves:
Third Reality 3RWS18BZ Water Leak Sensor - First commit

The device works but the battery percentage is reporting 1/2. Not sure why but it works well enough

### DIFF
--- a/devices/thirdreality/3RWS18BZ_water_sensor.json
+++ b/devices/thirdreality/3RWS18BZ_water_sensor.json
@@ -8,7 +8,7 @@
   "product": "3RWS18BZ",
   "sleeper": true,
   "status": "Bronze",
-  "path": "/devices/ThirdRealityWaterSensor.json",
+  "path": "/devices/thirdreality/3RWS18BZ_water_sensor.json",
   "subdevices": [
     {
       "type": "ZHAWater",
--- a/devices/thirdreality/3RWS18BZ_water_sensor.json
+++ b/devices/thirdreality/3RWS18BZ_water_sensor.json
@@ -1,14 +1,10 @@
 {
   "schema": "devcap1.schema.json",
-  "manufacturername": [
-    "Third Reality",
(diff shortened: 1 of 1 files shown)


zahti:
I just bought 2 of these sensors, and would like to start using them. Please let me know how I can help!

BabaIsYou:
A directory named /thrid_reality is already existing in /devices. May be this DDF could be deposit on that one too ? ;)

BabaIsYou:
> The device works but the battery percentage is reporting 1/2. Not sure why but it works well enough

You have to modify ` config/battery` that way because it reports 200 when batteries are full : 
```

        {
          "name": "config/battery",
          "parse": {
            "at": "0x0021",
            "cl": "0x0001",
            "ep": 1,
            "eval": "Item.val = Attr.val / 2"
          },
          "default": 0
        },
```

BabaIsYou (changes requested):
Suggestions : 
You have to modify ` config/battery` that way because it reports 200 when batteries are full : 
```

        {
          "name": "config/battery",
          "parse": {
            "at": "0x0021",
            "cl": "0x0001",
            "ep": 1,
            "eval": "Item.val = Attr.val / 2"
          },
          "default": 0
        },
```

Binding could also benefit to become to keep batteries longer: 
```
    {
      "bind": "unicast",
      "src.ep": 1,
      "cl": "0x0001",
      "report": [
        {
          "at": "0x0021",
          "dt": "0x20",
          "min": 7200,
          "max": 14400,
          "change": "0x00000002"
        }
      ]
    }
```
A directory named /third_reality is already existing in /devices. May be this DDF could be deposit on that one too ? ;)

BabaIsYou:
Think this PR could be closed as I included comments in #6826 and tested the DDF with three of these sensors on my own mesh.

PR #5757, opened 2022-02-06
DDF prevent load/store incomplete sensors with endpoint 0xFF

manup:
Auto generated DDF draft sensors — without a prior C++ implementation — have initially the ednpoint part set to FF in the uniqueid template.

This PR prevents loading and storing incomplete sensor of the database.
The problem was mostly even than the template was completed, two sensor resources existed with the FF one conflicting with the correct one.

One example is Wisers LK Switch: https://forum.phoscon.de/t/schneider-wiser-devices-not-working-in-dyna-version-2-14-00/1516


manup:
@Smanar there were other devices we've seen this, I think this PR should fix them as well.

PR #5758, opened 2022-02-06
Add DDF for Philips motion sensor SML003

SwoopX:


sn0wf4ke:
test procedure:
* deconz version: 14.4.0 (beta)
* copied to /usr/share/deCONZ/devices/philips/
* restarted container

* device showes up in phoscon, but with different icon then the "sml001":
image

* all clusters seem to work
* deleted and repaired -> works as expected
* looked for changes in rest api endpoints: 

image
image
image


good work!

manup:
Nice, since all seems to work we can also apply Gold status so it works in default setups which don't have Silver enabled.

Mimiix:
Great one to have in right before release. Well done guys

SwoopX:
> Nice, since all seems to work we can also apply Gold status so it works in default setups which don't have Silver enabled.

Ok, done.

PR #5806, opened 2022-02-16
Fixed REST response when setting a schedule's localtime attribute

### LINKED ISSUE #5787: Setting a schedule's localtime returns a defect result
## Describe the bug

When calling the API for setting a schedule's `localtime` (only), on success, the corresponding route's handler always tries to return/reflect `time` which is `null` in such a case, as it hasn't been part of the request. 
Also, the returned slug contains `"time"` instead of `"localtime"`.

## Steps to reproduce the behavior

Send this body via PUT method:
```
{"localtime": "W127/T12:00:00"}
```
Current behaviour:
```
[{"success":{"/schedules/1/time":null}}]
```

## Expected behavior

```
[{"success":{"/schedules/1/localtime":"W127/T12:00:00"}}]
```

## Environment

 - Host system: Raspberry Pi
 - Running method: Raspbian Buster
 - deCONZ version: 2.14.01
 - Python 3.7 REST client
 - Firefox developer tools HTTP client

## deCONZ Logs

/
## Screenshots 

/

## Additional context

I think the issue is caused by this line, that wrongly tries to reflect the `time` field of the request, even if it hasn't been given.
https://github.com/dresden-elektronik/deconz-rest-plugin/blob/80952d6c5bfc5a12d3527176602b8423948897de/rest_schedules.cpp#L669

Thank you and cheers!

meandthemachine:
This should fix #5787. Feel free to edit or let me know if I should change something. I haven't been programming C++ for 20 years. Thanks.

### DIFF
--- a/rest_schedules.cpp
+++ b/rest_schedules.cpp
@@ -666,9 +666,19 @@ int DeRestPluginPrivate::setScheduleAttributes(const ApiRequest &req, ApiRespons
 
             QVariantMap rspItem;
             QVariantMap rspItemState;
-            rspItemState[QString("/schedules/%1/time").arg(id)] = map["time"];
-            rspItem["success"] = rspItemState;
-            rsp.list.append(rspItem);
+            if (map.contains(QLatin1String("localtime")))
+            {
+                rspItemState[QString("/schedules/%1/localtime").arg(id)] = map["localtime"];
+                rspItem["success"] = rspItemState;
+                rsp.list.append(rspItem);
+            }
+            rspItemState.clear();
+            if (map.contains(QLatin1String("time")))
+            {
+                rspItemState[QString("/schedules/%1/time").arg(id)] = map["time"];
+                rspItem["success"] = rspItemState;
+                rsp.list.append(rspItem);
+            }
             rsp.httpStatus = HttpStatusOk;
         }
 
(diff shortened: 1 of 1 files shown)


meandthemachine:
@SwoopX Thanks a lot for reviewing. I will merge the `if` statements like you suggested. That's reasonable.

Regarding the `clear()`: I first tried without it, but if I set both, `localtime` as well as `time`, in the second `if` statement the map will then contain both key/value pairs and so falsely will the JSON response.

If you wonder why I even want to set both `time` and `localtime` in first place, that might be subject to another issue. `localtime` and `time` currently are not kept in sync (at least for my example of a weekly recurring "Wxxx/T" timer... seems its done only for absolute timers, probably because it's little easier there).
This means that if I only `PUT` one of them, the other one keeps its old value within the `jsonMap` that goes into the database. To me, that's irritating and kind of distractive when retrieving and inspecting a schedule's properties during development of my client.
Of course I could just ignore `time`, but in comparison, Phoscon app somehow seems to get it right - both properties are always in sync if I set a new time value via the web ui.

I'd like to suggest that we leave the `clear()` in for now and this way fix #5787 in a first step.
I then will post another issue on the syncing of `time` and `localtime` and some other little things I found in this method.

Somewhere in the code it says that "time is deprecated". Removing `time`, of course, would immediately get rid of the syncing issue and the `clear`. But that's something that needs to be decided by the maintainers and then also documented well (the API doc doesn't know `localtime` yet).

Thank you.

SwoopX:
Thanks for explaining. I guess the code could use a good overhaul, but regardless 🙂 Should be fine as it is now.

PR #5833, opened 2022-02-26
Securely generate network key when resetting adapters

puddly:
`qrand()` is not suitable for generating the network key when factory resetting an adapter.

The first call to `qrand()` sets the public PAN ID. This leaks enough RNG state to make bruteforcing the full state feasible from a single captured packet and thus allows the network key to be easily computed.

Unoptimized PoC: https://github.com/puddly/deconz-network-key-leak-poc

---

I can't figure out how to test this branch with the deCONZ Docker image so while the code compiles, I'm not certain of its correctness.  Feel free to close this PR and re-implement it yourself, the changes are pretty minor.

### DIFF
--- a/backup.cpp
+++ b/backup.cpp
@@ -17,6 +17,7 @@
 #include <deconz.h>
 #include "backup.h"
 #include "json.h"
+#include "crypto/random.h"
 
 #define EXT_PROCESS_TIMEOUT 10000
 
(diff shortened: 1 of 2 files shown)


manup:
Thanks, very nice good spot on the Mersenne Twister use. The fallback is likely not used much as in most installations OpenSSL should be available.

moovy2 (approved):

PR #5846, opened 2022-03-05
Add support for Mueller-Licht Tint remote 404049

kayhannay:
This adds support for the Müller-Licht Tint remote control 404049 as requested in #5841.


heinrich-foto:
dose this work? can it be merged?

PR #5619, opened 2021-12-29
Window Covering devices

ebaauw:
Update support for Aqara Roller Shade Driver E1 (`lumi.curtain.acn002`), see #5330:
- Fix `type`: now `Window covering device` instead of `Dimmable light`;
- Add write-only `state` attribute `lift_inc` step up (-1), stop (0), and step down (1).  Values < -1 are mapped to -1, values > 1 to 1;
- Use _Multistate Output_ cluster for `stop`, `open`, and `lift_inc`;
- Add write-only `state` attribute `alert` to blink the led;
- Add read/write `state` attribute `speed` to set the motor speed: 0 = low, 1 = medium, 2 = high.
- Handle additional attributes of Lumi-specific cluster.  Created a separate attribute set for this device;
- Handle additional tags for Xiaomi special attribute;
- Create ZHABattery `/sensors` resource to expose the battery and charging state.

Update support for Aqara Motor B1 (`lumi.curtain.hagl04`), see #4249:
- Use _Multistate Output_ cluster for `stop`, `open`, and `close`;
- Handle additional tags for Xiaomi special attribute;
- Bug fix: B1 can no longer be controlled through the API;
- Expose charging state on ZHABattery resource (as yet untested).

Update support for IKEA FYRTUR and KADRILJ:
- Add write-only `state` attribute `alert` to blink the led.

~~Remove deprecated `state` attributes `on`, `bri`, `bri_inc`, and `sat` from window covering devices.  Use `open`, `lift`, `stop`, and `tilt` instead, see https://github.com/dresden-elektronik/deconz-rest-plugin/issues/2475#issuecomment-609485739.~~

Remove workarounds for `deCONZ::ZclAttribute` bug handling `deCONZ::ZclSingleFloat` and `deCONZ::Zcl8BitEnum` values.

Update _Node List_ panel in the GUI when receiving new values for _Basic_ cluster attributes.

Added another Sunricher wireless switch, `ZG2833K4_EU06,` to `button_maps.json`.

Added manufacturer-specific attributes for Sunricher ZG9092 thermostat to `general.xml`, see #5409.

Bug fixes:
- Wrong API code when specifying an invalid value for `alert` when PUTting a lights state.
- Cannot set `sunriseoffset` nor `sunsetoffset` in `config` for built-in Daylight sensor.

Issue #5330, opened 2021-09-28
Aqara Roller Shade Driver E1


hastime:
## Device

- Product name: Roller Shade Driver E1
- Manufacturer: LUMI
- Model identifier: lumi.curtain.acn002
- Device type : Please remove all unrelated device types. 
 
  - Other: Battery Powered roller shade driver

## Screenshots

<img width="521" alt="Screenshot 2021-09-28 at 23 13 49" src="https://user-images.githubusercontent.com/67225476/135173063-460ea2e2-4928-4db5-bd84-aac77b9177aa.png">

### Basic

<img width="515" alt="Screenshot 2021-09-28 at 23 15 19" src="https://user-images.githubusercontent.com/67225476/135173265-609d4cff-ae44-433b-b7a9-89aab72a0f4f.png">

### Identify

<img width="524" alt="Screenshot 2021-09-28 at 23 16 01" src="https://user-images.githubusercontent.com/67225476/135173279-41d86430-d4e6-4c1f-8d4d-63ce9522a501.png">

### Alarms

<img width="530" alt="Screenshot 2021-09-28 at 23 17 30" src="https://user-images.githubusercontent.com/67225476/135173888-f2fa51e5-0fa1-4934-9a22-c034aff00b98.png">

### Device Temperature

<img width="525" alt="Screenshot 2021-09-28 at 23 18 01" src="https://user-images.githubusercontent.com/67225476/135173909-d781c239-8ffb-4eb6-ad6b-3be9b9d14cb0.png">

### Groups

<img width="527" alt="Screenshot 2021-09-28 at 23 19 00" src="https://user-images.githubusercontent.com/67225476/135173946-8c5e8fd9-5665-42b2-8f4d-9fd2e46a2541.png">

### Scenes

<img width="518" alt="Screenshot 2021-09-28 at 23 20 31" src="https://user-images.githubusercontent.com/67225476/135173958-1604ca35-e9be-417a-8bf1-c469f903244a.png">

### On/Off

<img width="520" alt="Screenshot 2021-09-28 at 23 20 59" src="https://user-images.githubusercontent.com/67225476/135173974-3b57a02c-759a-49a9-ab6b-fa2e6b403500.png">

### Other clusters that are not mentioned above

[...]

PR #5851, opened 2022-03-09
DDF for Aqara E1 roller shade (lumi.curtain.acn002)

manup:
Followup for the C++ implementation in PR https://github.com/dresden-elektronik/deconz-rest-plugin/pull/5619 , Issue https://github.com/dresden-elektronik/deconz-rest-plugin/issues/5330

The PR replaces a few hard coded parts in C++ with DDF mechanics.

The ZHABattery `config.temperature` value is way off, and reports 37°C in my setup :) perhaps we should ditch it?

Starting with this PR the sub resource "type" attribute will overwrite the one given by the C++ implementation, like here "Dimmable light" is replaces with "Window covering", this might be handy for other devices as well.

### DIFF
--- a/bindings.cpp
+++ b/bindings.cpp
@@ -2578,10 +2578,6 @@ void DeRestPluginPrivate::checkLightBindingsForAttributeReporting(LightNode *lig
         }
         else if (lightNode->manufacturerCode() == VENDOR_XIAOMI)
         {
-            if (lightNode->modelId().startsWith(QLatin1String("lumi.curtain.acn002")))
-            {
-                return;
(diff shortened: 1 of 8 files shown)


kvanisterdael:
Hi, I have beta 2.15.02 but my E1 roller shade still appears as a dimmable light. 
I can control it via the REST API by acting on the lift state without issue, so that works fine.

I had another question: is it possible to make the engine stop moving at the current position? For example: I send "open": true and after 5 seconds, I want to stop at that position. Is there a command for this?

PR #5857, opened 2022-03-11
Add DDF for Xiaomi WRS-R02 Wireless Remote Switch H1 (Double Rocker)

SwoopX:
Adds the DDF and removes most of the associated legacy code. It requires PR #5868

- Introduces/reuses `config/devicemode` to now (automatically) set the device mode, which is advisable to be applied for all Xiaomi devices. Possible values `compatibility` (default) and `zigbee`. This should not be changed unless you have some solid knowledge around zigbee standard!
- Introduces `config/clickmode` to be able to configure button behavior. Possible values `multiclick` (default) and `highspeed` (allows only single click, but is transmitted considerably faster). I'd suggest to use this resource item also for coupled/decoupled mode for switches supporting it.
- Allows setting the respective values for the abovementioned resource items via API using a state change. For the changes to take effect, it might be required to wake up the device up to 3 times with a 5x fast click on the left button.

This PR also sets the default mode for switches from `ModeTwoGroups` to `ModeScenes`. Majority of the button maps is written for `ModeScenes`, but the DDF cannot change the value as the legacy code does.


rest_sensors.cpp
```diff
@@ -766,6 +766,22 @@ int DeRestPluginPrivate::changeSensorConfig(const ApiRequest &req, ApiResponse &
...
+                        rsub->addStateChange(change);
+                        updated = true;
+                    }
+                }
+                else if (rid.suffix == RConfigClickMode && !data.string.isEmpty()) // String
+                {
+                    StateChange change(StateChange::StateWaitSync, SC_WriteZclAttribute, task.req.dstEndpoint());
+                    change.addTargetValue(rid.suffix, data.string);
+                    Resource *rsub = DEV_GetSubDevice(device, nullptr, sensor->uniqueId());
+                    rsub->addStateChange(change);
```
manup, line 783: To guard against SEGV, I'd propose the following:

```
Resource *rsub = DEV_GetSubDevice(device, sensor->prefix(), sensor->uniqueId());
if (rsub)
{
    rsub->addStateChange(change);
}
```

sensor.cpp
```diff
@@ -151,7 +151,7 @@ bool SensorFingerprint::hasOutCluster(quint16 clusterId) const
 Sensor::Sensor() :
     Resource(RSensors),
     m_deletedstate(Sensor::StateNormal),
-    m_mode(ModeTwoGroups),
+    m_mode(ModeScenes),
```
manup, line 154: Tricky ;)

I was considering this myself recently but hold back since I wasn't sure if that doesn't break some switches.

manup:
I'd like to merge this PR next can you please update, there seems to be a small conflict in rest_sensors.cpp.

SwoopX:
@manup It is done 🙂

PR #5893, opened 2022-03-20
Add DDF for Xiaomi H1 dual rocker switch (neutral wire) WS-EUK04

SwoopX:
Relies on #5857

Adds the DDF and removes the respective legacy code. For the created ZHASwitch resource, `config/devicemode` is added to automatically put the device in the required mode of operation. Furthermore, `config/ledindication` is added to ZHASwitch resource to be able to disable the LEDs during night time. Both is set on the first endpoint only, so the switch resource can be chosen to make the resource items available.

What's still to be clarified is setting the decoupled mode. As this is set per endpoint (there's 2), the light resources would probably make sense. However, they do not really support `config` resource items.


manup:
Looks good to me, but needs to be realigned on latest master.

PR #5895, opened 2022-03-21
Enable setting `config/delay` on IAS presence sensor for MOSZB-140

SwoopX:
Relies on #5857

### DIFF
--- a/rest_sensors.cpp
+++ b/rest_sensors.cpp
@@ -774,13 +774,20 @@ int DeRestPluginPrivate::changeSensorConfig(const ApiRequest &req, ApiResponse &
                 }
                 else if (rid.suffix == RConfigDelay) // Unsigned integer
                 {
-                    if (sensor->modelId().startsWith(QLatin1String("SML00")) ||   // Hue motion sensor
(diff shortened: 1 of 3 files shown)


rest_sensors.cpp
```diff
@@ -774,14 +774,21 @@ int DeRestPluginPrivate::changeSensorConfig(const ApiRequest &req, ApiResponse &
...
                     {
                         pendingMask |= R_PENDING_DELAY;
                         sensor->enableRead(WRITE_DELAY);
                         sensor->setNextReadTime(WRITE_DELAY, QTime::currentTime());
                     }
+                    else
+                    {
+                        if (devManaged && rsub)
+                        {
+                            change.addTargetValue(rid.suffix, data.uinteger);
```
manup, line 788: Is there something missing, the `change` is not defined?

manup: Oh I see that's in the other PR.

Before merging I need to fix another issue, this code assumes that "write" parameters are defined in the DDF.
If not this would cycle until the overall timeout is hit.

PR #5904, opened 2022-03-25
Update general.xml to allow modification of Backlight attribute on Sinopé TH11xxZB and Ouellet  OTH4000-ZB thermostats

BabaIsYou:
Reintroduce mfcode for ELKO thermostat specific for not superseed others (ie Sinope) (#5898 )
Change Backlight choice for Sinope specific "sensing" to "Always ON" according to Sinope documentation.


SwoopX:
This will break functionality of the Elko thermostat, so I'd recommend notmerging this PR. The root cause, from my perspective, lies in inappropriate handling of manufacturer codes in deconz core.

BabaIsYou:
> This will break functionality of the Elko thermostat, so I'd recommend notmerging this PR. The root cause, from my perspective, lies in inappropriate handling of manufacturer codes in deconz core.

I understand, but at this stage is there a solution?

Smanar:
And reversing the order is not enought ? First put the part for Sinope and after the part for elko ?
The sinope can't overwrite the elko part (because use the manufacture code check)

BabaIsYou:
BTW  basing perrenity solely on the position of a paragraph in a file, which is moreover an XML file whose purpose is precisely not to depend on this kind of thing, seems a little fragile to me. 

But it seems works ;-)

Smanar:
Lol, yes for me too, but it's a workaround, so much thing to improve.

Smanar:
@BabaIsYou if it work on your side, can you edit the PR according to your last changes ?

BabaIsYou:
> @BabaIsYou if it work on your side, can you edit the PR according to your last changes ?

Done ! ;-)

Smanar:
@SwoopX it's ok for you ?
As the code for Sinope use `mfcode="0x119c"` it have no impact on yours ?

SwoopX:
@Smanar To be honest, it is not. As a consequence, folks having the Elko Thermostat will (rightfully) start crying to revert this change, as it breaks using the respective 0x04XX attributes, which are NOT manufacturer specific in this case.

So the PR is fighting symptoms instead of the root case, which is in my view in deconz core. Presumably, the loop there breaks or continues on the cluster items after first hit without appropriately honoring the mfc.

BabaIsYou:
It’s clearly not my goal to introduce regression. But it seems weird tomme that deconz code could have this leak just for 0x4xx and not 0x4xxx for instance.

But if that is the case people with Eurotronic and Danfoss would have the same trouble …

Well, to be sure we have to ask someone with Elko to give a try like I did.

Le 27 mars 2022 à 15:46, SwoopX ***@***.***> a écrit :

﻿

@Smanar<https://github.com/Smanar> To be honest, it is not. As a consequence, folks having the Elko Thermostat will (rightfully) start crying to revert this change, as it breaks using the respective 0x04XX attributes, which are NOT manufacturer specific in this case.

So the PR is fighting symptoms instead of the root case, which is in my view in deconz core. Presumably, the loop there breaks or continues on the cluster items after first hit without appropriately honoring the mfc.

Smanar:
> @Smanar To be honest, it is not. As a consequence, folks having the Elko Thermostat will (rightfully) start crying to revert this change, as it breaks using the respective 0x04XX attributes, which are NOT manufacturer specific in this case.

But If m right, you can have 2 situations
- The device is an elko, so it don't use the Sinope command because it was manufacture specific, and use the elko one.
- The device is Sinope and use the Sinope command because of the order.

I don't have elko device to test, but for me there is no problem. The elko is not manufacturer specific but the sinope is.

BabaIsYou:
> But If m right, you can have 2 situations
> 
>     * The device is an elko, so it don't use the Sinope command because it was manufacture specific, and use the elko one.
> 
>     * The device is Sinope and use the Sinope command because of the order.

Hi folks,

Just to remind you that we are only talking here about attributes whose names are "unknown" for Elko thermostats or about adding the attribute 400 which does not exist in the specific part of Elko. So I don't think it can have an impact on their use.

BabaIsYou:
I closed it to re-open a new one. Less job to resolve conflicts.

PR #5910, opened 2022-03-27
Fix DDF handling of negative values and values above 2^52

manup:
For signed integers the C++ union is tricky as the correct member needs to be
extracted. In contrast to unsigned integers  num.s16 != num.s64. Since the
ZclAttribute already has the correct negative number in its internal QVariant
we can just use it.

Also expose 56-bit values as string since JS only has 2^52 integers.

The scopes where added to prevent stack corruption  (not actually reported but to be safe..).

### DIFF
--- a/device_js/device_js_wrappers.cpp
+++ b/device_js/device_js_wrappers.cpp
@@ -129,6 +129,7 @@ QVariant JsZclAttribute::value() const
         return attr->numericValue().u8 > 0;
     }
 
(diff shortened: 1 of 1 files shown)


ebaauw:
Does this also fix the power value -1 for the unisys D1 R?  See https://github.com/ebaauw/homebridge-deconz/issues/29#issuecomment-1073267829.

PR #5931, opened 2022-04-03
Add additional Xiaomi MFC for ZDP handler

SwoopX:
There's apparently a new Xiaomi related MFC, so we want to make sure that one gets a proper node descriptor response as well.

### DIFF
--- a/zdp/zdp_handlers.cpp
+++ b/zdp/zdp_handlers.cpp
@@ -289,10 +289,11 @@ struct MapMfCode
     quint16 serverMask;
 };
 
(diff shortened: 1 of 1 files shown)


manup:
This won't work ;) currently only the MAC address is checked to match an entry.

Not sure if at this stage we already have the MFC of the device at hand, the Node Descriptor request from the device might arrive before we have queried the device Node Descriptor.

manup:
I'd assume that the old MFC works either way since Xiaomi has likely gateways with old and newer MFC in the wild.

SwoopX:
I'll get the device to take a deeper look.

PR #5962, opened 2022-04-14
Add DDF for Sinope sp2610zb smart plug and TH1124ZB Thermostat

BabaIsYou:
Using eval function to get right values


BabaIsYou:
Adding DDF to have ZHAComsumption power updated and voltage divided by 10 for TH1124ZB.
If merged with PR #5904 will made available the modification for Sinope specific attribute 0x0402

manup:
PR looks fine to me, since DDF read/write/parse functions don't depend on the ZCLDB general.xml file this PR should work even without  #5904

PR #5968, opened 2022-04-20
Completing some Sinope specific thermostat attributes

BabaIsYou:
Based on excellent work of @claudegel in https://github.com/claudegel/sinope-zha


BabaIsYou:
Replaced by PR #6659

PR #5976, opened 2022-04-23
Set `config/delay` resource item default to 60

SwoopX:
A default of 0 can lead to irritation presence detection is not working properly. State jumps from `true` back to `false` within the second.


ebaauw:
For what device is this?  Please leave the default to 0 for the Hue motion sensors.  Also note that it’s read-only for the Trådfri motion sensors.

Issue #5974, opened 2022-04-22
Aquara Multi Sensors (Temperature/Pressure/Humidity) no longer reporting temperature


UppyJC:
## Describe the bug

Since upgrading to latest deCONZ version, Aquara Multi Sensors (Temperature/Pressure/Humidity) no longer reporting temperature. API reports null values, Phoscon reports 0 degrees. Pressure and humidity values fine.

Deleting and re-adding sensors makes no difference.

Working correctly on previous version, but can't remember which version that was!

## Steps to reproduce the behavior

Add any sensor of affected type to deCONZ. Observe null temperature value in API and 0 temperature value in Phoscon. Humidity and pressure values report correctly.

## Expected behavior

Temperature value reports correctly in API and Phoscon.

## Screenshots

<img width="541" alt="Screenshot 2022-04-22 at 14 49 32" src="https://user-images.githubusercontent.com/98237047/164728857-29befa92-6173-4d51-bdfd-25a9741996ed.png">

<img width="839" alt="Screenshot 2022-04-22 at 14 49 45" src="https://user-images.githubusercontent.com/98237047/164728895-797acc76-e82b-4d24-ad67-7b3e09708afc.png">

## Environment

 - Host system: PC
 - Running method: Windows
 - Firmware version: (26580700)
 - deCONZ version: (2.15.02)
 - Device: ConBee II
 - Do you use an USB extension cable: Yes
 - Is there any other USB or serial devices connected to the host system?: No

Mimiix:
What does the respective clusters show in deCONZ? 

Also, did you try this?

https://github.com/dresden-elektronik/deconz-rest-plugin/wiki/Xiaomi:-Lost-devices-or-functional-issues

UppyJC:
Yes, I have re-paired one of the sensors and read node descriptors and simple descriptors.

There is a reading in deCONZ as shown of the screenshot below, but the API and Phoscon both still report null/0 respectively.

<img width="542" alt="Screenshot 2022-04-22 at 15 38 37" src="https://user-images.githubusercontent.com/98237047/164736894-5e2d6986-1451-4d9e-9b14-8173ec758cd2.png">

I have 11 of these sensors and they're all exhibiting exactly the same behaviour from the same time (when I upgraded to the latest deCONZ version).

nunoary:
I see the same behavior. It was working fine with version 2.14.01 but after installing version 2.15.01 the sensors no longer report temperature.

[...]

PR #5983, opened 2022-04-26
Device code improve pairing

manup:
- In legacy `delayedFastEnddeviceProbe()` function don't interfere with Device code ZDP requests for Node Descriptor, Active Endpoints and Simple Descriptors.
- Consider already available data in Basic cluster attributes during pairing, which is needed for some Xiaomi sensors that send the modelid in a report.
- Fix handling of unsupported binding table status (Xiaomi sensors send a response which has only 2 bytes of payload).
- Check if there is polling needed after awake event, that's useful during pairing to quickly query all needed attributes.
- Sync NWK address changes from core to device `attr/nwkaddress` resource item.

Related issue https://github.com/dresden-elektronik/deconz-rest-plugin/issues/5974

### DIFF
--- a/de_web_plugin.cpp
+++ b/de_web_plugin.cpp
@@ -15272,6 +15272,11 @@ void DeRestPluginPrivate::delayedFastEnddeviceProbe(const deCONZ::NodeEvent *eve
 
         if (!hasNodeDescriptor)
         {
+            if (DEV_TestManaged())
+            {
+                return; // Device code handles this
+            }
+
             DBG_Printf(DBG_INFO, "[1] get node descriptor for 0x%016llx\n", sc->address.ext());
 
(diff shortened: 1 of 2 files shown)


SwoopX:
Nice one. Would it also make sense to prevent the request towards the basic cluster in `delayedFastEnddeviceProbe()` a bit further down? I'd assume that is managed by the DDF code as well.

manup:
Yes it makes sense to limit processing further :) Currently only the first steps of joining are considered to not accidently break anything. Before merging I'll add a small bit on top to bail out as soon as we know the device is controlled by a DDF (likely only Gold status for now).

In the sniffer joining now looks less noisy, the Xiaomi humidity sensor was a hit and miss to join since it didn't poll so often.

PR #6009, opened 2022-05-03
Create config_local_occupancy_item.json

BabaIsYou:
Creation of config item to change local occupancy. This enable thermostat to automaticallly switch to Unoccupied heat set point


BabaIsYou:
Already in #6011

PR #6011, opened 2022-05-03
Adding Sinope Thermostat TH11xxZB management of Occupancy specific attribute and Unoccupied heat set point

BabaIsYou:
Adding management via REST-API of specific Occupancy Sinope state attribut and Unoccupied heat set point.
Taht way thermostat can switch automatically to unoccupied heating set point or occupied heating set point depending on the value of Occupancy.


manup:
I think it should be fine, but like to have more opinions on that before adding new resource items. We need to care to being more streamlined than in the past.

SwoopX:
Here're my thoughts around this. Besides that in my very personal opinion occupancy and unoccupied heatsetpoint is pointless, resource items should be strictly lower case only. I'm also a fan to keep them as generic as possible so they can potentially be reused on other devices/for other purposes as well, simply to restrict a rising zoo.

For `local_occupancy`, that could be just `occupancy` while also having the benefit match generic attribute 0x0002 of the thermostat cluster. Expressing that it is local does not seem relevant in my view.
On the other hand, the aforementioned attribute is read only as per spec, the one in the PR is allows also to write (`config` vs `state` having the question around exposing it twice for its own purpose). And of course does the proposed shortened name create a certain room for misunderstanding and misinterpretation (e.g. motion sensors)...

So I see removal of the underscore a must (for at least some consistency). We also should/must make other aware for their API clients?

SwoopX (commented):
Here're my thoughts around this. Besides that in my very personal opinion occupancy and unoccupied heatsetpoint is pointless, resource items should be strictly lower case only. I'm also a fan to keep them as generic as possible so they can potentially be reused on other devices/for other purposes as well, simply to restrict a rising zoo.

For `local_occupancy`, that could be just `occupancy` while also having the benefit match generic attribute 0x0002 of the thermostat cluster. Expressing that it is local does not seem relevant in my view.
On the other hand, the aforementioned attribute is read only as per spec, the one in the PR is allows also to write (`config` vs `state` having the question around exposing it twice for its own purpose). And of course does the proposed shortened name create a certain room for misunderstanding and misinterpretation (e.g. motion sensors)...

So I see removal of the underscore a must (for at least some consistency). We also should/must make other aware for their API clients?

BabaIsYou:
The fact that you find it useless doesn't necessarily make it pointless. On the contrary, I think it's brilliant of the manufacturer to have provided this possibility. ;-)

So I can assign a different value of unoccupied setpoint for each room and just change the value of local_unoccupancy so that the thermostat takes automatisa lot the heating value of unoccupiedsetpoint instead of having to change the heatsetpoint for each thermostat from my home automation system.

Programmons and management of thermostats by script become simpler from my automation system.

As occupancy is readonly in Zigbee specification it seems that the manufacturer had to add another writable attribute.

That Béring sauf if the only issue is to change some attributes names to avoid confusion or to be compliant to some rules to have this PR accepted just tell me want name you prefer instead of local_occupancy (local refers to room).

Thanks

BabaIsYou:
@SwoopX  To be consistent with other config items (externalsensortemp and externalwindowsopen), what would you think about externaloccupancy ?

BabaIsYou:
Also, sometimes DDF file for Sinope are not loaded because of legacy code is overwriting  the real value reported by the manufacturer name attribute. Then we have to force reading of basic cluster as @SwoopX described it in support forum 
[issue 2121](https://forum.phoscon.de/t/ddf-is-not-loading-only-for-one-of-my-two-devices/2211)

Smanar:
For me, we can remove this part in code

```
    else if ((node->nodeDescriptor().manufacturerCode() == VENDOR_SINOPE))
    {
        sensorNode.setManufacturer("Sinope");
    }
```

- try to modify the model id or the manufacture name is the worst idea possible
- I don't think it can have an impact on previous integrated device.

manup:
Do we have any new considerations for the local_occupancy / occupancy naming?
I have to admit the thermostat topic is still a bit of a black box for me with all the variations floating around.

One note to the item description `generic/items/config_local_occupancy_item.json`: I think as a generic item which can be device specific the read, write and parse functions should be specified in the device DDF. For example here the 0x0400 attribute appears to be vendor specific.

BabaIsYou:
I totaly agree that the binding/location in cluster that are device/manufacturer specific are to be preferably located in devices DDF. But, as I remember, that does not make it writable until you bind it to a already existing writable item. At time we discussed it with @smanar there wasn’t such an existing config item with a name and a corresponding data type.

Le 28 août 2022 à 07:40, Manuel Pietschmann ***@***.***> a écrit :

﻿

Do we have any new considerations for the local_occupancy / occupancy naming?
I have to admit the thermostat topic is still a bit of a black box for me with all the variations floating around.

One note to the item description generic/items/config_local_occupancy_item.json: I think as a generic item which can be device specific the read, write and parse functions should be specified in the device DDF. For example here the 0x0400 attribute appears to be vendor specific.

Smanar:
>I think as a generic item which can be device specific the read, write and parse functions should be specified in the device DDF

I m not sure to understand, there is something missing ?

manup:
Related thermostat items streamline issue: https://github.com/dresden-elektronik/deconz-rest-plugin-v2/issues/13

BabaIsYou:
> I schedule this for the next beta after stable, so we need to work out a opinion and naming for these items :)

Seems we skipped some milestones ... Chance to have it un next one ?

BabaIsYou (commented):
Removing the local_occupancy item that seemed to generate dscussions and reused reset presence instead.

PR #6029, opened 2022-05-07
Add DDF: WOOX R7060 smart garden irrigation control

Finne75:
DDF for WOOX R7060 (TS0101) smart garden irrigation control device with working battery readings.


matze20211:
"irrigation control device with working battery readings." 
using 2.17.01 / 1.7.2022 FW:26720700 there ar no battery readings in fhem?

PR #6038, opened 2022-05-09
Add DDF for Xiaomi presence sensor lumi.sensor_motion.aq2 (RTCGQ11LM)

SwoopX:
Adds respective DDF, no legacy code to remove.

### DIFF
--- a/devices/generic/illuminance_cluster/lux_to_lightlevel.js
+++ b/devices/generic/illuminance_cluster/lux_to_lightlevel.js
@@ -1,0 +1,14 @@
+const tholddark = R.item('config/tholddark').val;
+const tholdoffset = R.item('config/tholdoffset').val;
+const lux = Attr.val;
+let ll = 0;
+
+if (lux > 0 && lux < 0xffff) {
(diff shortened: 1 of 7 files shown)


cyablo:
This change breaks update of light level on this sensor. It's stuck on (probably) the last value before the update.

SwoopX:
Hm, works fine over here

cyablo:
Tried 2.17 again,  even tried repairing a sensor, light level is stuck. Downgraded to 2.16.1 everything fine again. Conbee II with Firmware 26780700

Tscheurgen:
Unfortunately I have the same issue as cyablo. My lux value of ZHALightLevel is stuck at 0. Any hint to fix this?

Tscheurgen:
I have a bit investigated.
Measured Value of Cluster 0400 (Illuminance Measurement) does Change, but REST provides always the same value.

cyablo:
We got into that in the Forum: https://forum.phoscon.de/t/rtcgq11lm-light-level-broken-since-2-17/2104/7

Tscheurgen:
I'm sorry, Thank you for the link!

PR #6056, opened 2022-05-15
Fix: Guard against non-existent device to prevent SIGSEGV

SwoopX:
While changing the settings of the Daylight sensor, the code runs into a segmentation fault in `DEV_GetSubDevice()`, trying to iterate over the available sub devices of the device provided, although a nullptr was given.

In addition to the above, two further missing guards were added.


manup:
Evil pointers :)

PR #6057, opened 2022-05-16
Add DDF for Danalock V3

manup:
Compared to the legacy implementation the PR configures bindings and reporting for Power Configuration, Poll Control and Doorlock clusters.

This yields a bit more traffic towards the gateway and shows the battery symbol in the GUI.

### Connection get lost issue

There are several reports that the device has difficulties keeping the connection (this seems to also happen with other gateways). Looking at various sniffer logs one reason is when the lock changes the parent from a router to the coordinator, this wasn't handled correctly in all cases. The bug is fixed in the new ConBee || and RaspBee II firmware 0x26780700 ([firmware changelog](https://github.com/dresden-elektronik/deconz-rest-plugin/wiki/Firmware-Changelog)).

The lock runs fine now in my setup for a few days with the new firmware and DDF.

However the sniffer logs show another problem which needs to be addressed in the Danalock V3 firmware: The lock detects than a parent is not reachable (aka powered off) and correctly rejoins via another parent. But it doesn't detect if the parent router plays nice, when commands are send to the gateway via a router the lock doesn't rejoin the network if it doesn't receive APS ACKs or ZCL Default Reponses from the gateway for a few times. The fix should be fairly simple and bring robustness like the Philips Hue motion sensor has.

### DIFF
--- a/devices/danalock/danalock_v3.json
+++ b/devices/danalock/danalock_v3.json
@@ -1,0 +1,79 @@
+{
+  "schema": "devcap1.schema.json",
+  "manufacturername": "Danalock",
+  "modelid": "V3-BTZBE",
+  "product": "Danalock V3",
+  "sleeper": true,
+  "status": "Gold",
+  "subdevices": [
+    {
+      "type": "$TYPE_DOOR_LOCK",
+      "restapi": "/lights",
+      "uuid": [
+        "$address.ext",
+        "0x01"
+      ],
+      "items": [
+        {
+          "name": "attr/manufacturername"
+        },
+        {
+          "name": "attr/modelid"
+        },
+        {
+          "name": "attr/swversion",
+          "refresh.interval": 84000,
+          "read": {"cl": "0x0000", "at": "0x0006", "ep": 1, "fn": "zcl"},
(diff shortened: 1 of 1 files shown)


MarkoPaasila:
Should this DDF make the battery status show up in Home-assistant?

AnthonMS:
Great work! Has this been released yet? Been waiting for this so I don't come home and have to dig up a key because I forgot about it. 
Will we have to do anything like reconnect our devices to deconz after updating?

Weltraumpenner:
I'm thinking about replacing my ABUS HomeTec Pro with a Zigbee-Danalock.
Can someone tell me if it supports the states close, unlock and open being set over Home Assistant?
My ABUS doesn't support the unlock-command which is quite annoying.

AnthonMS:
> Can someone tell me if it supports the states close, unlock and open being set over Home Assistant? 

The only thing I have been missing is a battery status through deconz. But that seems fixed. So yes, it is fully supported in HA through Deconz now then.

sorano:
Yeah, lock/unlock and state of last change works for me.
However, I have updated everything that I could yet I still don't see any battery status in phoscon-GW/HA.
image
image
image

@manup Do you have any idea what's missing on my end?

Nazarbayevv:
Hello, 

I use RaspBee I. Can I wait update to 0x26780700 for my old RaspBee for lost connection fixing?

PR #6069, opened 2022-05-22
Add DDF for Aqara P1 presence sensor RTCGQ14LM

SwoopX:


manup:
Do I understand this right, the sensor knows the distance to the triggering object?

SwoopX:
@manup You're a little ahead of the PR description, haven't had the time yet.

Opened Pandorra's Box 🙂 Well, I wouldn't call it knowing the distance, but it apparently can distinguish between different ranges. Not sure how this is technically done precisely, maybe by adjusting measurement angles... However, they also describe it on their website, so nothing in terms of slow or hectic movement/number of movements (that's what I would understand on sensitivity).

The FP1 presence sensor interestingly handles both, true sensitivity and distance. We probably want to expose it cooked? Here, I did it uncooked, but for the FP1, it is by setting `near`, `medium` and `far`.

manup:
Ah, since this is only a PIR I don't think it has anything to do with distance, at most sampling sensitivity (imho). I'd prefer to use `config/sensitivity` to be consistent.

>We probably want to expose it cooked? Here, I did it uncooked, but for the FP1, it is by setting near, medium and far.

Yeah it's time we get this streamlined, normalized values here are much needed, the `near`, `medium` and `far` are nicer and clearer, but historically `config/sensitivity` is a number :) perhaps it's better to create a RFC issue to discuss with the other clients/devs for a proper approach how to streamline this without breaking anything?

For docs, nice comparison between the former version and this p1: https://www.aqara.com/en/product/motion-sensor-p1

The FP1 is super interesting with the radar, we haven't managed to get one yet :/

PR #6091, opened 2022-05-29
Fix: Make Hue motion sensors configurable again

SwoopX:
Basically, the sensors were missing the respective write functions to actually trigger the configuration change. On that occasion, the DDFs have been reworked:

- An attribute query loop has been resolved
- Polling and has been amended and minimized
- Manufacturer name for SML003 and SML004 has been corrected
- DDF for2nd gen outdoor sensor has been promoted to gold
- DDF for 1st gen Hue outdoor sensor has been added
- State change for `config/usertest` has been added

Based on this change, the associated legacy code could be removed. `config/pending` is effectively not used anymore and could be removed as well. However, that needs to be done cautiously in the respective code, as other sensors using occupancy cluster probably still rely on its usage.

### DIFF
--- a/devices/generic/items/config_ledindication_item.json
+++ b/devices/generic/items/config_ledindication_item.json
@@ -4,5 +4,6 @@
 	"datatype": "Bool",
 	"access": "RW",
 	"public": true,
-	"description": "Activates the LED indicator."
+	"description": "Activates the LED indicator.",
+	"default": false
 }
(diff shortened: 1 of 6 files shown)


manup:
Looks good to me, could you test this with one of the sensors?
I have the 1. Gen version here can make a test if needed.

manup:
Perfect, I'll run the PR tomorrow in my setup and add a few simple rules.

PR #6118, opened 2022-06-06
Add DDF for Ikea Starkvind

manup:
The device can be controlled via `config.mode` with the values which reflect the on device states:

```
"off", "auto", "speed_1", "speed_2", "speed_3", "speed_4", "speed_5"
```

The **read-only** `state.speed` attribute reflects the current speed as normalized number 0..100.
When the mode is changed manually with the knob the `config.mode` is updated automatically.

**Note 1:** The device target/current values might not always be in steps of 10, e.g. 10..20, 20..30, etc. in which case they are mapped to the best fitting `config.mode` value, so 3 &rarr; 10 aka `"step_1"`.

**Note 2:** the `state.airquality` string values might need to be changed to better reflect the numeric values.

JSON view:

```
{
    "config": {
        "mode": "off",
        "on": true,
        "reachable": true
    },
    "etag": "80ae0dff37324f2582044ec71f3755ae",
    "lastannounced": null,
    "lastseen": "2022-06-06T18:08Z",
    "manufacturername": "IKEA of Sweden",
    "modelid": "STARKVIND Air purifier",
    "name": "AirQuality 234",
    "state": {
        "airquality": "excellent",
        "airqualityppb": 19,
        "lastupdated": "2022-06-06T18:09:00.681",
        "speed": 0
    },
    "swversion": "1.0.033",
    "type": "ZHAAirQuality",
    "uniqueid": "0c:43:14:ff:fe:6e:65:55-01-fc7d"
}
```

This PR only contains the basics to make the Ikea Starkvind controllable, other features as locking and filter monitoring should be added in future.

Issue #5351, opened 2021-10-05
IKEA Starkvind Air purifier (E2007/E3007)


grrrmml:
## Device

- Product name: STARKVIND Air purifier (E2007 and E3007 [Table integrated])
- Manufacturer: IKEA of Sweden
- Model identifier: STARKVIND Air purifier
- Device type : 
  - Sensor: Air Quality (unknown if exposed and readable)
  - Other: Fan Control: Off/Low/Medium/High/On/Auto/Smart ( on the device, there is a wheel with levels Auto/1-5 and On/Off [push] )

## Screenshots
![e2007_node_info](https://user-images.githubusercontent.com/6319974/136112317-2eee0da4-c029-4bc8-955c-6cb3e066365d.png)

![e2007_overview](https://user-images.githubusercontent.com/6319974/136112328-729c9ac0-6482-4573-a312-120d35be031d.png)

### Basic
![e2007_basic_cluster](https://user-images.githubusercontent.com/6319974/136112019-dc8ace5e-e9af-42fd-8ebf-b5651e4b2ae4.png)

### Groups
![e2007_groups](https://user-images.githubusercontent.com/6319974/136112136-6f2bfeab-735f-4854-bda0-aff4f8c86491.png)

### Scenes
![e2007_scenes](https://user-images.githubusercontent.com/6319974/136112214-de941ae2-7a2d-4de4-b2cd-fedbbe979ad5.png)

### Fan Control
![e2007_fan_control](https://user-images.githubusercontent.com/6319974/136112034-b2e8cad5-6d00-493e-964c-fac923780995.png)

### Unknown (FC7D)
 nothing visible for both

dusopn:
> Waiting for this device request until I buy one 👍

Exactly me too 👍

sjorge:
I've been poking this device using z2m, sadly no gateway to compare but I did manage to get something out of 0xfc7d

This is the cluster definition I added to z2m, I'm not 100% sure PM25 but it's extremely likely. It follows the same attribute layout as far as I could tel for the Ember zigbee additional measurement clusters.

```
    manuSpecificIkeaPM25Measurement: {
        ID: 0xfc7d,
        manufacturerCode: manufacturerCode_1.default.IKEA_OF_SWEDEN,
        attributes: {
            measuredValue: { ID: 0x0000, type: dataType_1.default.uint16 },
            measuredMinValue: { ID: 0x0001, type: dataType_1.default.uint16 },
[...]

PR #6124, opened 2022-06-10
IKEA Starkvind Air Purifier

ebaauw:
Update for IKEA Starkvind Air Purifier, see #5351 and #6118.

- Add fingerprint;
- Change `type` to `ZHAAirPurifier` so clients can distinguish between air quality sensors and air purifiers;
- Expose attribute 0x0004, which reports PM2.5 in µg/m³, as `state.pm2_5` (instead of `state.airqualityppb`);
- Change thresholds for `state.airquality` as they're very different for PM2.5 vs TVOC;
- Base `state.airquality` directly on 0x0004 as I cannot get the `numtostr` on `state.pm2_5` to work.
- Expose attribute 0x0000 as `state.filterruntime`;
- Expose attribute 0x0001 as `state.replacefilter`;
- Expose attribute 0x0002 as `config.filterlifetime`;
- Expose attribute 0x0003 as `config.ledindication`;
- Expose attribute 0x0005 as `config.locked`;
- Expose attribute 0x0008 as `state.deviceruntime`;
- Setup reporting for all attributes, change reporting to 1/300/1 for all attributes.  Mains-powered device, so no battery considerations.

Resource now looks like:
```
{
  "config": {
    "filterlifetime": 256728,
    "ledindication": true,
    "locked": false,
    "mode": "auto",
    "on": true,
    "reachable": true
  },
  "ep": 1,
  "etag": "c4b3b77e0fc9b62f7f6b85031de61d98",
  "lastannounced": null,
  "lastseen": "2022-06-11T15:40Z",
  "modelid": "STARKVIND Air purifier",
  "name": "Starkvind",
  "state": {
    "airquality": "excellent",
    "deviceruntime": 185310,
    "filterruntime": 182857,
    "lastupdated": "2022-06-11T15:39:46.328",
    "pm2_5": 6,
    "replacefilter": false,
    "speed": 20
  },
  "swversion": "1.0.033",
  "type": "ZHAAirPurifier",
  "uniqueid": "cc:86:ec:ff:fe:6d:30:11-01-fc7d"
}
```

Note that `deviceruntime`,  `filterruntime`, and `filterlifetime` are in minutes.  To indicate that you've replaced the filter, update `filterlifetime`.  The default life time (as set by the IKEA app is 259200 (6 months, or: 6 * 30 * 24 * 60).


devices/generic/items/state_pm2_5_item.json
```diff
@@ -0,0 +1,8 @@
+{
+	"schema": "resourceitem1.schema.json",
+	"id": "state/pm2_5",
+	"datatype": "UInt16",
+	"access": "R",
+	"public": true,
+	"description": "Measured PM2.5 in µg/m³."
```
SwoopX, line 7: You might want to add `"default": 0`, so we have a default values for this item

SwoopX, line 7: You might want to add `"default": 0`, so we have a default values for this item

devices/ikea/starkvind_air_purifier.json
```diff
@@ -48,7 +61,16 @@
           "name": "attr/uniqueid"
         },
         {
-          "name": "config/on"
+          "name": "config/ledindication",
+          "parse": {"at": "0x0003", "cl": "0xfc7d", "ep": 1, "fn": "zcl", "mf": "0x117c", "eval": "Item.val = !Attr.val"},
+          "write": {"at": "0x0003", "cl": "0xfc7d", "dt": "0x10", "ep": 1, "fn": "zcl", "mf": "0x117c", "eval": "!Item.val"},
+          "default": true
```
SwoopX, line 67: The default value could be removed, as it comes already with the resource item. I'd recommend to add the read function as well, otherwise the state changes has probably no chance to succeed. If you've configured reporting for this attribute, it's also advisable to add `"refresh.interval"` with a duration time of max reporting interval + 60 (to allow some polling fallback)

SwoopX, line 67: The default value could be removed, as it comes already with the resource item. I'd recommend to add the read function as well, otherwise the state changes has probably no chance to succeed. If you've configured reporting for this attribute, it's also advisable to add `"refresh.interval"` with a duration time of max reporting interval + 60 (to allow some polling fallback)

devices/ikea/starkvind_air_purifier.json
```diff
@@ -48,7 +61,16 @@
...
+          "name": "config/ledindication",
+          "parse": {"at": "0x0003", "cl": "0xfc7d", "ep": 1, "fn": "zcl", "mf": "0x117c", "eval": "Item.val = !Attr.val"},
+          "write": {"at": "0x0003", "cl": "0xfc7d", "dt": "0x10", "ep": 1, "fn": "zcl", "mf": "0x117c", "eval": "!Item.val"},
+          "default": true
+        },
+        {
+          "name": "config/locked",
+          "parse": {"at": "0x0005", "cl": "0xfc7d", "ep": 1, "fn": "zcl", "mf": "0x117c", "eval": "Item.val = Attr.val"},
+          "write": {"at": "0x0005", "cl": "0xfc7d", "dt": "0x10", "ep": 1, "fn": "zcl", "mf": "0x117c", "eval": "Item.val"},
+          "default": false
```
SwoopX, line 73: See above 🙂

SwoopX, line 73: See above 🙂

devices/ikea/starkvind_air_purifier.json
```diff
@@ -61,22 +83,22 @@
...
         },
         {
-          "name": "state/airqualityppb",
-          "parse": {"at": "0x0004", "cl": "0xfc7d", "ep": 1, "fn": "zcl", "mf": "0x117c", "eval": "if (Attr.val != 65535) { Item.val = Attr.val; }"},
-          "read": {"fn": "zcl", "ep": 1, "cl": "0xfc7d", "at": "0x0004", "mf": "0x117c"},
-          "refresh.interval": 30
+           "name": "state/pm2_5",
+           "parse": {"at": "0x0004", "cl": "0xfc7d", "ep": 1, "fn": "zcl", "mf": "0x117c", "eval": "Item.val = Attr.val"},
+           "read": {"fn": "zcl", "ep": 1, "cl": "0xfc7d", "at": "0x0004", "mf": "0x117c"},
+           "refresh.interval": 30
```
SwoopX, line 96: Consider adjusting the refresh interval as mentioned above 🙂

SwoopX, line 96: Consider adjusting the refresh interval as mentioned above 🙂

SwoopX (commented):
Just a couple of minors from my end. Still strange that the previously used function to set `airquality` doesn't work for you. Maybe Manup has any idea around it.

devices/ikea/starkvind_air_purifier.json
ebaauw: The item defaults to false, where the Starkvind defaults to true.

manup:
> Base state.airquality directly on 0x0004 as I cannot get the numtostr on state.pm2_5 to work.

> Just a couple of minors from my end. Still strange that the previously used function to set `airquality` doesn't work for you. Maybe Manup has any idea around it.

If I remember correctly currently there might be a certain order required for the items, so the source item like `state/airqualityppb` or `state/pm2_5` needs to be in the JSON array before `state/airquality`. (This should be fixed in future so the order shouldn't matter).

Anyhow otherwise the PR looks  solid to me, I haven't followed the full discussion are the REST-API visible changes ok for everyone? If yes we I'd like to merge it for v2.17.1 :)

Thomas-Vos:
Is there any reason air quality state is not named e.g. `airqualitypm2_5` to keep it consistent with `airqualityppb` naming?

Thomas-Vos:
Another question: can we expect `pm2_5` to appear on `ZHAAirQuality` sensors and `airqualityppb` on `ZHAAirPurifier` sensors in the future? For example if support for new devices is added that use different measurements than the current devices.

ebaauw:
I don’t like `airqualityppb`, as it doesn’t indicate of what pollutant the level is being measured (in case of the Xioami: TVOC).  Ideally, we would name the attributes `tvoc_level` and `pm2_5_density`, leaving room for `tvoc_density` and `pm2_5_level`.  I haven’t seen them, but I can imagine CO sensors reporting the raw CO level or density, in addition to the binary CO detected.  Same for natural gas sensors or other pollutants.

Air quality is derived from a pollutant level or density (with different thresholds for different pollutants).  I don’t see any value in prefixing the pollutant attributes with `airquality`.

Ideally, I would have exposed the sensor function of the Starkvind as ZHAAirQuality resource, separately from the ZHAAirPurifier resource for the purifier function.  However, this would lead to a `uniqueid` conflict, as the Starkvind combines both functions in the same manufacturer-specific Zigbee cluster.  I would only expect a `tvoc_level` or `airqualityppb` on a ZHAAirPurifier for a device that also uses one cluster for both functions.  I would expect `pm2_5_density` or `pm2_5` on ZHAAirQuality, possibly even combined with `tvoc_level` or `airqualityppb` depending on what the device measures and how (in how many clusters) it reports these quantities.

Kane610:
Is this the new way with deconz to expose one device with all data? As this is similar to the other fan device which is a light as well as a fan, this is an air purifier with fan, these types of functionality becomes similar but different rather than being kept aligned between device types and on a separate e.g. fan endpoint

ebaauw:
No, it’s the old way to expose a resource per cluster, as reflected in the `uniqueid`.

manup:
From my perspective it's clean to separate into two sub resources as @ebaauw mentioned, for the `endpoint` and `cluster id` parts in the `uniqueid` we can just pick some other value, as long as we stick to `xxxxxxxxxxxxxxxx-aa-bbbb` format I don't expect problems in clients. Internally the DDF processing for one sub resource items already supports processing arbitrary endpoints and clusters.

ebaauw:
> for the endpoint and cluster id parts in the uniqueid we can just pick some other value, as long as we stick to xxxxxxxxxxxxxxxx-aa-bbbb format

Happy to change the DDF to expose two `/sensors` resources (ZHAAirPurifier and ZHAAirQuality) - this would actually make it easier on Homebridge deCONZ.  Any takers for the "cluster" value in the `uniqueid`?  I don't suppose Zigbee has an Air Quality Measurement cluster?  Maybe abuse _IAS Zone_ (cf some of the CO sensors), or maybe it's less confusing to just fake another endpoint?

Kane610:
Would you mind sharing an updated json output of the different endpoints?

ebaauw:
Damn, I had, but appeantly forgot to press _Comment_.  Anyways, new attempt:
```json
{
  "config": {
    "filterlifetime": 256728,
    "ledindication": true,
    "locked": false,
    "mode": "auto",
    "on": true,
    "reachable": true
  },
  "ep": 1,
  "etag": "fea6623ea3909029409fed7a6224e60b",
  "lastannounced": null,
  "lastseen": "2022-06-30T18:19Z",
  "modelid": "STARKVIND Air purifier",
  "name": "Starkvind",
  "state": {
    "deviceruntime": 212939,
    "filterruntime": 210486,
    "lastupdated": "2022-06-30T18:19:46.177",
    "replacefilter": false,
    "speed": 20
  },
  "swversion": "1.0.033",
  "type": "ZHAAirPurifier",
  "uniqueid": "cc:86:ec:ff:fe:6d:30:11-01-fc7d"
}
```
```json
{
  "config": {
    "on": true,
    "reachable": true
  },
  "ep": 1,
  "etag": "74eb5d8558a3895a39a3884189701c99",
  "lastannounced": null,
  "lastseen": "2022-06-30T18:20Z",
  "modelid": "STARKVIND Air purifier",
  "name": "Starkvind",
  "state": {
    "airquality": "excellent",
    "lastupdated": "2022-06-30T18:18:26.205",
    "pm2_5": 8
  },
  "swversion": "1.0.033",
  "type": "ZHAAirQuality",
  "uniqueid": "cc:86:ec:ff:fe:6d:30:11-02-fc7d"
}
```

ebaauw:
Yes.  They're pretty much mapped 1:1 to the Zigbee attributes, see https://github.com/dresden-elektronik/deconz-rest-plugin/issues/5351#issuecomment-968267581.

kimosak:
Running on v2.18.1-beta and still not able to control Starkvind Air Purifier. Only sensor (airquality) is reporting the state.
No option to change mode or speed? Can I do something?

hph304:
I have the exact same problem as @kimosak

fsedarkalex:
Same here.

Mimiix:
Not sure what you mean..is this about phoscon? If so, report on the forums as we don't maintain phoscon in this GitHub.

Rest api should be able to control the values you ask for.

oywino:
> Rest api should be able to control the values you ask for.
As far as I could determine, the IKEA STARKVIND Air Purifier only accepts PUT commands to change values. But according to the HA DOC's, the REST-API only supports GET and POST 
https://www.home-assistant.io/integrations/sensor.rest/

PR #6129, opened 2022-06-12
Update ikea-ota-download.py to work with pyservice

SNPerkin:
A colleague was trying to update Home Assistant OTA files with this script. We tried using pyscript, but I needed to modify urlopen and urlretrieve to be asynchronous functions for it to work. Similarly I needed to replace print statements with logging in order to get feedback from the script into Home Assistant's general logs. I am not sure if this is the intended use of the script, but this was at least a solution to my colleague's problem that may benefit others.

### DIFF
--- a/ikea-ota-download.py
+++ b/ikea-ota-download.py
@@ -4,6 +4,7 @@ Snipped to download current IKEA ZLL OTA files into ~/otau
 compatible with python 3.
 """
 
+import logging
 import os
 import json
 try:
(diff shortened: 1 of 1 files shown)


manup:
Oh, this script is not meant to be Home Assistant specific but just to be run via plain Python 2/3.
If I remember correctly the HA add-on has it's own different download script.

SNPerkin:
Okay no problem, thanks for the reply and clarification.

I figured that may be the case, but my colleague could only find this script and not the add-on specific script. I'll close this pull request, probably makes more sense to improve searchability for the home assistant add-on elsewhere.

Issue #6136, opened 2022-06-16
Frient/Develco WISZB-121 open/close sensor


paddyobrien:
## WISZB-121 open/close sensor

- Product name: Frient/Develco Entry Sensor
- Manufacturer: friend A/S
- Model identifier: WISZB-121
- Device type: 
  - Sensor

This appears to be a chopped down version of https://github.com/dresden-elektronik/deconz-rest-plugin/blob/master/devices/frient/wiszb-120_open_close_sensor.json, without the temperature sensor.

## Screenshots

<img width="592" alt="Screenshot 2022-06-16 at 16 49 09" src="https://user-images.githubusercontent.com/846372/174112453-fcc8f8c6-bc72-4626-8e53-3ea88724f46e.png">

<img width="365" alt="Screenshot 2022-06-16 at 16 47 39" src="https://user-images.githubusercontent.com/846372/174112195-89b58bfc-3131-40c8-bbfa-01e4151cbd5b.png">
/Devel

### Basic

<img width="587" alt="Screenshot 2022-06-16 at 16 48 56" src="https://user-images.githubusercontent.com/846372/174112498-e63bb3b8-71a6-4dfa-a8f3-14dece503ecf.png">

### Identify
<img width="586" alt="Screenshot 2022-06-16 at 16 50 26" src="https://user-images.githubusercontent.com/846372/174112933-7623cc81-1390-434a-822a-7b14d991de76.png">

### Power configuration

<img width="580" alt="Screenshot 2022-06-16 at 16 50 34" src="https://user-images.githubusercontent.com/846372/174113067-8d380f58-e5e7-4d66-82a2-727fcc3704e1.png">

### Binary Input (Basic)

<img width="586" alt="Screenshot 2022-06-16 at 16 50 50" src="https://user-images.githubusercontent.com/846372/174113150-3d923b95-9669-4bac-97ac-f9f2793ad0e7.png">

### Poll Control

<img width="570" alt="Screenshot 2022-06-16 at 16 51 02" src="https://user-images.githubusercontent.com/846372/174113221-66ca9c09-5508-4672-a00e-49afe49c2c65.png">

### IAS Zone

<img width="573" alt="Screenshot 2022-06-16 at 16 51 31" src="https://user-images.githubusercontent.com/846372/174113284-3f31bb45-eb29-4656-82b3-494bb602233a.png">

[...]

PR #6137, opened 2022-06-16
Add wiszb 121

paddyobrien:
For https://github.com/dresden-elektronik/deconz-rest-plugin/issues/6136

These are just copies of the existing DDF files for the WISZB-120 with the temp sensor item removed.

I'm not sure if this is the correct approach but seems like a reasonable starting point!

### DIFF
--- a/devices/develco/wiszb-121_open_close_sensor.json
+++ b/devices/develco/wiszb-121_open_close_sensor.json
@@ -1,0 +1,152 @@
+{
+    "schema": "devcap1.schema.json",
+    "manufacturername": "Develco Products A/S",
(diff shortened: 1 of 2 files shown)


paddyobrien:
I figured out how to test this locally and it appears to be working correctly:

Screenshot 2022-06-16 at 22 11 38

SwoopX:
Would you mind closing your PR in favor of mine? I've seen that you also have a small mistake in there (which shouldn't do any harm) 🙂

paddyobrien:
Oh great thanks for taking a look! Closing

PR #6145, opened 2022-06-20
Add missing json file for resource item `config/resetpresence`

SwoopX:
Adds the missing file and amends refresh interval for `swversion` of Aqara FP1 presence sensor.

### DIFF
--- a/devices/generic/items/config_resetpresence_item.json
+++ b/devices/generic/items/config_resetpresence_item.json
@@ -1,0 +1,9 @@
+{
+	"schema": "resourceitem1.schema.json",
+	"id": "config/resetpresence",
+	"datatype": "Bool",
+	"access": "RW",
+	"public": true,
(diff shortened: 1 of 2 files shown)


manup:
Noticed this as well while updating the docs, there is one more:

image


```
{
	"id": "config/unoccupiedheatsetpoint",
	"datatype": "Int16",
	"access": "RW",
	"public": true,
	"default": 0,
	"description": "Unoccupied target temperature of a thermostat.",
	"parse": {"at": "0x0014", "cl": "0x0201", "ep": 0, "eval": "Item.val = Attr.val;", "fn": "zcl"},
	"read": {"at": "0x0014", "cl": "0x0201", "ep": 0, "fn": "zcl"},
	"range": [500, 3200]
}
```

(not sure if I got this one right)

BabaIsYou:
unoccupiedsetpoint cames from non validated PR #6011 that is still waiting for an answer ... -)

PR #6151, opened 2022-06-26
Filter inappropriate illumination values for Aqara P1 presence sensor

SwoopX:
Apparently, when there's not much light, the sensor seems to report extremely high values not making any sense. While testing, the value lux value has not gone below 65531 lux. However, to allow for some buffer, the maximum allowed lux value has been set to 65519, anything above will be considered as 0.

It's unknown if that is a bug in the device firmware or maybe even a hardware issue. As of now, Xiaomi hasn't released an updated firmware.


manup:
If I remember correctly 0xFFFF is specified as invalid value in the spec.
Related: For the log10 issue there is now a PR https://github.com/dresden-elektronik/deconz-rest-plugin/pull/6152

SwoopX:
Don't be tempted to assume Xiaomi is sticking to the rules :)

Z2M pretty much implemented the same.

PR #6162, opened 2022-07-02
Update DDF for Aqara TVOC sensor

SwoopX:
Update includes:

- Promote DDF to Gold status
- Change processing of `state/airquality`, as the device reports it via attribute 0x0129 on the 0xfcc0 cluster. The applicable scale is the one of the WHO
- Change `attr/swversion`. Recent discovery is that item 0x08 of the Xiaomi special reporting holds the true firmware version (also displayed and confirmed by the Xiaomi App). In rare cases, it can also be item 0x0d on different devices
- Consider the device as awake, when `attr/modelid` is recieved. This usually happens when the device button is pressed (it does mac poll then, should be standard for all Xiaomi battery powered devices)
- Set a binding and attribute reporting for `state/airqualityppb`. Some people mentioned reporting would eventually stop otherwise. Has been tested for more than 10 days and reports were still coming in, the actual value hasn't changed much though

### DIFF
--- a/devices/xiaomi/xiaomi_airmonitor_acn01.json
+++ b/devices/xiaomi/xiaomi_airmonitor_acn01.json
@@ -76,11 +76,7 @@
           "name": "config/reachable"
         },
         {
-          "name": "config/temperature"
-        },
-        {
           "name": "state/airqualityppb",
-          "awake": true,
           "parse": {
             "at": "0x0055",
             "cl": "0x000c",
@@ -186,9 +182,6 @@
         },
         {
           "name": "config/reachable"
-        },
-        {
-          "name": "config/temperature"
         },
         {
           "name": "state/humidity",
--- a/devices/xiaomi/xiaomi_airmonitor_acn01.json
+++ b/devices/xiaomi/xiaomi_airmonitor_acn01.json
@@ -96,11 +96,18 @@
             "at": "0x0129",
             "cl": "0xfcc0",
             "ep": 1,
-            "eval": "if (Attr.val === 1) {Item.val = 'excellent'} else if (Attr.val === 2) {Item.val = 'good'} else if (Attr.val === 3) {Item.val = 'moderate'} else if (Attr.val === 4) {Item.val = 'poor'} else if (Attr.val === 5) {Item.val = 'unhealthy'}",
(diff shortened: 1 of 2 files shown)


manup:
>Change attr/swversion. Recent discovery is that item 0x08 of the Xiaomi special reporting holds the true firmware version (also displayed and confirmed by the Xiaomi App). In rare cases, it can also be item 0x0d on different devices

Interesting, does this apply to all Xiaomi devices? I'd prefer to have this in a separate .js file then, similar to Tuya version parsing in `devices/tuya/tuya_swversion.js`. I don't really understand what the JS does here? :)

```js
let b = Attr.val.toString(16);
c = [0,2].map(function(o) {if (b.length % 2) {b = '0' + b} return b.slice(o,o+2)});
f = parseInt(c[1], 16).toString();
x = ''; for (i = 0; i < (f.length % 4); i++) {x += '0'};
Item.val = '0.0.0_' + x + f;
```

Do you have an example what the input number for the version is and to what string it is converted?

SwoopX:
This definitively seems to apply for the newer Xiaomi devices, however, some break out of the pattern and my use 0x0d (iirc, this was the case for the wired H1 switch family).

E.g., the H1 switch ws-euk04 sends 0x0B0C (2838) before a firmware upgrade, Xiaomi app showed FW version 0.0.0_0022. In essence, the script takes the least significant Byte, checks how many 0s are required to have a length of 4 (in above example 0022) and constructs the version string.
After FW upgrade, the device sends 0x0B0D (2839), which is 0.0.0_0023. Confirmed all that with the TVOC sensor, the FP1 and the H1 switch (though it is index 0x0d here).

I guess, the script could also be rewritten a bit. Happy to move it to a seperate file.

SwoopX:
@manup Code is now moved to seperate js file

manup:
Thanks looks better :)

I've just merged the standard `String.padStart` polyfill in PR https://github.com/dresden-elektronik/deconz-rest-plugin/pull/6186, which might be handy here to simplify the padding of strings.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/padStart

```
let v = "123";

v = v.padStart(6, "0");

// v == "000123"
```

SwoopX:
Yup, that indeed works 🙂

Issue #6160, opened 2022-07-01
Hue tap dial switch


ebaauw:
## Device

- Product name: Hue tap dial switch
- Manufacturer: Signify Netherlands B.V.
- Model identifier: RDM002
- Device type : Please remove all unrelated device types. 
  - Remote

As mentioned on Discord, the device seems to be the love child of the 2021 Hue dimmer switch (#4287) and the Lutron Aurora (#2305).  Unlike the Aurora, it pairs with deCONZ immediately, and It reports button/rotary-events on 0xFC00 without any configuration.  It is using the same [scheme](https://github.com/dresden-elektronik/deconz-rest-plugin/issues/2305#issuecomment-584359012) as the Aurora for rotary events.  Like the Hue in-wall module, it doesn't seem to support hybrid mode (reporting on 0xFC00 and sending _On/Off_ and _Level Control_ commands at the same time),despite the client clusters, even after writing 0x0B to 0x0000/0x0031.  It won't be able to control the lights directly.

The device feels well constructed and pleasantly heavy.  However it's not quite heavy enough to move the dial while it's laying on the table: the whole device moves as there's not enough friction between the switch and the table.  When pressing down a bit while dialling, it works fine, though.  Like the Hue dimmer switch, the buttons give good feedback when pressed, as does the dial when turned.

Exposing the buttons and battery will be straightforward.  For the dial, we'd need to introduce a ZHARelativeRotary `/sensors` type.  And we'll need a trick to handle the `swversion`, probably just faking another endpoint, cf. the IKEA Starkvind.

## Screenshots
![Screenshot 2022-07-01 at 17 01](https://user-images.githubusercontent.com/22179355/176920282-3a78c607-9b43-481f-ad75-51358c392490.png)
![Screenshot 2022-07-01 at 17 05](https://user-images.githubusercontent.com/22179355/176920646-2f1726c8-274f-430d-9316-82163a2c9f1f.png)

### Basic
![Screenshot 2022-07-01 at 17 02](https://user-images.githubusercontent.com/22179355/176920302-1f4c11b4-34f5-4dfa-8980-644d595ef6dc.png)
### Power Configuration
![Screenshot 2022-07-01 at 17 02 1](https://user-images.githubusercontent.com/22179355/176920325-48e400aa-f049-41e1-8b59-470f37054305.png)
### Hue
[...]

PR #6164, opened 2022-07-03
Refactor Hue sensors and controllers.

ebaauw:
Refactor support for the following Hue devices:
- Hue motion sensor;
- Hue outdoor motion sensor;
- Hue motion sensor (2022);
- Hue outdoor motion sensor (2022);
- Hue dimmer switch;
- Hue dimmer switch (2021);
- Hue smart button;
- Hue wall switch module;
- Hue tap dial switch;
- Lutron Aurora Friends-of-Hue switch.

All devices are now fully support by DDF, and the C++ code specifically for these devices has been removed.

Changes:
- Add support for Hue tap dial switch (see #6160);
- New ZHARelativeRotary sensors type for Hue tap dial switch.  It should work the same as for the Hue API.
- ~~`mode` has been removed from the ZHASwitch resources for the controllers;~~
- `config.pending` has been removed from the ZHAPresence, ZHALightLevel, and ZHATemperature resources for the motion sensors;
- Bug fix: updating `config.offset` wouldn't change `state.temperature` immediately;

Note that the issue with Lutron Aurora not working with other gateways than the Hue bidge still exists.  Consequently it's exposed as a single ZHASwitch resource, mapping rotations of the dial to button events.


Kane610:
Is it the ```
For ubisys S1/S2, operation mode of the switch.
Supported values:
"momentary"
"rocker"
``` which has been removed?

**Comment by ebaauw** [comment]
Note that only the Hue devices are changed by this PR.

> Is it the mode ... which has been removed?

No, my bad, I meant the `mode` top-level attribute (the one with a numerical value), that for some (dresden?) switches indicates whether the switch in scenes or dimming mode (or something).  This is the mode used in `buttonmap.json`, which was never updatable through the API.  Updated the text above.

> Is all existence of config.pending removed?

Only for the Hue motion sensors.  It wasn't doing anything anymore with the DDF setup (i.e. it would remain `[]` even if you changed a `config` attribute through the API).  Since the Hue motion sensors are light sleepers, it wasn't really needed anyways.  I orginally copied that from the Hue API, back in the day when we thought we wanted to keep the deCONZ REST API as similar as possible.

**Comment by ebaauw** [comment]
Damn, `mode` is still there, even though not defined in the DDF?!  It's even listed in the issue - totally missed that.

It's not even a `ResourceItem`; it's hard-coded for ZHASwitch resources to be included in the GET response.  https://github.com/dresden-elektronik/deconz-rest-plugin/blob/42f07f6c21c01a17be6c5c9163804cfd6437609d/rest_sensors.cpp#L2867-L2871

File: de_web_plugin.cpp [hunk 1]
```diff
@@ -9564,20 +9466,6 @@ void DeRestPluginPrivate::updateSensorNode(const deCONZ::NodeEvent &event)
...
-                                    {
-                                        quint16 mask = item->toNumber();
-                                        mask &= ~R_PENDING_USERTEST;
-                                        item->setValue(mask);
-                                        Event e(RSensors, RConfigPending, i->id(), item);
-                                        enqueueEvent(e);
-                                    }
-                                    i->clearRead(WRITE_USERTEST);
-                                }
-
```
SwoopX, line 9580: In my view, the whole code for attributes 0x0032 - 0x0034 can be deleted here. I mean if all Philips devices get their own DDFs, there's probably no reason to keep it? Not sure about the lights though.

devices/generic/items/state_expectedeventduration_item.json
```diff
@@ -0,0 +1,8 @@
+{
+	"schema": "resourceitem1.schema.json",
+	"id": "state/expectedeventduration",
+	"datatype": "UInt16",
+	"access": "R",
+	"public": true,
+	"description": "The expected duration of the last received rotary event."
```
SwoopX, line 7: Might be beneficial to include also `"default": 0` for all the to be added resource items

occupancy_sensing.cpp
```diff
@@ -212,20 +212,6 @@ void DeRestPluginPrivate::handleOccupancySensingClusterIndication(const deCONZ::
...
-                {
-                    quint16 mask = item->toNumber();
-                    mask &= ~R_PENDING_SENSITIVITY;
-                    item->setValue(mask);
-                    Event e(RSensors, RConfigPending, sensor->id(), item);
-                    enqueueEvent(e);
-                }
-                sensor->clearRead(WRITE_SENSITIVITY);
-            }
-
```
SwoopX, line 228: Here, I'd also say the while case can be removed?

rest_sensors.cpp
```diff
@@ -803,14 +803,7 @@ int DeRestPluginPrivate::changeSensorConfig(const ApiRequest &req, ApiResponse &
...
                 {
-                    if (RConfigDeviceModeValues.indexOf(data.string) >= 0)
-                    {
-                        pendingMask |= R_PENDING_DEVICEMODE;
-                        sensor->enableRead(WRITE_DEVICEMODE);
-                        sensor->setNextReadTime(WRITE_DEVICEMODE, QTime::currentTime());
-                        updated = true;
-                    }
-                    else if (!data.string.isEmpty())
+                    if (!data.string.isEmpty() && RConfigDeviceModeValues.indexOf(data.string) >= 0)
```
SwoopX, line 806: The if statement could be deleted if the `isEmpty()` check is moved one level up.

rest_sensors.cpp
```diff
@@ -1039,6 +1012,8 @@ int DeRestPluginPrivate::changeSensorConfig(const ApiRequest &req, ApiResponse &
                 }
                 else if (rid.suffix == RConfigOffset) // Signed integer
                 {
+                    offset = data.integer - item->toNumber();
```
SwoopX, line 1015: Wouldn't `offset = data.integer` be more adequate here?

SwoopX (commented):
Just a couple of minors so far. However, I left out the DDFs for the time being, as I need some more time to take a look. In this regard, my gut feeling says there might be some impact on previous behavior/functionality, but hope, it will say just a feeling.

Interesing approach on the button events. Curious to hear Manuel's opinion on that one.

de_web_plugin.cpp
```diff
@@ -230,10 +230,6 @@ static const SupportedDevice supportedDevices[] = {
     { VENDOR_NYCE, "3014", emberMacPrefix }, // NYCE garage door/tilt sensor
     { VENDOR_NYCE, "3041", emberMacPrefix }, // NYCE motion sensor
     { VENDOR_NYCE, "3043", emberMacPrefix }, // NYCE ceiling motion sensor
-    { VENDOR_PHILIPS, "RWL02", philipsMacPrefix }, // Hue dimmer switch
-    { VENDOR_PHILIPS, "ROM00", philipsMacPrefix }, // Hue smart button
-    { VENDOR_PHILIPS, "RDM00", philipsMacPrefix }, // Hue wall switch module
-    { VENDOR_PHILIPS, "SML00", philipsMacPrefix }, // Hue motion sensor
```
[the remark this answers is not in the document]
ebaauw: Yes.  Did I miss that one?

rest_sensors.cpp
ebaauw: No, you need to undo the previous offset value and then apply the new offset.  The temperature/humidiy ResourceItem holds the adjusted value, not the value reported by the sensor.

de_web_plugin.cpp
ebaauw: I think you're right.  I'll do a search and destroy for any code handling these attibutes.
I haven't seen any Hue lights using these attributes.

occupancy_sensing.cpp
ebaauw: Indeed, both `HUE_SENSITIVITY` and `HUE_SENSITIVITY_MAX` are Hue-specific.

devices/generic/items/state_expectedeventduration_item.json
ebaauw: Wouldn't it default to `null`?

ebaauw:
> Interesing approach on the button events. Curious to hear Manuel's opinion on that one.

Stole the idea to set multiple ResourceItems from a single `parse` of the Zigbee attribute value from `devices/generic/illuminance_cluster/sml_lightlevel`.  It seems more efficient that parsing the same Zigbee attribute values multiple times.

The `global` comment is to satisfy jsonlint.  I use that for my Homebridge plugins to adhere to the [standard Javascript](https://standardjs.com) layout.

de_web_plugin.cpp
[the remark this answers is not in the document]
ebaauw: Did remove the entry, but the code is still needed, as it the entry in `buttonmaps.json`.

bindings.cpp
```diff
@@ -3358,13 +3207,6 @@ bool DeRestPluginPrivate::checkSensorBindingsForAttributeReporting(Sensor *senso
...
         else if (*i == VENDOR_CLUSTER_ID)
         {
-            if (sensor->modelId().startsWith(QLatin1String("RWL02")) || // Hue dimmer switch
-                sensor->modelId().startsWith(QLatin1String("ROM00")) || // Hue smart button
-                sensor->modelId().startsWith(QLatin1String("RDM00")) || // Hue wall switch module
-                sensor->modelId().startsWith(QLatin1String("Z3-1BRL"))) // Lutron Aurora Friends-of-Hue dimmer switch
-            {
-                val = sensor->getZclValue(*i, 0x0000); // button event
-            }
             if (sensor->modelId() == QLatin1String("de_spect")) // dresden elektronik spectral sensor
```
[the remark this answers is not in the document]
manup: The `de_spect` will be removed in a followup PR. It was a internal prototype spectral sensor for a student measuring rgb parts of the light spectrum. If necessary it can be added later as DDF.

bindings.cpp
```diff
@@ -3635,24 +3438,9 @@ bool DeRestPluginPrivate::checkSensorBindingsForClientClusters(Sensor *sensor)
     //quint8 srcEndpoint = sensor->fingerPrint().endpoint;
     std::vector<quint16> clusters;
 
-    if (sensor->modelId().startsWith(QLatin1String("RWL02")) || // Hue dimmer switch
-        sensor->modelId().startsWith(QLatin1String("ROM00")) || // Hue smart button
-        sensor->modelId().startsWith(QLatin1String("ElkoDimmer")) || // Elko dimmer
-        sensor->modelId().startsWith(QLatin1String("E1E-"))) // Sengled smart light switch
```
manup, line 3641: Do we have "ElkoDimmer" and "E1E-" already on DDF?

de_web_plugin.cpp
```diff
@@ -1292,14 +1287,7 @@ void DeRestPluginPrivate::apsdeDataIndication(const deCONZ::ApsDataIndication &i
...
         // case DE_CLUSTER_ID:
-            if (zclFrame.manufacturerCode() == VENDOR_PHILIPS)
-            {
-                handlePhilipsClusterIndication(ind, zclFrame, device);
-            }
-            else // Shouldn't we check for DE manufacturer code?
-            {
-                handleDEClusterIndication(ind, zclFrame);
-            }
+            handleDEClusterIndication(ind, zclFrame);
```
manup, line 1290: Feel free to remove the whole `handleDEClusterIndication()` function as it is basically a NOP wasting cycles :)

de_web_plugin.cpp
```diff
@@ -3059,7 +3043,9 @@ void DeRestPluginPrivate::addLightNode(const deCONZ::Node *node)
 
         if (existDevicesWithVendorCodeForMacPrefix(node->address(), VENDOR_PHILIPS))
         {
-            if (lightNode.manufacturer() != QLatin1String("Philips"))
+            if (lightNode.manufacturer() != QLatin1String("Philips") &&
+                lightNode.manufacturer() != QLatin1String("Signify Netherlands B.V.")
+               )
             { // correct vendor name, was atmel, de sometimes
```
manup, line 3049: Note for the future we need to get rid of this. The Device code already robustly queries the Basic cluster manufacturer name, no mac prefix magic needed.

manup (commented):
First small review, I'm still testing will report if I find anything.

bindings.cpp
ebaauw: Damn, I completely missed those.  I think we’d better keep the code for these.

de_web_plugin.cpp
ebaauw: Done

de_web_plugin.cpp
ebaauw: So leave it for now?

devices/generic/items/state_expectedeventduration_item.json
SwoopX: Yes. It's just beauty I suppose...

de_web_plugin.cpp
manup: Thinking about it a bit more, I'd suggest we can remove this code. The DDFs should use the actual exposed manufacturer name from the *Basic Cluster*, it can be cumbersome to have "Philips" entries in the database which should be "Signify Netherlands B.V.".

manup:
Yeah I think so too, did a bunch of tests pairing my 1. gen SML001 motion sensor. Together with the fixes in PR https://github.com/dresden-elektronik/deconz-rest-plugin/pull/6196 this works very well.

Next up, I'll test the 1. and 2. gen dimmer switches before merging. After that I'd like to debug and fix the issue with the rules not triggering proper (before the v2.18.0-beta release).

Mimiix:
> After that I'd like to debug and fix the issue with the rules not triggering proper (before the v2.18.0-beta release).

Can you please place this in the issue about it (that it's being worked on?).

Thanks!

manup:
Tested with SML001, RWL022 and RWL021 including #6196 works nicely.

Finne75:
I noticed that the battery status from cluster 0x0001 is reported every 300 seconds for Hue wall module. Are there any objections to set this to something higher and more battery saving like min 3600 max 86400?

ebaauw:
That setting was copied from the Hue bridge, so I wouldn’t worry about battery consumption.  Since, unlike the Hue motion sensor, the module doesn’t do any other periodic reporting, this setting basically determines the interval at which the module polls its parent router.  If you would change it, it might take longer for configuration changes to reach the module (if at all).

Finne75:
Ack, thanks for explaining! BTW I noticed that my wall modules regularly send DeviceAnnce messages, I wonder whether this is a way to increase reliability by avoiding network drops (?) or whether it is something else.

PR #6165, opened 2022-07-03
Adapt approach to auto-reset presence/vibration for DDF devices

SwoopX:
This PR is aimed at circumventing usage of `durationDue`, which is one of the reasons the new code ignoring the cluster handlers cannot be more widely applied (e.g. occupancy sensing cluster).

In essence, a new resource item, `config/autoreset`, is introduced, which shall indicate if e.g. the presence state of a device needs to be reset automatically. As this item will be only available on DDF supported devices, it shall also serve for distinction. The actual role of `durationDue` will be taken over by `item->lastZclReport()`. In conjuction with `config/duration`, which must be present as well if auto-reset is required, it is checked if the current time is larger than the last report time + duration.

This has successfully been tested with the Xiaomi presence sensor lumi.sensor_motion.aq2. Further devices which are affected by this are the Aqara vibration sensor lumi.vibration.aq1(#6167), Aqara high presicion sensor lumi.motion.agl04 (#6168) and Aqara P1 presence sensor lumi.motion.ac02.
The legacy code path is still followed for devices without DDF.

### DIFF
--- a/de_web_plugin.cpp
+++ b/de_web_plugin.cpp
@@ -1372,7 +1372,7 @@ void DeRestPluginPrivate::apsdeDataIndication(const deCONZ::ApsDataIndication &i
             break;
 
         case OCCUPANCY_SENSING_CLUSTER_ID:
-            if (!DEV_TestStrict()) { handleOccupancySensingClusterIndication(ind, zclFrame); }
+            if (!DEV_TestStrict() && !devManaged) { handleOccupancySensingClusterIndication(ind, zclFrame); }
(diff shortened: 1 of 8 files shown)


SwoopX:
After some discussion with ebaauw I'm not too sure anymore if `config/autoreset` is really required, as `config/duration` might already enough to indicate a reset is required. I'm not too sure though if that was consistently applied/used back in the days...

SwoopX:
No consensus

PR #6167, opened 2022-07-03
Update DDF for Aqara Vibration Sensor DJT11LM

SwoopX:
This PR depends on #6165.

This is a major overhaul of the DDF:
- Promotion to Gold status
- Removed <s>`config/temperature` and</s> `config/pending` resource items
- Add `config/duration` and `config/autoreset` resource items. The vibration status does apparently not get reset automatically, but now this can be defined to happen after the time defined in duration.
- The functions for parsing/writing had an inappropriate mfc set, which got corrected
- Sentitivity values got corrected based on a sniff of a native gateway
- Some minor adjustments

Respective legacy code shall be removed by a follow up PR, to allow easy fallback in case of errors.

### DIFF
--- a/devices/generic/items/state_tiltangle_item.json
+++ b/devices/generic/items/state_tiltangle_item.json
@@ -5,5 +5,6 @@
 	"access": "R",
 	"public": true,
 	"range": [0, 360],
-	"description": "The current tilt angle."
+	"description": "The current tilt angle.",
+	"default": 0
 }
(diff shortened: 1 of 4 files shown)


Thomas-Vos:
Could you explain why `config/temperature` was removed, please? Is there an alternative?

SwoopX:
> Could you explain why `config/temperature` was removed, please? Is there an alternative?

Xiaomi doesn't offer anything like this, that's why it got removed.

SwoopX:
No consensus

PR #6170, opened 2022-07-03
Add DDF support for Namron Zigbee 1 Channel Switch K2 Black (4512728)

SwoopX:
Yet another model ID.


yusijs:
I realize this is the wrong place to ask, but assume this way I'll disturb as few people as possible so will try here first of all:
How can I contribute to the device lists? I tried going into the container and changing the json-files, but nothing changed. Do I need to restart the rest-api service first (or in my case, rebuild the docker-container)?

PR #6174, opened 2022-07-05
Fix rare and device-related SIGSEGV

SwoopX:
Missing ZCL payload length check can cause a segmentation fault

### DIFF
--- a/de_web_plugin.cpp
+++ b/de_web_plugin.cpp
@@ -5583,11 +5583,11 @@ void DeRestPluginPrivate::checkSensorButtonEvent(Sensor *sensor, const deCONZ::A
                 {
                     ok = false;
 
-                    if (buttonMap.zclParam0 == zclFrame.payload().at(1))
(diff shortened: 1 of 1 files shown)


manup:
Another rabbit hole :D
Quick look over the code base for `payload().at(`  it looks we have more places (not related to this PR) which may have the same damage potential.

I think it might be good to add a safe wrapper around `payload().at(index)` like `ZclFrame::payloadAt(index)` which doesn't dive into NULL pointer territory and returns 0 for out of bounds indices.

ebaauw:
Would we need a similar check for the `ZclFrame.at()` in the DDF and included scripts?

SwoopX:
That function is truly a ticking bumb and it's remarkable that there weren't any notable shout outs yet. Haven't had another look around, but I share your feeling. Such a wrapper would make good sense and I assume it should also cover the DDFs, as Erik mentions.

PR #6175, opened 2022-07-05
Add further Danfoss specific resource items

SwoopX:
This PR adds a few missing resource item json files:
- `config/lastchange_source`
- `config/lastchange_time`
- `config/lastchange_amount`

and adds a few new resource items for thermostats (primarily Danfoss specific):
- `config/mode` -> Generic item, allowing for values 'off' and 'heat'. Set to 'off' when heating period ends to prevent unnecessary valve adjustments.
- `config/loadbalancing` -> Thermostat runs stand alone (false) or shares load with others in the same room (true).
- `config/radiatorcovered` -> Radiator is standing freely (false) or is covered (true). Set this to true for Danfoss Ally to use external temp sensor.
- `config/windowopendetectionenabled` -> Turns the window open detection feature on or off.


Prior99:
This would be really useful to have.

Didier3L:
Good evening, there is also `Heat Supply Request` which allows you to know if the radiator needs hot water
image

ebaauw:
Shouldn’t the `lastchange` attributes be in `state` rather than `config`?

SwoopX:
> Shouldn’t the `lastchange` attributes be in `state` rather than `config`?

Yeah, I guess so. Not sure if you had a reason to do it as is back in the days? 🙂 
Can change it once we have had the discussion on the thermostat resource items.

Didier3L:
these changes will be available when in the latest version ?

Smanar:
What is blocking for this PR ?

Pe-MaKer:
I would really appreciate these items to be available soon.
It's a shame that one has to set config/heatavailable in the UI for each thermostat twice a year if the solution just has to be merged.

Jonaslinderoth:
I would really appreciate  it if we could add the items discussed in as well #6872.
Allowing for reading the load estimate and writing to the load radiator load mean, will make it possible to calculate the average load estimate for all the radiators in the room, and forwarding it to each radiator. 

This will help distribute the load between the radiators which is helpfull when you have multiple radiators of different sizes in the same room.

manup:
Almost ready, can you please also change the remaining `{"fn": "zcl"}` cases so the validator won't scream :)

SwoopX:
> Almost ready, can you please also change the remaining `{"fn": "zcl"}` cases so the validator won't scream :)

That now took me a while to understand which ones you meant as I was focussed on Danfoss and totally forgot about the additions 😀 Done

Hencor93:
May a little late, but I just saw that the PR was merged now. 
I miss the Items for Config `Load Radiator Room Mean` and the `Load estimate on this radiator` as a State. Without these items the load balancing mode on the TRVs cannot be used. 
These two are present in the PR #6872 and was missed to merged into this PR I think.

Smanar:
I think the better solution is making a new PR just for them, it's not possible (more complex) to rollback a PR just for that.

PR #6188, opened 2022-07-10
Update Xiaomi FW version strings for DDF supported devices

SwoopX:
Devices will now get the same version string as shown in their native app.


manup:
Ready for v2.18.0-beta?

PR #6218, opened 2022-07-24
DDF for Terncy Smart Dial SD01

SwoopX:
Adds respective DDF and removes legacy code.

Note that newly added devices require a device firmware update in order to work properly.


manup:
Ouh this one sits here for a while ;) from my side the PR looks all right. Should be merged in v2.21.0-beta?

>Note that newly added devices require a device firmware update in order to work properly.

Did the older version work with the C++ code variant, breaking change?

SwoopX:
Yup, that one is sitting tight :) More than happy to get that one off the road.

It is indeed some sort of breaking change. With an old firmware, the device doesn't have a manufacturer name at all, so we have/had a hack for that in the legacy code. The new firmware corrects that and as I have the device here, I know it works. Nevertheless, it should be mentioned explicitly in the release notes (note that there's also a respective md file).

manup:
Ok, yeah I guess it's better to take the OTA update route and describe it as a breaking change than to carry around the hack.

I've added a new "Breaking Change" label which can be assigned to PRs, the changelog generator needs a little update and then is able to put it in the right section.

PR #6235, opened 2022-07-30
Fix state changes failing prematurely under rare conditions

SwoopX:
In case read and parse function for a resource item differ, the state change will fail, as the read attributes command cannot be issued. This leads to a condition where the state change will always have the state failed and gets deleted on the next occasion.

As of now, this seems to impact just a hand full of devices.


manup:
Good catch, the read function must only use the read parameters not the data from parse parameters.

PR #6252, opened 2022-08-09
Fix potentially jumpy reachable state for sensors

SwoopX:
When a sensor has a cluster in its fingerprint, which doesn't exist under the sensor's endpoint, function `checkSensorNodeReachable()` will set the respective sensor unreachable, as it checks for existence of **all** clusters.

While it makes sense to add those non-existent clusters to leverage some advantages on the legacy code, it apparently impacts the overall sensor status (falsely). To resolve this, the last check in the function to check for all clusters has been made lazy to just find one fit and then move on.

It's more unlikely this will occur on DDF supported devices, but can happen there as well (especially if the device is not newly paired). Eventually, this cluster check can be removed completely, as it seems to be a too tight precaution.

### DIFF
--- a/de_web_plugin.cpp
+++ b/de_web_plugin.cpp
@@ -4276,7 +4276,9 @@ void DeRestPluginPrivate::checkSensorNodeReachable(Sensor *sensor, const deCONZ:
                 reachable = true;
             }
 
-            // check that all clusters from fingerprint are present
+            bool found = false;
+
+            // check if any cluster from fingerprint is present
             for (const deCONZ::SimpleDescriptor &sd : sensor->node()->simpleDescriptors())
             {
                 if (!reachable)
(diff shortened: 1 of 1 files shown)


manup:
I'll have a closer look the coming days for review. Indeed DDF devices shouldn't be affected by this anymore. Hopefully we can get rid of the whole fingerprint stuff over time, causes more trouble than it solves :)

manup:
I think most of the cluster related code now does nothing? ;)

If `reachable = false` early on, the loop is finished right away. reachable only depends on lastRx() which is good and how the DDF code does it.

https://github.com/dresden-elektronik/deconz-rest-plugin/pull/6252/files#diff-d0aaee66bc2f60fdd243c0e5b86aebca01fa3eb36c7ec39578d5beeca47078abR4282-R4284

I'd suggest to throw out the now not needed cluster and endpoint code here to simplify things.

SwoopX:
Thanks, I'm all in 😄

PR #6264, opened 2022-08-13
Fix potentially unwanted behavior if DDF has `cmd` defined in zcl parser

SwoopX:
When a ZCL Frame should be parsed for a resource item through the respective `zcl` parse function, defining a command is optional. However, if it is defined and the intention is to only consider those commands, one might experience some unwanted side effects.

Something like `"eval": "Item.val = (ZclFrame.payloadSize == 1 && ZclFrame.at(0) == 0x00) ? false : true;"` will way to often result in true, as all incomming responses are evaluated and not pre-filtered by commandID. The PR now adds that pre-filtering to be able to bail out early.


manup:
Good catch!

PR #6268, opened 2022-08-15
Extend Xiaomi special report parsing

SwoopX:
The idea of this PR is to dedicate a non-public resource item to a Xiaomi resource, allowing to extract more than 1 item out of the special report (provided as string list), while setting the respective values of up to 3 resource items (could probably be extended to 4).

With the currently given read/write/parse functions, one has to decide how to obtain the current state values:
- either by reading an attribute (polling) or 
- by parsing the respective item of the special report

While the 2nd option is quite charming, as data arrives in more or less defined intervals, it actually gets tricky when a configuration setting should be changed by a state change (SC). The SC tries to write an attribute value, then reads that value, which is then evaluated by the parser. When now read and parse are async (read goes for attribute XYZ and parse uses the special reporting), the SC cannot succeed as a correctly written value cannot be confirmed. In case a write wouldn't succeed, deconz holds the assumed written value while it truly remained untouched on the device and that would eventually never be corrected.

The above is especially relevant for sleeping devices like sensors. Detaching the special report parsing from all or, more precisely, the most relevant resource items could aid in the SCs to work properly and succeed, reduce polling while allowing to use the device's awake time more efficiently and still update 3 resource items in one go.

This is certainly not the silver bullets to cater with all Xiaomi specifics and there could be more items of interest from the special reports, but I consider this as a low hanging fruit to improve device experience.

An example of how an amended DDF could look like is part of the PR (P1 motion sensor). Below the debug output of 3 resource items being set on one special report.

```
20:44:29:576 JsResourceItem.setValue(config/duration) = 25
20:44:29:577 54:ef:44:10:00:49:01:ac-01-0406/config/xiaomispecial expression: if (Attr.id == 0x69) { R.item('config/duration').val = Attr.val; } else if (Attr.id == 0x6b) { R.item('config/ledindication').val = Attr.val; } else if (Attr.id == 0x6a) { R.item('config/sensitivity').val = Attr.val; } --> 25

20:44:29:577 JsResourceItem.setValue(config/ledindication) = 0
20:44:29:578 54:ef:44:10:00:49:01:ac-01-0406/config/xiaomispecial expression: if (Attr.id == 0x69) { R.item('config/duration').val = Attr.val; } else if (Attr.id == 0x6b) { R.item('config/ledindication').val = Attr.val; } else if (Attr.id == 0x6a) { R.item('config/sensitivity').val = Attr.val; } --> 0

20:44:29:578 JsResourceItem.setValue(config/sensitivity) = 2
20:44:29:578 54:ef:44:10:00:49:01:ac-01-0406/config/xiaomispecial expression: if (Attr.id == 0x69) { R.item('config/duration').val = Attr.val; } else if (Attr.id == 0x6b) { R.item('config/ledindication').val = Attr.val; } else if (Attr.id == 0x6a) { R.item('config/sensitivity').val = Attr.val; } --> 2
```

**EDIT**
Small caviat on this approach is that the event emitter misses that the referred to items have been set. It only fires on `config/xiaomispecial`, but does not consider the 3 ones actually set above. Based on that, the change state machine cannot leverage any received values.

### DIFF
--- a/device_access_fn.cpp
+++ b/device_access_fn.cpp
@@ -1102,8 +1102,8 @@ bool parseXiaomiSpecial(Resource *r, ResourceItem *item, const deCONZ::ApsDataIn
     {
         return result;
     }
-
-    if (ind.clusterId() != 0x0000 && ind.clusterId() != 0xfcc0) // must be basic or lumi specific cluster
+    
+    if (!(ind.clusterId() == 0x0000 || ind.clusterId() == 0xfcc0) || zclFrame.payload().isEmpty()) // must be basic or lumi specific cluster
     {
         return result;
     }
@@ -1127,7 +1127,7 @@ bool parseXiaomiSpecial(Resource *r, ResourceItem *item, const deCONZ::ApsDataIn
         if (ind.clusterId() == 0xfcc0)
         {
             param.clusterId = 0xfcc0;
-            param.manufacturerCode = 0x115f;
+            //param.manufacturerCode = 0x115f;
         }
 
         if (map.contains(QLatin1String("ep")))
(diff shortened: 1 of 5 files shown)


SwoopX:
No consensus

PR #6279, opened 2022-08-19
DEV improve robustness of ZCL attribute polling

manup:
- Fix timeout didn't increase the retry counter of a poll item, machine was stuck in infinite retries.
- Also check that clusterId is the one from the request.
- Allow to optional specify `"noseq": true` in DDF item read parameters to ignore the response ZCL sequence number and only match against clusterId. This is needed for devices like Danfoss Ally which respond with a random sequence number.
- Fix potential rare SEGV when poll items are already empty.

### DIFF
--- a/de_web.pro
+++ b/de_web.pro
@@ -12,6 +12,7 @@ QMAKE_CXXFLAGS += -Wno-attributes \
 
 CONFIG(debug, debug|release) {
     LIBS += -L../../debug
+    DEFINES += DECONZ_DEBUG_BUILD
 }
 
 CONFIG(release, debug|release) {
--- a/de_web_plugin.cpp
+++ b/de_web_plugin.cpp
@@ -35,7 +35,6 @@
 #include <cmath>
 #include "air_quality.h"
 #include "alarm_system_device_table.h"
-#include "colorspace.h"
 #include "database.h"
 #include "device_ddf_init.h"
 #include "device_descriptions.h"
@@ -636,6 +635,10 @@ DeRestPluginPrivate::DeRestPluginPrivate(QObject *parent) :
 
     DEV_SetTestManaged(deCONZ::appArgumentNumeric("--dev-test-managed", 0));
 
+#ifdef DECONZ_DEBUG_BUILD
+    EventTestZclPacking();
+#endif
+
     pollManager = new PollManager(this);
 
     databaseTimer = new QTimer(this);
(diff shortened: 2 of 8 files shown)


SwoopX:
@manup Thanks a lot for the PR and the good discussion prior to it. I added the changes to my test version and so far I can say that any poll hammering I observed earlier has stopped. This definitively explains what I've observed every now and then in the sniffer 👍 

However, I feel I still should clarify a bit on what I'm working on, since the remark in your 3rd bullet about the Ally is not quite right. Still trying to get thermostat schedules to be covered solely by DDF, where I've prepared 3 functions for read/write and parse. Now, the main difference is that schedules on thermostats are writte **and** read by a cluster specific command, there's no attribute for it. Attributes btw do always seem to come back with the same seq# as far as I can tell.

For that reason, `"noseq": true` does not really have an effect with schedules, the read function sending a `get weekly schedules` request command gets a `get weekly schedules` response command and doesn't emit a `REventZclResponse` event. Reason for that is https://github.com/dresden-elektronik/deconz-rest-plugin/blob/aa82bc773289d3952f46957d90139392f1365f88/de_web_plugin.cpp#L1023-L1026 resulting in the `schedule` resource item being polled 3 times and then timing out. Yes, could be worse 🙂 

If I add the event emitter just for that very specific cluster command (`get weekly schedules` response command), the item is just polled once and then popped, also considering `"noseq": true`.

Honestly, I'm not sure desirable or even make the emitter no cluster and command specific. What I know is that there's probably only a handful of further/other cluster specific response commands with might be considerable at all (some on the door lock and IAS ACE cluster), so this "little challenge" might affect other devices as well (in future)..

manup:
@SwoopX Ha, took a while, thanks for the explanation now I finally understand. I added a few lines to emit a `REventZclResponse` for server-to-client incoming frames = assumed to be response.

It's a bit rough approach but should do for now — the event is also emitted for IAS Zone Status Change Notification and the like, but that shouldn't do any harm. We can refine this later on if needed.

With that we get the event in the Device poll state machine, however I'm not sure with what read function the initial request is send, there might be a missing `DA_ReadResult::ignoreResponseSequenceNumber := true` if the request isn't send by `readZclAttribute()` but that should be an easy fix.

SwoopX:
@manup Thanks a lot. Admittedly, this is not too easy if you're not in the middle of it. I also think your recent addition shouldn't cause any harm, but any potentially required fine tuning can happen over time.

One of that beloved edge cases. You got no idea how much headache the on-device schedules cause here 🙂 It's one small step after the over...

manup:
This is the way™ :smiley:  It's nice to get  these things ironed out piece by piece and can still yield generic solutions.
We got quite some internal improvements over the last releases without having to change the DDF interface too much.

Issue #4325, opened 2021-02-01
Woox R7060 Smart Garden Irrigation Control


EarlOfBurton:
## Device

- Product name: Woox R7060 Smart Garden Irrigation Control
- Manufacturer: _TZ_3210_eymunffl
- Model identifier: TS0101
- Device type : Please remove all unrelated device types. 
  - Other: Garden Watering Switch

Product page: https://wooxhome.com/products-c10/other-c4/woox-r7060-smart-garden-irrigation-control-p61

## Screenshots
![node_info](https://user-images.githubusercontent.com/16398996/106467752-60b43d00-649d-11eb-91a7-6fd24b78abcd.png)
![cluster_info](https://user-images.githubusercontent.com/16398996/106467762-6447c400-649d-11eb-8a4f-95c3c20b38cb.png)
![basic_cluster](https://user-images.githubusercontent.com/16398996/106467775-6742b480-649d-11eb-9bb0-af8a8aabed29.png)
![overview](https://user-images.githubusercontent.com/16398996/106467782-69a50e80-649d-11eb-9a0d-076aa8ec2fcd.png)

EarlOfBurton:
Now adding _all_ of the cards, since I know what is needed (sorry for that).
## Node Info
![00_node_info](https://user-images.githubusercontent.com/16398996/106476410-5434e200-64a7-11eb-8acc-6a357278c66a.png)
## Basic
![01_basic](https://user-images.githubusercontent.com/16398996/106476027-e25c9880-64a6-11eb-940e-260d3c1e59e8.png)
## Groups (1/2)
![02_groups1](https://user-images.githubusercontent.com/16398996/106476066-ec7e9700-64a6-11eb-9f09-2af0fe42de2c.png)
## Groups (2/2)
![03_groups2](https://user-images.githubusercontent.com/16398996/106476090-f30d0e80-64a6-11eb-9d27-614bf2bf64c1.png)
## Scenes (1/3)
![04_scenes1](https://user-images.githubusercontent.com/16398996/106476112-fb654980-64a6-11eb-8b8a-5256c231c86e.png)
## Scenes (2/3)
![05_scenes2](https://user-images.githubusercontent.com/16398996/106476123-0029fd80-64a7-11eb-9959-f4371332b6fd.png)
## Scenes (3/3)
![06_scenes3](https://user-images.githubusercontent.com/16398996/106476143-05874800-64a7-11eb-8695-182fc3cd243f.png)
[...]

PR #6295, opened 2022-08-28
DDF: Promote woox_r7060_ts0101 Gold

mkarnebeek:
Is it possible to promote the "Woox R7060 Smart Garden Irrigation Control" DDF to Gold? 

I've got 4 of these and they have two issues with the legacy implementation. Both are solved with using the DDF for this device:

- Battery status is not present in the legacy support, but is there in de DDF
- On reboot of deconz, the device is unresponsive to on/off from deconz, and requires a push of a button on the physical device for commands from deconz to work again. With the DDF, the device on/off calls responds as it should after a reboot.

I've been reading up on https://github.com/dresden-elektronik/deconz-rest-plugin/issues/4325 for a solution to both of these problems and I've solved it by just using this DDF

### DIFF
--- a/devices/woox/woox_r7060_ts0101.json
+++ b/devices/woox/woox_r7060_ts0101.json
@@ -5,7 +5,7 @@
   "vendor": "WOOX",
   "product": "TS0101",
   "sleeper": false,
-  "status": "Bronze",
+  "status": "Gold",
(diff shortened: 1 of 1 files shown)


manup:
Well if the DDF works as indented and better than the legacy code, lets raise the status to Gold. :)

PR #6296, opened 2022-08-29
Fix JS padStart polyfill

manup:
The prototype method is padStart not padString.


manup:
@SwoopX this should fix the error you've seen.

PR #6322, opened 2022-09-11
Add DDF for Develco/friend external meter interface ZHEMI101

SwoopX:


devices/develco/zhemi101_external_meter_interface.json
```diff
@@ -0,0 +1,223 @@
...
+          "name": "config/reachable"
+        },
+        {
+          "name": "state/consumption",
+          "refresh.interval": 360,
+          "parse": {
+            "at": "0x0000",
+            "cl": "0x0702",
+            "ep": 2,
+            "eval": "Item.val = Attr.val / 1000",
```
Qalculus, line 147: Hi!
After I updated to the new Deconz version in Home Assistant I noticed that the values from my sensor were suddenly a factor 1000 too small. I bought the EU version of the sensor and live in EU where the standard is 1000 pulses/kWh, which the sensor assumes. I managed to fix it by manually editing the DDF and removing the scaling, but I would suggest not scaling the value in the default DDF, since this will most likely break the values for many users.

Qalculus:
```suggestion
            "eval": "Item.val = Attr.val",
```

Qalculus:
I have created a pull request for this: #6402

PR #6323, opened 2022-09-11
Expose change timeout via optional write function attribute `change.timeout`

SwoopX:
This PR is intended to allow for "external" control over the state change lifetime, which might be especially relevant for deep sleeping enddevices.

The internal default settings are 180 seconds to have a state change completed. However, if you want to change the configuration on a battery Xiaomi device via API, this is not long enough if you don't want to be forced to push the button on the device to wake it up, as they're only able to receive data only once every 50 mins. With the new attribute, the lifetime can be set at will and the relevant Xiaomi DDFs will get a lifetime of 60 mins. That should be sufficient to complete the change (but does not necessarily omit the physical button press).


manup:
We should update the docs soon to not forget about these things 🙂

PR #6353, opened 2022-09-25
Adding Netvox z809ae3r smart plug DDF

clockbrain:
Works ok in my network.


clockbrain:
Hold this one for the moment. I think the current reading is out by a factor of 1000.

clockbrain:
I need to better understand DDF before I progress this one.

PR #6357, opened 2022-09-25
Add DDF for Ikea TRADFRI remote control E1810

SwoopX:


manup:
Are the buttons already handled by the button map?

SwoopX:
Yes, sir.

PR #6359, opened 2022-09-25
Add checks for command ID and MFC to `parseZclAttribute()`

SwoopX:
The PR adds a check for the command ID and the manufacturer code to the standard zcl parsing function. Those were just used once during function assignment, but not subsequently in regular operations.


manup:
Ha good catch!

PR #6370, opened 2022-09-30
Fix scrypt use of mixed signed/unsigned char*

### LINKED ISSUE #6369: Compile errors against openssl: invalid conversion from ‘const char*’ to ‘const unsigned char*’
## Describe the bug
Compiling on Ubuntu (Jammy) on aarch64 of deconz 2.18.02 (git - see below for hash)
The function signature of EVP_PKEY_CTX_set1_scrypt_salt() demands that the salt be of the type unsigned char *, (however the salt is treated as a string, and string functions earlier on in the processing, including using strlen()).

crypto/scrypt.cpp: In function ‘int scryptDerive(const char*, size_t, std::array<unsigned char, 64>&, int, int, int, const char*)’:
crypto/scrypt.cpp:89:60: error: invalid conversion from ‘const char*’ to ‘const unsigned char*’ [-fpermissive]
   89 |     if (result == 0 && EVP_PKEY_CTX_set1_scrypt_salt(pctx, salt, strlen(salt)) <= 0)
      |                                                            ^~~~
      |                                                            |
      |                                                            const char*
In file included from crypto/scrypt.cpp:29:
/usr/include/openssl/kdf.h:123:56: note:   initializing argument 2 of ‘int EVP_PKEY_CTX_set1_scrypt_salt(EVP_PKEY_CTX*, const unsigned char*, int)’
  123 |                                   const unsigned char *salt, int saltlen);
      |                                   ~~~~~~~~~~~~~~~~~~~~~^~~~
make[1]: *** [Makefile.Release:1114: release/scrypt.o] Error 1

Patch inbound to fix the observed issue, but I'm not sure if it is valid to treat the salt as a string in the first case.

## Steps to reproduce the behavior

 Compile deconz-rest-api using the instructions provided, against openssl 3.0.2-0ubuntu1.6 (Ubuntu Jammy)

## Expected behavior
Compile should succeed.

## Screenshots
See log output above.

## Environment
 - Host system: aarch64 Ubuntu Jammy
 - Running method: N/A
 - Firmware version: N/A
 - deCONZ version: git: commit hash 449ab29e1548af8e95631ae2d0c4ef14fe9d14f8
 - Device: N/A
 - Do you use an USB extension cable: N/A
 - Is there any other USB or serial devices connected to the host system? N/A

## deCONZ Logs
See above for compile logs.

Monofin:
Fixes #6369 - or at least silences the compiler.

### DIFF
--- a/crypto/scrypt.cpp
+++ b/crypto/scrypt.cpp
@@ -30,7 +30,7 @@
 
 /*! KDF to scrypt the \p input.
  */
-static int scryptDerive(const char *input, size_t inputLength, std::array<unsigned char, 64> &out, int N, int r, int p, const char *salt)
+static int scryptDerive(const char *input, size_t inputLength, std::array<unsigned char, 64> &out, int N, int r, int p, const unsigned char *salt, size_t saltlen)
 {
 #ifdef Q_OS_WIN
     QLibrary libCrypto(QLatin1String("libcrypto-1_1.dll"));
@@ -86,7 +86,7 @@ static int scryptDerive(const char *input, size_t inputLength, std::array<unsign
         result = -2;
     }
 
-    if (result == 0 && EVP_PKEY_CTX_set1_scrypt_salt(pctx, salt, strlen(salt)) <= 0)
+    if (result == 0 && EVP_PKEY_CTX_set1_scrypt_salt(pctx, salt, saltlen) <= 0)
     {
         result = -3;
     }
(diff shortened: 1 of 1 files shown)


manup:
Interesting looks like other OpenSSL versions have a slightly different signature for the function with `void *` for salt.

Monofin:
Yep, its a bit of an oddity - to my mind void * would be the preferred function signature, as it places no assumptions about the type or alignment of the passed in data (and the API doesn't care really...).

PR #6402, opened 2022-10-08
Remove rescaling of consumption in Develco/friend ZHEMI101 DDF

Qalculus:
This commit removes the division by 1000 when extracting the consumption value for the Develco/friend external meter interface ZHEMI101. This sensor comes in different flavors that are compatible with different standards. For example the EU version is compatible with the EU standard of 1000 pulses per kWh, so the value is already correct from the sensor.

### DIFF
--- a/devices/develco/zhemi101_external_meter_interface.json
+++ b/devices/develco/zhemi101_external_meter_interface.json
@@ -144,7 +144,7 @@
             "at": "0x0000",
             "cl": "0x0702",
             "ep": 2,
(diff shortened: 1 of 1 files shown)


rhgiter:
any news on this    - when do we see a new update ?

SwoopX (approved):

Qalculus:
This is my first PR and I am a bit unsure of the workflow. Do I need to mark it as done or something? 😊

PR #6413, opened 2022-10-11
Update smt402ad DDF to reflect correct endpoint

BabaIsYou:
Changing $TYPE_HUMIDITY_SENSOR uuid to respect the fact that it's on cluster 25 (0x19) and not 1 (0x01)


manup:
The 0x19 here is the endpoint I think :) but 0x19 looks correct according to the screenshot in the issue.

PR #6431, opened 2022-10-18
Add DDF for OTH4000-ZB thermostat

BabaIsYou:
Add support for Ouellet  OTH4000-ZB thermostat, based on Sinope TH1124ZB


BabaIsYou:
Still have to do some work on it … like bindings and perhaps merge it with th1124ZB DDF

PR #6482, opened 2022-10-29
Fix flooding the network with ZGP broadcasts

manup:
In a network with multiple ZGP devices (FoH switch,..) the ZGP proxy keep alive broadcast could have been send in the while loop without taking a break after each broadcast.

The fix adds a minimum delay of one second (idle timer delay) between each broadcast.


ebaauw:
Brought the changes over (all two lines ;-), compiled them and now running in my large production network.  It's the only one (currently) with ZGP switches.

ebaauw:
Looking good, checking the deCONZ log: only one ZGP broadcast per ~8 minutes instead of a dozen at the same time, and no more 0xD2 errors.  As far as I can tell, my ZGP switches still work.

PR #5260, opened 2021-08-28
Add support for innr SP 234 smart plug

SwoopX:
See #5256.

PR #6513, opened 2022-11-08
Add Innr SP-234 based on sp120

### LINKED ISSUE #5256: Innr SP 234
## Device

- Product name: Innr SP-234 (US only)
- Manufacturer: Innr
- Model identifier: SP 234
  - Switch
  - Other:  energy monitoring

### Basic

image

### Simple Metering
image

### Other clusters that are not mentioned above

Electrical measurement
image

xconverge:
Fixes #5256 

DDF for Innr SP-234 

This supercedes #5260 which can be closed

### DIFF
--- a/devices/innr/sp_234.json
+++ b/devices/innr/sp_234.json
@@ -1,0 +1,232 @@
+{
+  "schema": "devcap1.schema.json",
+  "manufacturername": "innr",
+  "modelid": "SP 234",
+  "vendor": "innr",
+  "product": "SP 234 smart plug",
+  "sleeper": false,
+  "status": "Gold",
+  "path": "/devices/innr/sp_234.json",
+  "subdevices": [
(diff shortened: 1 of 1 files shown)


manup:
>This supercedes https://github.com/dresden-elektronik/deconz-rest-plugin/pull/5260 which can be closed

@SwoopX looks allright? :)

SwoopX:
> > This supercedes #5260 which can be closed
> 
> @SwoopX looks allright? :)

Jep, looking good to me 👍

xconverge:
Also worth noting/putting it here that the ota file is floating around and available if anyone has any old devices. I was able to update with deconz otau plugin. I don't see anywhere these ota files are hosted so it would need to be sourced separately. The latest FW should be 3.1. The old FW for my devices was 2.8

PR #6518, opened 2022-11-09
Replace Qt Javascript engine with Duktape JS

manup:
Currently the Javascript in DDF files is handled by the Qt JS engine. The variety of Qt versions used behind deCONZ installations has shown different support for Javascript features but also problems with growing memory usage. The garbage collector is called quite often but doesn't seem to fix the problem, and is rather slow on light devices like the Raspberry Pi.

Duktape JS is a smaller engine suitable for embedded devices (could even run on the ConBee itself), it doesn't fully support newer features above ES5 but is more than enough for our purposes.

**Advantages** (over Qt JS engine)
* Tiny memory footprint with less than 220 KB of RAM usage, even on the Raspberry Pi 3 it fits into L2 cache
* No dynamic memory allocations, it just one continuous memory block handled by an arena allocator
* Constant memory usage, tests have shown we got rid of growing memory
* Garbage collection is done in under 1 millisecond, due a simple memory trick
* Statically compiled together with the plugin, so the same version is used across all installations
* For the usual JS code we have, the processing time is around 2-10 ms top
* Later on we can cache compiled JS bytecode and just reuse it without recompiling on each execution, which safes a few milliseconds

**Disadvantages**
* Missing newer JS features, here arrow functions would be nice to have:
   `(a) => a + 1;` vs. `function(a) { return a + 1; }`
* No `let` keyword, the PR adds a workaround to safely replace `let` by `var`

The PR has testing code which successfully compiles Javascript during loading from all DDFs (and also to apply the `let` workaround). If a JS expression or script can't be compiled it will be printed out during initial load. All DDF features which were in Qt JS version are ported to Duktape JS.

The Qt JS code is still present and the engine can be selected during compilation via C defines in `de_web.pro` QMake file:
```
# select Javascript engine
#DEFINES += USE_QT_JS_ENGINE
DEFINES += USE_DUKTAPE_JS_ENGINE
```


Mimiix:
Question: Would this require different dependencies to be installed ?

SwoopX:
No. I already run it in my production since a few days and it was compiled on a different machine.

manup:
Jop the engine sources are included within this PR (duktape.c, etc.).

Mimiix:
The reason i ask is for the home assistant addon.

PR #6531, opened 2022-11-14
Zusatzgerät für Lidl Silvercrest Gartenspieß mit 2 Steckdosen HG06620 

d-halle:
Added device for Lidl Silvercrest Garden Spike with 2 Sockets HG06620


d-halle:
The new power plug garden spike HG06620 from Lidl Silvercrest works with deconz as expected, just in the phoscon UI the manufacturer's name is not correctly recognized/displayed - it shows just the manufacturer's code - as shown in the screenshot.
Therefore I added the zigbee device's model id, the manufacturer's id, the full name of the manufacturer and the name of the model.
I ran all 4 workflow action checks in my fork with positive result.
grafik

PR #6565, opened 2022-11-22
Add DDF for Bosch Thermostat II

SwoopX:
So, the PR adds the respective DDF. Was able to identify the purpose of most of the relevant attributes (I guess). What currently isn't exposed is the "boost" Mode where the thermostat goes to full power. Also haven't played around with the groups cluster here.


SwoopX:
@ChrisHae I left the PR as draft intentionally, as there's quite some things to correct/update 🙂 

Regardless, update will come on weekend at the latest.

loefchen:
Any success on this? I think the install code problem still exists. Can't get the Bosch thermostat working, even with the last deCONZ update...

manup:
I think we can merge it for v2.21.0-beta after resolving merge conflict?

SwoopX:
Absolutely

besanur:
Is there any instruction how can we add the install code ?

jensflorian:
I second this. How do you pair the Bosch thermosat in phoscon and how to and the Install code?

chrissiboy:
I installed the version 2.21.01 and tried to pair the Bosch thermostat 2 with the deConz Conbee II but it failed. I get only two unknown new nodes in the map but the the phoscon app does not recognize a new sensor. The thermostat is alwas searching for a gateway and is blinking orange.

PR #6572, opened 2022-11-25
Fix `type` "Unknown" for `lights` resource for coordinator

ebaauw:
Don't restore the `type` value of a `lights` resource from the database, as it's already set when (re-) creating the resource on startup.  See #6552.  Instead the value is derived from the _Device Type_ on the _Simple Descriptor_, on evert startup, when the resource is (re-) created.

This should also prevent the need to remove and repair a light after we've patched the `type` in `setHaEndpoint()`.

Caveat: This might impact stale `type` values on `lights` resources for other devices that have been lingering in the database.  These resource will also show the value derived from the _Device Type_.  Also, it will no longer be possible to change the type by patching the database (`ritems` column of the `nodes` table, that holds a JSON will all the resource items).


manup:
Note for DDF based devices the type is taken from the DDF. This also allows to fix wrong types made up by legacy code by using a DDF. In future I'd like to remove the `setHaEndpoint()` function and its guess works in favor of the data available in the DDF.

https://github.com/dresden-elektronik/deconz-rest-plugin/blob/c6edb924934aa680fc18d941387dfcaa380451fa/device_ddf_init.cpp#L279-L287

ebaauw:
Unfortunately, you cannot (yet?) attach a DDF to the coordinator device.

manup:
The caveats make me a bit nervous, perhaps we should limit this to the coordinator for now?

```cpp
if (item->descriptor().suffix == RAttrType)
{
    // workaround for "Unknown" type of coordinator
    // https://github.com/dresden-elektronik/deconz-rest-plugin/pull/6572
    if (manufacturerCode() == 0x1135 && address().nwk() == 0x0000)
    {
        // type is set in setHaEndpoint()
        continue;
    }
}
```

ebaauw:
Or simply check for `"Unknown"`?

manup:
The "Unknown" check should do the trick 👍

ebaauw:
> The "Unknown" check should do the trick

Done.

PR #6577, opened 2022-11-28
DDF for Moes/Aubess temperature/humidity sensor with LCD (_TZ3000_ywagc4rj)

Finne75:
DDF for [this ](https://www.aliexpress.com/item/1005004506264679.html)device. It is advertised as Aubess, but the package names Moes as the vendor (?). In any case, it is another TS0201 _TZ3000 clone (_TZ3000_ywagc4rj), but with the twist that humidity value needs to be multiplied by 10 to get correct reports.


BabaIsYou:
Hi,

May be you could modify the ptath line according to the real name of the file, or delete it completly ?

PR #6599, opened 2022-12-07
Fix extra config/battery for ts0202 devices

Monofin:
Extra config/battery in the DDF was causing double battery level reporting, confusing deconz-homebridge.
The devices implement the power cluster and level reporting correctly meaning config/battery is not required.

Tested as working on current TS0202 devices.


BabaIsYou:
May be deconz-homebridge don’t need it but what will be the consequences on other  REST-API based interfaces that could use this attributes to create according sensor in there interface with third party apps ?

Monofin:
It’s not a field populated by the device, and leaving it there causes incorrect information to be exposed at the REST API.
I’m not sure how that would help other interfaces function.
You could potentially add a bind/parse to the DDF to populate this field, but that could also result in inconsistent data.

The correct source of battery data as I understand it is the power cluster, automatically handled by deCONZ and presented as an associated device at the interface.

BabaIsYou:
Well. That beeing said, perhaps making the attr exposing a updated value could do the trick and maid the device exposure consistent with all that kind of already existing type.

Something like could do the trick ? ;-)
{
"name": "config/battery",
"parse": {"fn": "zcl", "ep": 1, "cl": "0x0001", "at": "0x0021", "eval": "Item.val = Attr.val / 2"},
"read": {"fn": "zcl", "ep": 1, "cl": "0x0001", "at": "0x0021"},
"refresh.interval": 3600
},

manup:
> Well. That beeing said, perhaps making the attr exposing a updated value could do the trick and make the device exposure consistent with all that kind of already existing sensor types. Something like could do the trick ? ;-) { "name": "config/battery", "parse": {"fn": "zcl", "ep": 1, "cl": "0x0001", "at": "0x0021", "eval": "Item.val = Attr.val / 2"}, "read": {"fn": "zcl", "ep": 1, "cl": "0x0001", "at": "0x0021"}, "refresh.interval": 3600 },

Indeed if the device has the Power Configuration cluster and exposes this attribute this is the correct way.

SwoopX (changes requested):
Based on the PR comments, `config/battery` should be amended accordingly. If that is confirmed to be working alright, it is probably better to remove `state/lowbattery`.

manup:
Any news here for the comments?

BabaIsYou:
> Indeed if the device has the Power Configuration cluster and exposes this attribute this is the correct way.

If I believe the binding part of the DDF it does. But I didn’t get one of these devices to test.

Monofin:
Hi there: I can make the change to the attribute read and parse for the battery cluster, but I'd like to understand why that is the 'correct' way:

IIRC - The cluster binding requests the device to report changes to values in the cluster, and this applies to the battery cluster as well.
The battery cluster is handled by default by deCONZ's power cluster handling, and bound to the appropriate attribute, and then to the device as a whole, and reported as part of the overall device.

If there is a correct bind on that battery cluster, then surely explicit scheduled read/parse is not required.
In fact, this is also a sleepy device - the device will wake on schedule and report its battery level, and be processed appropriately.

IIRC, if the device is flagged as sleepy, it won't be considered as awake until either motion is detected, or the battery report is received. If it is a battery report, this then has the possibility of then triggering a second read of the battery cluster as the device is now considered awake.

Why would one need the secondary mechanism (i.e. the proposed read of the cluster) for battery status when the ts202 series of devices all (so far) seem to correctly respond with battery status with a binding, on schedule?

SwoopX:
@Monofin I'm afraid I don't get what you're trying to do here as it does the exact opposite from what I understand you want to have. You remove any battery percentage from the REST API.

manup:
@Monofin I guess this needs a bit of work based on Swoopx comment.

Monofin:
> @Monofin I'm afraid I don't get what you're trying to do here as it does the exact opposite from what I understand you want to have. You remove any battery percentage from the REST API.

That's exactly what I want to do: Because the battery cluster for the device seems to be being handled elsewhere in deCONZ.

Without any patch I get two instances of battery percentage reported at the REST API, one of which is a fixed incorrect value (due to the lack of the read).

I have tested that the patch you suggested leaves me with two battery battery percentages that some of the time do not actually match up, resulting in a (slightly improved) situation similar to the original bug.

With the patch originally submitted, I get fully working battery percentage presented at the REST API as the Battery for the accessory.

Somewhere in this, there is a correct behaviour; Somewhere, there is possibly a bug: I'll pull the REST API output an place it here if if might help the understanding.

BabaIsYou:
The point of all this is to ensure that when a device is supported by DFF it is fully supported so that it no longer depends on legacy code. It is therefore preferable to find the global solution through a DDF rather than remaining in a hybrid mode.

So it would be interesting to continue looking for a solution through the DDF, and to already understand why while the device is covered by a DDF, it seems that DeConz continues to manage the battery "independently"?

Monofin:
@BabaIsYou - Thank you for the explanation, that's much appreciated:

To clarify my understanding: What I think you are saying is that the behaviour that 'works' for me is a result of a hybrid-mode where the battery cluster is managed by legacy code, but the presence cluster is managed through the DDF. It's then presented as a single device with all functionality through the REST interface.


If that is the case, I think perhaps the complete DDF solution would sensibly bind to the battery cluster as well as the presence cluster. (I don't think it would use timed reads separately).

I'll do some investigation to see if I can get that configuration to behave correctly - I've tried before and ended up in the same place...
Will report back!

SwoopX:
@Monofin I fear we're jumping back and forth in arguing here 🙂 

> Without any patch I get two instances of battery percentage reported at the REST API, one of which is a fixed incorrect value (due to the lack of the read).

I do not believe this is possible (at least, it shouldn't), please provide us a proof from the REST API. Btw, doing a read on an attribute either once or regularly is pretty easy. However, I acknowledge were talking about Tuya here, so any surprise can possibly happen (or something serious has been hacked together in the past).

> What I think you are saying is that the behaviour that 'works' for me is a result of a hybrid-mode where the battery cluster is managed by legacy code, but the presence cluster is managed through the DDF.

In fact, it's the other way around. The power configuration cluster (battery) is managed solely by DDF and the presence related part by legacy code. Of course, the goal is to have that all covered by DDF related code only, but it's quite delicate, as the legacy code is sensitive. A pending PR of mine would be one piece of that puzzle, which also needs to be aligned. Another piece would require to come up with an alternative approach to temporarily store data for later processing; not as easy in that construct as one would think.

Monofin:
Okay, after much research I *think* I'm looking at a bug here?
So, after having a hack on the command line I have shown that at the REST interface, if I use the suggested configuration, there are two devices generated - one ZHABattery, bound to the 0001 cluster and one ZHAPresence associated with the 0005 cluster.

Removing the presence sensor from the interface also removes the battery, but re-adding it re-creates the same battery cluster, incorrectly, with the old device ID (I'd expect it to be created at the same time as the presence sensor is).

So every time I make a change to the DDF, remove the device and re-add it, it's somehow regaining the ZHABattery (which isn't in the DDF).... and therefore double reporting the charge level.

120638-    "81": {
120650-      "config": {
120668-        "on": true,
120688-        "reachable": true
120714-      },
120723-      "etag": "615e97a24ed36cf2cb19c89b4f31aec4",
120773-      "lastannounced": null,
120802-      "lastseen": "2023-02-24T13:10Z",
120841-      "manufacturername": "_TZ3000_otvn3lne",
120887-      "modelid": "TS0202",
120914-      "name": "Dining Room Motion",
120950-      "state": {
120967-        "battery": 100,
120991-        "lastupdated": "2022-09-29T11:56:15.779"
121040-      },
121049-      "type": "ZHABattery",
121077:      "uniqueid": "a4:c1:38:7f:7c:a3:ee:b1-ff-0001"
121129-    },

        "91": {
            "config": {
                "battery": 91,
                "duration": null,
                "on": true,
                "pending": [],
                "reachable": true
            },
            "ep": 1,
            "etag": "665a0b46c027dfb1720f4169bb71036d",
            "lastannounced": null,
            "lastseen": "2023-02-24T09:59Z",
            "manufacturername": "_TZ3000_otvn3lne",
            "modelid": "TS0202",
            "name": "Presence 91",
            "state": {
                "lastupdated": "2023-02-24T10:00:10.313",
                "presence": false
            },
            "swversion": "1.0.6",
            "type": "ZHAPresence",
            "uniqueid": "a4:c1:38:7f:7c:a3:ee:b1-01-0500"
        },

BabaIsYou:
> "a4:c1:38:7f:7c:a3:ee:b1-ff-0001

It seems you clearly fell on the [already referenced ](https://forum.phoscon.de/t/after-ddf-modification-got-duplicate-clusters-with-0xff-endpoint/2605)endpoint 0xff bug

Monofin:
I'll close this PR with a reference to the deconz issue for the 0xff endpoint bug?
(I have spotted an Aquara combo sensor or three in my network that have the same issues thanks to that reference)

BabaIsYou:
> If I delete the battery (in this case 81), then the other device (91) is also deleted for some reason.

Because there is actullay no more a way to juste delete a sensor. The delete act on a node/device, even with the REST-API delete sensor option.

>So every time I make a change to the DDF, remove the device and re-add it, it's somehow regaining the ZHABattery (which isn't in the DDF).... and therefore double reporting the charge level.

Did you try to restart DeCONZ between the moment you delete de device and the moment you adding it again ?

Monofin:
So far I haven't seen a referenced issue or pull request for fixes in this area, only the discussion on the phoscon forums - as (as far as I know!) this remains the only issue/PR coupling the two at the moment, I'm reluctant to close it until there's a descriptive Issue/PR in this codebase?
Am I missing something here?

PR #6623, opened 2022-12-21
Update and rename _TZ3000_d16y6col_water_sensor.json to _TZ3000_water…

flopp999:
…_sensor.json

added two manufactures

### DIFF
--- a/devices/tuya/_TZ3000_water_sensor.json
+++ b/devices/tuya/_TZ3000_water_sensor.json
@@ -1,7 +1,15 @@
 {
   "schema": "devcap1.schema.json",
-  "manufacturername": "_TZ3000_d16y6col",
-  "modelid": "TS0207",
+  "manufacturername": [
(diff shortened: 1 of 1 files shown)


SwoopX (approved):

SwoopX:
@flopp999 @manup Just wondering if it wouldn't make more sense to name the DDF file like `ts0207_water_sensor.json`, was the manufacturer is more likely to change...

flopp999:
> @flopp999 @manup Just wondering if it wouldn't make more sense to name the DDF file like `ts0207_water_sensor.json`, was the manufacturer is more likely to change...

flopp999:
@SwoopX yes, TS0207 which is the model I'd is much better to use.

manup:
Good question while the modelid often if the same for the Tuya variants of the same device, there are also some where even the modelid changes. But it seems that modelid is a lesser moving target than the manufacturer name.

PR #6655, opened 2023-01-08
Update and rename DDF for SONOFF SNZB-02 multi sensor

Finne75:


BabaIsYou:
Hi,
As _tuya_swversion.js_ does not exist in _/devices/sonoff_ are you sure it will work ?
Personnaly I use : 
```

         "parse": {
            "at": "0x0001",
            "cl": "0x0000",
            "ep": 255,
            "eval": "let v = Attr.val;\nItem.val = String((v & 192) >> 6) + '.' + String((v & 48) >> 4) + '.' + String(v & 15);",
            "fn": "zcl"
          }
```

Finne75:
Uh, good catch. Copy-paste error. The device does not expose SW build ID. Should we use "date code" instead?
image

Smanar:
If you prefer use the Date code

```
        {
          "name": "attr/swversion",
          "refresh.interval": 84000,
          "read": {"cl": "0x0000", "at": "0x0006", "ep": 1, "fn": "zcl"},
          "parse": {"cl": "0x0000", "at": "0x0006", "ep": 1, "eval": "Item.val = Attr.val", "fn": "zcl"}
        },
```

BabaIsYou:
> Should we use "date code" instead

Could do the job too. I use it for another device like this : 

```
       {
          "name": "attr/swversion",
          "refresh.interval": 86400,
          "read": {
            "at": "0x0006",
            "cl": "0x0000",
            "ep": 0,
            "fn": "zcl"
          },
          "parse": {
            "at": "0x0006",
            "cl": "0x0000",
            "ep": 255,
            "eval": "Item.val = Attr.val",
            "fn": "zcl"
          }
        },
```

Finne75:
Done (and tested)

PR #6679, opened 2023-01-17
Update ts0202_presence_sensor to support _TZ3000_6ygjfyll

djhworld:
Received device off Aliexpress that is exposing `_TZ3000_6ygjfyll` instead of `_TZ3040_6ygjfyll`

Adding this extra entry fixes the issue.


manup:
Hi almost there, the `modelid` array needs to be also extended by one item "TS0202".

djhworld:
Fixed!

PR #6702, opened 2023-01-27
Negative value for zclFrame.payload().at() on x86

ebaauw:
Use intermediate `quint8` variable for value for `zclFrame.payload().at()` to force an unsigned value, see #6670.

Found and fixed same issue for Qt Javascript engine as well.

Unfortunately, I cannot test as I don't have an x64 system running deCONZ.


manup:
Interesting case, I'm not 100% sure if the conversion in the change works. As alternative an explicit cast does the trick, I've used that in some other places. E.g. `(uint8_t)zf->payload().at(i)`

ebaauw:
Cool.  I tried to find the other places, but failed.  Updated.

PR #6708, opened 2023-01-28
Update ZY-M100 DDF (fix & adding precision - breaking change for triggerdistance)

BabaIsYou:
Adding precision on Far detection sensitivity configuration and fixing conversion on errorcode reporting (see #6517 )


manup:
This looks like a breaking change, I think the item values should use some generic approach which works for all sensors of this kind?

BabaIsYou:
A breaking change tested by people that asked for integration of this device less than a month ago seems not a so big deal 🤣… but you’re right and the better way to have such a generic approach would be to create a new attribute for that purpose from my point of vue.

PR #6718, opened 2023-02-01
Update DDF for battery bindings for Aubess/Tuya temperature and humidity sensors

Finne75:
Set longer read intervals, inspired by [this comment](https://github.com/dresden-elektronik/deconz-rest-plugin/issues/6697#issuecomment-1405171226) in #6697. Remove deprecated file for Aubess multisensor TZ3000.


BabaIsYou:
Hi @Finne75, what do you think about changing  "swversion"  "refresh.interval" in  snzb-02 DDF accordingly to 86400 like other DDF in this list ? ;-)

Finne75:
> Hi @Finne75, what do you think about changing "swversion" "refresh.interval" in snzb-02 DDF accordingly to 86400 like other DDF in this list ? ;-)

Sounds like a great idea :)

BabaIsYou:
Changes for SNBZ-02 are already included in PR #6699 that add support for another device. Then I suggest you to drop this file off this PR. Sorry about that

SwoopX:
I'd prefer not to make those changes to battery reporting, as this results in just a single report per 24h. Then you got other people complaining that the devices do not work properly anymore and they want to have more reports...

Finne75:
> Changes for SNBZ-02 are already included in PR #6699 that add support for another device. Then I suggest you to drop this file off this PR. Sorry about that

No problem. How do I do this, should I make a commit with the original version before your PR?

Finne75:
> > I'd prefer not to make those changes to battery reporting, as this results in just a single report per 24h.
> 
> ```
>           "at": "0x0021",
>           "dt": "0x20",
>           "min": 3600,
>           "max": 84600,
>           "change": "0x00000002"
> ```
> 
> means for me minimum 1 hour, maximum 1 day, or any change of 2 degree ... Am I wrong ?

How about maximum half (42300) a day as a compromise?

BabaIsYou:
> How about maximum half (42300) a day as a compromise?

With a `"refresh.interval": 3600,` on the battery resource I don't see why we should change this binding.

PR #6720, opened 2023-02-01
DDF for Loratap ZigBee 3.0 Wireless 4-Button Remote Control

sinus61:
Problem from this issue with fake lamps is solved with latest updates, so this is working now
https://github.com/dresden-elektronik/deconz-rest-plugin/issues/6156

https://www.loratap.com/products/ss6400zb-v2
https://www.loratap.com/products/ss9400zb-ya


BabaIsYou:
Hi.
is there a reason why you didn't add it in the already existing _TZ3000_ygvf9xzp_4gang_remote.json file that is already in use for this device ?

sinus61:
Was not aware of this file, but yes this could go there. Maybe better name that TS0044_4gang_remote.json, so that would be easier to recognize.

sinus61:
Have to say it's not working that good. There is much battery drain with this devices. Both Loratap devices _TZ3000_ufhtxr59 and _TZ3000_b7bxojrg send their battery state about every 3 minutes via the API. After a few hours or a day the battery is empty. Same goes also for the _TZ3000_abci1hiu that was already in the button_maps.json, but not in the ddf. Same also for the 2 gang _TZ3000_t8hzpgnd.

Otherwise the _TZ3000_dziaict4 works fine. It pushes the battery state once a day or if a button is pressed via the API and works a few weeks here now. Just what I expect from the DDF. Only the battery value is 0%, 

But I think you can't believe the values anyway with this tuya stuff. If I put a CR2032 in the _TZ3000_b7bxojrg it says always 95% and is empty later. Have added a external 3V power source now to the device and it says still 95%.

Looking aroud there are some discussions about this, but no solution for now. Some say it's working better with z2m, others say not. But it seems to work fine with the tuya gateways.

PR #6729, opened 2023-02-07
DDF for LIDL Smart watering timer (_TZE200_htnnfasr)

ChrisHae:
- On/Off and swversion are working
- make battery, frostlock and valve open duration available via RestAPI in separate sensor object (wrong values / no impact on actual device)

### DIFF
--- a/devices/tuya/_TZE200_htnnfasr_water_valve
+++ b/devices/tuya/_TZE200_htnnfasr_water_valve
@@ -5,7 +5,7 @@
   "vendor": "LIDL",
   "product": "Smart watering timer",
   "sleeper": false,
-  "status": "Bronze",
+  "status": "Silver",
   "subdevices": [
     {
       "type": "$TYPE_ON_OFF_OUTPUT",
@@ -34,69 +34,13 @@
           "name": "attr/name"
         },
         {
-          "name": "attr/swversion",
-          "refresh.interval": 84000,
-          "read": {
-            "at": "0x0001",
-            "cl": "0x0000",
-            "ep": 1,
-            "fn": "zcl"
-          },
-          "parse": {
-            "at": "0x0001",
-            "cl": "0x0000",
-            "ep": 1,
-            "fn": "zcl",
-            "script": "tuya_swversion.js"
-          }
+          "name": "attr/swversion"
         },
         {
           "name": "attr/type"
         },
         {
           "name": "attr/uniqueid"
-        },
-        {
-          "name": "config/battery",
-          "parse": {
-            "dpid": 11,
-            "eval": "Item.val = Attr.val * 50;",
-            "fn": "tuya"
-          },
-          "default": 0
-        },
-        {
-          "name": "config/duration",
-          "parse": {
-            "dpid": 5,
-            "eval": "Item.val = Attr.val;",
-            "fn": "tuya"
-          }
-        },
-        {
-          "name": "config/lock",
-          "parse": {
-            "dpid": 109,
-            "eval": "Item.val = Attr.val;",
-            "fn": "tuya"
-          }
-        },
-        {
-          "name": "config/mode",
-          "parse": {
-            "dpid": 108,
-            "eval": "Item.val = Attr.val;",
-            "fn": "tuya"
-          }
-        },
-        {
-          "name": "state/deviceruntime",
-          "parse": {
-            "dpid": 6,
-            "eval": "Item.val = Attr.val;",
-            "fn": "tuya"
-          },
-          "default": 0
         },
         {
           "name": "state/on",
@@ -118,6 +62,88 @@
         },
         {
           "name": "state/reachable"
+        }
+      ]
+    },
+    {
+      "type": "$TYPE_OPEN_CLOSE_SENSOR",
+      "restapi": "/sensors",
+      "uuid": [
+        "$address.ext",
+        "0x01",
+        "0x0500"
+      ],
+      "items": [
+        {
+          "name": "attr/id"
+        },
+        {
+          "name": "attr/lastannounced"
+        },
+        {
+          "name": "attr/lastseen"
+        },
+        {
+          "name": "attr/manufacturername"
+        },
+        {
+          "name": "attr/modelid"
+        },
+        {
+          "name": "attr/name"
+        },
+        {
+          "name": "attr/swversion"
+        },
+        {
+          "name": "attr/type"
+        },
+        {
+          "name": "attr/uniqueid"
+        },
+		{
+          "name": "config/battery",
+          "parse": {
+            "dpid": 11,
+            "eval": "Item.val = Attr.val * 50;",
+            "fn": "tuya"
+          },
+          "default": 100
+        },
+        {
+          "name": "config/duration",
+          "write": {
+            "dpid": 5,
+            "dt": "0x02",
+            "eval": "Attr.val = Item.val;",
+            "fn": "tuya"
+          },
+          "parse": {
+            "dpid": 6,
+            "eval": "Item.val = Attr.val;",
+            "fn": "tuya"
+          }
+        },
+        {
+          "name": "config/locked",
+          "write": {
+            "dpid": 109,
+            "dt": "0x01",
+            "eval": "Item.val == 1 ? 1 : 0;",
+            "fn": "tuya"
+          },
+          "parse": {
+            "dpid": 108,
+            "eval": "Item.val == 1 ? 0 : 1;",
+            "fn": "tuya"
+          },
+          "default": false
+        },
+        {
+          "name": "config/reachable"
+        },
+        {
+          "name": "state/lastupdated"
         }
       ]
     }
--- a/devices/tuya/_TZE200_htnnfasr_water_valve
+++ b/devices/tuya/_TZE200_htnnfasr_water_valve
@@ -5,7 +5,7 @@
   "vendor": "LIDL",
   "product": "Smart watering timer",
   "sleeper": false,
-  "status": "Silver",
+  "status": "Gold",
   "subdevices": [
     {
       "type": "$TYPE_ON_OFF_OUTPUT",
@@ -58,7 +58,8 @@
             "dpid": 1,
             "eval": "Item.val = Attr.val;",
             "fn": "tuya"
-          }
+          },
+	  "default": 0
         },
         {
           "name": "state/reachable"
@@ -105,13 +106,14 @@
           "name": "config/battery",
           "parse": {
             "dpid": 11,
-            "eval": "Item.val = Attr.val * 50;",
+            "eval": "Item.val = Attr.val;",
             "fn": "tuya"
           },
           "default": 100
         },
         {
           "name": "config/duration",
+          "description": "The duration the valve stays open.",
           "write": {
             "dpid": 5,
             "dt": "0x2b",
@@ -119,13 +121,14 @@
             "fn": "tuya"
           },
           "parse": {
-            "dpid": 6,
+            "dpid": 5,
             "eval": "Item.val = Attr.val;",
             "fn": "tuya"
           }
         },
         {
           "name": "config/locked",
+          "description": "Read: frost lock status. Write: frost lock reset.",
           "write": {
             "dpid": 109,
             "dt": "0x10",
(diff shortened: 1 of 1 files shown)


BabaIsYou:
Looking to [Tuya specific cluster definition](https://developer.tuya.com/en/docs/iot/Zigbee-Water-valve-controller-access-standard?id=Kbahvojbqzvgl) could probably help, for instance this doc says that the DP for battery is 7 and not 11 that is used in the DDF.

ChrisHae:
> Looking to [Tuya specific cluster definition](https://developer.tuya.com/en/docs/iot/Zigbee-Water-valve-controller-access-standard?id=Kbahvojbqzvgl) could probably help, for instance this doc says that the DP for battery is 7 and not 11 that is used in the DDF.

I have testet it with DP7. There is no difference how deCONZ shows the battery symbol. And also no information in the Rest API.

ChrisHae:
parkside_battery

So the battery Icon is alway near empty also i just changed the batteries. DP11 or 7 makes no difference.

BabaIsYou:
Have you tried also to change the eval calculation ? On another device the fact to multiply the Attr.Val by 50 made an out of range error. Have you some log (just with INFO and DDF) to share ?

sinus61:
I think as it paired only as a light there will be no battery reported anyway. And it creates no additionally sensor.

Have tried and you can set it on, but it doesn't report the state. So it switches off after a minute, but reports always on.

ChrisHae:
I have this information in the INFO logs: `12:27:48:628 TY_DATA_REPORT: seq 130, dpid: 0x0B, type: 0x02, length: 4, val: 90`
So I think  this could be the battery. The battery is brand new. So what would be the right multiplier?

> I think as it paired only as a light there will be no battery reported anyway. And it creates no additionally sensor.
Have tried and you can set it on, but it doesn't report the state. So it switches off after a minute, but reports always on.

The "Main Device" I have used in the DDF is an On/Off output. Should I try another one?
And I have tested to send any vaule to DP 0x05. This should affect the on-time. The result was with any value the valve stayed on "forever" at least what I have observed.

sinus61:
I have played also a little bit with this, seems to be reporting the on off state is working now. Send also values to DP5 in the deconz GUI, has worked 1 or 2 times and then it seems the internal timer was removed, as you describe. Can now just send on and off from my automation like I want.

Maybe for battery there have to added a sensor like the plugs have for their reporting.

sinus61:
This adds a battery sensor for this device

`    {
      "type": "$TYPE_BATTERY_SENSOR",
      "restapi": "/sensors",
      "uuid": [
        "$address.ext",
        "0x01",
        "0x0001"
      ],
      "items": [
        {
          "name": "attr/id"
        },
        {
          "name": "attr/lastannounced"
        },
        {
          "name": "attr/lastseen"
        },
        {
        },
        {
          "name": "attr/modelid"
        },
        {
          "name": "attr/name"
        },
        {
          "name": "attr/swversion"
        },
        {
          "name": "attr/type"
        },
        {
          "name": "attr/uniqueid"
        },
        {
          "name": "config/battery",
          "parse": {
            "dpid": 11,
            "eval": "Item.val = Attr.val;",
            "fn": "tuya"
          },
          "default": 0
        },
        {
          "name": "config/on"
        },
        {
          "name": "config/reachable"
        },
        {
          "name": "state/battery",
          "refresh.interval": 3600
        },
        {
          "name": "state/lastupdated"
        }
      ]
    }`

ChrisHae:
@sinus61 Exec "Reset to Dactory Defaults" from the Basic Cluster resets the timer to 1 minute again.
Since you also have the device I'm curious: is it normal behaviour that you have to press the on/off button on the device 2 times to open the valve? 1st press the battery indicator led lights up. And 2nd press the valve opens.

sinus61:
Ok, thanks, that brings the timer to 1 minute again. I have that behaviour also, but I think that was not always the same. Maybe some setting for this? 
Reporting is now also not working anymore. Really hard to control device.

BabaIsYou:
> This adds a battery sensor for this device

Don't really see the need to create a new battery sensor, can't you get battery level reported thru REST-API with just modifying the initial DDF with 
```
    {
      "name": "config/battery",
      "parse": {
        "dpid": 11,
        "eval": "Item.val = Attr.val;",
        "fn": "tuya"
      },
      "default": 0
    },
```
 ?

sinus61:
I think it ist already in the DDF, but it is not reported through the Api for "lights".

BabaIsYou:
> I think it ist already in the DDF, but it is not reported through the Api for "lights".

Ah ok, I didn't understand that, sorry. Thx

sinus61:
Yes, I think that may be working, could be that all that stuff have to be moved to the sensor device then. Although the discussion in your link mix "time_left", which is dp6 and I think only readable with "duration", which is dp5 and only writeable.

But it is not really good working in deconz GUI for me at this time, mostly it removes the timer and I have reset the device. I have expected 2 minutes this way:
grafik

ChrisHae:
> May be I'm wrong again, but it seems also that "timer_duration" could me manageable with https://github.com/zigpy/zha-device-handlers/issues/1674 but don't see write part in the DDF for "duration" attribute.

The featureset and Datapoints I used in this DDF I have found on this github page: https://gist.github.com/rforro/688a885bfdb09452ca31ffd795590a42

So there is DP5 which is used for setting the timer and DP6 for reading the time left I think.
Couldn't get it working in RestAPI though.

ChrisHae:
> Yes, I think that may be working, could be that all that stuff have to be moved to the sensor device then. Although the discussion in your link mix "time_left", which is dp6 and I think only readable with "duration", which is dp5 and only writeable.
> 
> But it is not really good working in deconz GUI for me at this time, mostly it removes the timer and I have reset the device. I have expected 2 minutes this way: grafik

I think you have to use type "value" not raw And then I don't know the unit of the value. I thought it was seconds. in hex numbers 60 seconds would be 0x3c. And then you have to add the length parameter. 01 for 1 Byte 02 for 2 bytes and so on...

But I also didn't get any results I would expect.

BabaIsYou:
> 

This screen is for "data request", not to write value into attribute. To write a value the only way using DDF is to add a "write" part like something like this : 

       {
          "name": "state/deviceruntime",
          "read": {
            "fn": "none"
          },
          "parse": {
            "dpid": 6,
            "eval": "Item.val = Attr.val;",
            "fn": "tuya"
          },
          "write": {
            "dpid": 6,
            "dt": "0x2b",
            "eval": "Attr.val = Item.val;",
            "fn": "tuya"
          },
          "default": 0
        },
Not quite sure of the DataType (dt) BTW

ChrisHae:
A Zigbee Data Request is to send data to a device. That is what the deCONZ GUI does. A Data Request can contain a payload but it does not have to. And yes using the "write" command in a DDF is to make the Data Request available in Rest API.

The datatypes should be:
```
// 0x00 	DP_TYPE_RAW 	?
// 0x01 	DP_TYPE_BOOL 	?
// 0x02 	DP_TYPE_VALUE 	4 byte unsigned integer
// 0x03 	DP_TYPE_STRING 	variable length string
// 0x04 	DP_TYPE_ENUM 	1 byte enum
```

BabaIsYou:
> A Zigbee Data Request is to send data to a device. That is what the deCONZ GUI does. A Data Request can contain a payload but it does not have to. And yes using the "write" command in a DDF is to make the Data Request available via Rest API.
> 
> The datatypes should be:
> 
> ```
> // 0x00 	DP_TYPE_RAW 	?
> // 0x01 	DP_TYPE_BOOL 	?
> // 0x02 	DP_TYPE_VALUE 	4 byte unsigned integer
> // 0x03 	DP_TYPE_STRING 	variable length string
> // 0x04 	DP_TYPE_ENUM 	1 byte enum
> // 0x05 	DP_TYPE_FAULT 	1 byte bitmap
> ```

These dt are for the GUI, but it seems that for write using DDF that is [that table](https://github.com/dresden-elektronik/deconz-rest-plugin/blob/11153d9eddf803927d4d0da07e6ca1375ddd3395/device_access_fn.cpp#L842)

ChrisHae:
@BabaIsYou Thx for the hint. These indeed seem to be the correct datatypes.

sinus61:
Tried this and setting duration is working for me, so thanks for this.

Just I have a different understanding for dpid 5 and 6.
5 is a permanent setting for the duration time, while 6 is a reading for time left. It counts from duration to zero.

So, for me, I have set read and write for duration to dpid 5 and added this:

        {
          "name": "state/seconds_remaining",
          "parse": {
            "dpid": 6,
            "eval": "Item.val = Attr.val * 60;",
            "fn": "tuya"
          }
        }

ChrisHae:
@sinus61 

> Tried this and setting duration is working for me, so thanks for this.

I don't have the device anymore for testing. So can you confirm that setting the duration is possible and you can set the device to lets say keep it open for 2 minutes? 

What parameters exactly do you send to the device (dpid, datatype, length and value) And can you set it using RestAPI?

> So, for me, I have set read and write for duration to dpid 5 and added this:
```
{
      "name": "state/seconds_remaining",
      "parse": {
        "dpid": 6,
        "eval": "Item.val = Attr.val * 60;",
        "fn": "tuya"
      }
    }
```
> "default": 0
> to state/on

This seems to make sense to me. Can you confirm this is working? Then I would update the DDF.

sinus61:
I send to the API:
`{"config": {"duration": 2}}`

set the device on and get back for the sensor
`{
  "config": {
    "battery": 100,
    "duration": 2,
    "locked": false,
    "on": true,
    "reachable": true
  },
[...]
  "state": {
    "lastupdated": "2023-02-28T11:53:25.941",
    "seconds_remaining": 120
  },`

a minute later I get back
`"seconds_remaining": 60`

Not sure about the battery reading, I get 100% with your ddf, but had with just "Item.val = Attr.val;" before 90%

BabaIsYou:
Just a global comment about this PR : As some attributes have been reused from their original setting, even one is using different dpip, then combine 2 tuya attributes in one  (`config/locked`), I suggest to add some specific description in the DFF to keep this undestrandable for further users.

ChrisHae:
> Just a global comment about this PR : As some attributes have been reused from their original setting, even one is using different dpip, then combine 2 tuya attributes in one (`config/locked`), I suggest to add some specific description in the DFF to keep this undestrandable for further users.

This was just an idea that I had. Because technicaly both DPIDs are for the same attribute. E.g. 108 is for reading the frost lock status und 109 is for writing it (reset frost lock). I'm not sure if this even work with DDF. But adding comments to the DDF is a good idea.

With duration I think I will go the way @sinus61 suggested.

ChrisHae:
> Not sure about the battery reading, I get 100% with your ddf, but had with just "Item.val = Attr.val;" before 90%

I'm also not sure. The eval function * 50 I got from another DDF for another Tuya device from this github repository.

BabaIsYou:
> > Not sure about the battery reading, I get 100% with your ddf, but had with just "Item.val = Attr.val;" before 90%
> 
> I'm also not sure. The eval function * 50 I got from another DDF for another Tuya device from this github repository.

As previously suggested, from my point of view the `* 50` lead to an out of range error then reporting default value that is (strangely ?!) set to 100.

Then I suggest to let the value unmultiplicated (like in the last DDF version) and without default value because https://github.com/dresden-elektronik/deconz-rest-plugin/blob/master/devices/generic/items/config_battery_item.json as already one.

sinus61:
I think battery is right now. Only the frost guard is not reporting, but the device locks and you can reset.
I would let it go now.

sinus61:
My device is still in Phoscon, as a light chain, but you don't see battery there, it's in a extra sensor device not visible in Phoscon.
If you you want control the device you have to check it via the Rest-API, if it's in the deconzGUI ist should be also in the API.

ChrisHae:
The File is missing the extension.
Adding .json to the file and it should work.
The "fairy light" in Phoscon App should be fixed. Perhaps it was not merged into the newest App release.

FranzToepfer:
Good morning,

thanks for the guidance. You're right, adding the .json extension to the DDF file inside the docker container did the trick. After file name change, had to restart the docker container to make changes effective. Next attempt to search the watering switch succeeded, directly included the battery state 🙂

Great improvement for my personal smart home, thanks for your efforts to realize this. 

Kind regards

joeBezunta:
Hi guys,

I have one of these for quite some time, and until today I only had the switch on/off button, but thanks to you guys, now I can see the battery on an extra sensor (openclose_XX), but I can't access the other attributes, like duration , state, etc.

How can I see the remaining attributes?

Best regards

Issue #6743, opened 2023-02-11
Lonsonho 16A plug  _TZ3000_dksbtrzs


wvuyk:
## Device

- Product name: Lonsonho 16A plug
- Manufacturer:  _TZ3000_dksbtrzs
- Model identifier: TS011F
- Device type : Please remove all unrelated device types. 
  - Light (plug)
  - Sensor (power/consumption/voltage)
 
Light/Plug infor is in the JSON. The sensor information for power is missing.

## Screenshots

![image](https://user-images.githubusercontent.com/30700296/218278603-de8fb5f2-6d4b-4a7d-b17c-8c9f02c7ab95.png)

![image](https://user-images.githubusercontent.com/30700296/218278631-00b85f4f-8d30-4706-adb0-47ffb12dadbd.png)

Basic info part 1
![image](https://user-images.githubusercontent.com/30700296/218278680-052f442c-54b9-44f0-883c-398ef791cdcc.png)
Basic info part 2
![image](https://user-images.githubusercontent.com/30700296/218278776-a2daa29e-b287-4a56-b517-8911abd26e1d.png)

Identify
![image](https://user-images.githubusercontent.com/30700296/218278802-1328d412-14aa-4f6f-adc8-7664a231f13a.png)

Groups
![image](https://user-images.githubusercontent.com/30700296/218278827-7f187f7a-ccf3-4e44-824f-cff046b4be58.png)

Scenes
![image](https://user-images.githubusercontent.com/30700296/218278862-8864e77d-b6ba-4121-9108-c09e54200353.png)

On/Off
![image](https://user-images.githubusercontent.com/30700296/218278883-6219e899-8df6-4ce2-b4c1-0865e9f2c781.png)

Simple Metering 
![image](https://user-images.githubusercontent.com/30700296/218278932-0b4855c7-ff0c-4bc6-9832-c007d1e9710e.png)

Electrical Measurement part 1
![image](https://user-images.githubusercontent.com/30700296/218278993-84048790-0b55-45a6-8783-aeda2a50650f.png)

Electrical Measurement part 2
![image](https://user-images.githubusercontent.com/30700296/218279036-c09b96bc-3419-4ad1-84c0-f22f65aba6e4.png)

E000 Unknown and empty
E001 Unknown and empty

OTAU
![image](https://user-images.githubusercontent.com/30700296/218279100-2fd2ac59-4d07-46ce-b1d0-bc77a4f5b0d4.png)

[...]

PR #6745, opened 2023-02-12
Add  _TZ3000_dksbtrzs Smart plug support

BabaIsYou:
See #6743


retrography:
@manup, is it possible that the Overload properties for this plug are rw rather than r?

There is a common issue with these plugs that their overload protection activates with very minor surges, sometimes due to errors in estimation. Being able to write this property can help resolve the issue. Note that not all variants of this plug have the overload protection feature.

https://github.com/Smanar/deconz-rest-plugin/blob/b3b53e51011703ebeb8b521e235db14b957217ab/general.xml#L4075C1-L4076C1

PR #6770, opened 2023-03-02
Adjusting swversion to a whole day refresh

BabaIsYou:
Why 8400 sec instead of a whole day ?


manup:
Good catch, was very likely a typo, a day should be sufficient.

PR #6781, opened 2023-03-05
DDF for Blitzwolf BW-IS4 multisensor (_TYZB01_hjsgdkfl)

sinus61:
https://zigbee.blakadder.com/BlitzWolf_BW-IS4.html


BabaIsYou (changes requested):
Suggestions : Is there a reason why swversion referesh.interval should be different than the one in generic attributes (84000 instead of 86400) ?

Are descriptions necessary as they seems the same than in generic attributes ?

PR #6791, opened 2023-03-09
Enable usage of zigbee install codes

SwoopX:
As deconz was missing a native way to use zigbee install codes, this PR adds the respective capability. The code to derive the required hash is ported from the zigpy implementation.

It has been tested with the reference codes from zigbee smart energy standard and also with some device codes on Ubuntu 20.04 LTS. Also, A Bosch Thermostat II has been successfully paired.

It is important to include the CRC for the correct calculation, so all characters of the code are required.

grafik

grafik

### DIFF
--- a/de_web.pro
+++ b/de_web.pro
@@ -64,7 +64,7 @@ macx {
 unix:LIBS +=  -L../.. -ldeCONZ
 
 unix:!macx {
-    LIBS += -lcrypt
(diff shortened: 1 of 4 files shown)


SwoopX:
> Awesome, will try this out after the stable release. I have a few small change requests in mind for loading of the library (to use dynamic linking) but overall the PR looks good.

Sure, just bring it on.

Issue #6775, opened 2023-03-03
SYMFONISK sound remote gen2


LiamTheCat:
## Device
- Product name: SYMFONISK gen2
- Manufacturer: IKEA of Sweden
- Model identifier: SYMFONISK sound remote gen2
- Device type : 
  - Switch

## Screenshots
![image](https://user-images.githubusercontent.com/126875522/222711210-efa9e392-1882-48a1-82fb-167922b5d7dd.png)

### Basic
![image](https://user-images.githubusercontent.com/126875522/222710249-0a98c2c2-019d-4d90-baf0-766032c8621e.png)

### Power Configuration
![image](https://user-images.githubusercontent.com/126875522/222710439-25bfbb83-f18a-486c-af6d-f7ffd64f3713.png)

### Identify
![image](https://user-images.githubusercontent.com/126875522/222710344-ba899d05-bda9-471e-b3d5-123abd575530.png)

### Poll Control
![image](https://user-images.githubusercontent.com/126875522/222710514-fd7c91d9-dd93-4e31-8d7b-52f609e00baf.png)

### ZLL Commissioning
![image](https://user-images.githubusercontent.com/126875522/222710594-867ce9ba-561a-499d-8f78-ecbaec2fc8e6.png)

### Identify
![image](https://user-images.githubusercontent.com/126875522/222710708-fc41ab6c-e5a3-4ec8-afe6-607ab47fbdc3.png)

### Group
![image](https://user-images.githubusercontent.com/126875522/222710764-c4336e7f-ebc5-47f7-91ea-3767a37aab52.png)

### On/Off 
![image](https://user-images.githubusercontent.com/126875522/222711371-8b4efdd7-3ae3-4bdf-859d-26829f4ccfb8.png)

### Level Control
![image](https://user-images.githubusercontent.com/126875522/222711447-7b218a50-6e83-411d-95d6-8504c4407bc7.png)

### OTAU
![image](https://user-images.githubusercontent.com/126875522/222710942-1164fdc3-842a-42e3-8d3e-3c3f67ad787b.png)

### ZLL Commissioning
![image](https://user-images.githubusercontent.com/126875522/222711064-addfdea6-fcab-4cf2-bb07-362a5d86dc12.png)

Mimiix:
Hi,

Looks like the values are not populated in the screenshots. can you please click "read" on all cluster screenshots and make sure the values are populated?

Kind regards,

[...]

PR #6803, opened 2023-03-15
IKEA SYMFONISK sound remote gen2

ebaauw:
Add support for IKEA SYMFONISK sound remote gen2,  see #6775.


ebaauw:
@manup, I really don't understand that I need to update the schema to add a new cluster and corresponding commands, especially as listing the hex values works perfectly fine.

manup:
> @manup, I really don't understand that I need to update the schema to add a new cluster and corresponding commands, especially as listing the hex values works perfectly fine.

It might not be the best solution, but it's there to catch errors with the related GitHub action which verifies the button_maps.json against the JSON schema file.

ebaauw:
Damn, the `button_map.json` from that PR is messed up, breaking both the gen-1 and gen-2 SYMFONISK controller.  My bad.

PR #6804, opened 2023-03-15
Fix DDF for IKEA cws light

ebaauw:
Bindings need to be defined at top level.

Thanks to @Zehir 's DDF validator for reporting this!  See Discord.


manup:
Thanks, that's a further place which should be checked by the GitHub actions in future.

PR #6813, opened 2023-03-20
Fix ubisys J1 auto calibration

SwoopX:
While the legacy code was able to check every couple of seconds if the cover is still moving, this code doesn't work anymore while using DDFs. As a quick workaround, it is assumed full open/close doesn't last longer than 30 secs.

With the amended time spans, auto calibration has been successfully tested. All in all not pretty, but easily implemented and functional.


manup:
The delays need a bit of patience, but I'd say it looks fine for now to include :)

PR #6825, opened 2023-03-24
Remove unused _TZE200_htnnfasr_water_valve file

BabaIsYou:
Suppress file without extension, useless since #6802


manup:
Not the first time we had the problem of forgetting the file extension. Would be nice to catch these via GitHub Actions in future.

PR #6832, opened 2023-03-29
DDF for HZC S900W-ZG water leak sensor

### LINKED ISSUE #6831: HCZ electric S900W-ZG Water leak sensor
## Device

- Product name: HCZ electric S900W-ZG Water leak sensor
- Manufacturer: Shyugj
- Model identifier: WaterLeakageSensor-ZB3.0
- Device type : 
  - Sensor

## Screenshots
image

### Basic
image

### Power Configuration
image

### Identify
image

### Poll Control
image

### IAS Zone
image
image

### Diagnostics
image

agramner:
Copied DDF from bosch water leak sensor and modified for this sensor.

These entities popped up in Home Assistant for the sensor (after a restart of HA):
image

Confirmed that `binary_sensor.water_x` works
Haven't confirmed the rest of the sensors.
`sensor.water_x_battery` is still 100% so can't be sure it works.
Not sure if `tampered` is relevant for this sensor. How could that trigger?

fixes #6831

### DIFF
--- a/devices/hzc_electric/s900w-zg_water_leak_sensor.json
+++ b/devices/hzc_electric/s900w-zg_water_leak_sensor.json
@@ -1,0 +1,118 @@
+{
+    "schema": "devcap1.schema.json",
+    "manufacturername": "Shyugj",
+    "modelid": "WaterLeakageSensor-ZB3.0",
+    "vendor": "HZC electric",
+    "product": "S900W-ZG Water leak sensor",
+    "sleeper": true,
+    "status": "Silver",
+    "subdevices": [
+      {
+        "type": "$TYPE_WATER_LEAK_SENSOR",
+        "restapi": "/sensors",
+        "uuid": [
+          "$address.ext",
+          "0x01",
+          "0x0500"
+        ],
+        "fingerprint": {
+          "profile": "0x0104",
+          "device": "0x0402",
+          "endpoint": "0x01",
+          "in": [
+            "0x0000",
+            "0x0001",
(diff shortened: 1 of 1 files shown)


BabaIsYou:
Nice and easy ! ;-)

Just a few comments : 
- Would have may be used constant "$TYPE_WATER_LEAK_SENSOR" instead of directly "ZHAWater" as type for subdevice
- Using fingerprint then perhaps have it complete  with all in clusters, a suggestion as it works like this too.
- Is it realistic to report battery every minute ? With the risk of ... draining battery ? Suggest a min of 3600 and max of 14400. BTW if battery changes for 1% it will reports.

About your question for tampered, not exposed by this device, could be deleted.

agramner:
Thanks for the suggestions :). Seems very reasonable.
What is fingerprint used for? Can't find any information about that.
The binding for cluster 0x500 doesn't have report. How does that work? When does it report changes?

BabaIsYou:
> Thanks for the suggestions :). Seems very reasonable. What is fingerprint used for? Can't find any information about that. The binding for cluster 0x500 doesn't have report. How does that work? When does it report changes?

What a bunch of good questions ;-)
I can make reference to already answered parts here : https://github.com/dresden-elektronik/deconz-rest-plugin/issues/6368#issuecomment-1263674660 ans there https://forum.phoscon.de/t/ddf-documentation/1401/5

agramner:
Thanks, pushed your suggested changes

Issue #6845, opened 2023-04-01
Philips Outdoor Sensor pairing uncomplete after OTA


olicooper:
## Describe the bug

I performed an OTA update on the sensor and after the update I noticed deconz was flashing green next to the device every few seconds. I found a related GitHub issue https://github.com/dresden-elektronik/deconz-rest-plugin/issues/1569 which states it is to do with pairing not being complete. I looked for logs related to this (see below). I tried a factory reset and repairing the device via the phoscon app but nothing changed.

## Steps to reproduce the behavior

Perform an update using this ota file: https://otau.meethue.com/storage/ZGB_100B_010D/bd3f218f-190b-4498-b6b3-69aea563fd9d/Sensor-ATmega_6.1.1.27575_0012.sbl-ota

This should update the sensor:
- **`Date Code`** `20180828` to `20190219`
- **`SW Build ID`** `6.1.0.25261` to `6.1.1.27575`

After the update has complete, you will see this behaviour.

## Expected behavior

Pairing should already be complete but if it has to pair again after the update then this should work without intervention.

## Environment

 - Host system: Raspberry Pi
 - Running method: Raspbian
 - Firmware version: 26720700
 - deCONZ version: 2.21.01
 - Device: ConBee II
 - Do you use an USB extension cable: yes

## deCONZ Logs

`16:11:37:489 No presence sensor found for 0x00178801XXXXXXXX, endpoint: 0x02`

ebaauw:
That's not what the green light means.  See https://github.com/dresden-elektronik/deconz-rest-plugin/wiki/deCONZ-for-Dummies#node-status-led.

olicooper:
Okay thank you for the quick response and for clarifying that.. I tired looking in the [deconz handbook PDF](https://www.dresden-elektronik.com/wireless/software/deconz.html?file=files/dresden-elektronik/content/downloads/dokumente/funktechnik/deCONZ-BHB-en.pdf) but the status LEDs are not mentioned in there.

Having left the device for a while the presence detection seems to have come back to life!

I am apprehensive to update my other sensors due to the log entries: `No presence sensor found for 0x00178801XXXXXXXX, endpoint: 0x02`... Should I keep this issue open to address a possible DDF problem?

ebaauw:
> There seems to be a lot of activity happening on the device since the update despite being on battery power.

[...]

PR #6849, opened 2023-04-02
Add `fingerprint` to DDFs for `sensors` resources

ebaauw:
Add `fingerprint` to the DDF sub devices for `sensors` resources of the following devices:
- Develco SMSZB-120 smoke detector;
- IKEA FYRTUR block-out roller blind;
- IKEA VINDSTYRKA Air Quality sensor;
- innr SP 120 smart plug;
- Hue smart button
- Hue dimmer switch (2021)
- Hue motion sensor;
- Hue outdoor motion sensor;
- Hue motion sensor (2022);
- Hue outdoor motion sensor (2022);
- Xiaomi Aqara FP1 presence sensor;
- Xiaomi Aqara Roller Shade Driver E1 

This is especially relevant to ZHAPresence sensors (like the Hue motion sensor family), where the fingerprint is used to find the `Sensor`, see #6845.  Not sure if this changes any functionality for the other resources, but it should get rid of "Missing cluster in sensor fingerprint" warnings in the log.

The devices in this PR are just the ones I have online.  I think this needs to be checked for all DDFs exposing `sensors` resources.


manup:
Side info: The fingerprint data in DDFs are currently extra baggage, needed for some devices, to keep parts of the old code working. I hope we can get rid of it in future entirely.

SwoopX:
While I really appreciate the effort, also to have the DDFs complete and consistent, the practical benefit of those additions is very limited as I recall. The DDFs their selves do not make the fingerprints (sort of) superfluous, but the parts which can be skipped in `apsdeDataIndication()` due to the result of the check if a device is managed or not.

As we see also with this PR, motion sensors are somewhat a special case in this regard, as the handler predominantly cannot be skipped. Devices exposing `config/delay` are save to skip it (Hue motion sensors, Develco/Frient), but all the others require it due to the `durationDue` related presence state reset.

Hue/Develco/Frient motion sensors "just" throw a, admittedly irritating, message that a respective sensor could not be found (true if no fingerprint is available) and exit the handler function. Same might happen for other clusters as well.

manup:
I think for now it's ok to include the fingerprint extras, once cleanup in C++ parts is more progressed and functionality which uses fingerprint can be handled by new DDF code we can remove these later on. On one side it's a uphill battle with baby steps, but I like the smaller steps so we don't mess up too eagerly :)

manup:
Random thought, we maybe should just remove the debug line in question, if it doesn't help us much other than being spammy?

PR #6851, opened 2023-04-02
DDF for GLEDOPTO RGBW

ebaauw:
DDF for my GLEDOPTO flood light, to eliminate the need to whitelist this device in Homebridge deCONZ.

The older GLEDOPTO devices all used `GLEDOPTO` for _Manufacturer Name_ and _Model Identifier_, but they would expose different endpoints.  This is for the RGBW lights, that expose a dimmable light on endpoint 0x0A and an extended color light on endpoint 0x0B, like the flood light that I have on my balcony.  I have some other devices (controllers) whitelisted in Homebridge Hue, but I'm not sure if I still have these.

Notes:
- I want to add a check for _SW Build ID_ of `1.0.2`, but that's currently not possible in a DDF.
- The max/min `ct` values are assumed; unfortunately, the light does not report the actual `ct`.  Note that `ct` is emulated using the RGB channels.

### DIFF
--- a/devices/gledopto/gledopto.json
+++ b/devices/gledopto/gledopto.json
@@ -53,10 +53,12 @@
                     "name": "cap/color/capabilities"
                 },
                 {
-                    "name": "cap/color/ct/max"
+                    "name": "cap/color/ct/max",
+                    "static": 500
                 },
                 {
-                    "name": "cap/color/ct/min"
+                    "name": "cap/color/ct/min",
+                    "static": 153
                 },
                 {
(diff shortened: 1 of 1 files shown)


manup:
Not sure if it helps here, but what already works with `matchexpr` ist testing for existing endpoints:

For example one DDF has: `"matchexpr": "(R.endpoints.indexOf(0x15) !== -1) && (R.endpoints.indexOf(0x16) !== -1);`

ebaauw:
Indeed, that's what I'm already using.

PR #6863, opened 2023-04-05
Add DDF generic item config/configured

manup:
Used by alarm systems and daylight sensor.


ebaauw:
Shouldn't that be read-only?

manup:
Hmm indeed it's only set internally, I'll add a followup PR.

PR #6864, opened 2023-04-05
Fix Javascript crash when setting unsupported JS value to item

manup:
The new code now prints the unsupported value. This can happen for example if NaN, null or undefined is returned, as well as JS objects and arrays.

@ebaauw this should prevent the crash in your network and give some info which item needs to be handled.

I'm not sure what we should do with NaN, null, or undefined in terms of setting a value suitable for a `ResourceItem`.

### DIFF
--- a/device_js/device_js_duktape.cpp
+++ b/device_js/device_js_duktape.cpp
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2022 dresden elektronik ingenieurtechnik gmbh.
+ * Copyright (c) 2022-2023 dresden elektronik ingenieurtechnik gmbh.
  * All rights reserved.
  *
  * The software in this package is published under the terms of the BSD
@@ -21,17 +21,16 @@
 #include "resource.h"
 #include "utils/utils.h"
 
-// before v2.19.2-beta DBG_JS isn't defined, can be removed afterwards
-#ifndef DBG_JS
-  #define DBG_JS DBG_INFO_L2
-#endif
-
+#ifdef DECONZ_DEBUG_BUILD
 #if _MSC_VER
   #define U_ASSERT(c) if (!(c)) __debugbreak()
 #elif __GNUC__
   #define U_ASSERT(c) if (!(c)) __builtin_trap()
 #else
   #define U_ASSERT assert
+#endif
+#else // release build
+  #define U_ASSERT DBG_Assert
 #endif
 
 static DeviceJs *_djs = nullptr; // singleton
@@ -801,7 +800,9 @@ static duk_ret_t DJS_SetItemVal(duk_context *ctx)
         }
         else
         {
-            U_ASSERT(0 && "unhandled value");
+            const char *str = duk_safe_to_string(ctx, 0);
+            DBG_Printf(DBG_JS, "%s: failed to set %s --> '%s' (unsupported)\n", __FUNCTION__, item->descriptor().suffix, str);
+            duk_pop(ctx); /* conversion result*/
         }
 
         if (!ok)


ebaauw:
Now running with changes to lines 803-805.  Fingers crossed.

ebaauw:
No crashes so far.  I think this is what we wanted to see?
```
Apr  6 09:11:56 pi5 deCONZ[21446]: 09:11:56:245 DJS_SetItemVal: failed to set state/airquality --> 'undefined' (unsupported)
Apr  6 09:12:06 pi5 deCONZ[21446]: 09:12:04:758 DJS result  undefined, memory peak: 147584 bytes
Apr  6 09:12:26 pi5 deCONZ[21446]: 09:12:26:153 DJS_SetItemVal: failed to set state/airquality --> 'undefined' (unsupported)
Apr  6 09:12:56 pi5 deCONZ[21446]: 09:12:56:058 DJS_SetItemVal: failed to set state/airquality --> 'undefined' (unsupported)
Apr  6 09:13:26 pi5 deCONZ[21446]: 09:13:25:969 DJS_SetItemVal: failed to set state/airquality --> 'undefined' (unsupported)
Apr  6 09:13:56 pi5 deCONZ[21446]: 09:13:55:875 DJS_SetItemVal: failed to set state/airquality --> 'undefined' (unsupported)
Apr  6 09:14:25 pi5 deCONZ[21446]: 09:14:25:782 DJS_SetItemVal: failed to set state/airquality --> 'undefined' (unsupported)
```

The only device with `state/airquality` I have connected to the test network is the Aqara TVOC Air Quality Monitor, `lumi.airmonitor.acn01`.  This is the temperature/humidity/TVOC density sensor with display.  I don't have any of these connected to any of my other networks.  It's exposed by DDF.  The definition for `state/airquality` reads:
```
        {
          "name": "state/airquality",
          "parse": {
            "at": "0x0129",
            "cl": "0xfcc0",
            "ep": 1,
            "eval": "Item.val = [excellent, good, moderate, poor, unhealthy][Attr.val]",
            "fn": "zcl",
            "mf": "0x115f"
          },
          "default": "excellent"
        },
```

The report received just before the message is:
```
Apr  6 09:11:56 pi5 deCONZ[21446]: 09:11:56:237 APS-DATA.indication srcAddr: 0xd62d, srcEp: 0x01 dstAddrMode: 2, profile: 0x0104, cluster: 0xFCC0, lqi: 255, rssi: -25
Apr  6 09:11:56 pi5 deCONZ[21446]: 09:11:56:239 	asdu: 1c5f112f0a29012005
```
So that's raw value `05` with is outside the range 0..4 supported by the JS array.  Note that the display shows TVOC as 0 to 5 leave-symbols, so it makes sense the sensor reports six values instead of five.

I see quite a few of the following message:
```
Apr  6 18:06:27 pi5 deCONZ[21446]: 18:06:20:614 DJS result  undefined, memory peak: 147584 bytes
```
messages that I hadn't noticed before.  Not sure if these are significant?

ebaauw:
Looking at the blame for the DDF file, the previous version used:
```
            "eval": "if (Attr.val === 1) {Item.val = 'excellent'} else if (Attr.val === 2) {Item.val = 'good'} else if (Attr.val === 3) {Item.val = 'moderate'} else if (Attr.val === 4) {Item.val = 'poor'} else if (Attr.val === 5) {Item.val = 'unhealthy'}",
```
This would suggest a rookie mistake: JavaScript array index starts at 0 instead of 1.  I don't think I've seen a value of 0 reported by the sensor, at least not in my logs of the last seven days (that's how long I keep them).

The fix is simple enough...

manup:
>Apr  6 18:06:27 pi5 deCONZ[21446]: 18:06:20:614 DJS result  undefined, memory peak: 147584 bytes
>messages that I hadn't noticed before. Not sure if these are significant?

These are there since the initial DuktapeJS introduction and logged due the DBG_JS option, mainly useful for me to determine how much memory is used while processing one "eval" run :)

PR #6867, opened 2023-04-06
Fix Aqara TVOC Air Quality Monitor

ebaauw:
Changes:
- Rename DDF to reflect `modelid`;
- Base `state/temperature`, `state/humidity`, `state/airqualityppb` (TVOC density), and `config/battery` on regualar  Zigbee attributes;
- ~~Setup attribure reporting for these regular attributes;~~
- ~~Base `state/airquality` on `state/airqualityppb` (TVOC density), as 0xFCC0/0x0129 doesn't support attribute reporting.~~
- Fix the `eval` for `state/airquality`, which caused deCONZ to crash due to out of bound array subscript (Javascript arrays start at index 0, where 0xFFC0/0x0129 starts at 1, see #6864);
- Replace script for `swversion` with inline Javascript.

The device is actually quite standard; the Xiaomi special attribute report is used only for `swversion`.

Note that the device wakes when pressing the physical button on top of it, ~~broadcasting _Identify Query_.~~


ebaauw:
Oh, this is fun.  I hadn't recevied any Xioami special report for a couple of hours, so reset and re-paired the device.  Now it seems to be in a different mode: it no longer supports configuring attribute reporting, and it sends the _Model Identifier_ and the XIaomi special attribute on pressing the button.  It does report temperature, humidiy, airqualityppb, airquality, and battery voltage thought both the XIaomi special attribute as well as through the regular attributes.

SwoopX:
Can we please revert the renaming of the DDF file and the changes on the `swversion`? Almost all files are named after their product ID and the `swversion` code will fail on numbers higher than 0x63.

I have to say I have a bad feeling about the attribute and reporting changes. Back in the days, it was a real hassle to get the device working somehow (see https://forum.phoscon.de/t/aqara-tvoc-zhaairquality-data/1160) and I would imagine it breaks running setups. Honestly, no real clue what to do here. My very personal impression on the firmware was already that it's fubar back in the days.

ebaauw:
I was surprised to see the attribute reporting for 0x000C/0x0055 in the old DDF.  Looking into that, I manages to setup bindings and reporting for the other attributes, much to my surprise.  I also noted more frequent reports as a result for that, but no longer saw the Xiaomi special attribute.  After reset/re-pair, the device reports all attributes (incl. 0x000C/0x0055) as not supporting attribute report configuration, but continued reporing.  It's probably one of those device settings: xiaomi mode vs zigbee mode, but I cannot find it.

Anyways until the `airquality` change and removed the bindings and attribute reporting.  My device still works after another reset/re-pair, with these changes.

ebaauw:
Now running device firmware 0.0.0_0029 (upgraded through the Aqara hub).  As far as I can tell, no functional changes, but they messed up the _Simple Descriptor_.

Added logic to parse the Xiaomi special attribute in JavaScript, so items are now updated by both regular attributes and Xiaomi special attribute.  Currently abusing `config/allowtouchlink` as hidden attribute for this, so I don't have to make any changes to the C++ code.  Will fix this, when cleaning up the C++ code.  Probably use hidden `attr/xiaomi` item, but I'mopen to suggestions.

PR #6878, opened 2023-04-10
Philips vs Signify

ebaauw:
I'm still getting reports of Hue devices incorrectly exposed with a `manufacturername` value of `"Philips"` instead of `"Signify Netherlands B.V."`, see e.g. https://github.com/ebaauw/homebridge-deconz/issues/131.  Normally, this should be corrected by force-reading the _Basic_ cluster attributes in the deCONZ GUI, using the _Cluster Info_ panel.  However, there are folks who don't have access to the GUI.

As far as I can tell, neither the deCONZ core nor the REST API plugin will read the _Manufacturer Name_ after a device has been paired and exposed.  I don't feel like messing with the C++ code to correct this, so instead opted to extend the DDFs for most (see below) Hue devices to match both values.  Also, I've updated the generic items for `attr/manufacturername` and `attr/modelid` to read the corresponding Zigbee attribute once a day.  This should cause the `manufacturername` to be updated.

By "most", I mean:
- The old Hue motion sensors and Hue dimmer switches are still matched only on `"Philips"`, as they haven't received a firmware upgrade to change the _Manufacturer Name_ (and I don't think they will).
- The gradient lights are matched only on `"Signify Netherlands B.V."` as (I think) they came out after deCONZ stopped force-exposing the `manufacturername` as `"Philips"`.
- Note that some of the newer ZB3 gamut-C lights still use `"Philips"`, but I expect this to change with their next firmware, so they're matched by both names.

I suppose we can undo the double matching in a couple of releases, when _Manufacturer Name_ will have been read through these DDFs.

I briefly considered force-exposing `manufacturername` as `"Signify Netherlands B.V."`, but decided against it.  I think it's good to see over the REST API what the device exposes, if only as a quick way to identify outdated firmware.  Lacking a firmware check in the DDF, this change will increase the chances of matching lights with older firmware, predating the _Hue Effects_ cluster.


BabaIsYou:
> I briefly considered force-exposing `manufacturername` as `"Signify Netherlands B.V."`, but decided against it. I think it's good to see over the REST API what the device exposes, 

I clearly agree this way of thinking ;-)

ebaauw:
Moved the `read` for `manufacturername` and `modelid` from the generic item files to the device DDF files (of the non-sleeper devices), at @manup 's request.  For sleepy end devices, we don't want these `read`s to take prio over the binding requests.

Issue #6871, opened 2023-04-07
Salus PS600 Pipe Temperature Sensor


Carl-mgb66:
## Device

- Product name: Salus PS600 Temperature Sensor
- Manufacturer: Computime
- Model identifier: SS909ZB
- Device type :
  - Sensor

## Clusters of the Node

![image](https://user-images.githubusercontent.com/75570551/230605157-70617780-339f-4a13-ab2a-361fa7adb529.png)

## Node Info Panel

![image](https://user-images.githubusercontent.com/75570551/230606240-e7f6a658-379f-4801-9c53-2f7ea93923f0.png)

### Basic

![image](https://user-images.githubusercontent.com/75570551/230605003-5f68d167-2bf5-40f6-a6ee-20a1baf87b23.png)

### Identify

![image](https://user-images.githubusercontent.com/75570551/230605261-fdadfb75-8895-444a-8946-2a7ac7c94eb0.png)

### Diagnostics

![image](https://user-images.githubusercontent.com/75570551/230605433-64932828-cc6e-4e53-b2eb-7abba80f59d9.png)

### Power Configuration

![image](https://user-images.githubusercontent.com/75570551/230605590-b341505a-24db-4876-b0f0-e26f1018216f.png)

### Poll Control

![image](https://user-images.githubusercontent.com/75570551/230605779-b550a89e-e98a-4680-bf1a-0622487ca7a9.png)

### Temperature Measurement

![image](https://user-images.githubusercontent.com/75570551/230605875-30eaa654-40eb-4c32-ac6f-4a985fa15b19.png)

### IAS Zone

![image](https://user-images.githubusercontent.com/75570551/230605946-0c88820f-b93b-41b9-9017-15af840348b4.png)

BabaIsYou:
Hi,

May be you have to create two files under the folder for DDF files on your system

The first one with a name like `ss909zb_temp_sensor.json `containing 

```

{
  "schema": "devcap1.schema.json",
  "manufacturername": "Computime",
  "modelid": "SS909ZB",
  "product": "Salus PS600 Temperature Sensor",
  "sleeper": true,
  "status": "Silver",
  "subdevices": [
    {
      "type": "$TYPE_TEMPERATURE_SENSOR",
      "restapi": "/sensors",
      "uuid": [
        "$address.ext",
        "0x09",
        "0x0402"
      ],
      "fingerprint": {
        "profile": "0x0104",
        "device": "0x0302",
[...]

PR #6882, opened 2023-04-11
DDF for Salus PS600 Pipe temperature sensor

Carl-mgb66:
Add JSON and associated JS files to enable support of Computime/Salus PS600 Pipe Temperature Sensor

Fix https://github.com/dresden-elektronik/deconz-rest-plugin/issues/6871

### DIFF
--- a/devices/computime-salus/computime_battery.js
+++ b/devices/computime-salus/computime_battery.js
@@ -1,0 +1,13 @@
+const vmin = 26;
+const vmax = 30;
+let bat = Attr.val;
+
+if      (bat > vmax) { bat = vmax; }
+else if (bat < vmin) { bat = vmin; }
+
+bat = ((bat - vmin) /(vmax - vmin)) * 100;
+
+if      (bat > 100) { bat = 100; }
+else if (bat <= 0)  { bat = 1; } // ?
+
+Item.val = bat;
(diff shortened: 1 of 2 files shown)


SwoopX:
A few questions around this:

- Some items have `"awake": true`. Has it been tested that the device is really awake when anything for the respective items is received?
- The device hold the IAS Zone cluster, does it make sense/is it necessary to enroll it. That would also involve adding 1 or 2 additional items.

Carl-mgb66:
In response to the questions, please note @BabaIsYou helped to write the files, hence I'm mentioning them here.

1. "awake": true - please advise how I could test this and provide the feedback?
2. IAS Zone Cluster - don't believe it is necessary to enroll it, it's working without and it doesn't really make much sense either.

Hope that helps.

manup:
The awake is a bit tricky to get right. It should be set to true, if the respective command in the "parse" function is followed by a MAC Data Request, unfortunately this can be seen only with a sniffer. Sometimes also with logging but only if the device has the coordinator as parent.

Carl-mgb66:
Thanks for the explanation, however I don't think I'll be able to make these checks myself. The sensor is operating reliably.

BabaIsYou:
> * "awake": true - please advise how I could test this and provide the feedback?

Hi @Carl-mgb66, I suggest to remove all the `"awake": true`, it won't change anything for that kind of device IMO.

> 
> * IAS Zone Cluster - don't believe it is necessary to enroll it, it's working without and it doesn't really make much sense either.

Agree with that.

Carl-mgb66:
I've removed the "awake": true statements as suggested.

PR #6895, opened 2023-04-14
Fix IKEA FYRTUR block-out roller blind

ebaauw:
Bug fix: `state.open` showed incorrect value.  Note that `state.open` is true when the blind if fully opened, and false otherwise.  Note that `state.lift` reports % closed, as per ZCL spec.  So `state.open` is true only when `state.lift` is 0.

Remove `state.on` and `state.bri`, which have been deprecated for over three years now.

### DIFF
--- a/devices/ikea/fyrtur_block-out_roller_blind.json
+++ b/devices/ikea/fyrtur_block-out_roller_blind.json
@@ -43,44 +43,16 @@
                     "name": "attr/uniqueid"
                 },
                 {
-                    "name": "state/bri",
-                    "parse": {
-                        "at": "0x0008",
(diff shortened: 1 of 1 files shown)


Armilar:
Has anything else been changed in the Ikea Fyrtur code? Since version 2.22.0 my Fyrtur no longer work in ioBroker. But before that everything was fine.

I periodically try the latest version and after I see that it doesn't work, I change the version with:


`sudo apt-mark hold deconz`

and

`sudo dpkg -i deconz-2.21.02-qt5.deb.3`

back to the last working 2.21.02

I would like to get a newer release, but that won't work as long as the Fyrtur doesn't work correctly.

PR #6903, opened 2023-04-16
Js to octed string

aryelevin:
This will allow sending Hex string to Octed String attributes, will enable full control of 0x41 data type attributes.
Together with my other PR #6884 which send Octed Strings as Hex string, this is allowing the other way around.


aryelevin:
I'm closing since we can now access the raw data by ZclFrame on the DDF system and we can use zcl:cmd to send hex string as data to ostrings values.

Issue #6904, opened 2023-04-17
Configure attribute reporting for `float` attributes


ebaauw:
## Describe the bug

Neither the GUI nor DDF definitions handle correctly the configuration of attribute reporting for attributes of type `float` (0x39).  The _Reportable Change_ needs to be encoded as a iEEE-754 16-bits floating point number (see e.g. the online [IEEE-754 Floating Point Converter](https://www.h-schmidt.net/FloatConverter/IEEE754.html)), but both the GUI and DDF encode the value specified as 16-bit integer.

When entering a _Reportable Change_ value of 1, the GUI sends a _Configure Reporting_ command with value `10 00 00 00`, which is 0x00000001 or 1 as integer, but 1.4013e-45 as floating point.  When receiving a _Read Reporting Configuration Response_, _Reportable Change_ remains empty.  
When entering 1065353216, the GUI sends `00 00 80 3f`, which is 0x3F800000 as integer, but 1 as as floating point.  When reading back the configuration, _Reportable Change_ still remains empty, but the Sniffer confirms that the correct value has been set.

In the DDF `binding[]/report/change` needs to be specified as hex string.  The logic behind the DDF also encodes this as an integer value, instead of a floating point value.  So for a value of 1, you need to specify `0x3F80000`.

## Steps to reproduce the behavior

See the Zigbee sniffer.

## Expected behavior

The GUI should display and accept numeric float values.

The DDF should accept JavaScript number values.

## Screenshots

n/a

## Environment

 - Host system: all
 - Running method: all
 - Firmware version: n/a
 - deCONZ version: all
 - Device: all
 - Do you use an USB extension cable: (yes / no) n/a
 - Is there any other USB or serial devices connected to the host system? If so: Which? n/a

## deCONZ Logs

n/a

## Additional context

ZCL v8 defines several _Concentration Measurement_ clusters with _Measured Value_ (0x0000) as `float` and reportable.  Notably _PM2.5 Measurement_ (0x040B) and IKEA-specific _TVOC Measurement_ (0xFC7E) on the IKEA VINDSTYRKA.

PR #6906, opened 2023-04-17
Update ZCLDB to add/update some Measurement clusters

BabaIsYou:
Complete general.xml to add Concentration Measurement clusters as Zigbee Cluster Library Specification Revision 8 and other measurement updates.

### DIFF
--- a/general.xml
+++ b/general.xml
@@ -3208,9 +3208,17 @@ Note: It does not clear or delete previous weekly schedule programming configura
       <description>The server cluster provides an interface to air pressure measurement functionality, including configuration and provision of notifications of air pressure measurements.</description>
       <server>
         <attribute-set id="0x0000" description="Pressure Measurement Information">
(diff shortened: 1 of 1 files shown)


BabaIsYou:
BTW need perhaps to be linked with the issue #6904 to be fully operational in GUI

ebaauw:
Damn, this PR breaks support for the Hue motion sensors sensitivity and sensitivity max.

BabaIsYou:
Look like the same trouble I already got because some Elko specific defintions hadn't mfcode specified in general.xml and were positionned before Sinopé specific definition. Then were loaded before specific Sinopé that had mfcode.

Changing the order into general.xml resolved the issue #6827

Then I guees that moving Hue-specific attributes 0x0030 and 0x0031 on Occupancy Sensing before the Zigbee specification will solve it.

BabaIsYou:
I know it works for having tested it before on another previous conflicted definition.

I’m also aware that’s just a workaround, but it works until other way of doing it (custom xml file with specific loading order should have been another solution … if it should have been functional; -/ )

PR #6913, opened 2023-04-19
Refactor MMO hash function

manup:
- Load OpenSSL function dynamically to not hard link the library.
- Replaced various std::vectors with simple C arrays.
- Move code from utils.cpp to crypto/mmohash.cpp
- Verify that that install code has valid sizes of 6, 8, 12, 16 plus two bytes CRC and contains actual hex data.
- Test if the CRC is valid and auto correct if not (there are some devices with invalid/twisted CRC)
- Verified against Zigbee Alliance test vectors.
- On Windows include msvcr100.dll to deCONZ installer, required for the OpenSSL libraries and might be missing on the host system.

This was also needed to get it compile for v2.22.0-beta Windows release (which didn't work out).

### DIFF
--- a/crypto/mmohash.cpp
+++ b/crypto/mmohash.cpp
@@ -39,7 +39,51 @@ static bool aesMmoHash(unsigned &length, unsigned char *result, unsigned char *d
     return true;
 }
 
-bool getMmoHashFromInstallCode(std::string hexString, std::vector<unsigned char> &result)
+#define CRC_POLY 0x8408
+static unsigned short ccit_crc16(unsigned char *data_p, unsigned short length)
+{
+    unsigned char i;
+    unsigned int data;
+    unsigned long crc;
+
+    crc = 0xFFFF;
+
+    if (length == 0)
+        return (~crc);
+
+    do {
+        for (i = 0, data = (unsigned int)0xff & *data_p++;
+             i < 8;
+             i++, data >>= 1) {
+            if ((crc & 0x0001) ^ (data & 0x0001))
+                crc = (crc >> 1) ^ CRC_POLY;
+            else
+                crc >>= 1;
+        }
+    } while (--length);
+
+    crc = ~crc;
+
+    return crc;
+}
+
+/*
+    Verification of offical alliance example:
+
+    curl -XPUT -H "Content-type: application/json" -d '{"installcode": "83FED3407A939723A5C639FF4C12"}' '127.0.0.1/api/12345/devices/999/installcode'
+
+    [
+      {
+        "success": {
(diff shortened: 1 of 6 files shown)


SwoopX:
@manup just to be sure: this is already part of the current release? Asking because Windows seems (still) to have issues, as there's no hash returned.

I've quickly tried it on a Windows VM and I wasn't able to get anything back 🤔

manup:
Hmm strange, I only checked in the debugger but not the release version. Will make some tests and changes before this goes life.

SwoopX:
@manup it seems like #6935 might also be related to this. My test env also refused to work at all if the required library is not installed (Ubuntu 22.04).

manup:
I made a few tests on my Windows box in debug mode with the test data from the Zigbee Alliance code, to verify the produced hash makes sense. Added a few goodies to verify and auto correct the install code CRC if needed (see updated PR description).

Gonna test the packaged version again and do some checks for the Ubuntu versions to verify loading of the library works later today.

manup:
@SwoopX something is weird. I've now tested the [deCONZ_Setup_Win32_V2_22_00.exe](https://deconz.dresden-elektronik.de/win/deCONZ_Setup_Win32_V2_22_00.exe) release on Windows 10 and 11, installed as user and as admin. So it doesn't look like DDLs are missing from installer or they aren't loaded.

Test works correctly for following 12-bytes install code (plus 2 bytes CRC):

```
curl -XPUT -H "Content-type: application/json" -d '{"installcode": "83FED3407A939723A5C639FF4C12"}' '127.0.0.1/api/12345/devices/999/installcode'
```

Producing the correct output:

```json
[
  {
    "success": {
      "installcode": "83FED3407A939723A5C639FF4C12",
      "mmohash": "58C1828CF7F1C3FE29E7B1024AD84BFA"
    }
  }
]
```

manup:
By testing on Ubuntu 22.04 I think to have found the problem, which is the switch from OpenSSL 1.x.y to 3.x.y.
While the new code is fine to load either via `QLibrary libCrypto("crypto");` The functions we are using are deprecated and I guess the Ubuntu OpenSSL library is compiled with deprecated functions excluded.

According to the [OpenSSL 3 migration guide](https://www.openssl.org/docs/man3.0/man7/migration_guide.html) the EVP API should be used instead low-level functions like `AES_encrypt()`.

The PR with recent changes should ensure loading OpenSSL 1.1.1 as long as it is installed.

```
wget http://mirrors.kernel.org/ubuntu/pool/main/o/openssl/libssl1.1_1.1.1f-1ubuntu2.19_amd64.deb
dpkg -i libssl1.1_1.1.1f-1ubuntu2.19_amd64.deb
```

Issue #263, opened 2017-11-10
Xiaomi Aqara flood


Kohle81:
I want to include this sensor with deCONZ to Homekit with homebridge-hue plugin.

Does it work or what can I do?

Greetings

manup:
To start with it would be nice to have a screenshot from the sensor with expanded cluster list, for that click on second circle on the sensor node.

Also the model id which the sensor reports in the `Basic Cluster` in the cluster info panel.
If it is empty you may need to rejoin the sensor whilte the network is open, for Xiaomi sensors there should be a button on the device.

Kohle81:
Ok, I will send a photo next days. Actually I have this picture on my smartphone from the sensor.
![27ace2a2-3277-4b48-81ee-201f16638515](https://user-images.githubusercontent.com/33038709/32663690-2dab747e-c62f-11e7-9585-755da390bbd4.jpeg)

manup:
That only shows the mac address (00158d...), for the integration the other infos are needed :)

oleg77s:
In my system:
![2017-11-10-190340_1824x984_scrot](https://user-images.githubusercontent.com/23496372/32666988-5da2aace-c64a-11e7-90c2-5bb8dc3ec137.png)

ebaauw:
Hm, that doesn’t look good: there should be another cluster where the sensor reports its state.  Could you try and repair the sensor while regularly pressing the sensor’s button (or hole), to force it to remain awake during pairing?

oleg77s:
yes, i try, but there is only basic cluster

Kohle81:
I have only the same fluster Info. There is no more Information.......

pietervandersluis:
I had the same issue with a aqara wall switch. Have you tried re-pairing the device?

Kohle81:
I have re-paired two times. Always the same...

Von meinem iPhone gesendet

oleg77s:
Xiaomi Smoke sensor: only basic cluster too :(

ebaauw:
See #152 for the smoke detector.

Reading back that issue, maybe Xiaomi don’t report the cluster in the device description (so deCONZ GUI won’t show it), but does send commands from the cluster anyways.  Wouldn’t be the first time they’re creative with the standard.  The bad news: we need to sniff the ZigBee traffic to figure out the flood sensor.

QempZor:
Sad news that it is still not working :
[...]

PR #6919, opened 2023-04-21
Xiaomi special attribute in DDF

ebaauw:
## Xiaomi special attribute
Create a number of dummy companion items (e.g `state/temperature_bis` for `state/temperature`), so the item can be updated from a regular Zigbee attribute and from the Xioami special attribute.  In time, the dummy items will become dynamic items, and can be removed from the C++ code.

Added logic to `handleSensorEvent()` to reset automatically `state/presence` or `state/vibration` on managed devices with `config/duration` > 0.  Note that `config/duration` should be exposed only on devices that don't send "no more presence" or "no more vibration" events.

I have been converting the DDFs for all Xiaomi devices in my possession, double-checking with the sniffer that these are handled correctly.  I won't touch DDF for devices not / no longer in my possession.

### `lumi.airmonitor.acn01`
This is the Aqara TVOC Air Quality Monitor VOCKQJK11LM from #6867.

Changes:
- Delete `fcc0_airmonitor_acn01.js` in favour of `eval` per dummy item.

### `lumi.motion.ac01`
This is the Aqara FP1 Human Presence Sensor RTCZCGQ11LM from #6900.

Changes:
- Remove `config/allowtouchlink` in favour of `config/reportgrid`, which can be updated from the API.  The grid data is not yet exposed by the API.

### `lumi.sensor_magnet.aq2`
This the square-ish Aqara Door and Window Sensor MCCGQ11LM that exposes the contact state through an _On/Off_ cluster, as well as through the special attribute.  The special attribute also reports device temperature and battery voltage.

Changes:
- Rename file to `lumi.sensor_magnet.aq2.json`;
- Parse special attribute to update `config/battery`, `config/temperature`, `state/open`;
- Base `swversion` on _Application Version_, in line with the Aqara app though the Aqara M2.

### `lumi.sensor_wleak.aq1`
This is the round Aqara Water Leak Sensor SJCGQ11LM from #263.  
It reports water leak detection through _IAS Zone_, but doesn't support enrolling.  
It also reports leak detection though the Xiaomi special attribute, as well as device temperature, battery voltage, and firmware version.

Changes:
- New DDF file.
- Remove the non-functional `config/enrolled`, `config/pending`, `state/lowbattery`, and `state/tampered` that the legacy code exposed.
- `swversion` now shows `0.0.0_0006`.

### `lumi.vibration.aq1`
This is the Aqara Vibration Sensor DJT11LM.

Changes
- Rename file to `lumi_vibration_aq1.json`;
- Normalise `config/sensitivity` from 0 (low) to 2 (high), with matching `config/sensitivitymax`;
- Add `config/duration` with fixed value of 65 to reset automatically `state/vibration`;
- Remove `config/pending`;
- Fix JavaScript to compute orientation and rename to `0101_orientation.js`.  Note that Javascript internally uses int32 for bitwise operators, hence the magic to extract x, y, and z as (signed!) int16 from the int48 attribute value.
- Set `state/tiltangle` to 230 on drop (same as legacy code);
- Fix generic items for `state/orientation_x`, `state/orientation_y`, `state/orientation_z` and remove `state/orientation`.

### `lumi.weather`
This is the square(-ish) Aqara Temperature and Humidity Sensor WSDCGQ11LM.  
It carries a Xiaomi special attribute, but without the software version and device temperature.  It also reports temperature, humidity, and air pressure over regular attributes, but no battery voltage.

Changes:
- Rename file to `lumi_weather.json`;
- Parse special attribute to update `config/battery`, `state/temperature`, `state/humidiy`, `state/pressure`; remove parse from `config/battery`
- Base `swversion` on _Application Version_, in line with the Aqara app though the Aqara M2.
- Remove `config/temperature` as device doesn't report this;
- Remove `awake` setting for `manufacturername` as it doesn't report _Manufacturer Name_ on pressing the button (only _Model Identifier_ and the Xiaomi special attribute).


BabaIsYou:
Hi. Is there a reason why `swversion` is still with `"refresh.interval": 84000` ? The generic item has already` "refresh.interval": 86400`

ebaauw:
> Hi. Is there a reason why swversion is still with "refresh.interval": 84000 ? The generic item has already "refresh.interval": 86400

Good one.   Actually, the generic item also already has the `parse` and `read` functions.

manup:
Finally in :) I've checked with a bunch of devices, looks to work well in my setup:

```
00:15:8d:00:07:e6:20:37-01-0006/state/open_bis expression: R.item('state/open').val = Attr.val --> true
```

PR #6933, opened 2023-04-28
Fix DDF for Hive

ebaauw:
This the the Hive battery-powered light by dresden elektronik, see https://phoscon.de/en/hive.  It supports ZCL v8, including power-on and execute-if-off settings.

The device firmware is still under development, notably the following limitations apply:
- It doesn't turn on nor off on _Move to Level (with On/Off)_;
- It doesn't report its supported colour gamut; the DDF just lists the default gamut;
- It doesn't support _Enhanced Hue/Saturation_ colour mode, nor _Color Loop_.  It does support _Enhanced Current Hue_, but this attribute is updated only in `xy` mode.
- When changing the colour on the device, it switches to `hs` mode.  Unfortunataly, it doesn't update `xy` when in `hs` or `ct` mode.  However, it does update `hue` and `sat` when in `xy` mode.
- The battery charing state isn't exposed.

Changes:
- Setup attribute reporting for _Color Control_, so that manually changing the colour (temperature) on the Hive is reflected in real-time in the GUI and API.  Note: still testing how the Hive behaves when outside network range; it does expect a _Default Response_ from the coordinator to _Report Attributes_.
- Combine reading the _Color Control_ attributes in one command;
- Add missing `capabilities` and `config` attributes;
- Remove `state/effect`, as Hive doesn't support colour loop.

Below the resulting `lights` resource.  Note that the battery is exposed as ZHABattery `sensors` resource, which is left unchanged by the PR.
```
{
  "capabilities": {
    "alerts": [
      "none",
      "select",
      "lselect",
      "blink",
      "breathe",
      "okay",
      "channelchange",
      "finish",
      "stop"
    ],
    "color": {
      "ct": {
        "max": 370,
        "min": 153
      },
      "modes": [
        "ct",
        "hs",
        "xy"
      ],
      "xy": {
        "blue": [
          0,
          0
        ],
        "green": [
          0,
          0.9961
        ],
        "red": [
          0.9961,
          0
        ]
      }
    }
  },
  "colorcapabilities": 25,
  "config": {
    "bri": {
      "execute_if_off": true,
      "startup": "previous"
    },
    "color": {
      "ct": {
        "startup": "previous"
      },
      "execute_if_off": true
    },
    "groups": [],
    "on": {
      "startup": "previous"
    }
  },
  "ctmax": 370,
  "ctmin": 153,
  "etag": "e06beda85b9f0de06a66116d392c1d1d",
  "hascolor": true,
  "lastannounced": null,
  "lastseen": "2023-04-28T09:05Z",
  "manufacturername": "Phoscon",
  "modelid": "Hive",
  "name": "Hive",
  "state": {
    "alert": "none",
    "bri": 196,
    "colormode": "xy",
    "ct": 370,
    "hue": 0,
    "on": true,
    "reachable": true,
    "sat": 254,
    "xy": [
      0.6936,
      0.2961
    ]
  },
  "swversion": "0401.0020001B",
  "type": "Extended color light",
  "uniqueid": "00:21:2e:ff:ff:0c:73:90-01"
}
```


ebaauw:
To test the Hive behaviour when out of reach, I shutdown deCONZ, but it seems that the coordinator device firmware handles the _Default Response_ independently from the gateway.  Next, I removed the RaspBee II.  Initially, the Hive resends the unanswered _Reeport Attributes_ commands, then it switches to broadcasting _Network Address Request_ for the coordinator.

After ~5 minutes, I re-inserted the ConBee II (without starting deCONZ).  Its firmware issues the _Network Address Response_.  The Hive goes back to _Report Attributes_ and the coordinator firmware responds with a _Default Response_.

@manup, as far as I can tell, setting up reporting to reflect the Hive state in real-time has no ill effects.  Are you happy with these results, or do you want me to do a longer test (e.g. disconnecting the ConBee II overnight)?

devices/generic/items/state_enhanced_hue_item.json
```diff
@@ -0,0 +1,15 @@
...
+  "datatype": "UInt16",
+  "access": "R",
+  "public": false,
+  "description": "The current color hue.",
+  "parse": {
+    "fn": "zcl",
+    "ep": 0,
+    "cl": "0x0300",
+    "at": "0x4000",
+    "eval": "R.item('state/hue).val = Attr.val"
```
manup, line 13: This throws an JS exception since `'state/hue`  is a unterminated string.
Note all JS code is test compiled during load and errors are reported via `--dbg-ddf=1`.

```
DDF failed to compile JS: /home/mpi/src/deCONZ_V2_XX/src/plugins/de_web/devices/dresden_elektronik/hive.json
SyntaxError: unterminated string (line 1, end of input)
```

The output is not too detailed but gives a hint.

To be honest I would propose not to add new `state/enhanced_hue` and `state/enhanced_colormode` items. The existing `state/hue` already reflects enhanced hue when available. For `state/colormode` we often map enhanced color mode values as well. Since we can redefine parsing directly in the DDF, if needed, we don't need extra items here imho.

manup:
Running the PR now since a few days without issues (state of May 16).

I'm still uneasy about the addition of `state/enhanced_hue` and `state/enhanced_colormode`, do we really need them (see https://github.com/dresden-elektronik/deconz-rest-plugin/pull/6933#discussion_r1195107316)?

ebaauw:
I'm not using `state/enhanced_colormode` at the moment, so I'll remove that one.  I'm using both `state/hue` and `state/enhanced_hue` for the Hive.  I'm happy to rename it to `state/hue_bis`, but that's only obfuscating what's happening under the hood.

manup:
I don't understand why it is needed. The `state/hue` "read" and "parse" functions only need to use the correct attribute for the device, or did I miss something?

ebaauw:
The Hive mostly uses _Current Hue_, but sometimes _Enhanced Current Hue_.  So I need to parse two Zigbee attributes for a single item (`state/hue`).

manup:
Ah ok I see, note that `"at"` parameter can also be an array, so for `state/hue` the following should do:

```json
"parse": {
  "fn": "zcl",
  "ep": "0x01",
  "cl": "0x0300",
  "at": ["0x0000", "0x4000"],
  "eval": "Item.val = (Attr.id === 0) ? Attr.val << 8 : Attr.val"
}
```

manup:
Cool thanks, DDF looks fine to me now, tested it a bit and all seems to work :+1: 

Only the `resource.[cpp|h]` files still contain the enhanced items.

ebaauw:
Damn.  Changed these.

But now I'm seeing changes to other files that I didn't make?  I did merge upstream/master the other day, but I don't understand why GitHub shows these files as changed?

manup:
I think this is just a GitHub UI hickup, when I merge locally it only shows the truly affected files.

ebaauw:
Scary.  I think `hive.json` is the only file left changed now.  I'll create a new PR with just that.

PR #6936, opened 2023-04-30
Some more Hue lights

ebaauw:
## Additional models
See Discord, **#restapi-issues** channel.

## `productname`
Expose `productname` for "all" Hue devices exposed through DDF:
- Read for light-oids, using 0xC0/0xC1 manufacturer-specific commands on the _Basic_ cluster.  The offset for the 0xC0 command is computed using the `modelid` and `manufacturername` items, instead of taking these from the 0xC1 response at offset 0.  This way, only a single 0xC0 needs to be issued.
- Hardcoded sensor-oids.  These have a dedicated DDF per model anyway, and 0xC0/0xC1 doesn't work for all sensor-oids (notably the Hue motion sensor family report themselves as Hue dimmer switch).

## `cmd` in DDF `read` and `write`
Add support for `cmd` (instead of `attr`) in `read` and `write` functions of item in DDF.  Both take an optional `eval` to specify the command payload as hex string.
- Add `ZCL_WriteAttribute()` and `ZCL_SendCommand()` to `zcl.cpp`;
- Move ZCL handling from `WriteZCLAttribute()` in `device_access_fn.cpp` to `ZCL_WriteAttribute()`;
- Call `ZCL_SendCommand()` from `ReadZCLAttribute()` and `WriteZCLAttribute()` to send the specified command instead of _Read Attributes_ or _Write Attributes_.


SwoopX:
@ebaauw Could you please also include a change for this particulat code here?

https://github.com/dresden-elektronik/deconz-rest-plugin/blob/91df7d74c6eb7a1bc7eeb6b18500748205da2700/thermostat.cpp#L331

and make it `item->setValue(s, ResourceItem::SourceDevice);` ?

This would allow using the read cmd for `config/schedule`. Without the suggested change, the command is send in an infinite loop, as the item timestamp is never correctly updated.

ebaauw:
> Could you please also include a change for this particular code here?

@SwoopX : [81376ca375823291359b0510969d8b86c88b4622](https://github.com/dresden-elektronik/deconz-rest-plugin/pull/6936/commits/81376ca375823291359b0510969d8b86c88b4622)

No idea what it does, but it compiles.

manup:
The `ResourceItem::SourceDevice` parameter tells the StateChange API where a value comes from (API, unknown, device).
It helps figuring out if a value set via the REST-API was actually applied.

device_access_fn.cpp
```diff
@@ -1638,15 +1642,15 @@ bool parseAndSyncTime(Resource *r, ResourceItem *item, const deCONZ::ApsDataIndi
 /*! A generic function to read ZCL attributes.
     The item->readParameters() is expected to be an object (given in the device description file).
 
-    { "fn": "zcl", "ep": endpoint, "cl" : clusterId, "at": attributeId, "mf": manufacturerCode, "noseq": noSequenceNumber  }
+    { "fn": "zcl:attr", "ep": endpoint, "cl" : clusterId, "at": attributeId, "noseq": noSequenceNumber  }
```
manup, line 1645: `"mf": manufacturerCode` is missing now

device_access_fn.cpp
```diff
@@ -1638,15 +1642,15 @@ bool parseAndSyncTime(Resource *r, ResourceItem *item, const deCONZ::ApsDataIndi
...
 
-    - endpoint, 0xff means any endpoint
+    - endpoint, (optional) 0xff means any endpoint
```
manup, line 1647: Ha interesting looks like a bug to me also in former version.

The endpoint is not optional, if it isn't specified it becomes 0xFF and the request goes into the void since we're doing unicasts. 0 is allowed, it's the `AutoEndpoint` which uses the endpoint from the items resource `.uiniqueid`

device_access_fn.cpp
```diff
@@ -1701,16 +1705,16 @@ static DA_ReadResult readZclAttribute(const Resource *r, const ResourceItem *ite
 /*! A generic function to write ZCL attributes.
     The \p writeParameters is expected to contain one object (given in the device description file).
 
-    { "fn": "zcl", "ep": endpoint, "cl": clusterId, "at": attributeId, "dt": zclDataType, "mf": manufacturerCode, "eval": expression }
+    { "fn": "zcl:attr", "ep": endpoint, "cl": clusterId, "mf": manufacturerCode, "at": attributeId, "dt": zclDataType, "eval": expression }
 
     - endpoint: (optional) the destination endpoint
```
manup, line 1710: Same as above, the endpoint isn't optional here. I think we should add checks to verify the parameters have the "ep".
This needs to be done in `readZclAttribute()` and `writeZclAttribute()` and `writeZclCommand()` since for `getZclParam()` it's valid that the "ep" is not present.

manup (commented):
Only a few minor comments, overall the PR looks pretty good to me.

device_access_fn.cpp
```diff
@@ -480,15 +480,19 @@ bool parseNumericToString(Resource *r, ResourceItem *item, const deCONZ::ApsData
...
-    {"fn": "zcl", "ep": endpoint, "cl": clusterId, "at": attributeId, "mf": manufacturerCode, "eval": expression}
+    {"fn": "zcl:attr", "ep": endpoint, "cl": clusterId, "mf": manufacturerCode, "at": attributeId, "eval": expression}
 
     - endpoint: (optional) 255 means any endpoint, 0 means auto selected from the related resource, defaults to 0
     - clusterId: string hex value
-    - attributeId: string hex value or array of string hex values
-    - manufacturerCode: (optional) string hex value, defaults to "0x0000" for non manufacturer specific commands
-    - expression: Javascript expression to transform the raw value
+    - manufacturerCode: (optional) string hex value
+    - attributeId: string hex value
```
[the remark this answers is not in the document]
ebaauw: Updated comment

device_access_fn.cpp
ebaauw: changed comment

device_access_fn.cpp
ebaauw: Updated comment for `readZclAttribute()`, `sendZclCmd()`, `writeZclAttribute()`, `writeZclCommand()`.

PR #6947, opened 2023-05-04
DDF - Add GiEXperience SMART Water Valve

### LINKED ISSUE #6944: GiEXperience SMART Water Valve (_TZE200_a7sghmms)
## Device

- Product name: GiEXperience SMART Water Valve
- Manufacturer: _TZE200_a7sghmms
- Model identifier: TS0601
- Device type : 
  - Other: Water Valve
- Link to product
[GiEXperience SMART Water Valve](https://www.aliexpress.com/item/1005002767720399.html?spm=a2g0o.productlist.main.23.2a167054om2ca6&algo_pvid=6cac93dd-63d8-4b5e-9e19-2381e9e3a8fa&algo_exp_id=6cac93dd-63d8-4b5e-9e19-2381e9e3a8fa-11&pdp_npi=3%40dis%21CHF%2165.38%2144.46%21%21%21%21%21%40211bf3f116830512797731945d080f%2112000022089158096%21sea%21CH%210&curPageLogUid=S5NqvJVz8CTR)

### Node Info
img1
img2
img2a
### Basic
img3
img3a
img3b
### Groups
img4
img4a
img4b
### Scenes
img5
img5a
img5b
img5c
img5d
img5e
### Other: Tuya specific
img6
### Other: OTAU
img7
img7a
img7b
### Other: Time
img8

BEskandari:
close #6944

### DIFF
--- a/devices/tuya/_TZE200_TS0601_water_valve.json
+++ b/devices/tuya/_TZE200_TS0601_water_valve.json
@@ -1,0 +1,192 @@
+{
+	"schema": "devcap1.schema.json",
+	"manufacturername": ["_TZE200_a7sghmms", "_TZE200_sh1btabb"],
+	"modelid": ["TS0601", "TS0601"],
+	"vendor": "GIEXperience",
+	"product": "Smart water valve",
+	"sleeper": false,
+	"status": "Silver",
+	"subdevices": [{
+			"type": "$TYPE_ON_OFF_OUTPUT",
+			"restapi": "/lights",
+			"uuid": [
+				"$address.ext",
+				"0x01"
+			],
+			"items": [{
+					"name": "attr/id"
+				},
+				{
+					"name": "attr/lastannounced"
+				},
+				{
+					"name": "attr/lastseen"
+				},
+				{
+					"name": "attr/manufacturername"
+				},
+				{
+					"name": "attr/modelid"
+				},
+				{
+					"name": "attr/name"
+				},
+				{
+					"name": "attr/swversion",
+					"parse": {"fn": "zcl", "ep": 1, "cl": "0x0000", "at": "0x0001", "script": "tuya_swversion.js"},
+					"read": {"fn": "zcl", "ep": 1, "cl": "0x0000", "at": "0x0001"}
+				},
+				{
+					"name": "attr/type"
+				},
+				{
+					"name": "attr/uniqueid"
+				},
+				{
+					"name": "state/on",
+					"refresh.interval": 300,
+					"read": {
+						"fn": "tuya"
+					},
+					"write": {
+						"dpid": 2,
+						"dt": "0x10",
+						"eval": "Item.val == 1 ? 1 : 0;",
+						"fn": "tuya"
+					},
+					"parse": {
+						"dpid": 2,
+						"eval": "Item.val = Attr.val;",
+						"fn": "tuya"
+					},
+					"default": 0
+				},
+				{
+					"name": "config/tuya_unlock"
+				},
+				{
+					"name": "state/reachable"
+				}
+			]
+		},
+		{
+			"type": "$TYPE_CONSUMPTION_SENSOR",
+			"restapi": "/sensors",
+			"uuid": [
+				"$address.ext",
+				"0x01",
+				"0x0702"
+			],
+			"items": [{
+					"name": "attr/id"
+				},
+				{
+					"name": "attr/lastannounced"
+				},
+				{
+					"name": "attr/lastseen"
+				},
+				{
+					"name": "attr/manufacturername"
+				},
+				{
+					"name": "attr/modelid"
+				},
+				{
+					"name": "attr/name"
+				},
+				{
+					"name": "attr/swversion",
+					"parse": {"fn": "zcl", "ep": 1, "cl": "0x0000", "at": "0x0001", "script": "tuya_swversion.js"},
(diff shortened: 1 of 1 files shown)


manup:
@SwoopX @ebaauw I'm not so sure here if the usage of `config/duration` `config/delay` is appropriate for this device type?

(besides the JSON needs to be reformatted to "standard" format)

SwoopX:
Well, it's a little bit mixed up for my personal taste and somehow a good example that even the best attempts for reasonable recycling of resource items can hit the limits quite quickly. Strickly speaking, items and device types available do not really fit the device at hand. Not sure if a 0x0702 consuption devices would be expected by anyone here 🤔  Tricky...

devices/tuya/_TZE200_TS0601_water_valve.json
```diff
@@ -0,0 +1,195 @@
...
+					"name": "config/battery",
+					"parse": {
+						"dpid": 108,
+						"eval": "Item.val = Attr.val;",
+						"fn": "tuya"
+					},
+					"default": 0
+				},
+				{
+					"name": "config/duration",
```
ebaauw, line 119: I don't think we have anything for last irrigation duration.  I don't think it should be on the `sensors` resource, but on the `lights` resource instead, where you handle the on/off state.  Also, it should be a `state` item, as it's not configurable.

devices/tuya/_TZE200_TS0601_water_valve.json
```diff
@@ -0,0 +1,195 @@
...
+					"description": "Last irrigation duration",
+					"parse": {
+						"dpid": 114,
+						"eval": "Item.val = Attr.val;",
+						"fn": "tuya"
+					}
+				},
+				{
+					"name": "state/consumption",
```
ebaauw, line 128: Using ZHAConsumption's for water consumption (rather than electricity) is new for me, but seems logical.  However, I don't think it's a good idea to abuse `state/consumption`.  Instead, we should be using `state/measured_value` and the corresponding `cap` items to indicate the substance (water), quantity (volume), and unit (litres or m3?), as well as min/max values.  
Typically `state/consumption` reports the life time consumption (since last factory reset), not the consumption during last run.

devices/tuya/_TZE200_TS0601_water_valve.json
```diff
@@ -0,0 +1,195 @@
...
+						"eval": "Attr.val = Item.val;",
+						"fn": "tuya"
+					},
+					"parse": {
+						"dpid": 104,
+						"eval": "Item.val = Attr.val ;",
+						"fn": "tuya"
+					}
+				}, {
+					"name": "config/mode",
```
ebaauw, line 178: `config/mode` (and similar like `config/devicemode` etc) is becoming a nightmare.  I think we really need a `cap` resource to list the possible modes (value and description?).

devices/tuya/_TZE200_TS0601_water_valve.json
```diff
@@ -0,0 +1,195 @@
...
+					"name": "config/on"
+				},
+				{
+					"name": "config/reachable"
+				},
+				{
+					"name": "state/lastupdated"
+				},
+				{
+					"name": "config/delay",
```
ebaauw, line 147: Using `config/delay` for a cycle interval doesn't seem right.  Closest we have resembling this is the schedule on thermostats, but that's way too elaborate for this.

devices/tuya/_TZE200_TS0601_water_valve.json
```diff
@@ -0,0 +1,195 @@
...
+						"fn": "tuya"
+					},
+					"parse": {
+						"dpid": 104,
+						"eval": "Item.val = Attr.val ;",
+						"fn": "tuya"
+					}
+				},
+				{
+					"name": "state/seconds_remaining",
```
ebaauw, line 163: We expose (writeonly) `state/ontime` on `lights` resources to set the target "duration" for the light to be on (this sends an _On with Timed Off_ command).  I think that could be used here, to set the irrigation duration, but it should be on the _On/Off Output_ `lights` resource.  
We don't (yet?) expose the _OnTime_ attribute (the currently remaining "duration", which counts down to 0s, before the light turns off).  The issue is that it's not reportable and there's no way to force the API to issue a _Read Attributes_ right now, so the value will always be outdated hopelessly (by minutes at least).  Unless the water valve sends push notifications for the remaining time, it doesn't make sense to expose `seconds_remaining`.  Also, it should be exposed on the `lights` resource, because that's where you control the on/off state.  Maybe expose the expected end time instead?  Although, we don't have anything for that either.

ebaauw (changes requested):
Hm, tricky.  Added quite a few comments in the code.

devices/tuya/_TZE200_TS0601_water_valve.json
BEskandari: The DDF was created through @BabaIsYou advice (see [#6944](https://github.com/dresden-elektronik/deconz-rest-plugin/issues/6944)). I was able to make the device working (on/off) and have added also some printscreen of eve app on the issue request.

ebaauw: I see I'm already quoted on that issue.

To get remaining duration in Eve (and even in Apple's Home app), expose the device as a _Valve_ accessory.  Homebridge deCONZ does expect `state/ontime` to support that, though.  Never tried to combine that with consumption, though.

BEskandari:
Waited too long! newer DDF are available.

PR #6949, opened 2023-05-05
DDF for Tuya TRV ( _TZE200_2ekuz3dz )

cagnulein:
following this https://github.com/dresden-elektronik/deconz-rest-plugin/issues/6269#issuecomment-1535207386


cagnulein:
Screenshot 2023-05-05 at 10 10 31

cagnulein:
Screenshot 2023-05-05 at 10 10 56

cagnulein:
@Smanar done!

cagnulein:
@Smanar one question: this thermostat is taking the time from Deconz but it always set a chinese timezone. Do you know if it's something fixable? @manup

cagnulein:
thanks for the quick response!

for example the deconz time is 2:34pm and the real time is 2:34pm so deconz is fine 

the thermostat shows 10:34pm so the minutes are fine but the hours is not, the offset is constant so it seems just the time zone 

so i guess the tuya sync doesn't work?

Smanar:
Long time I haven't check this code part.
From what I m reading your device use the tuya synchronisation request, but on your issue, your device have the Time cluster.
The Time cluster use values you can set on the GUI, when selectionning the Gateway (the blue) then the time cluster (in blue too).

For the tuya Time, I don't see settings, for it just use the host time, so don't use timezone.

You can play with them to see wich one your device is using, but you have this issue only on this TRV ? or don't have more ?

Smanar:
Ok this device have a bugged firmware > https://github.com/Koenkk/zigbee2mqtt/issues/14037#issuecomment-1271994929
Need to patch the result with 8 hours.
Problem this is not possible using DDF .....

cagnulein:
> You can play with them to see wich one your device is using, but you have this issue only on this TRV ? or don't have more ?

Ok Thanks i will try this, I have only this thermostat with a time

cagnulein:
> Ok this device have a bugged firmware > https://github.com/Koenkk/zigbee2mqtt/issues/14037#issuecomment-1271994929
> 
> Need to patch the result with 8 hours.
> 
> Problem this is not possible using DDF .....

Can't we subtract the time before sending to it?

Smanar:
>Can't we subtract the time before sending to it?

All is possible, the problem is "hack" are now rejeted in c++ code (I mean special code just for one device).
There is absolutely nothing in DDF core to do something on the tuya synchronisation, and I m almost sure adding a field "config/timezone" just for one bugged device will be refused too.


Making a hack just for you is easy, but you will need to recompile the lib, and again every time you will want to update deconz.

cagnulein:
Got it, I guess I can leave without the timing sync :) thanks anyway

PR #6957, opened 2023-05-07
Use Tuya unlock sequence for Moes 2-gang switch (_TZ3000_18ejxno0)

BabaIsYou:
1. It's not battery powered device but wired with or without neutral (changing name of the DDF file)
2. Newer version need tuya unlock sequence to toggle switch individualy.

Thx to @tosko in Discord to mention this.


manup:
Almost there, can you please replace tabs with two spaces in the JSON file.

BabaIsYou:
> Almost there, can you please replace tabs with two spaces in the JSON file.

Done ! ;-)

PR #6971, opened 2023-05-13
DDF for Tradfri bulb E14 WS Opal 600 lm

dimitripb:
DDF for Tradfri bulb E14 WS Opal 600 lm


manup:
Hi thanks heaps for all the work on new Ikea DDF files :+1: 

Can you please remove the "path" and "description" fields on the items. These are occasionally included by the DDF editor although they shouldn't, only if the description differs from what is already in the `generic/items`.

PR #6977, opened 2023-05-15
Add DDF + Button Map for _TZ3000_ja5osu5g Tuya Smart Button

Xyaren:
My Try at a DDF for a smart button I recently bought:

https://de.aliexpress.com/item/1005005238316199.html?spm=a2g0o.order_list.order_list_main.11.49a15c5fjVVmas&gatewayAdapt=glo2deu

This is a single button, battery powered (button cell).

The button has two modes, you can switch between them by 3xPressing the button very fast.

### Scenario Mode
image

Single Press
Double Press
Long Press

**I added the button events for this mode. They are working very reliable.**

### Remote Mode
image

_Not very much info yet._

Single Press to turn on, double to turn off.

When on: press and hold to dim.
It seems to increase, then decrease and cycle again.
I was not able to identify a "start" event for dimming, although a "LEVEL_CONTROL STOP" event was sent.

Single press when on: Seems to cycle between 4 Presets of Level & Color Control based on the received button events (When on) I have no dimmable light to test this using a direct binding.
Could also just be some kind of Tuya-Magic.
The button can be bound directly to a plug to control on/off (tested)

### Notes
Testing is quite challenging. When playing aroung with bindings, I managed to put the button in some kind of state, that prevented the mode switch by 3x pressing the button.
Even after re-pairing this state remained.
Only after I brute-force played around with the unbinding in the GUI I managed to revert it backt to its initial behaviour.

### DIFF
--- a/button_maps.json
+++ b/button_maps.json
@@ -1372,7 +1372,7 @@
         },
         "TuyaSmartButton": {
             "vendor": "Tuya",
-            "doc": "Smart Button with Scene Control (scene control not mapped)",
+            "doc": "Smart Button with two modes (remote mode not mapped)",
             "modelids": ["_TZ3000_ja5osu5g"],
             "map": [
                 [1, "0x01", "ONOFF", "0xfd", "0", "S_BUTTON_1", "S_BUTTON_ACTION_SHORT_RELEASED", "Short"],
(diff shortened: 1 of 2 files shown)


Xyaren:
In case this PR is not enough or is going to be abandoned for whatever reason I just created #6979 with all the details

PR #6989, opened 2023-05-20
Add DDF validator action

Zehir:


Zehir:
I just updated the validator 2.2.0, there is no need to change this PR to use the lastest version

PR #7005, opened 2023-05-23
DDF move NEO DDF to manufacturer folder

BabaIsYou:
Move NEO Immax NEO Smart controller DDF to /neo manufacturer subfolder


BabaIsYou:
> You really don't like refresh interval in DDFs 🤣

@manup I don’t like them if they are useless because the same or close to the one already present in generic item. It charges DDF for nothing ;-)

PR #7023, opened 2023-05-30
Update DDF for H2L-ZBPH-RS roller shutter

dimitripb:
This device was placed under manufacturer Philips. It's actually a Hue shutter which was branded H2L-ZBPH-RS, just like the model ID. It's from the Dutch webshop Home2Link. The DDF was not working (anymore) with manufacturer Philips.


manup:
Merging for now. I think we can also remove the file `philips/h2l-zbph-rs.json` and add the `$MF_PHILIPS` into this DDF to only have one DDF later on.

PR #7024, opened 2023-05-30
Delete h2l-zbph-rs.json

dimitripb:
This DDF should be removed. I replaced it with https://github.com/dresden-elektronik/deconz-rest-plugin/pull/7023/commits/c62ebcb45002a0feec7ddb4e2a45bb9fca5b0290 There was some confusing about the manufacturer. With manufacturer $MF_PHILIPS it does not work. It seems to be a rebranded Philips/Sunricher device to H2L (Home2Link). The new DDF is placed under its own subdirectory.


manup:
Ahh ok so there was no device with `$MF_PHILIPS` at all?
The other PR is now merged.

dimitripb:
Yes you are right. Things got messed up on my site. This seemed the best way to me to correct it.

PR #7034, opened 2023-06-01
DDF for Eva Powermeter

### LINKED ISSUE #6343: EVA Han Powermeter
## Device

- Product name: The device name as shown on the product or package.
- Manufacturer: EVA
- Model identifier: Meter REader
  - Sensor

## Screenshots

image
image
image
image
image
image

Another user posted a request for this device here
https://forum.phoscon.de/t/eva-metering-han-powermeeter/1941

Link to device:
https://shop.evasmart.no/produkt/smarthus/maleravleser

sevrinrobstad:
Fix #6343 .

### DIFF
--- a/devices/eva/powermeter.json
+++ b/devices/eva/powermeter.json
@@ -51,15 +51,14 @@
         },
         {
           "name": "state/current",
-          "refresh.interval": 30,
-          "default": 0
+          "refresh.interval": 360
         },
         {
           "name": "state/lastupdated"
         },
         {
           "name": "state/power",
-          "refresh.interval": 30,
+          "refresh.interval": 360,
           "read": {
             "at": "0x0304",
             "cl": "0x0b04",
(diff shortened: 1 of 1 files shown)


BabaIsYou:
Hi,
To avoid some isuues it's preferable that `refresh.interval` would be higher than `max` in binds, as a last resort if bind failed. 
And these values (same for  `default": 0`) are already into generic items for "power", "current" and "voltage" then useless into the DDF.

sevrinrobstad:
> Hi, To avoid some isuues it's preferable that `refresh.interval` would be higher than `max` in binds, as a last resort if bind failed. And these values (same for `default": 0`) are already into generic items for "power", "current" and "voltage" then useless into the DDF.

OK so just remove the refresh.interval and default:0 all together?

BabaIsYou:
Exactly, or just default and put a higher value for refresh like 360 for instance ;-)

sevrinrobstad:
Updated.

Issue #6756, opened 2023-02-18
Ikea Vindstyrka


vinyli85:
## Device

- Product name: Ikea Vindstyrka
- Manufacturer: Ikea of Sweden
- Model identifier: VINDSTYRKA
- Device type :
  - Sensor

### Basic
![basic](https://user-images.githubusercontent.com/45328686/219870433-70a06225-78e2-4a75-bc93-9819cfa5385c.jpg)

### Identify
![indentify](https://user-images.githubusercontent.com/45328686/219870455-7fb65686-7afd-429d-a5a1-40b8b24e39f7.jpg)

### Groups
![group1](https://user-images.githubusercontent.com/45328686/219870469-3a43f823-1d9e-4165-83c6-e43eee4cc847.jpg)
![group2](https://user-images.githubusercontent.com/45328686/219870472-c91d358d-ac7f-4f66-99df-d2144f1ad074.jpg)

### Temperature Measurement
![temp](https://user-images.githubusercontent.com/45328686/219870507-72410b35-8998-465d-a04e-7d55b1a2a457.jpg)

### Relative humidity measurement
![hum](https://user-images.githubusercontent.com/45328686/219870528-0b170dc6-ae6b-4eae-807b-6937d8f80ec1.jpg)

### Ikea specific
![ikea_spec](https://user-images.githubusercontent.com/45328686/219870550-346afd50-b487-4c3a-986d-55f7ce243bef.jpg)

### UNKNOWN (3 times)
Empty Cluster Info

ebaauw:
Where did you get this device?  I would love to get my hands on one, but It's only expected to come out earliest in April, here in the Netherlands.

Supporting the temperature and humidity sensors is rather straighforward, but we need to reverse engineer the custom clusters for air quality support.  For that we would have to pair the device to an IKEA gateway, and sniff the Zigbee traffic to see what commands the gateway sends on what app actions.  The Startvind uses cluster 0xFC7D for Air Purfier control and reports PM2.5 density on attribute 0x0004 on that cluster.  deCONZ computes the quality label ("good", "fair", ...) from that.  The Vindstyrka should report both PM2.5 density and TVOC level, if I read the documentation correctly.  I wonder if it reports the coloured label or if we need to compute that ourselves.  And if there's separate labels for PM2.5 vs TVOC.

vinyli85:
Hi Erik,

got mine today at an Ikea store in Germany. There were maybe 2-3 left. 
[...]

PR #7042, opened 2023-06-08
First draft of WWAHu cluster into ZCLDB

BabaIsYou:
First attempt to include "Work with all hubs" cluster (WWAHu).

Zigbee devices must have implemented with ‘Works With All hubs’ (WWAHu) cluster in order to be certified for and receive the Works with Alexa (WWA) badge. 

It seems it is currently implemented into  specific manufacturer cluster 0xfc57 attached to Amazon lab but should already been defined  into document 17-01067-022 from Zigbee Alliance.

But as I couldn't gain access to this document I'm quite sure that's not complete (especially regarding commands and their eventual payload).

Extrapolation from https://infocenter.nordicsemi.com/index.jsp?topic=%2Fsdk_tz_v4.2.0%2Fgroup___z_b___z_c_l___w_w_a_h.html

### DIFF
--- a/general.xml
+++ b/general.xml
@@ -6230,6 +6230,201 @@ Writing 1 to 0x0202 and 0x0302 makes 0x0300 writable. Then writing 1 to 0x0300 s
       <client>
       </client>
     </cluster>
+   		<!-- Amazon Lab -->
+    <cluster id="0xfc57" name="Amazon lab WWAHu Specific cluster" mfcode="0x1217">
+			<description>This cluster provides an opportunity to use various features of the Work With All Hubs cluster.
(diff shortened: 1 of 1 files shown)


manup:
Do we have devices actually using this?
I know ubisys writes about it but I can't find that cluster on a ubisys C4 with recent version, but there is a 0xFC02 client cluster.

BabaIsYou:
Found this looking for cluster FC57 used by IKEA (#6756) may be it’s the same. Perhaps @ebaauw has some hint about this.

ebaauw:
Dunno what that cluster is for.  The 0xFCxx clusters are manufacturer specific, so I would think that the Zigbee alliance wouldn’t define these.

manup:
It's a Amazon specific cluster, I had a look at it a while ago to be frank I don't know why we'd need it / it doesn't seem to provide useful functionality for us.

Issue #7045, opened 2023-06-10
Iluminize 5128.10


AndreKoepke:
## Device

- Product name: Zigbee 3.0 roller shutter actuator
- Manufacturer: iluminize
- Model identifier: 5128.10
- Device type : 
  - Other: roller shutter actuator

# Special
This device appears to be a normal dimmer for lights. When the dimmer changed to ...
* 0-30%: roller shutter will close
* 30-70%: roller shutter will stop
* 70-100%: roller shutter will open

In deCONZ-gui, it also appears as "window covering" and all functions works like a charm.

## Screenshots

<img width="241" alt="image" src="https://github.com/dresden-elektronik/deconz-rest-plugin/assets/15613049/d64f577c-5db1-470e-8420-0826b5edccee">

### Basic
<img width="511" alt="image" src="https://github.com/dresden-elektronik/deconz-rest-plugin/assets/15613049/fb93044e-2e6a-45fa-b0ef-079dce36cb82">

# Clusterinfos

### Basic (0x0000)
<img width="510" alt="image" src="https://github.com/dresden-elektronik/deconz-rest-plugin/assets/15613049/0c2dbe28-d95a-41c0-81d9-50e61536506d">

###  Identify (0x0003)
<img width="513" alt="image" src="https://github.com/dresden-elektronik/deconz-rest-plugin/assets/15613049/6cbe1c7b-eebc-471e-8768-6eef8d1e8f3d">

### Groups (0x0004)
<img width="499" alt="image" src="https://github.com/dresden-elektronik/deconz-rest-plugin/assets/15613049/9decb0a4-abb6-4a9c-a2bc-a70f07ad45f1">

### On/Off (0x0006)
<img width="497" alt="image" src="https://github.com/dresden-elektronik/deconz-rest-plugin/assets/15613049/afc9f2e4-4c6b-4156-a45e-7df189221a3c">

### Level control (0x0008)
<img width="493" alt="image" src="https://github.com/dresden-elektronik/deconz-rest-plugin/assets/15613049/4342ab64-24a7-441b-9083-a51eae9860b7">

### Window covering
<img width="509" alt="image" src="https://github.com/dresden-elektronik/deconz-rest-plugin/assets/15613049/b928df8e-f6ec-4267-afdf-cb9b35329461">

### Diagnostics (0x0b05)
[...]

PR #7046, opened 2023-06-10
DDF for iluminize motion sensor (5128.10)

AndreKoepke:
See #7045


AndreKoepke:
I tried it with in my local home. This ddf seems to be working (setting lift/tilt works without any problems).

It was the first time, that I wrote a ddf. I just copied it and changed the adresses.

BabaIsYou:
Did not really understand binding on cluster "0x0102" "`cl": "0x0102"` on endpoint 6  `"src.ep": 6`, because device descibes on referenced issue doesn't have endpoint 6

AndreKoepke:
@BabaIsYou  
I am very new to DDF and I guessed the values. But from your comment I learned that `"cl"` stands for `cluster`. :D  
Feel free to edit. I can test it with my devices.

BabaIsYou:
Think you should have a look to https://github.com/dresden-elektronik/deconz-rest-plugin/wiki/DDF-cheat-sheet then would understand what I meant ;-)

AndreKoepke:
I will read it and fix the JSON. ^^  
It turns out that after a while, that the value-report don't works as expected. :/

BabaIsYou:
As the device is quite simple, perhaps it would be better to start from scratch using DDF editor and compare the final DDF with the one you did ;-)

AndreKoepke:
:D  
I tried to write the plain JSON while there is a DDF editor. omg. 

I fix the JSON and I will change it here, when the tests are successful and my slow train-vpn-forwarded-x11-ssh-connection allows my to save.

BabaIsYou:
I did the changes from EP 6 to EP 1 into the DDF but you have to test before merging and close #7045

AndreKoepke:
Ok. I dropped the settings via DDF-editor and it from the default-values. Maybe the endpoint came from there.

On the first view, it seems to be working. I will report again, when I'm back home.  
Thanks for the change. :)

AndreKoepke:
Seems good. ready4merge from my site.

manup:
@BabaIsYou all good to merge?

PR #7073, opened 2023-07-02
Bug fix: fp1 triggerdistance not updated from special atttribute

ebaauw:
Typo (wrong tag) in handling `config/triggerdistance` for Aqara FP1 from Xioami special attribute, resulting in errors "failed to evaluate expression for 54:ef:44:10:00:4d:7b:c1-01-0406/config/triggerdistance_bis: R.item('config/triggerdistance').val = ['far', 'medium', 'near'][Attr.val], err: TypeError: failed to set Item.val"


ebaauw:
I think there's something wrong with the check: it's checking (all?) 498 DDF files, and finding errors in 14, none of which are touched by this PR.

ebaauw:
I forgot to create a new branch for this PR - will create a new PR with the change.

Issue #7061, opened 2023-06-26
Xiaomi Aquara Water leak sensor (SJCGQ11LM) falsely reports trouble on deConz 2.22.2 (HA addon)


bennydiamond:
## Describe the bug
I'm running deConz HA addon.
I updated addon to version 6.20.0 (which includes deConz 2.22.2). One of my water leak sensor SJCGQ11LM started reporting trouble right away.

I tried to delete and re-pair the device, device will behave normally for a couple of seconds and then will fall back to reporting trouble.

I tried replacing the battery in the device, same issue.

I downgraded deConz HA addon to previous version (6.19.0). The problem went away. SJCGQ11LM sensor behaves as intended.

Only 1 of my 5 * SJCGQ11LM sensor exhibited the problem but I suspect this is due to the fact the others might have been sleeping and thus did not report to deConz while I was witnessing the issue.

## Steps to reproduce the behavior

Pair "SJCGQ11LM" with deConz HA addon 6.20.0
Once paired, press the button on SJCGQ11LM sensor to trigger sending its attributes
Device will report trouble/wet and will stay in that state indefinitely

## Expected behavior

 If no water is physically touching the sensor, the sensor should **not** report trouble/wet.

## Environment

 - Host system: USFF PC, HP ProDesk Mini G3 400, running Debian 12.
 - Running method: Home Assistant deCONZ Add-on (6.20.0)
 - Firmware version: 26400500
 - deCONZ version: (2.22.2)
 - Device: ConBee I
 - Do you use an USB extension cable: yes
 - Is there any other USB or serial devices connected to the host system? Aeotec Z-Wave stick Gen5, 2 * RTL-SDR receiver dongles

## Additional context

I reverted to HA Addon 6.19.0 and did not save any logs. If necessary, I could update again to the faulty version and generate logs.

I highly suspect the issue originates from this change: https://github.com/dresden-elektronik/deconz-rest-plugin/pull/6919

Mimiix:
@SwoopX can you check?

Larry0ua:
Same here. 3 of 4 aquara water sensors started reporting water shortly after update to HA addon 6.20.0 yesterday (deCONZ 2.21.2->2.22.2) running with ConBee II

If additional logs are necessary I can extract them

ebaauw:
[...]

PR #7091, opened 2023-07-07
Bug fix: false water detection for lumi_sensor_wleak_aq1

ebaauw:
Remove `state/water_bis` as Xiaomi special attribute doesn't always reflect the water detected state, see #7061

### DIFF
--- a/devices/xiaomi/lumi_sensor_wleak_aq1.json
+++ b/devices/xiaomi/lumi_sensor_wleak_aq1.json
@@ -100,17 +100,6 @@
             "fn": "ias:zonestatus",
             "mask": "alarm1"
           }
-        },
-        {
-          "name": "state/water_bis",
-          "awake": true,
-          "parse": {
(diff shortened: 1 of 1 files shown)


manup:
Interesting approach :) if it works we should give it a try.

manup (approved):

ebaauw:
> Interesting approach

I suppose that's as good a description as any ;-)

Been running this on my test environment for two days now, and `state/lastupdated` is updated as intended.

dxa13:
Hi Erik @ebaauw , I tried using your fix with the 2.23.0 beta branch, and my one leak sensor with firmware 0.0.0_0004 (all other 9 I have have firmware 6) is still reporting a false leak (from issue #7061). Is that expected behavior after the fix, or is something just remembering the last state? Is there a way to force the false leak go away or be cleared (i.e. update the state) without physical interaction with the sensor (i.e. shorting the leads, or having to re-pair it)?

PR #7119, opened 2023-07-20
Add generic measured_value items

manup:
Related to https://github.com/dresden-elektronik/deconz-rest-plugin/pull/6673#issuecomment-1635627969

The main item `state/measured_value` provides a generic way to expose a measurement as floating point value e.g a temperature. The related `cap/measured_value/*` items describe the details — unit, min and max — about the value.

`state/measured_value`
`cap/measured_value/min`
`cap/measured_value/max`
`cap/measured_value/unit`

### DIFF
--- a/devices/generic/items/cap_measured_value_max.json
+++ b/devices/generic/items/cap_measured_value_max.json
@@ -1,0 +1,9 @@
+{
+  "schema": "resourceitem1.schema.json",
+  "id": "cap/measured_value/max",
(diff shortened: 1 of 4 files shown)


manup:
@Zehir can you please add the `Double` value type to the validator?

Zehir:
> @Zehir can you please add the `Double` value type to the validator?

Done, you should be able to re run the github action

PR #7131, opened 2023-07-26
Fix macOS opening of button maps file

aryelevin:
This resolves the macOS button_maps.json file location on the app bundle


manup:
I've just checked the compiled bundle, no need for change the button_maps.json file is already copied to the correct location, due a higher level CMakeLists.txt.

aryelevin:
@manup This is not about the button_maps.json file itself, just the plugin's ability to locate the file, without this modification, the file isn't loaded and all the node's clusters are just Unknown (0) state.
I'm using the macOS version now for my main setup, so I can be somewhat a tester for you if you want.

manup:
Ah damn missed it, yes that's correct thanks.

PR #7133, opened 2023-07-28
DDF for OWON AC201 thermostat

aryelevin:
This is an integration of OWON AC201 AC IR  Blaster over def including the removal of legacy c code.
I've added the ability to set the IR commands set from the API using the config/pulseconfiguration setting.
I have 3 units of this product, I tested it for full functionality.


Zehir:
Hello @aryelevin I made the DDF validator and I have some errors reported by this PR, could you take a look at it ?

https://github.com/dresden-elektronik/deconz-rest-plugin/actions/runs/5688976117/job/15419686665#step:3:19

They are all about the item `config/pulseconfiguration`.
```
devices/owon/ac201.json:Validation error: 
```
The value is "00" but hexadecimal values are expected, it's should be "0x00"
```
Invalid hexadecimal value at "subdevices[0].items[15].read.cmd";
Invalid hexadecimal value at "subdevices[0].items[15].parse.cmd";
```

I have no idea what is the `fc`property, it's value is 5 what this does ? Also the eval property is an empty string is that valid ?
```
Unrecognized key(s) in object: 'fc' at "subdevices[0].items[15].read";
```

The item is missing the data type should be an hexadecimal value with 2 numbers for example "0x00"
```
Required at "subdevices[0].items[15].write.dt"
```

aryelevin:
Hi @Zehir ,

Thanks for your comment first.
OK, so, for items[15], please remember, its a cluster command item (zcl:cmd), not an attribute (zcl:attr/zcl[deprecated]), thus:
1. "fc" property is frame control of the zcl cmd, you can look at "device_access_fn.cpp" file to see more.
2. The eval of the read is empty since its a raw data input that set to the config.
3. The hex vs dec data strings: I will verify the functionality of the configuration after fixing, and will update.
4. The missing data type is just irrelevant to this type of item (zcl:cmd).

Sorry for my late reply, you know, post July-August... lol

aryelevin:
I checked, works perfect with hex commands.
@ebaauw already committed and merged this fix, so no need for anything right now.

Zehir:
> I checked, works perfect with hex commands.
> @ebaauw already committed and merged this fix, so no need for anything right now.

Yes Ebaauw fixed a lot à DDF recently but thanks for your reply ;)

PR #7155, opened 2023-08-06
Add items_optional to subdevice generics

Zehir:


Zehir:
There is a subjet about the Door lock, how do we clean them up ?

We have constants ;
```json
{
        "$TYPE_DOOR_LOCK_CONTROLLER": "Door lock controller",
        "$TYPE_DOOR_LOCK": "Door Lock",
}
```
But we have `door_lock.json` that don't match ;
```json
{
    "schema": "subdevice1.schema.json",
    "type": "ZHADoorLock",
    "name": "Door Lock"
}
```
It's should be :
```json
{
    "schema": "subdevice1.schema.json",
    "type": "$TYPE_DOOR_LOCK",
    "name": "Door Lock"
```
But the name either `Door Lock` or `ZHADoorLock`.

Currently the `danalock_v3.json` is using `$TYPE_DOOR_LOCK` so `Door Lock`.
But in the C code there is reference to `ZHADoorLock` and `Door Lock`.

We could split up the types to `$TYPE_DOOR_LOCK` and `$TYPE_DOOR_LOCK_ZHA` ?

Zehir:
It's also missing the resource item for `cap/measured_value/dimension` who is used in some subdevices `ZHACarbonDioxide`, `ZHAFormaldehyde` and `ZHAParticulateMatter`

Zehir:
There is also #7146 who fix errors reported by the validator

manup:
Uff large PR :see_no_evil: sorry but this needs to be split up in smaller chunks, one PR per sub-device. That way it can be reviewed by others developers / clients with one discussion per device. Which items are are optional or mandatory should have proper review, personally for many device types I can't tell but devs which added these sure have an opinion.

For the `cap/measured_value/dimension` we still need to figure out how to name it see https://github.com/dresden-elektronik/deconz-rest-plugin/pull/7146

Imho items can only be in either `"items"` aka mandatory or in `"items_optional"` but not both.

Zehir:
Sure I can make multiple PR

About the `items_optional` I made it that way to be able to still display it in the DDF editor as he will not handle this new key

Zehir:
@manup I splited up this PR in many PR, see above list

PR #7165, opened 2023-08-12
[subdevice] Add type ZHAAncillaryControl

Zehir:
Part 2 of #7155


BabaIsYou:
IMO some items could be optional because we coud imagine that some ACE could be USB powered or don't have a tampered switch. I mean : 
        "config/battery",
        "state/lowbattery",
        "state/tampered"

PR #7166, opened 2023-08-12
[subdevice] Add type Level control switch

Zehir:
Part 3 of #7155


ebaauw:
I think this is a `lights` resource?

manup:
Oh yes indeed, needs to be changed

Zehir:
> Oh yes indeed, needs to be changed

I do an other PR ?

PR #7168, opened 2023-08-12
[subdevice] items_optional for ZHAAirQuality

Zehir:
Part 5 of #7155


ebaauw (approved):
For now, let's go ahead.  In future, we might need to revisit.

I think maybe `state/airquality` should be mandatory.

VOC sensors have `state/airqualityppb`; PM2.5 sensors have `state/pm2_5`; so both are optional.

PR #7170, opened 2023-08-12
[subdevice] items_optional for Color dimmable light

Zehir:
Part 7 of #7155


Zehir:
@ebaauw suggested that this type should be removed

ebaauw:
Because it's used for lights with only colour temperature, only colour, and both colour and colour temperature.  This is the old (pre-Zigbee 3) ZHA type.  Most Zigbee-3 lights use the ZLL types: _Color Temperature Light_, _Color Light_, _Extended Color Light_.  _Color Dimmable Light_ should be mapped to one of these three, depending on the colour capabilities it supports.  Note that the legacy code did the mapping for some whitelisted lights, but not for all.

ebaauw:
Changed the DDFs of the three _Color Dimmable Light_ in https://github.com/dresden-elektronik/deconz-rest-plugin/pull/7221, so we could simply delete this type.

Zehir:
I updated the PR to remove the json references of the `Color Dimmable Light` but there is still some references to it in the C code

ebaauw (approved):

PR #7171, opened 2023-08-12
[subdevice] for Color light

Zehir:
Part 8 of https://github.com/dresden-elektronik/deconz-rest-plugin/pull/7155

### DIFF
--- a/devices/generic/subdevices/color_light.json
+++ b/devices/generic/subdevices/color_light.json
@@ -1,0 +1,34 @@
+{
+    "schema": "subdevice1.schema.json",
+    "type": "$TYPE_COLOR_LIGHT",
+    "name": "Color light",
+    "restapi": "/lights",
+    "order": 10,
+    "uuid": ["$address.ext", "0x01"],
+    "items": [
+        "cap/color/xy/blue_x",
+        "cap/color/xy/blue_y",
+        "cap/color/xy/green_x",
+        "cap/color/xy/green_y",
+        "cap/color/xy/red_x",
+        "cap/color/xy/red_y",
+        "state/alert",
+        "state/bri",
+        "state/colormode",
+        "state/on",
+        "state/reachable",
+        "state/x",
+        "state/y"
+    ],
+    "items_optional":[
+        "config/on/startup",
+        "config/bri/startup",
+        "config/color/ct/startup",
+        "cap/alert/trigger_effect",
+        "cap/bri/move_with_onoff",
+        "config/bri/execute_if_off",
+        "config/color/execute_if_off",
+        "cap/color/capabilities",
+        "cap/on/off_with_effect"
+    ]
+}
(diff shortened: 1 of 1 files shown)


devices/generic/subdevices/color_light.json
```diff
@@ -0,0 +1,36 @@
...
+        "cap/color/xy/green_x",
+        "cap/color/xy/green_y",
+        "cap/color/xy/red_x",
+        "cap/color/xy/red_y",
+        "state/alert",
+        "state/bri",
+        "state/colormode",
+        "state/on",
+        "state/reachable",
+        "state/x",
```
ebaauw, line 20: I do think we might have a couple of lights that don't support all colour capabilities.  Best make the following items optional:
- `state/x`, `state/y` are mandatory only when `cap/color/capabilities` includes `xy`;
- `state/hue`, `state/sat` are mandatory only when `cap/color/capabilities` include `hs`;
- `state/effect` is mandatory only when `cap/color/capabilities` includes `effect`.

Note that either `state/x` and `state/y` or `state/hue` and `state/sat` are mandatory (or the light would be a _Dimmable Light_), but that might be too complex for validator.

ebaauw (changes requested):

devices/generic/subdevices/color_light.json
```diff
@@ -0,0 +1,36 @@
...
+        "state/y"
+    ],
+    "items_optional":[
+        "cap/alert/trigger_effect",
+        "cap/bri/move_with_onoff",
+        "cap/color/capabilities",
+        "cap/groups_not_supported",
```
ebaauw, line 27: `cap/groups/not_supported`

devices/generic/subdevices/color_light.json
```diff
@@ -0,0 +1,36 @@
...
+        "state/colormode",
+        "state/on",
+        "state/reachable",
+        "state/x",
+        "state/y"
+    ],
+    "items_optional":[
+        "cap/alert/trigger_effect",
+        "cap/bri/move_with_onoff",
+        "cap/color/capabilities",
```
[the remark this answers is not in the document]
Zehir: Probably added it to optional because I got a light that did not have it, we will see what DDF we need to update later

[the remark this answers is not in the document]
Zehir: Moved back to items list

devices/generic/subdevices/color_light.json
Zehir: Fixed

devices/generic/subdevices/color_light.json
Zehir: This case is now handle from the validator rule

ebaauw (approved):
Still many more optional items, but I guess that's for later.

PR #7173, opened 2023-08-12
[subdevice] items_optional for Dimmable light

Zehir:
Part 10 of https://github.com/dresden-elektronik/deconz-rest-plugin/pull/7155

### DIFF
--- a/devices/generic/subdevices/dimmable_light.json
+++ b/devices/generic/subdevices/dimmable_light.json
@@ -6,14 +6,19 @@
     "order": 11,
     "uuid": ["$address.ext", "0x01"],
     "items": [
-        "cap/alert/trigger_effect",
-        "cap/bri/move_with_onoff",
-        "cap/on/off_with_effect",
-        "config/bri/execute_if_off",
-        "config/bri/startup",
-        "config/on/startup",
         "state/alert",
         "state/bri",
         "state/on"
+    ],
+    "items_optional":[
+        "config/on/startup",
+        "config/bri/startup",
+        "config/color/ct/startup",
+        "cap/alert/trigger_effect",
+        "cap/bri/move_with_onoff",
+        "config/bri/execute_if_off",
+        "config/color/execute_if_off",
+        "cap/color/capabilities",
+        "cap/on/off_with_effect"
     ]
 }
--- a/devices/generic/subdevices/dimmable_light.json
+++ b/devices/generic/subdevices/dimmable_light.json
@@ -11,14 +11,16 @@
         "state/on"
     ],
     "items_optional":[
-        "config/on/startup",
-        "config/bri/startup",
-        "config/color/ct/startup",
         "cap/alert/trigger_effect",
         "cap/bri/move_with_onoff",
-        "config/bri/execute_if_off",
-        "config/color/execute_if_off",
         "cap/color/capabilities",
-        "cap/on/off_with_effect"
+        "cap/groups_not_supported",
+        "cap/on/off_with_effect",
+        "cap/transition_block",
+        "config/bri/execute_if_off",
+        "config/bri/startup",
+        "config/color/ct/startup",
+        "config/color/execute_if_off",
+        "config/on/startup"
     ]
 }
--- a/devices/generic/subdevices/dimmable_light.json
+++ b/devices/generic/subdevices/dimmable_light.json
@@ -8,19 +8,16 @@
     "items": [
         "state/alert",
         "state/bri",
-        "state/on"
+        "state/on",
+        "state/reachable"
     ],
     "items_optional":[
         "cap/alert/trigger_effect",
         "cap/bri/move_with_onoff",
-        "cap/color/capabilities",
-        "cap/groups_not_supported",
+        "cap/groups/not_supported",
         "cap/on/off_with_effect",
-        "cap/transition_block",
         "config/bri/execute_if_off",
         "config/bri/startup",
-        "config/color/ct/startup",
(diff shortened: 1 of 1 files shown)


manup:
Ha interesting, @ebaauw if we have a light which doesn't support `cap/on/off_with_effect` is it better to not have the item and implicitly treat it as `false`, or have it and explicitly set it to `false`?

ebaauw:
My idea was to include `capabilities` items only when the light supports the corresponding capability, similar to only including supported `config` items.  Not sure if that’s best, but it seems simpler and maybe cleaner to me.  On the other hand, making these mandatory would force people to think about them when creating a DDF.

Note that the `color` `capabilities` and `config` items don’t apply to a dimmable light, and shouldn’t even be optional.  DDF should overwrite the device type (to color temperature light or (extended) color light) when the device does support these.  Also the old ZHA color dimmable light should not be used in a DDF, in favour of (extended) color light.

Zehir:
Maybe we should add a third array with a list of forbidden attributes ?

ebaauw:
> Maybe we should add a third array with a list of forbidden attributes ?

That’s redundant if you have optional attributes.

Zehir:
> 

Or maybe a rule that say if there is the attribute `capabilities/color` you should not use type `$TYPE_DIMMABLE_LIGHT`

ebaauw:
Huh?  Either we specify mandatory and optional items, implying the rest are not allowed, or we specify mandatory and forbidden item, implying the rest are optional.  I don't see why we need to specify mandatory, optional, and forbidden.

Either we specify the items per device type, or the device types per item.  I don't see why we want to specify mandatory items per device type, but forbidden devices types per item.

devices/generic/subdevices/dimmable_light.json
```diff
@@ -6,14 +6,21 @@
     "order": 11,
     "uuid": ["$address.ext", "0x01"],
     "items": [
```
ebaauw, line 8: `state/reachable` should be mandatory.

devices/generic/subdevices/dimmable_light.json
```diff
@@ -6,14 +6,21 @@
...
+        "state/alert",
+        "state/bri",
+        "state/on"
+    ],
+    "items_optional":[
         "cap/alert/trigger_effect",
         "cap/bri/move_with_onoff",
+        "cap/color/capabilities",
```
ebaauw, line 16: Not for _Dimmable Light_.

devices/generic/subdevices/dimmable_light.json
```diff
@@ -6,14 +6,21 @@
...
+        "cap/groups_not_supported",
         "cap/on/off_with_effect",
+        "cap/transition_block",
         "config/bri/execute_if_off",
         "config/bri/startup",
-        "config/on/startup",
-        "state/alert",
-        "state/bri",
-        "state/on"
+        "config/color/ct/startup",
```
ebaauw, line 22: Not for _Dimmable Light_.

devices/generic/subdevices/dimmable_light.json
```diff
@@ -6,14 +6,21 @@
...
+        "config/color/execute_if_off",
```
ebaauw, line 23: Not for _Dimmable Light_.

devices/generic/subdevices/dimmable_light.json
```diff
@@ -6,14 +6,21 @@
...
+        "state/bri",
+        "state/on"
+    ],
+    "items_optional":[
         "cap/alert/trigger_effect",
         "cap/bri/move_with_onoff",
+        "cap/color/capabilities",
+        "cap/groups_not_supported",
         "cap/on/off_with_effect",
+        "cap/transition_block",
```
ebaauw, line 19: Not for _Dimmable Light_.

ebaauw (changes requested):

devices/generic/subdevices/dimmable_light.json
```diff
@@ -6,14 +6,21 @@
...
     "items": [
+        "state/alert",
+        "state/bri",
+        "state/on"
+    ],
+    "items_optional":[
         "cap/alert/trigger_effect",
         "cap/bri/move_with_onoff",
+        "cap/color/capabilities",
+        "cap/groups_not_supported",
```
ebaauw, line 17: `cap/groups/not_supported`

ebaauw (changes requested):

devices/generic/subdevices/dimmable_light.json
Zehir: Done

devices/generic/subdevices/dimmable_light.json
Zehir: Removed

devices/generic/subdevices/dimmable_light.json
Zehir: Removed

devices/generic/subdevices/dimmable_light.json
Zehir: Removed

devices/generic/subdevices/dimmable_light.json
Zehir: Removed

devices/generic/subdevices/dimmable_light.json
Zehir: Fixed

ebaauw (approved):

PR #7176, opened 2023-08-12
[subdevice] items_optional for Extended color light

Zehir:
Part 13 of https://github.com/dresden-elektronik/deconz-rest-plugin/pull/7155

### DIFF
--- a/devices/generic/subdevices/extended_color_light.json
+++ b/devices/generic/subdevices/extended_color_light.json
@@ -7,16 +7,17 @@
     "uuid": ["$address.ext", "0x01"],
     "items": [
         "cap/color/capabilities",
+        "cap/color/ct/max",
+        "cap/color/ct/min",
         "state/alert",
         "state/colormode",
+        "state/ct",
         "state/on",
         "state/reachable"
     ],
     "items_optional":[
         "cap/alert/trigger_effect",
         "cap/bri/move_with_onoff",
-        "cap/color/ct/max",
-        "cap/color/ct/min",
         "cap/color/xy/blue_x",
         "cap/color/xy/blue_y",
         "cap/color/xy/green_x",
@@ -32,7 +33,6 @@
         "config/color/execute_if_off",
         "config/on/startup",
         "state/bri",
-        "state/ct",
         "state/effect",
         "state/hue",
         "state/sat",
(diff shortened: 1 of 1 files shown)


ebaauw (changes requested):

I do think we might have a couple of lights that don't support all colour capabilities.  Best make the following items optional:
- `state/x`, `state/y` are mandatory only when `cap/color/capabilities` includes `xy`;
- `state/hue`, `state/sat` are mandatory only when `cap/color/capabilities` include `hs`;
- `state/effect` is mandatory only when `cap/color/capabilities` includes `effect`.

Note that either `state/x` and `state/y` or `state/hue` and `state/sat` are mandatory (or the light would be a _Color Temperature Light_), but that might be too complex for validator.

devices/generic/subdevices/extended_color_light.json
```diff
@@ -6,8 +6,6 @@
     "order": 10,
     "uuid": ["$address.ext", "0x01"],
     "items": [
```
ebaauw, line 8: `state/reachable` should be mandatory.

Zehir: `state/reachable` is already in the items list

devices/generic/subdevices/extended_color_light.json
```diff
@@ -6,9 +6,15 @@
...
+        "state/alert",
+        "state/colormode",
+        "state/on",
+        "state/reachable"
+    ],
+    "items_optional":[
         "cap/alert/trigger_effect",
         "cap/bri/move_with_onoff",
-        "cap/color/capabilities",
         "cap/color/ct/max",
```
ebaauw, line 18: `cap/color/ct/max` and `cap/color/ct/min` are mandatory, as _Extended Color Light_ also has mandatory `state/ct`.

devices/generic/subdevices/extended_color_light.json
```diff
@@ -17,20 +23,18 @@
...
+        "cap/transition_block",
         "config/bri/execute_if_off",
         "config/bri/startup",
         "config/color/ct/startup",
         "config/color/execute_if_off",
         "config/on/startup",
-        "state/alert",
         "state/bri",
-        "state/colormode",
         "state/ct",
```
ebaauw, line 35: `state/ct` is mandatory (or it would be a _Color Light_).

ebaauw (changes requested):

ebaauw (approved):

PR #7178, opened 2023-08-12
[subdevice] items_optional for ZHAOpenClose

Zehir:
Part 15 of https://github.com/dresden-elektronik/deconz-rest-plugin/pull/7155

### DIFF
--- a/devices/generic/subdevices/openclose_sensor.json
+++ b/devices/generic/subdevices/openclose_sensor.json
@@ -6,9 +6,11 @@
     "order": 20,
     "uuid": ["$address.ext", "0x01", "0x0500"],
     "items": [
-        "config/on",
         "config/reachable",
         "state/open",
         "state/lastupdated"
+    ],
+    "items_optional": [
+        "config/on"
     ]
 }


manup:
From my understanding the `config/on` is mandatory on all sensor types. It seems to be only used for enabling sensors in rules processing.

Zehir:
> From my understanding the `config/on` is mandatory on all sensor types. It seems to be only used for enabling sensors in rules processing.

Then this one is having an issue, it's the only one that dont have the `config/on`
https://github.com/dresden-elektronik/deconz-rest-plugin/blob/dd4bb528ff1061dd4adbc8bae931a4bdfa6ae8e0/devices/tuya/_TZE200_htnnfasr_water_valve.json#L69-L161

manup:
Indeed imho `config/on` should be added to the device.

Zehir:
> Indeed imho `config/on` should be added to the device.

Then we can just close this PR, I will let somebody else update the DDF

ebaauw (changes requested):
`config/on` is mandatory for all `sensors` resources.  It's not mapped to any Zigbee attribute.  The REST API uses it to enable/disable rules triggered by this sensor.

Zehir:
> 

ok then no need to merge this PR because there is no changes

PR #7181, opened 2023-08-12
[subdevice] items_optional for ZHAThermostat

Zehir:
Part 18 of https://github.com/dresden-elektronik/deconz-rest-plugin/pull/7155

### DIFF
--- a/devices/generic/subdevices/thermostat.json
+++ b/devices/generic/subdevices/thermostat.json
@@ -6,14 +6,16 @@
     "order": 20,
     "uuid": ["$address.ext", "0x01", "0x0201"],
     "items": [
+        "config/on",
+        "config/reachable",
+        "state/lastupdated"
+    ],
+    "items_optional": [
         "config/heatsetpoint",
         "config/mode",
         "config/offset",
-        "config/on",
-        "config/reachable",
         "state/on",
         "state/temperature",
-        "state/valve",
-        "state/lastupdated"
+        "state/valve"
     ]
 }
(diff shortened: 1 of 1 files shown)


manup:
Are our thermostats really that terrible that we don't really have a meaningful mandatory item like `config/heatsetpoint` or `config/temperature`? :)

Zehir:
> 

Well, this one is missing the `config/heatsetpoint`
https://github.com/dresden-elektronik/deconz-rest-plugin/blob/dd4bb528ff1061dd4adbc8bae931a4bdfa6ae8e0/devices/legrand/cable_outlet.json#L9-L68
But having a `$TYPE_THERMOSTAT` for Cable Outlet ?

And `config/temperature` there is no more errors reported for this one, probably fixed since the PR was open.

ebaauw:
To expose a thermostat to HomeKit, I need current and target temperature, as well as current and target heating cooling mode.  Until now we’ve used for these: `state/temperature`, `config/heatsetpoint`, `state/on` and `config/mode`.  Not too sure about the mode, but a heating device without target temperature is no thermostat.  In HomeKit target mode if off, heat, cool, auto (between heat and cool); and current mode is off, heating, cooling.  I think `config/mode` might also have been (ab)used for manual vs schedule, but that should have been replaced by `schedule_on` (which is missing from optional items, as is the `schedule` itself).  The cursed Eurotronic Spirit has a mode where you can set target valve position, but we never exposed that over the API.

Note read-only `state` for current vs  read/write `config` for target (or write-only `action` for groups).  `config/battery` should be deprecated in favour of a separate ZHABattery subdevices, and `config/temperature` (for Xiaomi device temperature) should probably never have been exposed, or as separate ZHATemperature subdevices for “substance” device (instead of air or water).

ebaauw (approved):

Issue #7147, opened 2023-08-04
IKEA RODRET Kabelloser Dimmer/Schalter


TheGreatNemo:
## Device
- Product name: RODRET Kabelloser Dimmer/Schalter - E2201
- Manufacturer: IKEA of SWEDEN
- Model identifier: RODRET Dimmer
- Device type :
  - Switch

### Basic
![image](https://github.com/dresden-elektronik/deconz-rest-plugin/assets/62012121/42db0208-9478-48d9-90ea-c281a9048af5)
![image](https://github.com/dresden-elektronik/deconz-rest-plugin/assets/62012121/abf79d4c-a026-4003-ba3c-fd3728d5e46e)

### Power Configuration
![image](https://github.com/dresden-elektronik/deconz-rest-plugin/assets/62012121/677f8beb-4dd3-45d3-b15e-7d334b923586)
![image](https://github.com/dresden-elektronik/deconz-rest-plugin/assets/62012121/a96f765c-2c64-4c44-bb80-ff21095c9f31)

### Identify
![image](https://github.com/dresden-elektronik/deconz-rest-plugin/assets/62012121/69ab2ea0-f8f4-45a9-9a8f-0e47bd8f6d12)
![image](https://github.com/dresden-elektronik/deconz-rest-plugin/assets/62012121/5654d3e8-731f-46fd-967f-49df880dc593)

### Poll Control
![image](https://github.com/dresden-elektronik/deconz-rest-plugin/assets/62012121/b5a40bbf-30af-41a9-a7d1-2a93e02a4007)

### IKEA specific
![image](https://github.com/dresden-elektronik/deconz-rest-plugin/assets/62012121/026702cf-4005-4b59-b842-0a85c58f57dc)

## Node Panel
![image](https://github.com/dresden-elektronik/deconz-rest-plugin/assets/62012121/c0e72720-ddb5-4842-a58e-51326e4cc7a6)

DIYtechie:
I also have this request. Will not seem to pair.

Mimiix:
Guys, it's not supported yet. That's why it's not showing up in phoscon. As long as it has the hex names it wont work. If you can't get it in deconz, it's a different issue and perhaps you need to ask for help on the forums with that.

BabaIsYou:
As suggested by ebaauw in Discord, will begin by reusing DDF for Styrbar (adapted of course)

```
 {
  "schema": "devcap1.schema.json",
[...]

PR #7199, opened 2023-08-25
DDF for Ikea RODRET Remote Control

TheGreatNemo:
Added DDF for Ikea RODRET Remote Control as sugested in Issue #7147


TheGreatNemo:
Seams to me, like the failed verification, is caused by DDFs I did not touch and which are unrelated to the made changes. 
So from my point of you, it looks like the masterbranch needs to be fixed, before the verification can succeed?
Is there anything I need to do to the PR?

Issue #7200, opened 2023-08-26
Aqara LED Strip T1


ebaauw:
## Device
- Product name: LED Strip T1
- Manufacturer: Aqara
- Model identifier: lumi.light.acn132
- Device type: Light

The base product includes a 24V, 1.5A power adapter; a wired 3-button control unit, incl. a microphone for music sync; and a 2m LED strip consisting of 10 segments.  Each segment contains 5 RGB, 5 cool white, and 5 warm white LEDs, 7 resistors, a capacitor, and a controller chip.  The 1m Aqara LED Strip T1 Extension Kit adds 5 segments.  According to documentation, the strip can be extended to 10m (adding 8 extension kits).  The strip can be cut at a segment border.

It pairs without any issues to my Aqara M2 hub at first attempt.  It was shipped with firmware 0.0.0_2626, but the Hub upgraded it to 0.0.0_2627, according to the app, to "fix known problems".

The Aqara app provides settings for
- Power-Off Memory: always on, always off; restore last status;
- Transition Settings: Off to On and On to Off dimming duration, from 0 to 30s, with 500ms intervals;
- Dimming range: a slider.  I think the app renders the screen incorrectly (the right hand side seems missing), and I cannot see any value nor scale;
- Strip length: to set the actual length of the strip after adding an extension kit or cutting the strip, from 1m to 10m at 0.2m intervals.

A note under the Strip length mentions that the white light is not affected by the length setting.  This seems consistent with the five connections between the sections, labeled: +24V, DI/DO, GND, WW, CW.  So the strip has traditional channels for the whites, and uses a serial protocol (with fixed clock) for RGB LEDs, probably similar to Neopixel.

The Aqara app allows for control of on/off, brightness, setting pre-defined RGB colours, setting RGB from a spectrum (in Windows #rrggbb notation), setting colour temparature from 2700K to 6500K (suggesting `ct` values from 153 to 370).

It comes with eight pre-defined RGB effects and four predefined tuneable white effects.  You can create custom effects.  For static effects, you can set an RGB colour per segment.  The brightness seems a single setting for all segments, I assume they control that through the +24V signal.  For dynamic effects, you can select up to eight colours, a "sync effect" (pattern, I think), and the speed (from 1% to 100%).

Finally, Music Sync can be turned on or off.  It has a setting for high vs low sensitivity, and a choice of four sync effects.

## Screenshots
[...]

PR #7201, opened 2023-08-26
Aqara LED Strip T1

ebaauw:
Add support for the Aqara LED Strip T1, see #7200

### DIFF
--- a/devices/generic/constants.json
+++ b/devices/generic/constants.json
@@ -2,6 +2,7 @@
     "schema": "constants1.schema.json",
 
     "manufacturers" : {
+        "$MF_AQARA": "Aqara",
         "$MF_BOSCH": "Bosch",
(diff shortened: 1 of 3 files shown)


Cyrille63:
Is it possible to set strip lenght (without aqara hub) if I add an 1m extension kit ? If yes how ? Thk

Sumd84:
@Cyrille63 > Is it possible to set strip lenght (without aqara hub) if I add an 1m extension kit ? If yes how ? Thk

On the DeconzGUI, select the Led Strip and look for the parameter "Segments" at address 0x051B in the Lumi Specific section:
The value corresponds to the number of 20cm segments. By default, the value is 10, corresponding to the 2m starter kit. 
The value only affects RGB LEDs. Warm and Cold white Leds are independent of the parameter.

Issue #1935, opened 2019-10-05
config.sensitivity values not normalised


ebaauw:
The REST API plugin doesn't normalise the values of `config.sensitivity`.  For the Hue motion sensors, the highest value of `config.sensitivity` (exposed as `config.sensitivitymax`) indicates the most sensitive setting, whereas for the Xiaomi vibration sensor (`lumi.vibration`), the highest value is the least sensitive setting.  For the Hue motion sensors, the lowest value is 0; for the Xioami vibration sensor it's 1.  I think that for all devices, value `0` should be the least sensitive setting, and value from `config.sensitivitymax` should be the most sensitive setting.

**Background**

`config.sensitivity` and `config.sensitivitymax` were copied from the Hue API, which uses these for the Hue motion sensor.  The values of `config.sensitivity` range from 0 to 2, the value of `config.sensitivitymax`, with 0 being the least sensitive and 2 being the most sensitive.  The recently introduced Hue outdoor motion sensor reports 5 for `config.sensitivitymax`.  The Hue motion sensors use manufacturer specific attributes for `config.sensitivity` and `config.sensitivitymax`, whose raw values map 1:1 to the API values.

The Xioami vibration sensor uses a different manufacturer specific attribute for setting sensitivity.  According to a comment in the code, this attribute uses raw values 0x15 for low sensitivity, 0x0b for medium sensitivity, and 0x01 for high sensitivity.  Currently, the REST API exposes `config.sensitivity` as the raw value of this attribute, see #1396; I don't know if the vibration sensor actually supports the full range of values between 1 and 21.

I think the REST API should expose raw value 0x01 as `config.sensitivity` value 2, 0x0b as 1 and 0x15 as 0, and expose `config.sensitivitymax` as 2.  If the intermediate values are useable, we could expose `config.sensitivitymax` as 20 (!) and map raw values 0x01, ..., 0x15 to API values 20, ..., 0.

JHenzi:
Does this issue still stand, there is no real way to get the sensor to return vibration: true based on sensitivity? Like others I want to detect the running of the washer and dryer, issue I have is that the value is always false, otherwise I get a vibrationstrength value as a number. Seems that once the dryer or washer stops it stops reporting and

ebaauw:
Bump

ebaauw:
Bump

dubocr:
[...]

PR #7208, opened 2023-09-01
Fix sensitivity for lumi_vibration_aq1

ebaauw:
See Discord, upon sniffing the Aqara app / Hub M2, it appears that the Xiaomi vibration sensor uses Zigbee values 1: High, 2: Medium, 3: Low.

The device seems to accept and report back any value, even higher then 21.  This was carried over from the C++ code when the sensor was supported initially, but, apparently, never verified.

See also:
- https://github.com/dresden-elektronik/deconz-rest-plugin/issues/1396#issuecomment-478644832
- https://github.com/dresden-elektronik/deconz-rest-plugin/issues/1935#issue-502937054
- https://github.com/dresden-elektronik/deconz-rest-plugin/pull/6919

### DIFF
--- a/devices/xiaomi/lumi_vibration_aq1.json
+++ b/devices/xiaomi/lumi_vibration_aq1.json
@@ -92,7 +92,7 @@
             "cl": "0x0000",
             "mf": "0x115F",
             "at": "0xFF0D",
-            "eval": "Item.val = 2 - Math.floor(Attr.val / 10)"
+            "eval": "Item.val = 3 - Attr.val"
           },
           "read": {
             "fn": "zcl",
@@ -109,7 +109,7 @@
             "mf": "0x115F",
             "at": "0xFF0D",
             "dt": "0x20",
-            "eval": "10 * (2 - Item.val) + 1"
+            "eval": "3 - Item.val"
           },
           "values": [
             [
--- a/devices/xiaomi/lumi_vibration_aq1.json
+++ b/devices/xiaomi/lumi_vibration_aq1.json
@@ -101,7 +101,7 @@
             "mf": "0x115F",
             "at": "0xFF0D"
           },
-          "refresh.interval": 600,
+          "refresh.interval": 3600,
           "write": {
             "fn": "zcl",
             "ep": 1,
@@ -109,8 +109,10 @@
             "mf": "0x115F",
             "at": "0xFF0D",
             "dt": "0x20",
-            "eval": "3 - Item.val"
-          },
+            "eval": "3 - Item.val",
+            "state.timeout": 2,
+            "change.timeout": 3600
+            },
           "values": [
             [
               0,


SwoopX (changes requested):
I'd advise some changes here:

1. Change `refresh.interval` from 600 to 3600 - We only want a forced read if the special reporting didn't come in. Most likely, the following special report will provide the values anyway before the actual read kicks in. A smaller value does not make much sense either way as the device is a deep sleeper
2. Add `"state.timeout": 2` to the write function - This is barely documented, I know, but it speeds up the state change when the device is awake and chances are high that the whole state change can finish within the few seconds the device is awake.
3. Add `"change.timeout": 3600` to the write function - This grants the state change a life time of 1h rather than the default 2 or 3 mins. So, whenever you change sensitivity and do not explicitly wake the device, the change will be applied when it wakes up later.

Other than that, the values changes look fine to me 👍

ebaauw:
> Change `refresh.interval` from 600 to 3600

Done.  I suppose we would need to check this for other Xiaomi DDFs for this as well, as they all do periodic reporting only on the special attribute.

> Add "state.timeout": 2 to the write function

Done.  Not sure I understand what this does exactly.  What operation is timing out here, and how does the code react to this?  In what circumstances, for devices with what characteristics, should be we adding this?

> Add "change.timeout": 3600 to the write function

Done.  This makes very much sense, and, I suppose should be added to all deep sleepers with changeable `config` items.  Notably the Hue sensors and controllers, with periodic reporting per 5 minutes.

manup:
Note to myself: It's time to better document how the `StateChange` works and is related to "read" and "write" functions in DD, and also add a state diagram to better visualize what's going on (internally it is a small state machine). The logic behind it is not too complex and stabilized since a while.

Special care needs to be taken for the `awake` keyword on items in the DDF. For sleeping devices it acts as a trigger to process open tasks.

For example a Xiaomi device that sleeps for one hour and sends the special report, should have the `awake = true` set for the respective "special" item. When the special report is received the `const char *REventAwake = "event/awake";` is triggered and handled in `device.cpp` to proceed with polling and `StateChanges`.

But here comes the hard part: To know that a device is truly awake when a certain ZCL command is received can only be seen in the sniffer! Awake is **only** true if a device sends a ZCL command **and** does a *MAC Data Request* (aka parent poll) afterwards to check if there is any response/ message to receive.

This is not always the case, some Xiaomi devices report sensor values without doing a *MAC Data Request*, which means sending anything to the device will timeout because it's sleeping.

That's the reason why not every item should have `awake = true`, because we would send into the void.

---------

The `StateChange` is designed to survive any error that may happen, like sending to a device that is sleeping, routing errors, or the device simply being powered off. It will retry as long as the `change.timeout > 0`, this is triggered by `REventAwake` and `REventPoll`. Having proper `awake` knowledge helps to prevent useless requests to a device which will timeout anyway as the device is sleeping (which the legacy code does a lot).

ebaauw:
`awake` is (and already had been) specified on items based on the Xiaomi special attribute, and not on other items.  I can confirm that the special attribute report is followed by a data request to the parent, also when triggered by pushing the button on the device.  The XIaomi app actually prompts you to do so, when changing the sensitivity, unless you're happy to wait for some time.

Note: instead of sniffing, you could also get this info from the deCONZ log, if the coordinator device is the parent of the sensor (which is probably achieved easiest when its the only router in the network).

raducotescu:
@ebaauw, without this issue being fixed, is there a way to correctly configure the sensitivity? The Internet is full of the 1-21 value range, where 1 is considered the most sensitive, most likely based on https://dresden-elektronik.github.io/deconz-rest-doc/devices/xiaomi/aq_vibration_sensor/#config-attributes.

I'm asking because I recently had to replace one of the sensors I have at home and the new one seems to randomly trigger. Since in my case this is connected to a security system, I'd *really* like to avoid false positives, otherwise my neighbours will hate me.

Thank you! (and sorry for not actually contributing to the issue)

ebaauw:
You can write any raw value to the corresponding Zigbee attribute (0xFF0D in the _Basic_ cluster) from the deCONZ GUI, bypassing the API altogether.  Press the button to wake the device and have deCONZ send the _Write Attributes_ command.  Press the button again, and read back the attribute, to make sure the value sticks.

As I've mentioned on Discord:
- The Xiaomi app only uses values 1 (high), 2 (medium), and 3 (low);
- The device seems to accept and report back values larger than 21 just as happily.  The documenation is likely based on the abovementioned comment, but I have not been able to find out on what this comment was based.
- I don't have the equipment to test in a controlled way whether the device actually changes it sensitivity for settings higher than 3.

I would love to hear your experience, at what setting the false positives stop if they still appear at 3.

manup (approved):

bcutter:
Could someone please summarize if this fix creates the need of users to actually change something?
E. g. if one uses `vibrationstrength` values of 6, 10 and 29 - will everything work as before?

cracyfloyd:
Iam a little bit confused. Which values can i now use for sensitivity ??. I need less sensitivity. Before this change i use 11.

bcutter:
> Iam a little bit confused. Which values can i now use for sensitivity ??. I need less sensitivity. Before this change i use 11.

Never got answered here unfortunately. But according to the (updated?) docs here it seems the answer to your question is: `0` (zero).

https://dresden-elektronik.github.io/deconz-rest-doc/devices/xiaomi/lumi_vibration_aq1/#config-attributes

PR #7213, opened 2023-09-08
DDFs for Trådfri bulbs

ebaauw:
Updated the DDFs for the Trådfri bulbs I still had lying around in a drawer.

### TRADFRI bulb E27 W opal 1000lm
- Updated existing DDF for _Dimmable Light_
- Added `capabilities`: `alert/trigger_effect`;
- Added `config` items: `bri/execute_if_off`, `bri/on_level`, `bri/onoff_transition_time`, `bri/startup`, `on/startup`;
- Updated attribute reporting settings to 1/300/1 and changed `refresh.interval` to 305.

### TRADFRI bulb E27 WS opal 980lm
- New DDF for _Color Temperature Light_.
- ~~This light reports a weird character in the _Manufacturer Name_, causing deCONZ to report it as hex string.  The DDF should correct this.~~
- `capabilities` items: `alert/trigger_effect`, `bri/move_with_onoff`, `color/capabilities`, `color/ct/max`, `color/ct/min`, `on/off_with_effect`, `transition_block`;
- `config` items: `bri/execute_if_off`, `bri/on_level`,`bri/onoff_transition_time`, `bri/startup`, `color/ct/startup`, `color/execute_if_off` `on/startup`;
- Attribute reporting settings to 1/300/1 and changed refresh.interval to 305.

### TRADFRI bulb E27 CWS opal 600lm
- Renamed existing DDF file and bumped status to Gold.
- This light identified as _Color Dimmable Light_.  Despite it's name, it doesn't support colour temperature, so it's exposed as a _Color Light_.
- `capabilities` items: `alert/trigger_effect`, `bri/move_with_onoff`, `color/capabilities`, `color/xy/...`, `on/off_with_effect`, `transition_block`;
- We don't know the actual colour gamut, so we report the default colour gamut;
- `config` items: `bri/execute_if_off`, `bri/on_level`,`bri/onoff_transition_time`, `bri/startup`, `color/execute_if_off`, `on/startup`;
- Updated attribute reporting settings to 1/300/1 and changed refresh.interval to 305.

### TRADFRI bulb E27 CWS 806lm
- Renamed DDF file.
- `capabilities` items: `bri/move_with_onoff`, `color/capabilities`, `color/xy/...`, `on/off_with_effect`;
- This is the first IKEA light that doesn't contain the `transition_block` firmware bug.  But it doesn't support _Trigger Effect_.
- `config` items: `bri/execute_if_off`, `bri/on_level`,`bri/onoff_transition_time`, `bri/startup`,  `color/ct/startup`, `color/execute_if_off`, `on/startup`;
- Updated attribute reporting settings to 1/300/1 and changed refresh.interval to 305.


manup:
Interesting, the hex conversion happens because the string isn't UTF-8 encoded. It is actually *"TRADFRI bulb E27 WS opal 980lm"*

`0x545241444652492062756c6220453237205753` this part is valid UTF-8 *"TRADFRI bulb E27 WS"*.

But hex `0xa0` in binary is `1010 0000` which in UTF-8 is the second, third, or fourth byte. But the previous byte is `0x53` which is the 1-byte ASCII char `S` so this is not a multi-byte character sequence.

Digging a bit deeper the string is actually latin1 encoded! `0xa0` is "non-breaking space" here.

The deCONZ ZCL attribute reader plays strict UTF-8 here (which implicitly includes ASCII) and just returns the whole sequence as hex string. Perhaps I should expand it to alternatively try to latin1 decode if UTF-8 fails, we have seen other strings which where also returned as hex string, perhaps the same issue.

manup:
image


------

It is allowed to use non UTF-8 but should be stated in the complex descriptor, which I've never seen available in a product (can be seen in the Node Info panel).

ebaauw:
From my pet shark on the wire:
```
ZigBee Cluster Library Frame, Command: Read Attributes Response, Seq: 29
    Frame Control Field: Profile-wide (0x18)
    Sequence Number: 29
    Command: Read Attributes Response (0x01)
    Status Record, String: TRADFRI bulb E27 WS�opal 980lm
        Attribute: Model Identifier (0x0005)
        Status: Success (0x00)
        Data Type: Character String (0x42)
        String: TRADFRI bulb E27 WS�opal 980lm
```

```
0000   1e 54 52 41 44 46 52 49 20 62 75 6c 62 20 45 32   .TRADFRI bulb E2
0010   37 20 57 53 a0 6f 70 61 6c 20 39 38 30 6c 6d      7 WS.opal 980lm
```

When I got the light, it didn't include the weird character, but it appeared after a firmware update.

Last time I checked, the Hue bridge chokes on this as well, and shows an empty `modelid` in the API.

manup:
I've extended the attribute reader to detect and convert Latin1 strings to UTF-8 and only contain printable characters.
Also as an opinionated implementation soft-hyphens will be changed to simple minus `-` and "non breakable space" to simple ASCII space `0x20` since these are easier to type in a DDF and impossible to spot visually when just looking at a string.

There fore the DDF only needs to contain the ASCII string: `"TRADFRI bulb E27 WS opal 980lm"` (with normal space after WS).

https://github.com/dresden-elektronik/deconz-lib/commit/1940240e8560a656c3e2b7d72f15d137d3520d39

manup (approved):

PR #7218, opened 2023-09-10
Improve/Fix DFF for IKEA RODRET Dimmer and other IKEA controllers

ebaauw:
### RODRET Dimmer
While this device looks like half a STYRBAR, it acts differently.  It picks a group ID on reset, reports this through _Touchpoint Commissioning_, and creates a non-removable binding from the client _On/Off_ cluster to this group.  This means `config.group` should match this group ID, and we need to create a binding of the client _Level Control_ cluster to this group as well.
Note that the device also sends an _Add Group_ to the coordinator for the group it picked on reset.
- Renamed DDF;
- Fixed indentation;
- Add `productid` to show the product ID (Exxxx), sourced from 0x0000/0x000A.
- Add `config/alert`;
- `config/battery`:
  - Change reporting to 300/2700/0x01; 
  - Changed `"fn": "zcl"` to `"fn": "zcl:attr"`l
  - Add `read` with `refresh.interval` of 86400;
  - Add `awake`;
- `state/buttonevent`: add `awake;
- Expose `cap/group` (to read the group from the _Touchpoint Commissioning_ cluster and `config/group` (to create the group bindings).  Through some C++ magic, the _Get Group Identifier Response_ is used to populate `config/group`.  While the API exposes this as read/write, changing it will not work, as the bindings to the _Touchpoint Commissioning_ group cannot be removed.
- Changed order of bindings to do the client clusters first.  Device becomes incommunicado without any client cluster bindings.
- Use group bindings for client clusters.
```
{
  "capabilities": {
    "group": "23998"
  },
  "config": {
    "battery": 100,
    "group": "23998",
    "on": true,
    "reachable": true
  },
  "ep": 1,
  "etag": "96ae43c63f3bedd3685532de8bb2df5c",
  "lastannounced": null,
  "lastseen": "2023-09-16T09:44Z",
  "manufacturername": "IKEA of Sweden",
  "mode": 1,
  "modelid": "RODRET Dimmer",
  "name": "Rodret",
  "productid": "E2201",
  "state": {
    "buttonevent": 1002,
    "lastupdated": "2023-09-16T09:11:04.895"
  },
  "swversion": "1.0.47",
  "type": "ZHASwitch",
  "uniqueid": "b4:35:22:ff:fe:4b:b5:c1-01-1000"
}
```

### Remote Control N2
[...]

PR #7230, opened 2023-09-16
DDF for IKEA RODRET Dimmer

ebaauw:
While this device looks like half a STYRBAR, it acts differently.  It picks a group ID on reset, reports this through _Touchpoint Commissioning_, and creates a non-removable binding from the client _On/Off_ cluster to this group.  This means `config.group` should match this group ID, and we need to create a binding of the client _Level Control_ cluster to this group as well.
Note that the device also sends an _Add Group_ to the coordinator for the group it picked on reset.

Changes:
- Renamed DDF;
- Fixed indentation;
- Add `productid` to show the product ID (Exxxx), sourced from 0x0000/0x000A.
- Add `config/alert`;
- `config/battery`:
  - Change reporting to 300/2700/0x01; 
  - Changed `"fn": "zcl"` to `"fn": "zcl:attr"`l
  - Add `read` with `refresh.interval` of 86400;
  - Add `awake`;
- `state/buttonevent`: add `awake;
- Expose `cap/group` (to read the group from the _Touchpoint Commissioning_ cluster and `config/group` (to create the group bindings).  Through some C++ magic, the _Get Group Identifier Response_ is used to populate `config/group`.  While the API exposes this as read/write, changing it will not work, as the bindings to the _Touchpoint Commissioning_ group cannot be removed.
- Changed order of bindings to do the client clusters first.  Device becomes incommunicado without any client cluster bindings.
- Use group bindings for client clusters.

```
{
  "capabilities": {
    "group": "23998"
  },
  "config": {
    "battery": 100,
    "group": "23998",
    "on": true,
    "reachable": true
  },
  "ep": 1,
  "etag": "96ae43c63f3bedd3685532de8bb2df5c",
  "lastannounced": null,
  "lastseen": "2023-09-16T09:44Z",
  "mode": 1,
  "modelid": "RODRET Dimmer",
  "name": "Rodret",
  "productid": "E2201",
  "state": {
    "buttonevent": 1002,
    "lastupdated": "2023-09-16T09:11:04.895"
  },
  "swversion": "1.0.47",
  "type": "ZHASwitch",
  "uniqueid": "b4:35:22:ff:fe:4b:b5:c1-01-1000"
}
```

### Generic
- Add `cap_group_item.json`;
- Add `touchlink_commissioning_cluster/1000_get_group_identifiers.js` to parse the _Get Group Identifiers Response_ into `cap/group`.

### Legacy Code
- `binding.cpp`: Bug fix: legacy code would still create a `config/group` item and corresponding `groups` resource on DDF managed devices.
- `resource.cpp`, `resource.h`: Add `RCapGroup`;
- `rest_sensors.cpp`: Expose `capabilities` and `productid` on `sensors` resources that support these;
- `general.xml`: Rename _ZLL Commissioning_ to _Touchpoint Commissioning_, in line with ZCL v8 spec.

This PR was part of #7218.


ebaauw:
Note `config/alert` was missing - fixed in #7238.

PR #7236, opened 2023-09-16
DDF for TRADFRI open/close remote

ebaauw:
New DDF, in line with DDFs for other IKEA controllers.

Controller no longer supports groups.

```
{
  "config": {
    "alert": "none",
    "battery": 100,
    "on": true,
    "reachable": true
  },
  "ep": 1,
  "etag": "6edbf2215f774c46b5fa62561593d499",
  "lastannounced": "2023-09-16T15:02:20Z",
  "lastseen": "2023-09-16T15:02Z",
  "mode": 1,
  "modelid": "TRADFRI open/close remote",
  "name": "Open/Close",
  "productid": "E1766",
  "state": {
    "buttonevent": 1002,
    "lastupdated": "2023-09-16T11:26:15.709"
  },
  "swversion": "2.3.079",
  "type": "ZHASwitch",
  "uniqueid": "00:0d:6f:ff:fe:e3:f4:90-01-0102"
}
```

### DIFF
--- a/devices/ikea/tradfri_open_close_remote.json
+++ b/devices/ikea/tradfri_open_close_remote.json
@@ -73,6 +73,9 @@
           "name": "attr/uniqueid"
         },
         {
+          "name": "config/alert"
+        },
+        {
           "name": "config/battery",
           "awake": true,
           "parse": {
--- a/devices/ikea/tradfri_open_close_remote.json
+++ b/devices/ikea/tradfri_open_close_remote.json
@@ -1,0 +1,131 @@
+{
+  "schema": "devcap1.schema.json",
+  "manufacturername": "$MF_IKEA",
+  "modelid": "TRADFRI open/close remote",
+  "product": "TRADFRI open/close remote - E1766",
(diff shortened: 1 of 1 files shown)


manup:
Tested with my unit works here as well.

Just noticed that the switch makes x002 and x003 events but not x001. Not a problem of the DDF and also stated as such in the button_maps.json but might be worth to check in the sniffer if we can't extract x001 as well in future.

manup (approved):

ebaauw:
No can do.  The remote sends _Open_ (or _Close_) on press/release and on press/hold; and _Stop_ on release after hold.  This makes sense from a blind control perspective: just press/release opens (closes) the blind; press hold moves the blind until released.
Note that the IKEA blinds don't provide idem potency on _Open_ nor on _Close_; when a second command is received while the blind is moving, it stops the movement.  Again this makes sense from the control perspective, but not so much from an automation perspective.  Also means that the remote will seem to behave differently, when bound to non-IKEA window covering devices.

PR #7241, opened 2023-09-16
Add DDF for Aqara T1 Temperature and Humidity Sensor WSDCGQ12LM

SwoopX:
Adds initial support for the device. Verification of the special reporting is pending, this is currently an educated guess.


manup:
There are also additions of `"mf": "0x115F"` in other devices parse functions, could you test that these work?
Perhaps these should be separate "Fix ... " PRs so they are also listed in the changelog :)

The new `xiaomi_wsdcgq12lm_t1_temp_hum_sensor.json` seems to contain invalid JSON?

```
Error: Parse error on line 94:
...mperature",        },        {       
----------------------^
Expecting 'STRING', got '}'
```

SwoopX:
> The new `xiaomi_wsdcgq12lm_t1_temp_hum_sensor.json` seems to contain invalid JSON?
> 
> ```
> Error: Parse error on line 94:
> ...mperature",        },        {       
> ----------------------^
> Expecting 'STRING', got '}'
> ```

Ah, nice catch. Thought, the validator would notify on this one as well 🙂 

> There are also additions of "mf": "0x115F" in other devices parse functions, could you test that these work?

Well, I guess this is more of a documentary addition rather than necessary. The Xiaomi parse function is currently not explicitly requesting the correct mfc.
As I recall, all those occurrances came from me anyway and were based on some sniffing, so I'm pretty confident it is all working alright (for the devices I own, where this doesn't count to).

manup:
> Ah, nice catch. Thought, the validator would notify on this one as well 🙂

It did :) output is a bit noisy due the errors not related to this DDF, but there was a inline box complaining about JSON.

> 
> > There are also additions of "mf": "0x115F" in other devices parse functions, could you test that these work?
> 
> Well, I guess this is more of a documentary addition rather than necessary. The Xiaomi parse function is currently not explicitly requesting the correct mfc. As I recall, all those occurrances came from me anyway and were based on some sniffing, so I'm pretty confident it is all working alright (for the devices I own, where this doesn't count to).

Ah ok, well then lets give it a try

PR #7243, opened 2023-09-16
DDF for IKEA TRADFRI motion sensor

ebaauw:
New DDF, in line with DDFs for other IKEA controllers (!).

The DDF handles both the gen-1 (E1525) and gen-2 (E1745) motion sensors.  They use the same `modelid`; but the `productid` to tell them apart.

Note that these are indeed controllers (not sensors), sending _On with Timed Off_ on detecting motion.  Light can be controlled directly (without deCONZ running) by adding them to the associated group.

The sensors won't detect next motion for the period from _On Time_, so we set `config.duration` based on that parameter.  This is a change with respect to the C++ code, that used `config.delay` for that.  However the DDF logic to reset `state/presence` only works for `config/duration`, and it should be set to `0` (so the C++ code used `config/delay` instead) or to the `config/delay` value anyways.

The duration can set between 60 and 600 seconds on the gen-1 model, using the dial on the device, and is fixed to 180 seconds on the gen-2 model.  The day/night setting on the device needs to be on night, or `state/dark` will always be reported as `true`.

Note that the C++ code will still add `config/delay` when motion is detected, but this is removed when deCONZ restarts.

```
{
  "config": {
    "alert": "none",
    "battery": 100,
    "delay": 60,
    "duration": 60,
    "group": "10461",
    "on": true,
    "reachable": true
  },
  "ep": 1,
  "etag": "292eb36a4d8bea64772e663dde57870f",
  "lastannounced": null,
  "lastseen": "2023-09-16T23:03Z",
  "modelid": "TRADFRI motion sensor",
  "name": "TRADFRI motion sensor",
  "productid": "E1525",
  "state": {
    "dark": false,
    "lastupdated": "2023-09-16T23:04:46.969",
    "presence": true
  },
  "swversion": "1.2.214",
  "type": "ZHAPresence",
  "uniqueid": "00:0b:57:ff:fe:22:cf:5f-01-0006"
}
```

```
{
  "config": {
    "alert": "none",
    "battery": 100,
    "delay": 180,
    "duration": 180,
    "group": "20004",
    "on": true,
    "reachable": true
  },
  "ep": 1,
  "etag": "0a91afd665efbbe9544fb2810405f52a",
  "lastannounced": null,
  "lastseen": "2023-09-16T23:04Z",
  "modelid": "TRADFRI motion sensor",
  "name": "Presence 130",
  "productid": "E1745",
  "state": {
    "dark": true,
    "lastupdated": "2023-09-16T23:04:45.342",
    "presence": true
  },
  "swversion": "2.0.022",
  "type": "ZHAPresence",
  "uniqueid": "14:b4:57:ff:fe:3e:ad:28-01-0006"
}
```

### DIFF
--- a/devices/ikea/0006_presence.js
+++ b/devices/ikea/0006_presence.js
@@ -1,0 +1,5 @@
+/* global Item, R, ZclFrame */
+
+Item.val = true
+R.item('state/dark').val = (ZclFrame.at(0) & 0x01) === 0x00
+R.item('config/duration').val = Math.round((ZclFrame.at(1) | ZclFrame.at(2) << 8) / 10)
--- a/devices/ikea/tradfri_motion_sensor.json
+++ b/devices/ikea/tradfri_motion_sensor.json
@@ -1,0 +1,153 @@
+{
+  "schema": "devcap1.schema.json",
+  "manufacturername": "$MF_IKEA",
+  "modelid": "TRADFRI motion sensor",
+  "product": "TRADFRI motion sensor - E1525, E1745",
+  "sleeper": true,
+  "status": "Gold",
+  "subdevices": [
+    {
+      "type": "$TYPE_PRESENCE_SENSOR",
+      "restapi": "/sensors",
+      "uuid": [
+        "$address.ext",
+        "0x01",
+        "0x0006"
+      ],
+      "fingerprint": {
+        "profile": "0x0104",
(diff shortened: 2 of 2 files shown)


manup (approved):

manup:
Nice, looks like that we can remove quite some Ikea C++ code soon :+1:

PR #7245, opened 2023-09-17
New DDF for all IKEA blinds

ebaauw:
This new DDF exposes the IKEA FYRTUR, KADRILJ, PRAKTLYSING, and TREDANSEN blinds from one DDF.
It replacing the individual DDFs for FYRTUR, and for KADRILJ and the combined DDF for PRAKTLYSING and TREDANSEN.

I've only been able to test it on the FYRTUR as I don't posses any of the other blinds
I've downgraded the existing DDFs to status `Draft`, instead of deleting them.
This way, there's an easy fallback in case the new DDF doesn't work: mark it `Draft` and set the old DDF to `Gold`.
Once the new DDF has been confirmed to work with the other blinds, the existing DDFs can be deleted.
Note that the DDF for the KADRILJ was set to `Silver`.

Changes:
- Add `productid`;
- Add `cap/groups/not_supported`.  While the blinds support the _Groups_ cluster, deCONZ cannot send groupcasts to these, as it sends groupcasts to the _Receiver On When Idle_ broadcast address and the blinds don't have this.  With the latest firmware, the TRADFRI open/close remote no longer supports groupcasts either.
- Remove long deprecated `state/bri` and `state/on` from the KADRILJ, PRAKTLYSING, and TREDANSEN.  They were already removed from the FYRTUR;
- Change the order of bindings to bind the _Window Covering_ cluster before the _Power Configuration_ cluster;
- `state/battery`: Standardise on attribute reporting settings of 1/3600/0x01, with refresh interval of 3660;
- Don't read the `attr` items on the ZHABattery, as they're already read for the _Window covering device_

```
{
  "etag": "087d96ca9e2219df5961bfc4b81d2feb",
  "hascolor": false,
  "lastannounced": null,
  "lastseen": "2023-09-17T08:09Z",
  "modelid": "FYRTUR block-out roller blind",
  "name": "Bathroom Blind",
  "productid": "E1757-140",
  "state": {
    "lift": 0,
    "open": true,
    "reachable": true
  },
  "swversion": "2.3.088",
  "type": "Window covering device",
  "uniqueid": "84:fd:27:ff:fe:e1:01:61-01"
}
```
Note the `-140` in the `productid`, which indicates the width of the blind.

### DIFF
--- a/devices/ikea/blind.json
+++ b/devices/ikea/blind.json
@@ -1,0 +1,216 @@
+{
+  "schema": "devcap1.schema.json",
+  "manufacturername": "$MF_IKEA",
+  "modelid": [
+    "FYRTUR block-out roller blind",
+    "KADRILJ roller blind",
+    "PRAKTLYSING cellular blind",
+    "TREDANSEN block-out cellul blind"
+  ],
+  "vendor": [
+    "IKEA of Sweden",
+    "IKEA of Sweden",
+    "IKEA of Sweden",
+    "IKEA of Sweden"
+  ],
+  "sleeper": false,
+  "status": "Gold",
+  "subdevices": [
+    {
+      "type": "$TYPE_WINDOW_COVERING_DEVICE",
+      "restapi": "/lights",
+      "uuid": [
+        "$address.ext",
+        "0x01"
+      ],
+      "items": [
+        {
+          "name": "attr/id"
+        },
+        {
+          "name": "attr/lastannounced"
+        },
+        {
+          "name": "attr/lastseen"
+        },
+        {
+          "name": "attr/manufacturername"
+        },
+        {
+          "name": "attr/modelid"
+        },
+        {
+          "name": "attr/name"
+        },
+        {
+          "name": "attr/productid",
+          "parse": {
+            "fn": "zcl:attr",
+            "ep": 1,
+            "cl": "0x0000",
+            "at": "0x000A",
+            "eval": "Item.val = Attr.val"
+          },
+          "read": {
+            "fn": "zcl:attr",
+            "ep": 1,
+            "cl": "0x0000",
+            "at": "0x000A"
+          },
+          "refresh.interval": 86400
+        },
+        {
+          "name": "attr/swversion"
+        },
+        {
+          "name": "attr/type"
+        },
+        {
+          "name": "attr/uniqueid"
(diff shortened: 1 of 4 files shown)


jan666:
Nice unification! With firmware 24.4.11 (of fyrtur at least) Ikea changed how the battery state is handled. To be correct it needs to be devided by 2 - so I always change the `state/battery` to `"eval": "Item.val = Attr.val / 2". Is it possible to add this to the DDF, depending on the firmware?

ebaauw:
IKEA are finally adhering to the Zigbee standard, reporting _Battery Level_ in units of 0.5%.  Newer devices do so, and for some older devices, this is fixed in new firmware.  Technically, it might be possible to take the firmware version into consideration while parsing _Battery Level_ into `state/battery`, but that’s tricky, especially since we’d need a semver comparison of firmware versions. I think it’s better to base the DDF on the latest firmware.

About that, I cannot find 24.4.11.  It seems like a weird version, my FYRTUR is on 2.3.88.  What’s the _Date Code_ reported by 24.4.11?  What’s the _File Version_?  How did you upgrade: through the Dirigeria or through the Trådfri hub?  Or through deCONZ, but where did you get the file?

jan666:
I updated via deCONZ OTAU plugin at the end of july. We discussed it here: https://forum.phoscon.de/t/fyrtur-battery-percentage-remaining-166/3842

ebaauw:
So the new firmware isn't (yet?) listed on https://github.com/Koenkk/zigbee-OTA, but IKEA published it to http://fw.ota.homesmart.ikea.net/, where the `ikea-ota-download.py` script downloads it from.  Note to self: I want to check that location from my `otau` command-line utility as well.

Next to the blinds, I also see newer firmware for the five-button remote, shortcut button, on/off switch, and open/close remote.  Presumably fixing the _Battery Level_ for these controllers as well.  Upgrading...  Will check once the upgrade has finished and create a new PR to make sure `state/battery` or `config/battery` is reported correctly.  They all seem to upgrade from a 2.3.x version to a 24.x.x, so we could do a `R.item('attr/fwversion').val.startsWith('24')` check in the DDF to handle the old firmwares as well.

jan666:
Maybe it's better to check for the old version(s)? They might update to 25.xx...

PR #7251, opened 2023-09-17
Rename state{x,y} item generic file

Zehir:
For consistency rename the file to match the item id.

The X item don't have a `read` method, is that expected ? Maybe in that case add a `read` with a `"fn": "none"` ?


ebaauw:
The x does a read for both, but each needs their own parse.

PR #7253, opened 2023-09-18
Support `DataTypeReal`

ebaauw:
This PR is based on #7252.

Add API and web socket notification support for `DataTypeReal`.

This is a tactical solution, awaiting the refactor of `ResourceItem`.  In particular, support for rule conditions based on a `DataTypeRead` item is still missing.


resource.cpp
```diff
@@ -1178,13 +1190,28 @@ bool ResourceItem::setValue(const QVariant &val, ValueSource source)
...
+
+        if (ok)
         {
-            DBG_Printf(DBG_ERROR, "todo handle DataTypeReal in %s", __FUNCTION__);
-        }
+            m_lastSet = now;
+            m_doublePrev = m_double;
+            m_flags |= FlagNeedPushSet;
 
+            if (m_double != d)
```
manup, line 1204: The exact comparison of floating point values can be a bit tricky at times, but might be ok in this case.

Another thing that we should be aware of are NaN, +-Infinity and negative zero values. Not sure if we need to have any limitations here.

resource.cpp
```diff
@@ -1178,13 +1190,28 @@ bool ResourceItem::setValue(const QVariant &val, ValueSource source)
             return true;
         }
     }
-    else
+    else if (m_rid->type == DataTypeReal)
```
manup, line 1193: Do we need to also support min/max ranges as for integer values?

manup:
I've added a few notes. For the JS engine the `DJS_GetItemVal()` function needs to be extended to support `var a = Item.val`. The setter should already work.

I would have preferred this PR to only contain the real code with the measured value items in a follow up PR ;) but it's ok for now but title should mention it too.

My main concern is with the special values of floats/doubles, guess only time will tell how to deal with these. In our JS / C++ bridging code this is currently also a weak spot that is not really handled properly (which is likely fine in most cases).

ebaauw:
> I would have preferred this PR to only contain the real code with the measured value items in a follow up PR

It’s the other way round: this PR is a fellow-up to the measured_value PR.  Since both touch `resource.cpp` and `resource.h`, I based the branch for this PR on the `measured_values` branch, to avoid a merge conflict.

resource.cpp
ebaauw: I don’t think we have any API writeable `DataTypeReal` items yet, so these non-real values would have to come from the Zigbee attribute through the JS `eval`.  Once we’ll do `messured_value` in CLIP sensors, we would need a check to only specify real values in the API, setting the item.

resource.cpp
ebaauw: See above.  I don’t think we have any API writeable `DataTypeReal` items yet.

manup (approved):

manup:
Ok, I have this PR running now since yesterday no problems so far. From my perspective it's ready for merging.
I'll add the Duktape addition in small followup PR.

PR #7272, opened 2023-09-30
Update DDFs with current function names

SwoopX:
Replace all occurrences of legacy function name "zcl" with current name "zcl:attr" in DDFs and generic resource items.


manup:
Oha 263 changed files :D
The PR itself looks good as far as I can tell and can be merged in v2.24.1-beta.

We may run into merge conflicts on other PR if they change/rename same files.

I've noticed that we have a few "read" and "parse" functions which don't specify ''fn", which default's to "zcl" (now "zcl:attr"). Perhaps we should streamline this too to be explicit and check by validator in future..

SwoopX:
Ah, yes, the curse of freedom :)

PR #7279, opened 2023-10-05
Correct battery reporting attribute for Aqara E1 open/close sensor MCCGQ14LM

SwoopX:
In practice, the incorrectly set attribute has no negative effect on the device. Battery values were processed and updated accordingly. However, this might be coincidence, as the underlying Xiaomi parse function(s) do not strickly (or better to say not at all) check the set attribute in the DDF.


ebaauw:
The changed file doesn't pass DDF validation, as the validator doesn't accept a `cl` with a `xioami:special` function.  Not sure if that's a bug in the DDF or in the validator, but by the looks of it, no other DDF includes `cl` with `xioami:special`.

PR #7286, opened 2023-10-08
Update error messages to be more precise

bluemoehre:
Hue has added more effects than Colorloop. Now we also have Candle & Fireplace. So we have to be more generic with this message.

### DIFF
--- a/rest_lights.cpp
+++ b/rest_lights.cpp
@@ -1229,7 +1229,7 @@ int DeRestPluginPrivate::setLightState(const ApiRequest &req, ApiResponse &rsp)
         }
         else if (taskRef.lightNode->isColorLoopActive())
         {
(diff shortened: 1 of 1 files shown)


bluemoehre:
Also we need to adapt the corresponding variable & function names.

ebaauw:
This is not going to work.  The `isColorLoopActive()` checks only for colorloop.  The special effects work very differently on Zigbee level, and are manufacturer-specific (Hue implements them differently from Mueller and from the LIDL Xmas light strip).  In hindsight, I half regret exposing both using `state.effect`.  Also note that `colormode` is `hs` for colorloop, but `effect` for the special effects.

I think setting the color(temperature) while a special effect is active actually terminates the effect, where this is simply ignored while colorloop is active (hence the check).

bluemoehre:
If I activate the Candle or Fireplace effect and after PUT something with `sat` or `bri` I get the response error "colorloop is active". 
The effect is not terminated.

mattreim:

Issue #7293, opened 2023-10-11
Philips Hue Xamento L does not accept effect colorloop anymore


bluemoehre:
### Does the issue really belong here?

- [X] I definitively want to report a bug within deCONZ or its REST-API

### Is there already an existing issue for this?

- [X] I have searched the existing issues and there is none for the bug at hand

### Describe the bug

Today my Xamento L was updated to firmware 1.93.7 and the colorloop effect does not work anymore.

API reports:
```
[
    {
        "error": {
            "address": "/lights/41/state/effect",
            "description": "Internal error, 951",
            "type": 901
        }
    }
]
```

I have to admit I didn't try it for a long time using this effect. It may already broke with a previous firmware.

In the HUE app I see the new effect "Prism" listed for this device. It wasn't there before and has the same look like colorloop.

If I try some of the new effects which already work for the GU10 bulbs, I get different errors:

```
[
    {
        "error": {
            "address": "/lights/41/state/effect",
            "description": "invalid value, candle, for parameter, effect",
            "type": 7
        }
    }
]
```

### Steps to reproduce the behavior

1. Pick a Hue which is running firmware 1.7x
2. Update firmware to 1.9x
3. Try to activate any effect

### Expected behavior

- colorloop is still working
- candle is now working
- fireplace is now working

### Environment

- Host system: PC
- Running method: Ubuntu / deCONZ docker
- Firmware version: 26720700
- deCONZ version: 2.22.02 / 19.9.2022
- Device: ConBee II

ebaauw:
> my Xamento L was updated to firmware 1.93.7

I'm not familiar with this particular model, and as far as I can tell, it's not listed in any DDF.  That would mean it's exposed by the legacy code, which would explain why the Hue manufacturer-specific effects, like `candle` and `prism` don't work for this model.  You can check the supported effects in the API response to a GET of the `lights` resource, in `capabilities/color/effects`.

Firmware 1.93.7 seems like an old version to me.  What's the _Image Type_ for this model?  Was the firmware upgraded by the Hue bridge or Hue app over BLE; or by deCONZ?

> the colorloop effect does not work anymore.

[...]

PR #7295, opened 2023-10-12
Add Hue Xamento L to DDF

ebaauw:
Add support for Hue Xamento L, see #7293.


Thomas-Vos:
Thanks Erik! For future reference:

image

PR #7246, opened 2023-09-17
DDF for IKEA VINDSTYRKA

ebaauw:
Note: this PR depends on #7252.

This updated DDF for the IKEA VINDSTYRKA sensor introduces the new `state/measured_value` and corresponding `capabilities` items.  Unfortunately, the sensor reports temperature and humidity at a coarse resolution, so we cannot see the floating point values in action.

### ZHATemperature
- Add `state/measured_value`, as float value in °C.  Unfortunately, the sensor's granularity is limited to integer values;
- Add corresponding `cap/measured_value` items.  `min` and `max` are reported by the sensor; `quantity`, `substance`, and `unit` are defined in the DDF;
- Deprecate `state/temperature` (integer value in 0.01 °C);
- Not sure what to do with `config/offset` (also integer value in 0.01 °C).  I do apply it to `state/measured_value`, but ultimately, we want this as a float values as well.  Maybe introduce `config/measured_value/offset` and deprecate `config/offset`?
```
{
  "capabilities": {
    "measured_value": {
      "max": 50,
      "min": -10,
      "quantity": "temperature",
      "substance": "air",
      "unit": "degC"
    }
  },
  "config": {
    "offset": 0,
    "on": true,
    "reachable": true
  },
  "ep": 1,
  "etag": "89dc3816a6ddd73d103a8e2cc4b61caa",
  "lastannounced": null,
  "lastseen": "2023-09-23T14:05Z",
  "manufacturername": "IKEA of Sweden",
  "modelid": "VINDSTYRKA",
  "name": "Vindstyrka",
  "productid": "E2112",
  "state": {
    "lastupdated": "2023-09-23T14:02:06.470",
    "measured_value": 23,
    "temperature": 2300
  },
  "swversion": "1.0.11",
  "type": "ZHATemperature",
  "uniqueid": "94:34:69:ff:fe:bb:f4:92-01-0402"
}

```

### ZHAHumidity
- Add `state/measured_value`, as float value in %.  Unfortunately, the sensor's granularity is limited to integer values;
- Add corresponding `cap/measured_value` items.  `min` and `max` are reported by the sensor; `quantity`, `substance`, and `unit` are defined in the DDF;
- Deprecate `state/temperature` (integer value in 0.01%);
- Not sure what to do with `config/offset` (also integer value in 0.01%).  I do apply it to `state/measured_value`, but ultimately, we want this as a float values as well.  Maybe introduce `config/measured_value/offset` and deprecate `config/offset`?
```
{
[...]

PR #7297, opened 2023-10-13
 Bug fix IKEA VINDSTYRKA: Translate VOC index to tVOC level.

ebaauw:
The IKEA VINDSTYRKA doesn't report tVOC level, but some VOC index instead, see #7246.

This PR translates the VOC index to tVOC level, using the translation table from: 
https://learn.kaiterra.com/hs-fs/hubfs/Email%20Series/IAQ%20Series/AQI%20tvoc.png?width=1932&name=AQI%20tvoc.png:


SwoopX:
@ebaauw Maybe I can convince you on the Develco approach, as it is a bit leaner?

```
{
          "name": "state/airqualityppb",
          "awake": true,
          "parse": {
            "at": "0x0000",
            "cl": "0xfc03",
            "ep": 38,
            "fn": "zcl:attr",
            "mf": "0x1015",
            "eval": "if (Attr.val != 65535) { Item.val = Attr.val; }"
          }
        },
        {
          "name": "state/airquality",
          "awake": true,
          "parse": {
            "fn": "numtostr",
            "srcitem": "state/airqualityppb",
            "op": "le", "to": [65, "excellent", 220, "good", 660, "moderate", 5000, "unhealthy", 65535, "out of scale"]
          }
        },
```

ebaauw:
> Maybe I can convince you on the Develco approach, as it is a bit leaner?

Sorry, I don't understand what you mean?
- The VINDSTYRKA doesn't report tVOC density in ppb, but some index value instead, which needs to be translated to the tVOC density, following the listed table.
- I thought we decided we wanted to move towards `measured_value` and corresponding `capabilities`?

LeoeLeoeL:
Please check this:
image
The "excellent/good/moderate etc." seems to be updated non correctly.

ebaauw:
> Please check this:

I have no idea what "this" is; please list the ZHAAirQuality API resource, after reading the VINDSTYRKA's 0xFC7E cluster attributes in the GUI.  And please double-check that you're using the DDF from this PR, having installed both the DDF and the JS script file to `/usr/share/deCONZ/devices/ikea`.

Note that `measured_value`, `airqualityppb` and `airquality` are updated when reading or receiving a report from 0xFC7E/0x0000.  Also note that PM2.5 has a different scale, which is used for `airquality` in the ZHAParticulateMatter resource.

It should report the same `airquality` values as the other ZHAAirQuality sensors, since I simply copied the `numtostr` logic from those:

From | To | AirQuality
-- | -- | --
0 | 65 | excellent
66 | 220 | good
221 | 660 | moderate
661 | 5000 | unhealthy
5001 | 65535 | out of scale

I don't know the source of this mapping; in particular, I don't know where the 65 threshold comes from, nor why unhealthy doesn't go up to 5500.  Last time I checked, this is a jungle, with different organisations / government bodies from different countries recommending/prescribing different thresholds, labels or even number of labels used.

ebaauw:
@LeoeLeoeL, please open new issues for the PR2.5 mapping, including the source for this mapping, and the `airquality` issue.  This PR is about the tVOC index for the Vindstyrka.

PR #7315, opened 2023-10-21
Remove value limits for `state/lift` and `state/tilt` resource items

SwoopX:
The currently set limits for the items (min=0, max=100) are set too tight, so that any higher values reported by an arbitrary attribute are not processible. In such cases, an error is thrown while setting the item value.

```
JS failed to set Item.val for state/lift
failed to evaluate expression for 00:1f:ee:00:00:00:ab:22-01/state/lift: Item.val = Attr.val;, err: TypeError: failed to set Item.val
```

As devices must not necessarily use values from 0 - 100, the DDF is responsible for checking/translating values. This also allows devices to express a potentially unconfigured state.


manup:
For dynamically created items (the ones that are not specified in C++) the range is also extracted from the JSON file. But there is a catch, this can only be done once per item, so no local per device different ranges, which I guess isn't too bad for consistency.

manup (approved):

PR #7316, opened 2023-10-22
DDF for Tuya ZP01 motion sensor (_TZ3000_lf56vpxj)

ulikoenig:
Add Support for Tuya _TZ3000_lf56vpxj from here https://de.aliexpress.com/item/1005006010327384.html


pepz-de:
Got the sensor today. Would be great If it will be added

manup:
Hi, the `modelid` array need to be extended as well since this is treated as a pair to identify the device by (manufacturer name, modelid).

manup (changes requested):
See validator message, the `modelid` array needs to be extended as well.

Zehir:
@manup While resolving the validation issue I came to this and look like the new tuya name is already added from an other PR. I think we can close this one

manup:
Good catch, yes indeed it already exists.
Closing as completed.

Issue #7319, opened 2023-10-26
TRADFRI bulb GU10 CWS 345lm


bluemoehre:
### Is there already an existing issue for this?

- [X] I have searched the existing issues and there is none for my device

### Product name

TRADFRI LED 345lm

### Manufacturer

IKEA of Sweden

### Model identifier

TRADFRI bulb GU10 CWS 345lm

### Device type to add

Light

### Node info

![Screenshot 2023-10-26 140210](https://github.com/dresden-elektronik/deconz-rest-plugin/assets/1735386/b73389c3-735f-40d3-ac8e-a38207f2ba2d)

### Endpoints and clusters

![Screenshot 2023-10-26 140527](https://github.com/dresden-elektronik/deconz-rest-plugin/assets/1735386/d12862d5-386d-4079-8ee2-c38b92ee27e7)

### Basic

![Screenshot 2023-10-26 140707](https://github.com/dresden-elektronik/deconz-rest-plugin/assets/1735386/2cb47d6a-a036-4625-b83c-2ab564afac5e)

### On/Off

![Screenshot 2023-10-26 140935](https://github.com/dresden-elektronik/deconz-rest-plugin/assets/1735386/ccd3505f-d4dd-409c-b48b-a451c8c2eb40)

### Level Control

![Screenshot 2023-10-26 141042](https://github.com/dresden-elektronik/deconz-rest-plugin/assets/1735386/e2c24bd2-8ec9-4a5e-a312-b383d1d0483a)

### Color Control

![Screenshot 2023-10-26 141530](https://github.com/dresden-elektronik/deconz-rest-plugin/assets/1735386/7a23990b-e628-4902-a9d4-0b0a60d13ccc)

ebaauw:
What's the _Image Type_ of this light?

bluemoehre:
Yes, it's `0x2802`. But if you just merge it into the other light I am afraid we cannot fill the field `product` correctly. That's another point why I wrote in my linked PR #7320 that I prefer a JSON for each device. But maybe I am wrong?

bluemoehre:
bump

PR #7320, opened 2023-10-26
Add DDF for Tradfri bulb GU10 CWS 345lm #7319

bluemoehre:
First time adding DDF by myself - I hope everything is fine =)

File is based on TRADFRI bulb E27 CWS 806lm since this bulb reports same OTAU image and looks identical in deCONZ.
So I just changed product & model.

### DIFF
--- a/devices/ikea/tradfri_bulb_gu10_cws_345lm.json
+++ b/devices/ikea/tradfri_bulb_gu10_cws_345lm.json
@@ -1,0 +1,396 @@
+{
+  "schema": "devcap1.schema.json",
+  "manufacturername": "$MF_IKEA",
+  "modelid": "TRADFRI bulb GU10 CWS 345lm",
+  "product": "TRADFRI bulb GU10 CWS 345lm",
+  "sleeper": false,
+  "status": "Gold",
+  "subdevices": [
+    {
+      "type": "$TYPE_EXTENDED_COLOR_LIGHT",
+      "restapi": "/lights",
+      "uuid": [
+        "$address.ext",
+        "0x01"
+      ],
+      "items": [
+        {
+          "name": "attr/id"
+        },
+        {
+          "name": "attr/lastannounced"
+        },
+        {
+          "name": "attr/lastseen"
+        },
+        {
+          "name": "attr/manufacturername"
+        },
+        {
+          "name": "attr/modelid"
+        },
+        {
+          "name": "attr/name"
+        },
+        {
+          "name": "attr/productid",
+          "parse": {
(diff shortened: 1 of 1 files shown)


ebaauw:
> File is based on TRADFRI bulb E27 CWS 806lm since this bulb reports same OTAU image and looks identical in deCONZ.

In that case, I would simply add the `modelid` to that DDF, cf. what I did for Hue lights.  Of course, we'd need a better filename for the DDF.

bluemoehre:
> Of course, we'd need a better filename for the DDF.

Actually I was irritated a lot by the different DDF filenames and the mingling of 1:1 and 1:n configs. It's not very intuitive to find out if a device is already configured or not. 
I can imagine that a base that can be extended, such as "Tradfri color light", "Tradfri white light" and "Tradfri controller", could reduce duplication, but on the other hand it increases complexity. 
If it is guaranteed that the functionality is the same on all devices using the same image, we could group them according to the image ID. 

For now I would go ahead and just keep it simple with a 1:1, due to its very clear, is easy to fix and all the redundancy should not be a huge deal with search/replace in the IDE. 
This also corresponds to the [docs](https://dresden-elektronik.github.io/deconz-dev-doc/modules/ddf/#directory-structure):

> The current implementation uses one directory per vendor and one file per product

But it is up to you guys what u prefer. Of course, I can also change the PR and add the model into the other file.

**BTW:**
Is there a way to test these changes without compiling the whole thing? I didn't find any documentation about the full workflow. Maybe I missed something?

ebaauw:
> Is there a way to test these changes without compiling the whole thing? I didn't find any documentation about the full workflow. Maybe I missed something?

Just copy the DDF to `/usr/share/deCONZ/devices`, or wherever deCONZ installs these, and restart deCONZ.  Only new items in the API require changes in the C++ code and re-compilation.

bluemoehre:
I added the JSON and now I can control the bulb completely via Phoscon - nice! So I would say DDF works.

Just let me know if I should change this PR to merge the DDFs into a single file.

manup:
>In that case, I would simply add the modelid to that DDF, cf. what I did for Hue lights. Of course, we'd need a better filename for the DDF.

Hi yes merging the manufacturer name / model id combo into the existing DDF should be fine.

manup:
The arrays are only needed (and supported) for `modelid` and `manufacturername`.
The product and descriptions fields are merely for the generated documentation to show some human readable text :)

bluemoehre:
Files are merged & renamed.

PR #7331, opened 2023-11-02
Create lumi.switch.b1laus01.json

konistehrad:
Create a DDF for `lumi.switch.b1laus01`, which is the US [no-neutral Aqara switch](https://www.amazon.com/dp/B081ZQWX1F?psc=1&ref=ppx_yo2ov_dt_b_product_details). It's currently missing a couple config options, namely:

- Flip LED status indicator
- Switch "mode" between "quick" and "anti-flicker"

Both are [exposed in Z2M](https://github.com/Koenkk/zigbee-herdsman-converters/blob/master/src/devices/xiaomi.ts#L1006), though I'm not familiar enough with the DDF structure to implement those endpoints. This does make the switch work, however, so that's nice.

### DIFF
--- a/devices/xiaomi/lumi.switch.b1laus01.json
+++ b/devices/xiaomi/lumi.switch.b1laus01.json
@@ -1,0 +1,169 @@
+{
+  "schema": "devcap1.schema.json",
+  "manufacturername": "$MF_LUMI",
+  "modelid": "lumi.switch.b1laus01",
+  "vendor": "Xiaomi",
+  "product": "Aqara smart wall switch (no neutral wire) WS-USC01",
+  "sleeper": false,
+  "status": "Bronze",
+  "subdevices": [
+    {
+      "type": "$TYPE_ON_OFF_LIGHT",
+      "restapi": "/lights",
+      "uuid": [
+        "$address.ext",
+        "0x01"
+      ],
+      "items": [
+        {
+          "name": "attr/id"
+        },
+        {
+          "name": "attr/lastannounced"
+        },
+        {
+          "name": "attr/lastseen"
+        },
+        {
+          "name": "attr/manufacturername"
+        },
+        {
+          "name": "attr/modelid"
+        },
+        {
+          "name": "attr/name"
+        },
+        {
+          "name": "attr/swversion",
+          "parse": {
+            "at": "0x00f7",
+            "ep": 1,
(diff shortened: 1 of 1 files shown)


devices/xiaomi/lumi.switch.b1laus01.json
```diff
@@ -0,0 +1,169 @@
...
+            "eval": "if (Item.val == 'decoupled') { 0 } else if (Item.val == 'coupled') { 1 } else { 'unknown' }",
+            "fn": "zcl:attr",
+            "mf": "0x115f"
+          },
+          "values": [
+            ["\"decoupled\"", "Decoupled Mode"],
+            ["\"coupled\"", "Coupled Mode"] 
+          ],
+          "default": "coupled"
+        },
```
SwoopX, line 153: You might want to add the following code for the LED:

```
        {
          "name": "config/ledindication",
          "refresh.interval": 300,
          "parse": {
            "at": "0x00F0",
            "cl": "0xfcc0",
            "ep": 1,
            "eval": "Attr.val ? Item.val = false : Item.val = true",
            "fn": "zcl:attr",
            "mf": "0x115f"
          },
          "read": {
            "at": "0x00F0",
            "cl": "0xfcc0",
            "ep": 1,
            "fn": "zcl:attr",
            "mf": "0x115f"
          },
          "write": {
            "at": "0x00F0",
            "cl": "0xfcc0",
            "dt": "0x20",
            "ep": 1,
            "eval": "Item.val ? false : true",
            "fn": "zcl:attr",
            "mf": "0x115f"
          },
          "values": [
            [true, "Device LED always on"],
            [false, "Device LED off from 9pm - 9am"] 
          ],
          "default": true
        },
```

SwoopX (changes requested):
Hi,

the DDF looks ok so far. However, you might want to consider and test the suggested changes. One thing that would be still missing though is the respective button map. As I don't know the device endpoints and closters, I'd assume nevertheless that there's a Multistate Input Cluster on endpoint 0x29?

If yes, attribute 0x0055 may yield the following button press values:
41: 'top', 42: 'bottom', 51: 'both'

konistehrad:
My apologies, my Conbee has been pressed into service adding ZGP to Zigpy so I’m not sure when or if I’ll be able to validate those changes. If anyone else would like to pick up where I left off however please feel free!

PR #7338, opened 2023-11-07
Add `config.group` to 2nd-gen Hue motion sensors

ebaauw:
See Discord: the second-generation Hue motion sensors, SML003 and SML004, support direct binding of the client _On/Off_ clusters to devices and/or groups.  They'll send _On with Timed Off_ on detecting motion, with a fixed _On Time_ of 5 minutes.  They don't seem to use _Accept Only When On_, even though they have a light sensor.

This PR adds `config.group` to these sensors.  I'm not quite sure whether that's a good idea though.  This will result in a broadcast (groupcast) each time motion is detected, which might be challenging on the Zigbee network.  As it is known, the Hue motion sensor family expects a _Default Response_ on the _Report Attributes_.  When it doesn't see any, it will broadcast the _Report Attributes_ as fallback (while blinking red), and I'm seeing these broadcasts in the sniffer.  Not sure, but I think the _On with Timed Off_ broadcast might be interfering with the (timing off the) _Default Response_.

Note the first-generation Hue motion sensors, SML001 and SML002, don't seem to support this.  Maybe related to the fact that they carry two endpoints, and the client _On/Off_ cluster is on a different endpoint than the server clusters?


manup:
Good point, while on one side it seems useful and always nice to have configurable group bindings, it indeed might be troublesome in larger networks if the interval is too small?

SwoopX:
Erik, I'd recommend to leave the modelID and manufacturer name alone here. Just want to stay on the safe side 🙂

PR #7361, opened 2023-11-19
Fix LIDL smart plug consumption value multiplier

Pirionfr:
multiply by 10 consumption

### DIFF
--- a/devices/lidl/hg08673.json
+++ b/devices/lidl/hg08673.json
@@ -174,7 +174,7 @@
           "name": "config/reachable"
         },
         {
-          "name": "state/consumption"
+          "name": "state/consumption",
           "refresh.interval": 420,
           "parse": {
             "at": "0x0000",
--- a/devices/lidl/hg08673.json
+++ b/devices/lidl/hg08673.json
@@ -180,7 +180,7 @@
             "at": "0x0000",
             "cl": "0x0702",
             "ep": 0,
-            "eval: "item.val = 10 * Attr.val"
+            "eval": "item.val = 10 * Attr.val"
           },
           "default": 0
         },
--- a/devices/lidl/hg08673.json
+++ b/devices/lidl/hg08673.json
@@ -175,14 +175,13 @@
         },
         {
           "name": "state/consumption"
-          "refresh.interval": 300,
+          "refresh.interval": 420,
           "parse": {
             "at": "0x0000",
             "cl": "0x0702",
-            "ep": 0,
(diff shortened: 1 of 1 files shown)


devices/lidl/hg08673.json
```diff
@@ -175,6 +175,14 @@
         },
         {
           "name": "state/consumption"
+          "refresh.interval": 300,
+          "parse": {
+            "at": "0x0000",
+            "cl": "0x0702",
+            "ep": 0,
+            "eval: "item.val = 10 * Attr.val"
```
manup, line 183: Invalid JSON :) there is a missing quote after `"eval`

manup (changes requested):

devices/lidl/hg08673.json
```diff
@@ -175,6 +175,14 @@
         },
         {
           "name": "state/consumption"
+          "refresh.interval": 300,
+          "parse": {
+            "at": "0x0000",
+            "cl": "0x0702",
+            "ep": 0,
```
SwoopX, line 182: Could you please set the endpoint here (to 1)?

devices/lidl/hg08673.json
```diff
@@ -175,6 +175,14 @@
...
+            "eval": "item.val = 10 * Attr.val"
+          },
+          "default": 0
```
SwoopX, line 185: The default can be safely removed, as it will be set already automatically. When removing, please do not forget the comma in the line above 😉

devices/lidl/hg08673.json
```diff
@@ -175,6 +175,14 @@
         },
         {
           "name": "state/consumption"
+          "refresh.interval": 300,
```
SwoopX, line 178: The refresh interval should be roughly 60 seconds larger than the max reporting interval of the attirbute. You have defined 360 max reporting, so it should be 420 up here.

As a side note, the vast majority of devices uses a max reporting interval of 300 secs and a respective refresh interval of 360.

SwoopX (changes requested):

SwoopX (approved):

devices/lidl/hg08673.json
```diff
@@ -175,6 +175,13 @@
         },
         {
           "name": "state/consumption"
```
manup, line 177: Small issue here, missing comma after `"state/consumption"`

manup (changes requested):

PR #7371, opened 2023-11-23
Enable setting/viewing write functions in DDF editor

SwoopX:
This PR finally adds the possibility to view/set write functions in the DDF editor. Set parameters in the DDF are displayed correctly and newly set parameters appear as expected in the preview tab.

Additionally, the split of the previously available `zcl` function into `zcl:attr` and `zcl:cmd` should be completed. Both functions are now available to have better distinction.

grafik


manup:
Nice did a quick check locally works alright.

PR #7384, opened 2023-11-26
Stop unnecessary attribute read requests for many Philips/Signify devices

SwoopX:
Stumbled upon that accidentally while the sniffer was running. Due to an unfortunate DDF amendment, many devices were queried for their modelID as fast as possible. Especially for battery powered devices, this could lead to a more rapid battery drain and eventually have some impact on network resposiveness.

The PR adds either throttling (one query per day) or removes the attribute read requests completely, where appropriate.

### DIFF
--- a/devices/philips/light_zb3_C.json
+++ b/devices/philips/light_zb3_C.json
@@ -152,7 +152,8 @@
             "ep": "0x0b",
             "cl": "0x0000",
             "at": "0x0005"
-          }
+          },
+          "refresh.interval": 86400
         },
         {
           "name": "attr/name"
--- a/devices/philips/light_zb3_C_gradient.json
+++ b/devices/philips/light_zb3_C_gradient.json
@@ -132,7 +132,8 @@
             "ep": "0x0b",
             "cl": "0x0000",
             "at": "0x0005"
-          }
+          },
+          "refresh.interval": 86400
         },
         {
           "name": "attr/name"
--- a/devices/philips/light_zb3_white.json
+++ b/devices/philips/light_zb3_white.json
@@ -64,7 +64,8 @@
             "ep": "0x0b",
             "cl": "0x0000",
             "at": "0x0005"
-          }
+          },
+          "refresh.interval": 86400
         },
         {
           "name": "attr/name"
--- a/devices/philips/light_zb3_white_ambiance.json
+++ b/devices/philips/light_zb3_white_ambiance.json
@@ -76,7 +76,8 @@
             "ep": "0x0b",
             "cl": "0x0000",
             "at": "0x0005"
-          }
+          },
+          "refresh.interval": 86400
         },
         {
           "name": "attr/name"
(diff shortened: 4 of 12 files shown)


ebaauw (commented):
Cool, I saw the same, but suspected the legacy code.  Didn't yet look into it.

Shouldn't we just put the `refresh.interval`, `read`, and `parse` in the generic item files, and leave the device files with only the `name` (as you've done for the motion sensors?).

SwoopX:
Well, good question. This shouldn't even be necessary, as the current code base is taking care of querying the manufacturer name and model ID as part of the pairing procedure. That's the reason why the generic items do not contain any functions as far as I can tell.

As some of the Philips DDFs have introduced a productname resource item, which apparently requires the content of manufacturer name and modelID, I didn't want to touch anything there. Maybe that part ought to be revised? Not sure if prepopulated manufacturer name and modelID would work out here.

I personally would prefer not to add the functions to the generic item, as I consider the noise to benefit ratio to be negative.

ebaauw:
> As some of the Philips DDFs have introduced a product name resource item, which apparently requires the content of manufacturer name and modelID

Strictly, it doesn't.  0xC0 and 0xC1 return a byte stream, encoding the manufacturer name, model ID, and product name, amongst others, in 40-byte chunks.  0xC0 is the request, specifying the offset, and 0xC1 is the response.  I use the length of the manufacturer name and model ID to specify the offset, so the payload of the respomse starts with the product name.  In theory, we could simply start at offset 0, but then we need to send multiple 0xC0 commands and stitch the chunks from multiple 0xC1 responses together and parse that.  The parsing is easy, and already implemented in `0000_capabilities.js`, but I wouldn't know how to handle the dialogue of multiple request and response commands from DDF.

> Not sure if prepopulated manufacturer name and modelID would work out here.

Only if we do a DDF per model ID (or strictly speaking: per length of the model ID) times "Philips" vs "Signify Netherlands B.V.".  Not looking forward to maintain that.

Anyway, I would think both manufacturer name and model ID will have been read already, or the DDF wouldn't even have been matched?

ebaauw:
Looks like basically I screwed up not specifying the `refresh.interval` on `attr/modelid`?   In that case I'd better double check the other DDFs I created.

SwoopX:
Yeah, as always, tiny detail with potentially large impact.

I'd say let's keep it as it'll be after the PR. You rightfully concluded that the DDFs get assigned through the intergrated queries for manufacturer name and modelid. And the procedure you described above is not really a desired state (meaning amendment to end up in maintenance hell), better not touch it and leave a handful of DDFs with added queries.

manup (approved):

manup:
From my side PR looks good to merge as well :+1:  please feel free to squash merge if nothing more needs to be added.

manup:
@ebaauw @SwoopX ready to merge? :)

PR #7396, opened 2023-11-28
Update xiaomi_ssm-u01_t1_switch.json

MSL-DA:
There is no need for the ZHASwitch resource


SwoopX:
I'm afraid I do not agree to this change. It e.g. removes the ability to configure the device. If one does not desire the switch resource, the DDF can be copied to the locally persistent location with individual changes.

SwoopX (commented):
I'm afraid I do not agree to this change. It e.g. removes the ability to configure the device. If one does not desire the switch resource, the DDF can be copied to the locally persistent directory with individual changes.

MSL-DA:
> switch resource

Can you tell what opportunities are lost by omitting the switch resource?

SwoopX:
Please check the code you've removed for that. I can also be that the device can generally offer more options, which haven't been exposed yet, either due to the fact they're unknown or they're suspected to be available but not tested.

PR #7405, opened 2023-11-29
Restore `parse` functions for `maufacturername` and `modelid`

ebaauw:
As discussed in Discord (**general** channel).  `parse` functions were removed in #6878.  This causes issues paring devices, as DDFs are matched an `manufacturername` and `modelid`.  The corresponding Zigbee attribites are read from C++ code alright, so a `read` function isn't needed.  But the resource items were no longer populated, due to the `parse` function missing.

### DIFF
--- a/devices/generic/items/attr_manufacturername_item.json
+++ b/devices/generic/items/attr_manufacturername_item.json
@@ -8,7 +8,7 @@
   "description": "Manufacturer name of the device.",
   "parse": {
     "fn": "zcl:attr",
-    "ep": 0,
+    "ep": 255,
     "cl": "0x0000",
     "at": "0x0004",
     "eval": "Item.val = Attr.val"
--- a/devices/generic/items/attr_manufacturername_item.json
+++ b/devices/generic/items/attr_manufacturername_item.json
@@ -5,5 +5,12 @@
   "access": "R",
   "public": true,
   "implicit": true,
-  "description": "Manufacturer name of the device."
+  "description": "Manufacturer name of the device.",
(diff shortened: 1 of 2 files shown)


manup:
Almost there, did a test but the `"ep": 0`  needs to be `"ep": 255`  like in the version before #6878 (also known as any/broadcast endpoint) otherwise the incoming request won't be matched since there are no sub-resources created before DDF matching, from which the `0` endpoint is usually resolved.

After that my test device joined and was created fully within 2 seconds.

ebaauw:
Hm, there are plenty of other generic items with `"ep": 0`.  I suppose these need to be 255 as well?

manup:
> Hm, there are plenty of other generic items with `"ep": 0`. I suppose these need to be 255 as well?

No this case here is special, the 255 is needed since these are evaluated before the DDF is loaded during join.
The 0 (auto) endpoint derives the endpoint from the sub-resource `uniqueid`.

ebaauw:
I do see some more `"ep": 255.  Should these be 0, then?  Generic items for `swversion`, `state/x`, and `state/y`, and a couple of device files.

manup:
Should be fine for `swversion` we might even add this to the initial queries before DDF loading.
For `state/x` and `state/y` 0 makes sense, **but** it also requires that the item is added to the sub-resource for which the incoming command has the same source endpoint as the one in the `uniqueid`. So 255 is not too bad here since it can be used regardless of the endpoint, and for lights we usually only have one color endpoint.

PR #7422, opened 2023-12-09
Add the SNZB-02P variant of the sensor to the DDF

Monofin:
Sonoff have aded further variations of the same temp/humidity sensor.
The existing DDF works well with them.


Monofin:
Tested as working with deCONZ 2.24.2, (devices detected by the Phoscon interfaces), correctly showing temperature and humidity.

SwoopX (approved):

PR #7445, opened 2023-12-11
DDF for Ikea Parasoll open/close sensor

SwoopX:
Mostly standard device. However, it still needs to be verified if it can work with groups, as mentioned in the respective issue.

### DIFF
--- a/devices/ikea/parasoll_open_close_sensor.json
+++ b/devices/ikea/parasoll_open_close_sensor.json
@@ -1,0 +1,143 @@
+{
+  "schema": "devcap1.schema.json",
+  "manufacturername": "$MF_IKEA",
+  "modelid": "PARASOLL Door/Window Sensor",
+  "product": "PARASOLL Door/Window Sensor",
+  "sleeper": true,
+  "status": "Gold",
+  "subdevices": [
+    {
+      "type": "$TYPE_OPEN_CLOSE_SENSOR",
+      "restapi": "/sensors",
+      "uuid": [
+        "$address.ext",
+        "0x02",
+        "0x0500"
+      ],
+      "fingerprint": {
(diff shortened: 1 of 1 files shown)


ebaauw:
Picked one up today.  My observations:
- It can work with groups alright: just bind the client _On/Off_ cluster on endpoint 0x01 to the group.  Consequently, `config/group` should be exposed;
- It supports _Idendity_, so `config/alert` should be exposed;
- I still see `config/pending` - maybe that's the legacy code interfering?
- Did you actually see _Battery_ reported in the _Zone Status_ or is `state/lowbattery` an assumption?
- It doesn't send supervisory reports, so we need to configure periodic reporting on _Zone Status_.
- _Image Type_ is 0x3277.

SwoopX:
@ebaauw Feel free to set up a follow up PR to improve the device. I don't own it, so it was the "typical" remote integration purely based on the respective issue.

As for `config/pending`, this is still a must for the IAS code to work properly, `state/lowbattery` is indeed an assumption. In my view, we wouldn't need any periodic zone status reporting, as the device would go with `config/duration` which I just note by writing these lines I have forgotten to include. If memory serves, durationdue would take care of resetting the state after the period of `duration` has passed.

VincentvdH-91:
Hi there, I'm quite new to using DDF files so sorry for asking 'stupid' questions. I've tried to find instructions how to use them, but am not able to find anything.

I've just tried to add the Parasoll sensor to deCONZ (running on RPI 3B with Domoticz 2024.2), I've added the parasoll.json DDF file into the "devices" folder on deCONZ, but the sensor won't show up either in Phoscon or Domoticz. I've tried to add it via the 'new sensor' function in Phoscon, but still it doesn't show up.

Thanks in advance!

PR #7452, opened 2023-12-14
DDF support "cmd": "any" wildcard

manup:
The `"any"` wildcard allows forwarding any ZCL command ID on a cluster to the same Javascript handler in `"zcl:cmd"`. Note the command is still filtered for ClusterID. This can be useful for switch button handlers until a more lightweight solution without Javascript will be available.

Warning: This should **not** be used as a default  way to parse different commands leading to large Javascript blobs.

### DIFF
--- a/device_access_fn.cpp
+++ b/device_access_fn.cpp
@@ -16,6 +16,8 @@
 #include "resource.h"
 #include "zcl/zcl.h"
 
+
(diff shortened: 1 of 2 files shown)


Zehir:
> @Zehir can you please add the "any" value for the "cmd" key to the validator, so it supports both: hex values and "any".

I have updated the validator and the github action

PR #7486, opened 2023-12-23
DDF for innr SP 240

jmue:
add DDF for innr SP 240 plug with consumption an power sensor


ebaauw (changes requested):
Finally, a new innr plug with power monitoring.  And, according to its specs, it supports up to 16A.  Ordered a couple to try them out, and see if they'll handle my washing machine and dryer.

As for the DDF (somehow, I cannot enter comments in-line):
- The refresh interval for `state/on` should be 360:  the periodic reporting interval plus 60s;
- Doesn't the plug report voltage and current (amperage) on 0x0B04.  The SP120 did;
- It doesn't make sense to create a binding and attribute reporting for _Level Control_ (0x0008), since _Current Level_ isn't reported through the API.  Typically plugs only include _Level Control_ to be able to react to _Move to Level (with On/Off)_; they's don't report _Current Level_, other than 0 or 254, corresponding to _On/Off_.
- The SP120 also supported attribute reporting on 0x0B04 and 0x0702.

jmue:
Fixed, ​could you please explain why I should add 60 seconds?

jmue:
I'm new to DDF, attributes and stuff. Could you point me in the right direction as to what should be added, what should be changed and why it is necessary.
Thank you.

jmue:
Yes, the plug reports current, power, voltage and consumption.

ebaauw:
See the DDF for the SP 120 as example; I would expect the SP 240 to be similar.

> Fixed, ​could you please explain why I should add 60 seconds?

As safety margin.  Ideally, the plug reports its attributes and deCONZ won’t need to poll the device.  Polling is configured only as fallback, in case the reporting fails.

ebaauw (approved):
I think this should do it.

SwoopX (approved):

PR #7493, opened 2023-12-26
Fix `config/mode` for ubisys devices S1/S2(-R)/J1(-R)

SwoopX:
So, apparently, the code is stepping on its own foot here. For a DDF to be able to really use config/mode, a special meta item needs to be set, which is missing for all ubisys devices offering a ZHASwitch resource. However, the legacy code for setting the mode is still available, but is never reached due to code precedence.

The PR satisfies now all requirements for the DDF code and works with the respective configuration values. Additionally, the respective legacy code is removed, as it wasn't used anyway anymore.


manup:
@Zehir looks like the validator needs a small addition, the `"fc"` keyword here is an optional hex value that can be given to read, write, parse functions to specify a non default Zigbee frame control value.

Zehir:
@manup I have updated the validator, you can re run the checks

manup:
> @manup I have updated the validator, you can re run the checks

Cool thanks, all green now :+1:

PR #7497, opened 2023-12-27
Force classic join/re-join for devices

manup:
This mainly affects ConBee III but could also affect ConBee II when the setup is downgraded from ConBee III.

Force old school Zigbee join/re-join without APS level per device link keys for now via Node Descriptor server mask. Since the random generated per device link keys aren't stored or backuped anywhere, this caused problems for rejoining devices of some brands and perhaps battery drain. For proper support of per device APS link keys we need further code to store them in database and forward to firmware.


xxLeoxx93:
I'm facing the high battary drain using Conbee 3 (and downgrade to 2) in combination with Homeassistant/ZHA. Is this now fixed wit hthe latest firmware?

PR #7521, opened 2024-01-07
DDF for Hue lightstrip plus (LCL006)

jmue:
Include Hue Lightstrip Plus (LCL006)


ebaauw:
Best add it under `"Philips"` as well.

PR #7522, opened 2024-01-07
DDF for Osram Plug 01

jmue:
add ddf for osram plug 01

### DIFF
--- a/devices/osram/plug_01.json
+++ b/devices/osram/plug_01.json
@@ -42,12 +42,11 @@
           "name": "attr/uniqueid"
         },
         {
-          "name": "state/alert",
-          "default": "none"
+          "name": "state/alert"
         },
         {
           "name": "state/on",
-          "refresh.interval": 5
+          "refresh.interval": 360
         },
         {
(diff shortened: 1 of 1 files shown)


devices/osram/plug_01.json
```diff
@@ -0,0 +1,74 @@
...
+        },
+        {
+          "name": "attr/type"
+        },
+        {
+          "name": "attr/uniqueid"
+        },
+        {
+          "name": "state/alert",
+          "default": "none"
```
SwoopX, line 46: The default here can be deleted :)

devices/osram/plug_01.json
```diff
@@ -0,0 +1,74 @@
...
+        },
+        {
+          "name": "state/on",
+          "refresh.interval": 5
```
SwoopX, line 50: I'd recommend to set the interval to 360 👍

SwoopX (changes requested):
Really 2 minors and then we're good to go here 👍

SwoopX (approved):

PR #7530, opened 2024-01-14
Fix LIDL smart plug consumption reporting 0

Pirionfr:
fix consumption reporting 0 because of missing semicolon

### DIFF
--- a/devices/lidl/hg08673.json
+++ b/devices/lidl/hg08673.json
@@ -180,7 +180,7 @@
             "at": "0x0000",
             "cl": "0x0702",
             "ep": 1,
-            "eval": "item.val = 10 * Attr.val"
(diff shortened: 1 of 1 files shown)


manup:
Hmm sure about this? It the semicolon shouldn't make any difference here.

Pirionfr:
I added it in my ddf and data was available in the api

Pirionfr:
after deconz plugins  restart is home assistant it's doesn't increase over the api. so you are right

Pirionfr:
image
but in the API
image

mattreim:
Please try this: "Item.val = Attr.val * 10;"

I think it's because of the small "i".

Pirionfr:
it's as been fixes in last beta version.

Issue #7490, opened 2023-12-24
New version of Hue Wall Switch (RDM004, 9290030171A)


koenheemskerk:
### Is there already an existing issue for this?

- [X] I have searched the existing issues and there is none for my device

### Product name

Philips Hue Wall Switch (2023)

### Manufacturer

Philips

### Model identifier

RDM004

### Device type to add

Switch

### Node info

![IMG_0546](https://github.com/dresden-elektronik/deconz-rest-plugin/assets/39374992/42f98019-fa19-4883-baf9-276ecb02329c)

### Endpoints and clusters + Basic

<img width="797" alt="Screenshot 2023-12-29 at 22 41 46" src="https://github.com/dresden-elektronik/deconz-rest-plugin/assets/39374992/cd0ed6e1-1c3b-4a4e-bbc0-d1a9d6cb1384">

<img width="492" alt="Screenshot 2023-12-29 at 22 45 06" src="https://github.com/dresden-elektronik/deconz-rest-plugin/assets/39374992/76a405ff-6dc5-4f57-8b71-28d451d5caa6">

<img width="482" alt="Screenshot 2023-12-29 at 22 45 17" src="https://github.com/dresden-elektronik/deconz-rest-plugin/assets/39374992/33af41e5-23bc-4623-8536-eebdc9e22574">

Mimiix:
Can you please add all screenshots?

ebaauw:
Where did you get it?  I cannot seem to find any mention of a new model.  Particularly interested to know if this model does support hybrid mode (as the previous model doesn’t).

koenheemskerk:
Added the screens!

Syorn630:
I've bought two of these RDM004 Philips hue wall switch modules this week as well and faced the same issue when trying to connect them to deconz via the Phoscon web app. Searching for a solution finally brought me here.

Has anybody an idea how to get the new version working? At least a quick fix? I've tried to copy the DDF file of the RDM001 version and just changed the Model ID. This got me one step further: the switch module now shows up in Phoscon but unfortunately it doesn't transmit anything when the switch is toggled.

When looking at the API events, every 5 minutes the following two events are received:

```
"16:26:27:436": {
    "attr": {
        "id": "6",
        "lastannounced": "2023-12-31T15:26:27Z",
        "lastseen": "2023-12-31T15:25Z",
        "manufacturername": "Signify Netherlands B.V.",
        "modelid": "RDM004",
[...]

PR #7567, opened 2024-02-02
New `modelid` values for Hue smart button and Hue wall switch module

ebaauw:
No changes to the devices themselves, just new _Model Identifier_ values.

See also https://hueblog.com/2024/01/12/philips-hue-wall-switch-module-new-revision-available/.

See #7490.


kitas1304:
thanks it works! this wall switch cost me 5 hours of my life.

Issue #7551, opened 2024-01-27
Philips Hue White Ambiance (LTW001) change state to „on“ automatically


FlatTV:
### Does the issue really belong here?

- [X] I definitively want to report a bug within deCONZ or its REST-API

### Is there already an existing issue for this?

- [X] I have searched the existing issues and there is none for the bug at hand

### Describe the bug

After switching off the lamp using the Deconz API, the status in the API changes back to “On” after a few seconds. In reality the lamp is still off.

### Steps to reproduce the behavior

1. Switch lamp to „off“ in deconz-API
2. The lamp switch to „off“ and the state in deconz-API switch to „off“.
3. wait and do nothing…
4. The deconz API state switch to „on“ but the lamp is still „off“.

### Expected behavior

just nothing

### Screenshots

[https://forum.phoscon.de/t/ltw001-light-shows-on-after-vew-seconds-in-deconz-api-but-the-real-state-is-off/4378](url)

### Environment

- Host system: (Raspberry Pi)
- Running method: (Raspbian)
- Firmware version: (26720700)
- deCONZ version (not Home assistant Addon version!): (2.25.1)
- Device: (ConBee II)
- Do you use an USB extension cable: (no)
- Is there any other USB or serial devices connected to the host system? If so: Which? (no)

SwoopX:
Hm, interesting case. Could you please fetch some debug logs?

#### Option 1:
To get debug logs in a GUI installation, open deCONZ and click `Help`. After that, click `Debug View`. The following debug levels need to be enabled for proper logging: `INFO`, `INFO_L2`, `ERROR`, `ERROR_L2`, `APS`, `APS_L2`.
Please note that in the HA Addon and Docker webVNC, you can't copy paste. You need to use a Native VNC Client. [HA Addon How-To](https://community.home-assistant.io/t/deconz-addon-how-to-enable-vnc-set-port-in-network-configuration/137597), [Docker Environment Vars](https://github.com/deconz-community/deconz-docker#environment-variables).

#### Option 2:
To get debug logs in a headless installation / ssh session, use the following commands:

```
sudo systemctl stop deconz-gui
sudo systemctl stop deconz
/usr/bin/deCONZ -platform minimal --dbg-aps=2 --dbg-info=2 --dbg-error=2 | tee debug.txt
```

[...]

PR #7576, opened 2024-02-05
Fix handling Hue Effects state attribute

ebaauw:
Fixes parsing the _State_ attribute of the _Hue Effects_ cluster, see #7551
 - Add support for mode 0x0007 with values for `on`, `bri`, and `ct`;
- Make sure to return value for `on`, also for unsupported modes.

(Dunno why it lists all these commits, but only one changed file).


manup:
Looks fine but I wouldn't include the eslint comments.

ebaauw:
I would, obviously, but removed the comment.

I would hope the JS is parsed and compiled only once?  Or is it compiled at every invocation (in which case, we might want to remove white space, comments, use short variable names, etc)?

manup:
Thanks, imho such things better go into a $EDITOR /eslint user.preferences or we provide a global one.

>I would hope the JS is parsed and compiled only once? Or is it compiled at every invocation (in which case, we might want to remove white space, comments, use short variable names, etc)?

Right now it is actually *parsed, compiled and executed* on each invocation. I've looked into DuktapeJS docs and it is indeed possible to safe/reload the generated bytecode which would safe the repeated parsing and compile steps (still much slower than a V8 engine JIT, but safes a few cycles). I'm planning to test this out more but first some other refactoring needs to be done to cut more drastic bottle necks.

You can get a feeling on the performance by enabling MEASURE and JS debug logging. Even with the non optimized approach it's surprisingly fast on average and mostly uses less than 250 KB on RAM in total. We currently have a lot of other more heavier performance hogs in the C++ side of the code base.

But as said earlier for me JS should be kept to a minimum and really boring noob safe code :)
The main purpose is to do things like `Item.val = Attr.val * 10`. It's not always possible but 95% of our JS code should be like this and not much more complex, until a better solution is found for the remaining 5%.

From my point of view the more complex JS scripts we have are only temporary to figure out generic ways later on to tackle the tasks they do in a declarative way.

Issue #7454, opened 2023-12-14
IKEA Vallhorn motion sensor device request 


MagnusKazen:
### Is there already an existing issue for this?

- [X] I have searched the existing issues and there is none for my device

### Product name

IKEA Vallhorn wireless motion sensor

### Manufacturer

IKEA of Sweden

### Model identifier

Vallhorn wireless motion sensor

### Device type to add

Sensor

### Node info

![image](https://github.com/dresden-elektronik/deconz-rest-plugin/assets/48179782/0e0a4c17-50dc-407b-846e-dca6a6707e5e)

### Endpoints and clusters

![image](https://github.com/dresden-elektronik/deconz-rest-plugin/assets/48179782/cc7cdc19-a34b-4ad0-a437-d104b078b955)

### Basic

![image](https://github.com/dresden-elektronik/deconz-rest-plugin/assets/48179782/06e86fa8-6316-4b62-8e49-aa30f17a2461)

### Power Configuration
![image](https://github.com/dresden-elektronik/deconz-rest-plugin/assets/48179782/c3d6ae48-784e-4975-b8db-6db1943cc2cc)

### On/Off
![image](https://github.com/dresden-elektronik/deconz-rest-plugin/assets/48179782/e2d1ce5f-263e-41b5-81b9-7d3813ad2a90)
Just blank, seems to be no info here.
### Level Control
None
### Color Control
None
### Thermostat
None
### Simple Metering
None
### Electrical Measurement
None
### Any other cluster of relevance/interest
![image](https://github.com/dresden-elektronik/deconz-rest-plugin/assets/48179782/3348bc6d-283d-41c1-b777-4bd9ee73d605)
![image](https://github.com/dresden-elektronik/deconz-rest-plugin/assets/48179782/06864be7-1dbb-403e-a970-e0429829f1b2)
![image](https://github.com/dresden-elektronik/deconz-rest-plugin/assets/48179782/8797a38b-acb6-489a-819c-0680d86dbfe8)
![image](https://github.com/dresden-elektronik/deconz-rest-plugin/assets/48179782/a6fba5d4-d499-4b9e-954c-a46934558ce3)
![image](https://github.com/dresden-elektronik/deconz-rest-plugin/assets/48179782/05f12053-9389-455f-b111-68edcda6b285)

Just let me know if there is anything more You need. I'm pretty new at this...

hyper2910:
any idea when Vallhorn is implemented?

[...]

PR #7584, opened 2024-02-09
IKEA Vallhorn add missing bindings for light

noud-github:
initial PR missed some bindings see

https://github.com/dresden-elektronik/deconz-rest-plugin/issues/7454


devices/ikea/vallhorn_wireless_motion_sensor.json
```diff
@@ -239,7 +243,8 @@
             "cl": "0x0400",
             "at": "0x0000",
             "script": "0400_lightlevel.js"
-          }
+          },
+          "refresh.interval": 86400
```
SwoopX, line 247: Could you please remove the refresh interval again? It isn't considered anyway here and is therefore superfluous. And please do not forget the comma above 😉

SwoopX (changes requested):
Hey, thanks for taking action already. Just some cosmetics from my side: Would be greate to remove the `path` stanza in line 8 and to take the comment into account. With that, we're ready to go here 👍

SwoopX (approved):
Perfect, thanks!

PR #7602, opened 2024-02-11
Fix unnecessary read requests for some Xiaomi devices

SwoopX:
Some read functions were without `refresh.interval`, potentially resulting in excessive unwanted read requests.


mattreim:
Is it the same with "xiaomi_ssm-u01_t1_switch"?

mattreim:
ssm-u01_t1_clickmode

ssm-u01_t1_devicemode

There seem to be quite a few queries.

SwoopX:
That is perfectly fine. An interval is defined here.

mattreim:
Ok

Issue #7546, opened 2024-01-22
OSRAM bulbs turn on randomly and unexpected


danube:
### Does the issue really belong here?

- [X] I definitively want to report a bug within deCONZ or its REST-API

### Is there already an existing issue for this?

- [X] I have searched the existing issues and there is none for the bug at hand

### Describe the bug

I have a network with 38 nodes, including several E27 lights. Five of them are from OSRAM and give me (and my wife) a serious headache, as they turn on unexpectedly.

These OSRAM bulbs are installed already since a while and have worked without problems, but since about end of December, I'm experiencing this issue. Maybe some software update raised it, but I really can't say which.

I changed from ConBee II to III but things didn't change.

I attach a log where the event happens today 2024-01-22 at 8:25:17 am. I also attach a ZSHARK capture from within the same time period. Both cropped to 5 seconds +/-. Lamp's UID is `7c:b0:3e:aa:00:af:6b:90-03`.

I referred to https://forum.phoscon.de/t/lights-randomly-turning-on/1679/3, but the creator moved to a Hue system without finding a solution.

I'll proceed capturing the event in the hope of any valuable information. If there is more I could do or try, let me know.

[debug.txt](https://github.com/dresden-elektronik/deconz-rest-plugin/files/14007369/debug.txt)
[nodered.json](https://github.com/dresden-elektronik/deconz-rest-plugin/files/14007374/nodered.json)
[capture.pcapng.gz](https://github.com/dresden-elektronik/deconz-rest-plugin/files/14007388/capture.pcapng.gz)

### Steps to reproduce the behavior

no idea... install OSRAM E27 bulbs and wait?

### Expected behavior

lamps don't turn on unexpectedly

### Environment

- Host system: Raspberry Pi 4
- Running method: Raspbian with Node-RED
- Firmware version: 264F0900
- deCONZ version: 2.24.3
- Device: ConBee III
- Do you use an USB extension cable: yes, 1m
- Is there any other USB or serial devices connected to the host system? no

danube:
Attached are the logs of another event. Lamp UID here is `7c:b0:3e:aa:00:af:6b:90-03`.

May I ask one question. What is that meaning of:
`INSERT INTO resource_items (sub_device_id,item,value,source,timestamp) SELECT id, 'state/on', 'true', 'dev', 1705908316 FROM sub_devices WHERE uniqueid = '7c:b0:3e:aa:00:af:6b:90-03'`

[...]

PR #7608, opened 2024-02-15
Fix OSRAM Classic A60 RGBW crash

manup:
The light does crash when ZCL attribute reporting is configured and the binding table is queried. Likely resulting in a watchdog reboot after which the lights turns on in full brightness.

The fix here is to use polling instead of attribute reporting as well as to disable query of the binding table.

Related https://github.com/dresden-elektronik/deconz-rest-plugin/issues/7546

### DIFF
--- a/devices/osram/classic_a60_rgbw.json
+++ b/devices/osram/classic_a60_rgbw.json
@@ -4,6 +4,7 @@
   "modelid": "Classic A60 RGBW",
   "product": "Classic A60 RGBW",
   "sleeper": false,
+  "supportsMgmtBind": false,
   "status": "Gold",
   "subdevices": [
     {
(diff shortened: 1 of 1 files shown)


ebaauw:
Is the crash caused by (one time) creating the bindings and reporting configuration to by (multiple times) querying these?  In other words: does the crash occur since the introduction of the DDF and final state machine actively checking the binding and reporting?  In that case, wouldn’t it be better to do a passive check (see if reports arrive), as is done for the polling as safety net?

manup:
The creation of the bindings and also the reporting does work, in fact my test light still sends the reports without crashing.
As I understand the sniffer logs the querying of the binding table makes trouble, and also the verification of the ZCL reporting configuration lets the light crash after a while.

With the addition `{ "supportsMgmtBind": false }` to the DDF the binding table queries are suppressed, and the state machine works like the legacy passive checking for reports (which it always does).

However the verification of the ZCL reporting configuration is still done to check that the values are indeed the ones specified in the DDF. This is usually only done once after startup and in case the passive checking detects that something isn't setup. Normally a nice feature that also allows to update a DDF with new saner values if needed.
This can't be deactivated, and currently I don't plan to do so as it would complicate the code just to workaround a few broken devices (so far only Osram and here only a selection of them).

I think using polling here is a ok'ish tradeoff for now. I don't think the old E27 and GU10 lights will be fixed by Osram but current lights on the market, if they show the problems too, should be reported to Osram so they can work on a bug fix firmware.

Finne75:
I think this will solve a long-standing mistery of lights switching on unsolicitadely :) - apologies for non-technical comment

PR #7614, opened 2024-02-15
Add `config/setvalve` item to generic

mattreim:
valve_set


mattreim:
The name "valve_set" fits better to "windowopen_set".

mattreim:
With setvalve I somehow expect this object.

Setvalve

manup:
While the original name doesn't have the ideal consistency this PR would be a breaking change for little gain (I think).

There are a few existing legacy devices out there which were joined with the `config/setvalve` item.

image

Further it changes the API for clients which would need to support with old vs. new deCONZ version.
To be honest I wouldn't do that.

If such a change is to be made it should deprecate the old item name while adding a new one and support both for some time until the old one is removed.

mattreim:
If the effort is too high, I can of course change it back to "setvalue". But it would still be good to use it again in DDF`s.

SwoopX (commented):
Puh, difficult... I guess we should have paid more attention to this one back in the days. However, if we now modify this resource item, we'll probably have more pain than gain in the end.

Clearly, we also have a discrepancy on the capability standing behind that functionality. The given item is missing an item specific json file if I'm not mistaken and expecting bools whereas the changed version here is also of type bool, but intended/meant for integers? On a personal note, I wouldn't consider setting the valve open percentage manually a good idea, the thermostat should do that for me imho.

If it might be about either detecting if a window is open to close the valve or enable such a feature, we do already have a resource item for both cases. So it can be added to the device if missing.

To sum it up: I'd also consider the suggested modification as a breaking change, but adding functionality where we already have items for should be doable.

mattreim:
I've changed it again now.

mattreim:
@manup Should I close this pull request?

SwoopX:
More out of curiosity: what does this function exactly do? I was initially assuming it lets you control the value position in percent, but that's obviously not the case.

PR #7621, opened 2024-02-18
Fix consumption reporting for Lidl SilverCrest (_TZ3000_j1v25l17)

### LINKED ISSUE #7603: LIDL smart plug: no update on consumption value (hg08673.json)
### Does the issue really belong here?

- [X] I definitively want to report a bug within deCONZ or its REST-API

- [X] I have searched the existing issues and there is none for the bug at hand

### Describe the bug

This commit breaks update on consumption value with TS011F
[e2413b8](https://github.com/dresden-elektronik/deconz-rest-plugin/commit/e2413b85ba4c487bd457e9c5ceb720962e5c5538)

### Steps to reproduce the behavior

I fix it locally that way and it works
[https://pastebin.com/z4gNQxWb](https://pastebin.com/z4gNQxWb)

### Expected behavior

.

### Screenshots

.

### Environment

- Host system: (Raspberry Pi)
- Running method: (Raspbian)
- Firmware version: (26xxyy00)
- deCONZ version (not Home assistant Addon version!): (2.26.0)
- Device: (ConBee II)

### deCONZ Logs

_No response_

### Additional context

_No response_

Fischelsberger:
I've create a PR for this issue, @FlatTV provided the fix in #7603.

I've used a little bit different approach, because the missing ";" seemed as an issue compare to a different DDF of a Tuya Device.
Seems like the ";" was missing and also the "Item" was written Lowercase.

resolves #7603 

old:
```
"eval": "item.val = 10 * Attr.val"
```
new: 
```
"eval": "Item.val = Attr.val * 10;"
```

### DIFF
--- a/devices/lidl/hg08673.json
+++ b/devices/lidl/hg08673.json
@@ -175,7 +175,7 @@
         },
         {
           "name": "state/consumption",
-          "refresh.interval": 300,
+          "refresh.interval": 420,
           "parse": {
             "at": "0x0000",
             "cl": "0x0702",
--- a/devices/lidl/hg08673.json
+++ b/devices/lidl/hg08673.json
@@ -180,7 +180,7 @@
             "at": "0x0000",
             "cl": "0x0702",
             "ep": 1,
-            "eval": "item.val = 10 * Attr.val"
+            "eval": "Item.val = Attr.val * 10;"
           }
         },
         {


FlatTV:
I think the lowercase of „Item.val“ is the only problem here

manup:
The former 420 interval is also better since it needs to be higher than the respective reporting interval.

Fischelsberger:
sorry, I think I screwed up my commit now...
Wanted to change the 300 back to 420, and guess that not worked out...

Not used to the usage of git in general...

But it seems fine now (with 2 additional commits...)?

manup:
Multiple commit are fine, we mostly do squash merge the PRs so they get reduced to one commit :)

PR #7630, opened 2024-02-24
DDF for ubisys S1(-R) corrected

Markkuuss:
Currently, the light can only be switched on by push-button and can no longer be switched off if the S1(-R) sensor is configured to "mode": "momentary".

According to https://www.ubisys.de/wp-content/uploads/ubisys-s1-technical-reference.pdf, the values for "InputActions Attribute" are not correct. With the default values on page 29 for push-buttons (momentary), it works correctly again:

image

### DIFF
--- a/devices/ubisys/s1_5501_s1r_5601.json
+++ b/devices/ubisys/s1_5501_s1r_5601.json
@@ -126,7 +126,7 @@
             "cl": "0xFC00",
             "cmd": "0x01",
             "ep": "0xE8",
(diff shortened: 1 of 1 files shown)


SwoopX (approved):
Either the bytes I've copied back in the days were incorrect or I fat-fingered here. Trusting the changes are currently in use by the requestor, as I don't have the device 🙂

torstenmarek:
I think this change broke my setup. When upgrading to a version that included this change, my S1s ended up as "momentary" switches, or maybe that had always been set to "momentary", but didn't behave like it. After changing them to "rocker" via the REST API, everything is back to normal

Markkuuss:
> I think this change broke my setup. When upgrading to a version that included this change, my S1s ended up as "momentary" switches, or maybe that had always been set to "momentary", but didn't behave like it. After changing them to "rocker" via the REST API, everything is back to normal

As you can see in the code, only the value for momentary has changed. the value for the rocker setting has remained unchanged. Whether the S1(-R) is operated in momentary or rocker mode has always had to be set via the REST API. I don't really understand your problem, if you have installed a rocker switch, set it to rocker via the REST API and it works, then everything is fine. The standard for S1(-R) - even on delivery - is momentary push button.

PR #7635, opened 2024-02-25
Philips Hue ZLL/ZB3 match DDF only for newer firmware

manup:
The existing DDFs require that the lights have a recent firmware which provides the 0xFC03 cluster and otherwise won't function fully for polling and updating the light state.

The PR no longer matches DDFs for older firmware lights, and falls back to the legacy code for them.

Later on we can introduce DDFs which also handle the old firmware with standard clusters instead of 0xFC03 cluster.
~~Note this PR only affects the ZLL lights, I'm not sure if also ZB30 lights need this for older firmware?~~

### DIFF
--- a/devices/philips/light_zb3_C.json
+++ b/devices/philips/light_zb3_C.json
@@ -111,6 +111,7 @@
   "product": "Hue white and color ambiance light",
   "sleeper": false,
   "status": "Gold",
+  "matchexpr": "R.hasCluster(0x0b,0xfc03)",
   "subdevices": [
     {
       "type": "$TYPE_EXTENDED_COLOR_LIGHT",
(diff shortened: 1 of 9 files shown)


ebaauw (approved):
I think we might also need this for the ZB3 lights (except the Gradient and Festavia), where effects were introduced at at later stage for the older models.  Note that these do support attribute reporting, unlike the ZLL lights.

manup:
Good point, ok I'll add the line to all light DDFs which use the 0xFC03 cluster.

In a separate PR there will also be an automatic refresh of the ZDP active endpoints and simple descriptors to prevent the false positives due the wrongly added clusters. This will be done in a lazy manner like refresh once per startup (and later on after OTA), so it might require one more reboot to fix old lights to not use the DDFs.

manup:
I've the `matchexp` also to Festavia and Gradient, since the cluster is used in the DDF, or was it always there?

ebaauw:
I won’t hurt, and keeps the DDFs similar.  But I think the cluster has been on those from the introduction, as that’s the way to set gradients.

PR #7650, opened 2024-02-29
Fix temperature reporting for Aqara Single Switch Module T1 (With Neutral) SSM-U01

mattreim:
Aqara Single Switch Module T1 (With Neutral):

Aqara Single Switch Module T1

### DIFF
--- a/devices/xiaomi/xiaomi_ssm-u01_t1_switch.json
+++ b/devices/xiaomi/xiaomi_ssm-u01_t1_switch.json
@@ -125,7 +125,14 @@
           "name": "config/reachable"
         },
         {
-          "name": "config/temperature"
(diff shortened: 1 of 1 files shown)


SwoopX:
Hi,

I'm afraid this one should have never made it into the DDF, this one slipped me. We currently do not have any resource item for device temperatures and using a `config/` item for a state value is not the intended way. Of course, locally, you're absolutely free to do as you please, one of the DDF benefits I might add.

@manup In my view, we should revert this PR.

mattreim:
In this case it was already there and I just repaired it.

mattreim:
device_temperature

Do you now want to remove all device temperatures?

Issue #7579, opened 2024-02-08
deconz V2.25.3 - Signify LTW012 show ON (in phoscon) even though it is OFF


ParanoidiordnA:
### Does the issue really belong here?

- [X] I definitively want to report a bug within deCONZ or its REST-API

### Is there already an existing issue for this?

- [X] I have searched the existing issues and there is none for the bug at hand

### Describe the bug

After Updating from version 2.24.3 to version 2.25.3 i experienced that two (of two) Signify LTW012 (FW 1.108.7) light bulbs appear in phoscon as ON even though they aren't. Turning the devices OFF in phoscon doesn't help. After a short time phoscon shows the devices as ON again.
Reverting back to 2.24.3 brings back normal behavior of the devices.

### Steps to reproduce the behavior

Connect a Signify LTW012 to phoscon, use deconz 2.25.3, wait for a few seconds

### Expected behavior

Same behavior of the devices as in deconz 2.24.3

### Environment

- Host system: Raspberry Pi
- Running method: Raspbian 
- Firmware version: 26720700
- deCONZ version 2.25.3
- Device: ConBee II
- Do you use an USB extension cable: yes 
- Is there any other USB or serial devices connected to the host system? NO

ebaauw:
Could you please check what happens when:
- You read the attributes of the _Hue Effects_ cluster in the GUI?
- You read the attributes of the  _On/Off_ cluster in the GUI?

If these don’t correct the state, please attach screenshots of these clusters.

ebaauw:
That confirms the light reports as off (and 25% brightness and `ct` value of 400), as would be expected.

My LTW012 (E14 candle) is (still?) on 1.108.5; I didn't realise there's a newer firmware, but looking at the screenshots, that part hasn't changed.

Does the GUI show the LTW012 as supported by DDF?  Do you see any errors in the deCONZ log, when enabling ERROR, DDF, and JS in the debug view?  What's the debug log when reading the _Hue Effects_ cluster?

Just to be sure: did you check other API clients than Phoscon (to rule out browser cache issues)?

SwoopX:
@ParanoidiordnA Please change the issue name to something more suited to describe what you want to report. It should allow the devs as well as other users to find and follow what's discussed here. Thanks! 🙂

Mimiix:
[...]

PR #7652, opened 2024-03-02
Mode in State attribute in Hue Effects cluster as bitmap

ebaauw:
As mentioned https://github.com/dresden-elektronik/deconz-rest-plugin/issues/7579#issuecomment-1934275146, the `mode` part of the _State_ attributes in the _Hue Effects_ cluster appears to be a bitmap.  This PR simplifies the Javascript that parses the _State_ attribute accordingly.

~~I want to run and test this for a while, before finalising the PR.~~

I have been running my installation with the changed fc03_state.js for two weeks now, and haven't seen any issues.


manup (approved):
Looks good and is likely also a bit faster to process.
Bit juggling is always fun to decipher :)

manup:
Changed the mile stone since v2.26.1-beta is already released.

PR #7672, opened 2024-03-17
Improve rules database handling

manup:
- Only store rules than needed and on application exit.
- Fix store/reload `timestriggered` and `lasttriggered` which was always "none" previously.


sadpixel:
Could persisiting these values across service restart be offered as an option? Maybe toggled by a new entry under /config? (Or perhaps it would make more sense as a new CLI flag…)

manup:
Reverted the loading at startup in PR #7678 for my purpose it's enough to have the last session recorded in the database.

PR #7681, opened 2024-03-26
Add cooling mode to Bosch room thermostat II (BTH-RM)

mattreim:
image

### DIFF
--- a/devices/bosch/room_thermostat2.json
+++ b/devices/bosch/room_thermostat2.json
@@ -401,7 +401,7 @@
             "at": "0x0000",
             "cl": "0x0405",
             "ep": 1,
-            "eval": "Item.val = Attr.val + R.item('config/offset').val;",
+            "eval": "Item.val = (Math.round(Attr.val/100)*100) + R.item('config/offset').val;",
             "fn": "zcl:attr"
           }
         },
--- a/devices/bosch/room_thermostat2.json
+++ b/devices/bosch/room_thermostat2.json
@@ -92,8 +92,7 @@
         },
         {
           "name": "config/coolsetpoint",
-          "refresh.interval": 360,
-          "default": 7
+          "refresh.interval": 360
         },
         {
           "name": "config/checkin",
@@ -150,7 +149,7 @@
             "at": "0x4042",
             "cl": "0x0201",
             "ep": 1,
-            "eval": "Item.val = Attr.val;",
+            "eval": "Item.val = Attr.val",
             "fn": "zcl:attr",
             "mf": "0x1209"
           },
@@ -158,8 +157,7 @@
         },
         {
           "name": "config/heatsetpoint",
-          "refresh.interval": 360,
-          "default": 5
+          "refresh.interval": 360
         },
         {
           "name": "config/locked",
@@ -208,14 +206,14 @@
             "cl": "0x0201",
             "dt": "0x28",
             "ep": 1,
-            "eval": "Item.val / 10;",
+            "eval": "Item.val / 10",
             "fn": "zcl:attr"
           },
           "parse": {
             "at": "0x0010",
             "cl": "0x0201",
             "ep": 1,
-            "eval": "Item.val = Attr.val * 10;",
+            "eval": "Item.val = Attr.val * 10",
             "fn": "zcl:attr"
           }
         },
@@ -278,7 +276,7 @@
             "at": "0x5000",
             "cl": "0x0201",
             "ep": 1,
-            "eval": "Item.val = Attr.val;",
+            "eval": "Item.val = Attr.val",
             "fn": "zcl:attr",
             "mf": "0x1209"
           }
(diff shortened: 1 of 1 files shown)


manup:
Tricky, I don't have any experience with that device, but it looks like the PR introduces a breaking change due the renamed values? Is the removal of "auto" and "heat" really needed or is it better to only extend the existing modes with the new ones?

mattreim:
ioBroker_room_thermostat


I think renaming auto is not necessary, but heat in manual is more understandable here. Incorrect operation could occur if "heat" is used twice.

devices/bosch/room_thermostat2.json
```diff
@@ -28,7 +28,7 @@
       },
       "meta": {
         "values": {
-          "config/mode": {"auto": 0, "heat": 1, "off": 5}
+          "config/mode": {"auto": 0, "manual": 1, "off": 5}
```
SwoopX, line 31: Please do not change the names here. Those are aligned with the respective zigbee attribute and REST API clients do not expect different values. We should stay consistent here.

devices/bosch/room_thermostat2.json
```diff
@@ -148,19 +178,19 @@
             "cl": "0x0201",
             "dt": "0x30",
             "ep": 1,
-            "eval": "if (Item.val == 'auto') { 0 } else if (Item.val == 'heat') { 1 } else if (Item.val == 'off') { 5 };",
+            "eval": "if (Item.val == 'auto') { 0 } else if (Item.val == 'manual') { 1 } else if (Item.val == 'off') { 5 };",
```
SwoopX, line 181: Same as above

devices/bosch/room_thermostat2.json
```diff
@@ -90,12 +90,42 @@
...
+          },
+          "parse": {
+            "at": "0x001C",
+            "cl": "0x0201",
+            "ep": 1,
+            "eval": "if (Attr.val == 0) { Item.val = 'off' } else if (Attr.val == 3) { Item.val = 'cool' } else if (Attr.val == 4) { Item.val = 'heat' };",
+            "fn": "zcl:attr"
+          },
+          "default": "heat"
+        },
```
SwoopX, line 128: Is this confirmed to be working? Bosch already seems to use a manufacturer specific attribute to set the heating modes and apparently disregards the standard attribute.

devices/bosch/room_thermostat2.json
```diff
@@ -148,19 +178,19 @@
...
             "at": "0x4007",
             "cl": "0x0201",
             "ep": 1,
-            "eval": "if (Attr.val == 0) { Item.val = 'auto' } else if (Attr.val == 1) { Item.val = 'heat' } else if (Attr.val == 5) { Item.val = 'off' };",
+            "eval": "if (Attr.val == 0) { Item.val = 'auto' } else if (Attr.val == 1) { Item.val = 'manual' } else if (Attr.val == 5) { Item.val = 'off' };",
             "fn": "zcl:attr",
             "mf": "0x1209"
           },
-          "default": "heat"
+          "default": "manual"
```
SwoopX, line 193: Same as above

SwoopX (changes requested):
Not too sure if we're there yet. Please take the comments into account, thanks!

devices/bosch/room_thermostat2.json
mattreim: Why? I don't understand. Here are two screenshots of unchanged devices.

Eurotronic SPZB0001:
Eurotronic_SPZB0001

Bosch radiator thermostat II:
Bosch_radiator_thermostat_II

devices/bosch/room_thermostat2.json
mattreim: At least it works for me. The two channels "heat/cool" then also use the normal modes: auto, manual(heat) and off.


The attribute “devicemode” was also obvious to me.

mattreim: But what do I do here with heat and the off(450) problem?

devices/bosch/room_thermostat2.json
mattreim: How about if I use mode/preset instead of devicemode/mode?

mode --> heat, cool and off
preset --> manual, auto and holiday

devices/bosch/room_thermostat2.json
mattreim: How about if I use mode/preset instead of devicemode/mode?

mode --> heat, cool and off
preset --> manual, auto and holiday

SwoopX:
I'm sorry, but I cannot say that the recent changes can get my vote. My perception is that thing got even more complicated and drift away from common user experience. Is there any particular reason why you chose to ignore the requested changes?

If I haven't overlooked anything, the amendments you've made over time state that all the required mode changes can be done through attribute 0x001C, being the zigbee intended way. That makes we wonder why not to use it?

mattreim:
Sorry, maybe I didn't make myself clear. Here are the device clusters

0x001C:

0x001C

0x4007:

0x4007

mattreim:
The different states:

(Mode --> devicemodes)

heat --> manual, auto and off

cool --> manual, auto and off

off--> off

mattreim:
They're just names and it works like that.

mattreim:
Inspired by ZHA;

Home_Assistant

mattreim:
@SwoopX Are the changes OK now?

PR #7696, opened 2024-04-07
Improve DDF for dresden elektronik's kobolt

ebaauw:
- Add missing `capabilities` and `config` items;
- Use attribute reporting for `state/on` and `state/bri` - any reason why this was disabled in favour of polling?

### DIFF
--- a/devices/dresden_elektronik/kobold.json
+++ b/devices/dresden_elektronik/kobold.json
@@ -43,42 +43,51 @@
           "name": "attr/uniqueid"
         },
         {
+          "name": "cap/bri/move_with_onoff"
+        },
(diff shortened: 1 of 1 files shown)


ebaauw:
Hm, seeing this every five minutes, and by the looks of it, also on change reports:
```
 Warning duplicated zcl.cmd: 0x0A, cluster: 0x0006 with zcl.seq: 57 for dresden elektronik / Kobold
```
The `zcl.seq` value changes between messages, not sure what is duplicated.  Do I need to sniff the traffic, or is this a known issue?
Kobolt on firmware 0102.100000026

PR #7701, opened 2024-04-07
Update _TZ3000_bmfw9ykl_vibration_sensor.json

ringelbaer:
Added new Manufacturer of TS0210 after local testing. _TZ3000_lqpt3mvr
Screenshot 2024-04-07 222111

Seller: https://de.aliexpress.com/item/1005006503967266.html


mattreim:
The corresponding modelid must always be added to a manufacturername.

SveLil:
Would you mind including "_TZ3000_lzdjjfss"  as another manufacturer id for the same sensor? Same modelid (TS0210), same look

PR #7704, opened 2024-04-10
Fix state of Hue lights not always updated correctly

ebaauw:
Damn, #7652 is not working as intended.  Bloody operator precedence in Javascript.

I missed this during testing, since, apparently, my lights were configured with bindings and attribute reporting for _Color Control_.  Apparently, deCONZ won't remove old bindings no longer defined in the DDF?

### DIFF
--- a/devices/philips/fc03_state.js
+++ b/devices/philips/fc03_state.js
@@ -12,23 +12,23 @@ if (attrid === 0x0002) {
       const mode = ZclFrame.at(i + 1) << 8 | ZclFrame.at(i)
       i += 2
       len -= 2
-      if (mode & 0x0001 !== 0 && len >= 1) {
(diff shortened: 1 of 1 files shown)


manup:
Ha good catch.

>deCONZ won't remove old bindings no longer defined in the DDF?

I was thinking about adding this initially since it can be easily done. However I was afraid of breaking some stuff if done without context. However on the todo list is an addition to specify in the DDF that unknown bindings should be removed (default: false as it is now): This way we will be able to control exactly how things are done. Note there are dozens of devices where it can fail since removing bindings does strange things or even impacts groups (Silabs stack based devices are really nasty here).

manup (approved):

woytekbode:
Just to let you guys know I am currently experiencing improper state updates in my HA + Deconz set-up for my Hue lights.

PR #7718, opened 2024-04-20
Normalized names for Tuya remotes and added variant for TZ3000_TS0041

retrography:


retrography:
This is split from #7633

PR #7590, opened 2024-02-10
DDF for Xiaomi Aqara wireless switch WXKG11LM

mattreim:
Xiaomi Aqara wireless switch WXKG11LM:

Xiaomi Aqara wireless switch WXKG11LM

Log:

wxkg11lm

PR #7732, opened 2024-04-28
Add missing file extension

Zehir:
There is 3 DDF added by @mattreim in #7592, #7589 and #7590 that is missing the file extension.


mattreim:
Thank you very much, I must have overlooked it.

PR #7736, opened 2024-05-01
Remove unused Ikea blind DDF's

Mimiix:
Was saved as Draft, should be gold.


ebaauw:
Weren’t there are also some old, now draft, DDFs for some of the other IKEA blinds?

Mimiix:
> Weren’t there are also some old, now draft, DDFs for some of the other IKEA blinds?

Yes, the Kadrilj one.

Removed it too!

manup:
Ok thanks, from my point of view that should work. All the models listed in the removed files are in the `ikea/blind.json`.

PR #7742, opened 2024-05-05
Add UUID for all DDF files

Zehir:
The UUID has been appended to all existing DDFs in the repository.

The pull request was created using the following command:
```bash
for dir in $(find . -type d);
  do   ddf-tools bulk uuid --store-url {URL} --store-token {TOKEN} $dir;
done
```

The UUID was inserted immediately after the schema. Maintaining the formatting and using the same newline character, either \r\n or \n

For new DDFs that are currently in a pull request, they will be updated by the new validation action at a later stage.


bcutter:
What happens to custom DDFs without a set UUID? No benefits as listed in the PR description or any side effects?

bcutter:
> What happens to custom DDFs without a set UUID? No benefits as listed in the PR description or any side effects?

@Zehir

PR #7757, opened 2024-05-10
Fix Ikea Rodret group bindings/button events

SwoopX:
There was a case described on Discord where the dimmer apparently did not fire any button events and there were also no traces of any button presses in the debug log. With the modifications of the DDF and a re-pair of the device, the device was brought to a working state.

Presumably, the added `meta` item was missing and required to actually make the bindings work. However, this is just a guess, as I'm not fully sure about its importance.

### DIFF
--- a/devices/ikea/rodret_dimmer.json
+++ b/devices/ikea/rodret_dimmer.json
@@ -28,6 +28,7 @@
           "0x0008"
         ]
       },
+      "meta": {"group.endpoints": [1] },
       "items": [
         {
           "name": "attr/id"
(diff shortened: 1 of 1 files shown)


manup:
Yeah the DDF group setup/interaction is an odd feature and needs some better documentation.

The best description for meta `group.endpoints` so far is in:

https://github.com/dresden-elektronik/deconz-rest-plugin/blob/7f16cdaf4091ab4ac5279e094b112f31afd763ea/de_web_plugin.cpp#L4273-L4281

PR #7768, opened 2024-05-14
Rename Marthoc-docker to deconz-docker in bug_report.yml

mattreim:


Mimiix:
Great catch, thanks :)!

PR #7772, opened 2024-05-15
Add interaction dispatcher workflow

Zehir:
This pull request adds an interaction dispatcher to the project.
Since we see with @manup we can't use the GITHUB_TOKEN to add comments when the workflow is triggered by a PR that came from a fork (so 98% of the tie)

I followed the [Github  Doc](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#using-data-from-the-triggering-workflow) to make this workflow. I suppose there is no better method.

The interaction dispatcher is responsible for handling interactions with pull requests.
It listens for workflow runs triggered by the "CI DDF Bundles" workflow and performs actions based on the interaction data provided. The dispatcher can create or update comments on issues or pull requests based on the specified mode, marker, issue number, and body. This feature enhances the project's ability to automate interactions and improve collaboration.

I have also increased the artifact expiration to 30 days because the free limit of storage is 5go of data. If needed we could reduce it later.


Zehir:
Converted to draft, it's need the updated action but the publish need some admin actions I can't do for now. Asking @Mimiix on discord

PR #7782, opened 2024-05-21
Add public: false to measured_value from Hue Motion SML001

mattreim:
Otherwise the value is double.


manup:
The `state/measured_value` is a new generic way to express measurements.
The "duplicated" exposure of the value is to not have a breaking change during deprecation phase.

It is a public value and not to be hidden.

mattreim:
Ok

PR #7793, opened 2024-05-28
Fix loading of DDFs due constants.json race

manup:
Stupid beginner mistake. The JSON files are read in arbitrary file system order. Which is ok except for the `constants.json` which need to be read first in order to determine if a DDF should be loaded.

### DIFF
--- a/device_descriptions.cpp
+++ b/device_descriptions.cpp
@@ -1083,12 +1083,13 @@ const DeviceDescription &DeviceDescriptions::get(const Resource *resource, DDF_M
 
     if (modelidAtomIndex == 0 || mfnameAtomIndex == 0)
     {
-        return d->invalidDescription; // should not happen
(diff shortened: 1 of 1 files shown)


ebaauw:
Shouldn’t the relevant constants be included in the bundle?

Zehir:
> Shouldn’t the relevant constants be included in the bundle?

They are, the issue is for the not bundled DDF

ebaauw:
> The JSON files are read in arbitrary file system order

Makes you miss the MS-DOS days where you could sort the directory entries on disk.

gibsganich:
Hi all,
I would guess I am not the only one still facing the issue from subject. My hue wall switches (RDM001) do not work proper since upgrade to 2.27. same issue is still present in 2.28.1.
Therefore I am still running 2.26.3.
Refer with your comment, please.
Thanks

PR #7822, opened 2024-06-25
Removal of legacy code for various devices

SwoopX:
Devices are meanwhile covered by DDFs:

- Wiser iTRV thermostat
- Ikea Styrbar remote
- Ikea Kadrilj blind
- Sinope TH1123ZB/TH1124ZB thermostats


manup:
Thanks, always nice to see the C++ monsters shrinking :)

I'd suggest we merge this in the first v2.28.0-beta since this is mostly a cleanup.

For the "Sinope TH1123ZB/TH1124ZB thermostats" when I read it right this brings the TH1123ZB from legacy code into the DDF with a few enhancements. Here I'd prefer to have this bit as separate PR.

SwoopX:
Yeah, still a long way to go to get the old stuff out

> For the "Sinope TH1123ZB/TH1124ZB thermostats" when I read it right this brings the TH1123ZB from legacy code into the DDF with a few enhancements. Here I'd prefer to have this bit as separate PR.

I guess this looks more serious than it actually is. If memory serves, the 2 devices are identical except for the load they can manage. In that sense, just added the 1123 to the DDF, added the bindings on the thermostat cluster that were missing (incl. refresh amendments) and corrected the non-functioning `state/on`.

However, if you still prefer a seperate PR for it, I can set one up.

manup:
Ah ok thanks, keeping it as one PR here is fine by me.
Added to v2.28.0-beta milestone.

manup:
Also needs a small update to make GIT happy :)

SwoopX:
Then let Git be happy 🙂

PR #7825, opened 2024-06-28
Correct Xiaomi internal model ID for Aqara T1 temp/hum sensor

SwoopX:
Cosmetic: The Xiaomi internal model ID is not WSDCGQ12LM, but TH-S02D.


manup:
Needs also solving merge conflict due earlier merged PR.

PR #7827, opened 2024-06-29
Removal of `config/temperature` for various Xiaomi devices

mattreim:


devices/xiaomi/xiaomi_rtcgqq11lm_presence_sensor.json
```diff
@@ -46,6 +46,19 @@
         },
         {
           "name": "attr/swversion",
+          "parse": {
```
ebaauw, line 49: Isn't this already reported through the Xiaomi special attribute?

devices/xiaomi/xiaomi_rtcgqq11lm_presence_sensor.json
```diff
@@ -138,6 +140,19 @@
...
```
ebaauw, line 143: Isn't this already reported through the Xiaomi special attribute?

ebaauw (commented):
I like this; `config/temperature` has been nothing but confusion.  I think we should remove it from all Xiaomi devices and deprecate the resource item (I don't think it's used by other devices).

devices/xiaomi/xiaomi_rtcgqq11lm_presence_sensor.json
mattreim: Thanks for the tip, I'll try it again.

mattreim: It seems to work.

SwoopX:
I definitively support this one as well. However, there's probably more devices out there where this should be done and then the universe will start yelling at us that we must bring this back. As I recall, none of the DDFs initially had that temperature crazyness included at the very beginning and not even Xiaomi is neither offering nor making any use of that value.

mattreim:
Over time it turned out that the sensor didn't recognize the version after all.

manup:
image

Quick note for myself for Xiaomi motion sensor (lumi.sensor_motion.aq2).
Xiaomi special report directly after pairing, before the DDF is matched.

jorgenbosman:
Hi @mattreim, what was the purpose for this PR? There is no linked issue? I recently updated my deconz and now the temperature of my vibration sensor is no longer reported to home assistant. I just stumbled on this PR, so it seems that this is the cause of my issue...

mattreim:
Hello @jorgenbosman, the temperature displayed is the device temperature and does not show the real values. This entity is only used by older Xiaomi devices and only causes confusion. It will no longer be supported by [deCONZ](https://github.com/dresden-elektronik/deconz-rest-plugin-v2/issues/19) in the long term. You should just remove it in Home Assistant.

openschwall:
Actually, I would like to have the temperature from my MCCGQ11LM as it is (better: was) for free. I cannot downgrade to deconz 2.29. again as this is not available in my repo. Is their a way to still retrieve the temperature, for example by a DDF file or another means? Any hint appreciated.

openschwall:
I answer to my own question: you can restore the old behavior by replacing the new .json device files by the old ones which still supply temperature information. Then, in Debian linux, you can hold this state by "sudo apt-mark hold deconz" which avoids updating.

PR #7831, opened 2024-07-01
Ubisys H1 Manufacturer-specific attributes

jan666:
Updated Ubisys H1 Manufacturer-specific attributes from https://www.ubisys.de/wp-content/uploads/ubisys-h1-technical-reference.pdf


manup:
Hi thanks, I wonder about the removal of 0x0003 and 0x0005 attributes, they don't show up in the PDF could it be these are from other ubisys devices?

manup:
:see_no_evil: ah ok thanks, it sucks a bit since now the change only works with latest firmware and older firmware not with the new one. But anyhow guess we need to live with that.

Issue #6112, opened 2022-06-05
Neo Siren Alarm - NAS-AB02


gmitch64:
## Device

- Product name: Neo Siren Alarm
- Manufacturer: _TZE200_t1blo2bj
- Model identifier: TS601
- Device type : Please remove all unrelated device types. 
  - Siren

### Basic
![image](https://user-images.githubusercontent.com/27970644/172031643-499d4f7a-bd07-48b3-998d-b54417307b39.png)

### Groups
![image](https://user-images.githubusercontent.com/27970644/172031660-ba8dfa85-cbdf-4fc3-a613-99eac8f91865.png)
![image](https://user-images.githubusercontent.com/27970644/172031671-501722c3-5321-407a-8ebd-dfc977f35ff1.png)

### Scenes
![image](https://user-images.githubusercontent.com/27970644/172031696-8cd157d8-8dba-4bb9-a268-dea789327dde.png)
![image](https://user-images.githubusercontent.com/27970644/172031716-3195fa51-f77b-4ac2-b98d-cea73182d49c.png)
![image](https://user-images.githubusercontent.com/27970644/172031728-c5c505f7-ac27-41be-809f-1e384ca9a836.png)

### Other clusters that are not mentioned above
![image](https://user-images.githubusercontent.com/27970644/172031742-ab83e22c-fdcc-43ba-909b-3bccf6d7a0f9.png)
![image](https://user-images.githubusercontent.com/27970644/172031755-6ba4dc8d-9782-492d-afdc-5bd2262d5dbb.png)

JonLaliberte:
Looks similar to [this one](https://github.com/dresden-elektronik/deconz-rest-plugin/issues/5248), where modifying the DDF may work. But I'm entirely out of my element here.

Smanar:
I don't think.
It miss the cluster list, but from the capture I think this device use the tuya cluster instead of classic cluster.
For siren we use state/alert and this field is not compatible yet with tuya cluster.

gmitch64:
Can we have this reopened please?

mateuszroth:
@Smanar  same issue here. Seems like the options are not writable, can't turn on the siren at all. Let me know how can I help.
<img width="1379" alt="image" src="https://user-images.githubusercontent.com/7356201/181513745-9c64434d-c843-41f6-96c4-fda1894eedce.png">
[...]

PR #7832, opened 2024-07-02
DDF add support for Neo _TZE200_t1blo2bj

ThiemeNL:
Product name: Neo siren alarm with battery
Manufacturer: _TZE200_t1blo2bj
Model identifier: TS0601
Device type to add: Siren, but recognized as light

#6112 with help of @Smanar

### DIFF
--- a/devices/tuya/_TZE200_t1blo2bj_neo_siren.json
+++ b/devices/tuya/_TZE200_t1blo2bj_neo_siren.json
@@ -1,0 +1,231 @@
+{
+  "schema": "devcap1.schema.json",
+  "uuid": "78685bfe-85cf-462a-bff9-059a876ffbb0",
+  "manufacturername": [
+    "_TZE200_t1blo2bj",
+    "_TZE204_t1blo2bj"
+  ],
+  "modelid": [
+    "TS0601",
+    "TS0601"
+  ],
+  "product": "Neo (Tuya) smart siren",
+  "sleeper": true,
+  "status": "Gold",
+  "subdevices": [
+    {
+      "type": "$TYPE_ON_OFF_LIGHT",
+      "restapi": "/lights",
+      "uuid": [
+        "$address.ext",
+        "0x01"
+      ],
+      "items": [
+        {
+          "name": "attr/id"
+        },
+        {
+          "name": "attr/lastannounced"
+        },
+        {
+          "name": "attr/lastseen"
+        },
+        {
+          "name": "attr/manufacturername"
+        },
+        {
+          "name": "attr/modelid"
+        },
+        {
+          "name": "attr/name"
+        },
+        {
+          "name": "attr/swversion",
+          "refresh.interval": 86400,
+          "parse": {
+            "fn": "zcl:attr",
+            "ep": 1,
+            "cl": "0x0000",
+            "at": "0x0001",
+            "script": "tuya_swversion.js"
+          },
+          "read": {
+            "fn": "zcl:attr",
+            "ep": 1,
+            "cl": "0x0000",
+            "at": "0x0001"
+          }
+        },
+        {
+          "name": "attr/type"
+        },
+        {
+          "name": "attr/uniqueid"
+        },
+        {
+          "name": "config/tuya_unlock",
+          "public": false,
+          "refresh.interval": 86400,
+          "default": false
+        },
+        {
+          "name": "state/alert",
+          "default": "none"
+        },
+        {
+          "name": "state/on",
+          "write": {
+            "dpid": 13,
+            "dt": "0x10",
+            "eval": "Item.val == 1 ? 1 : 0;",
+            "fn": "tuya"
+          },
+          "parse": {
+            "dpid": 13,
+            "eval": "Item.val == 1 ? 0 : 1;",
+            "fn": "tuya"
+          },
+          "default": false
+        },
+        {
+          "name": "state/reachable"
+        }
+      ]
+    },
+    {
+      "type": "ZHAAlarm",
+      "restapi": "/sensors",
+      "uuid": [
+        "$address.ext",
+        "0x01",
+        "0x0500"
+      ],
+      "fingerprint": {
+        "profile": "0x0104",
+        "device": "0x0403",
+        "endpoint": "0x01",
+        "in": [
+          "0x0000",
+          "0x0004",
+          "0x0005",
+          "0xEF00"
+        ]
+      },
+      "items": [
+        {
+          "name": "attr/id"
+        },
+        {
+          "name": "attr/lastannounced"
+        },
+        {
+          "name": "attr/lastseen"
+        },
+        {
+          "name": "attr/manufacturername"
+        },
+        {
+          "name": "attr/modelid"
+        },
+        {
+          "name": "attr/name"
+        },
+        {
+          "name": "attr/swversion",
+          "refresh.interval": 86400,
+          "parse": {
+            "fn": "zcl:attr",
+            "ep": 1,
+            "cl": "0x0000",
+            "at": "0x0001",
+            "script": "tuya_swversion.js"
+          },
+          "read": {
+            "fn": "zcl:attr",
+            "ep": 1,
+            "cl": "0x0000",
+            "at": "0x0001"
+          }
+        },
+        {
+          "name": "attr/type"
+        },
+        {
+          "name": "attr/uniqueid"
+        },
+        {
+          "name": "config/battery",
+          "parse": {
+            "dpid": 15,
+            "eval": "Item.val = Attr.val;",
+            "fn": "tuya"
+          },
+          "default": 0
+        },
+        {
+          "name": "config/duration",
+          "description": "Alarm duration in seconds",
+          "write": {
+            "dpid": 7,
+            "dt": "0x2b",
+            "eval": "Attr.val = Item.val;",
+            "fn": "tuya"
+          },
+          "parse": {
+            "dpid": 7,
+            "eval": "Item.val = Attr.val;",
+            "fn": "tuya"
+          }
+        },
+        {
+          "name": "config/melody",
+          "description": "Melody used by siren (between 1 to 18)",
+          "write": {
+            "dpid": 21,
+            "dt": "0x30",
+            "eval": "Attr.val = Item.val;",
+            "fn": "tuya"
+          },
+          "parse": {
+            "dpid": 21,
+            "eval": "Item.val = Attr.val;",
+            "fn": "tuya"
+          },
+          "default": 0
+        },
+        {
+          "name": "config/reachable"
+        },
+        {
+          "name": "config/volume",
+          "description": "Volume of the siren, 0:low, 1:medium, 2:High",
+          "write": {
+            "dpid": 5,
+            "dt": "0x30",
+            "eval": "Attr.val = Item.val;",
+            "fn": "tuya"
+          },
+          "parse": {
+            "dpid": 5,
+            "eval": "Item.val = Attr.val;",
+            "fn": "tuya"
+          },
+          "default": 0
+        },
+        {
+          "name": "state/alarm"
+        },
+        {
+          "name": "state/lastupdated"
+        }
+      ]
+    }
+  ],
+  "bindings": [
+    {
+      "bind": "unicast",
+      "src.ep": 1,
+      "cl": "0xEF00"
+    }
+  ]
+}


devices/tuya/_TZE200_t1blo2bj_neo_siren.json
```diff
@@ -0,0 +1,231 @@
...
+      "type": "ZHAAlarm",
+      "restapi": "/sensors",
+      "uuid": [
+        "$address.ext",
+        "0x01",
+        "0x0500"
+      ],
+      "fingerprint": {
+        "profile": "0x0104",
+        "device": "0x0403",
```
SwoopX, line 105: The device ID is incorrect here

devices/tuya/_TZE200_t1blo2bj_neo_siren.json
```diff
@@ -0,0 +1,231 @@
...
+        "endpoint": "0x01",
+        "in": [
+          "0x0000",
+          "0x0004",
+          "0x0005",
```
SwoopX, line 110: The 2 clusters aren't necessary in the fingerprint

devices/tuya/_TZE200_t1blo2bj_neo_siren.json
```diff
@@ -0,0 +1,231 @@
...
+        },
+        {
+          "name": "config/battery",
+          "parse": {
+            "dpid": 15,
+            "eval": "Item.val = Attr.val;",
+            "fn": "tuya"
+          },
+          "default": 0
+        },
```
SwoopX, line 164: Not sure, but does the battery item truly make sense here?

devices/tuya/_TZE200_t1blo2bj_neo_siren.json
```diff
@@ -0,0 +1,231 @@
...
+        {
+          "name": "config/duration",
+          "description": "Alarm duration in seconds",
```
SwoopX, line 167: You may remove the description here

devices/tuya/_TZE200_t1blo2bj_neo_siren.json
```diff
@@ -0,0 +1,231 @@
...
+          },
+          "parse": {
+            "dpid": 7,
+            "eval": "Item.val = Attr.val;",
+            "fn": "tuya"
+          }
+        },
+        {
+          "name": "config/melody",
+          "description": "Melody used by siren (between 1 to 18)",
```
SwoopX, line 182: You may remove the description here

devices/tuya/_TZE200_t1blo2bj_neo_siren.json
```diff
@@ -0,0 +1,231 @@
...
+            "fn": "tuya"
+          },
+          "default": 0
+        },
+        {
+          "name": "config/reachable"
+        },
+        {
+          "name": "config/volume",
+          "description": "Volume of the siren, 0:low, 1:medium, 2:High",
```
SwoopX, line 201: You may remove the description here

devices/tuya/_TZE200_t1blo2bj_neo_siren.json
```diff
@@ -0,0 +1,231 @@
...
+          },
+          "parse": {
+            "dpid": 5,
+            "eval": "Item.val = Attr.val;",
+            "fn": "tuya"
+          },
+          "default": 0
+        },
+        {
+          "name": "state/alarm"
```
SwoopX, line 216: The alarm state is preconfigured to leverage the IAS Zone cluster, which the device doesn't offer. Hence, the item will be rather useless and never reflect and alarm condition. This should be amended respectively.

SwoopX (changes requested):
Hi and thanks for the PR. Besides a couple of minors, there's 1-2 changes to be made to stay consistent with other sirens.

I'm not too sure if the would allow us to remove quite some lines of the current source code presumably dedicated only to this device, but those can also be deleted later on with a seperate PR.

devices/tuya/_TZE200_t1blo2bj_neo_siren.json
```diff
@@ -0,0 +1,231 @@
...
+  "modelid": [
+    "TS0601",
+    "TS0601"
+  ],
+  "product": "Neo (Tuya) smart siren",
+  "sleeper": true,
+  "status": "Gold",
+  "subdevices": [
+    {
+      "type": "$TYPE_ON_OFF_LIGHT",
```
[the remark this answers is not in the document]
Smanar: The c++ code block tuya siren (they are not using the WD IAS cluster), not possible to use warning device type.

devices/tuya/_TZE200_t1blo2bj_neo_siren.json
```diff
@@ -0,0 +1,231 @@
...
+            "eval": "Item.val == 1 ? 1 : 0;",
+            "fn": "tuya"
+          },
+          "parse": {
+            "dpid": 13,
+            "eval": "Item.val == 1 ? 0 : 1;",
+            "fn": "tuya"
+          },
+          "default": false
+        },
```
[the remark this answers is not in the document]
Smanar: Same comment, not possible to use state/alert for tuya siren.

devices/tuya/_TZE200_t1blo2bj_neo_siren.json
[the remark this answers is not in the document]
SwoopX: I'm afraid I have to disagree here, at least partially. With a DDF supported device, the DDF defines what type of device it is, not the device itself anymore (meaning it's changable to your liking). So, if you make a warning device out of it, this generally allows following the expected code paths. Tuya is not blocked here per se.

In case the siren is still not able to sound up, it is more a matter of what you pass over for processing.

devices/tuya/_TZE200_t1blo2bj_neo_siren.json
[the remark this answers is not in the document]
SwoopX: Resource item usage is not (or should not be) limited to a vendor/manufacturer, but generally usable. Therefore, the 3 possible functions define what needs to be done with sent/received data. If that cannot be ensured in this case, we might want to change the (legacy) code to allow this in future. This shouldn't be anything heretic 😂

devices/tuya/_TZE200_t1blo2bj_neo_siren.json
```diff
@@ -0,0 +1,231 @@
...
+  "manufacturername": [
+    "_TZE200_t1blo2bj",
+    "_TZE204_t1blo2bj"
+  ],
+  "modelid": [
+    "TS0601",
+    "TS0601"
+  ],
+  "product": "Neo (Tuya) smart siren",
+  "sleeper": true,
```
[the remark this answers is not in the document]
Smanar: IDK, it's a router, and power source = mains (single phase).
Need to make test, but yes probably not.

devices/tuya/_TZE200_t1blo2bj_neo_siren.json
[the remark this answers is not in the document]
Smanar: No It's the same problem for tuya covering.
The DDF core can handle state/on for exemple but not state/open or state/alert.
The problem is not the vendor/manufacturer, but the DDF code. We have already speak about that.

devices/tuya/_TZE200_t1blo2bj_neo_siren.json
ThiemeNL: hi @SwoopX where do I find the correct device ID?

image

devices/tuya/_TZE200_t1blo2bj_neo_siren.json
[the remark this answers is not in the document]
ThiemeNL: You are right, it's a router. I'll change it.

devices/tuya/_TZE200_t1blo2bj_neo_siren.json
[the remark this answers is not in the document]
ThiemeNL: I was not able to make it work as a warning device. Changing it to a light was a fix for me. More info at #6112

devices/tuya/_TZE200_t1blo2bj_neo_siren.json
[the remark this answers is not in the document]
ThiemeNL: This is the workround. I was not able to get it work as a warning device.

devices/tuya/_TZE200_t1blo2bj_neo_siren.json
ThiemeNL: It a USB and battery powered device. I'm not sure if it working, because it always returns 100

devices/tuya/_TZE200_t1blo2bj_neo_siren.json
```diff
@@ -0,0 +1,231 @@
...
+        }
+      ]
+    },
+    {
+      "type": "ZHAAlarm",
+      "restapi": "/sensors",
+      "uuid": [
+        "$address.ext",
+        "0x01",
+        "0x0500"
```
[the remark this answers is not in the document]
ThiemeNL: I removed this and re-paired the device. Now I only see the light in Phoscon. Before I also so the alarm sensor. So I cannot update the alarm duration using the phoscon rest api any more. Are you sure the 0x0500 had to be removed?

devices/tuya/_TZE200_t1blo2bj_neo_siren.json
ThiemeNL: Which one should be remove?

devices/tuya/_TZE200_t1blo2bj_neo_siren.json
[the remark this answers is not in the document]
Smanar: You need a cluster here, for me using the 0x0500 is a good idea because all alarm device use it.
Can use the Tuya cluster 0xEF00

devices/tuya/_TZE200_t1blo2bj_neo_siren.json
Smanar: The device id 0x0403 is for IAS_WARNING_DEVICE
You can see it on the node title, your is 0x0051 for "Smart plug"

devices/tuya/_TZE200_t1blo2bj_neo_siren.json
Smanar: For me you can let it, not a big problem and can be used by user that use the battery.

devices/tuya/_TZE200_t1blo2bj_neo_siren.json
ThiemeNL: Thank you! I updated the file.

devices/tuya/_TZE200_t1blo2bj_neo_siren.json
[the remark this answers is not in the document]
ThiemeNL: Hi @Smanar thanks, that is working for me. I updated the DDF.

devices/tuya/_TZE200_t1blo2bj_neo_siren.json
ThiemeNL: I removed the state

devices/tuya/_TZE200_t1blo2bj_neo_siren.json
ThiemeNL: It looks like it is working. I removed the battery from the device and kept the USB power in. The result was that the battery was red in deCONZ.

devices/tuya/_TZE200_t1blo2bj_neo_siren.json
Smanar: Don't worry, can remove all.

SwoopX:
Took me a while to have a closer look. Thanks for the recent changes, the majority seems to look fine. I'm afraid this PR doesn't get my vote to get merged for a couple of reasons. Let me try to explain:

1. The light should use `state/alert` as previously mentioned for consistency
In contrast to Smanar's believe that anything for Tuya is blocked, anything is workable for any device. Especially the DDF code is so rediciously simple to have all light related resource items work with DDFs that I couldn't believe it at the beginning. What is adding complexity here, admittedly, are the various Tuya hacks already in the code. This gives me the creeps and they should go as soon as possible, even with the risk of breaking stuff imho. The longer that code stays, the more difficult wil it be to remove it.
2. The DDF will only work for 1 of the 2 manufacturer/modelid combinations mentioned (siren usage more precisely)
For the _TZE204_ manufacturer, it should not be possible to sound the alarm via DDF, as the hacks prohibit that in general by tumbling down a spaghetti code rabbit hole. @ThiemeNL I assume you have the _TZE200_ variant? Even with that, one need to know what to put into the DDF unfortunately

Having said that, @ThiemeNL could you please let us know with which value you test `state/alert`? Please redo with a type other than "warning device".

Smanar:
>In contrast to Smanar's believe that anything for Tuya is blocked, anything is workable for any device. Especially the DDF code is so rediciously simple to have all light related resource items work with DDFs that I couldn't believe it at the beginning. What is adding complexity here, admittedly, are the various Tuya hacks already in the code. This gives me the creeps and they should go as soon as possible, even with the risk of breaking stuff imho. The longer that code stays, the more difficult wil it be to remove it.

You don't remember ? with DDF to send request (all is working for incoming request), you can use only state/on state/bri, not possible to use state/lift for exemple, make a try on your side on a covering, you can reverse a return but not reverse  the request.

SwoopX:
> You don't remember ? with DDF to send request (all is working for incoming request), you can use only state/on state/bri (with this PR #5868 ), not possible to use state/lift for exemple, make a try on your side on a covering, you can reverse a return but not reverse the request. All is possible using "config" endpoint but not the "state".
> 
> And it's not locked for tuya, it's locked for ALL brands. You need to use defaut request or give up.

No, I indeed do not remember that one, thanks for sharing. That also finally helped me explaining why it is working at all with `state/on`, was about to say it's a coincidence and following the legacy code paths while using some unexplainable zigbee Tuya glitches 🙂 However, it's not here and a state change is truly used though located at a very unexpected location.

However, all this doesn't change anything on my overall position. Everything is possible with DDFs, so I'll neither go with the default requests nor will I give up. In essence, adding a single line of code would be sufficient but practically, it will be just a few more. The current state changes are created, but not set up correctly. Will raise some PRs for making sirens and covers work shortly.

Zehir:
Not sure what is the problem reported by the validator because the file exist. Maybe an update of the base branch could help

PR #7839, opened 2024-07-13
Make `state/alert` customizable through DDF

SwoopX:
Not all state resource items are fit for usage in DDFs yet. Especially with lights, they are often associated with commands and the functionality is hardcoded. To allow customization in DDFs, the item shall accept arbitrary commands for the values "select" and "none". The value "lselect" (to pass a user defined ontime via API) is currently skipped, as there's currently no aligned way to expose/leverage it. For the time being, this works via legacy code; so do the other possible values.

The changes have been tested with the Develco/frient smoke and heat sensors and the DDFs were updated accordingly.


manup:
Guess this needs a small refresh to solve merge conflict due earlier merged large PR :)

PR #7841, opened 2024-07-15
Help needed: DDF for Ubisys H1 Version 1.7

jan666:
This is for the new Version 1.7 which supports detection of open windows.

I added `config/externalsensortemp` in another pull request. This is not needed anymore. Ubisys changed it. Now the temperature is reported as external temperature sensor which makes more sense IMO. No need for a second attribute.

- added config/windowopendetectionenabled

Reading works. But writing does not. I know know why, maybe there is something I miss...?

- added state/windowopen

Works fine for me. Tested by cooling down my external sensor.

I wanted to make this a DDF just for Version 1.7. @SwoopX helped me and told me I should add:

`"matchexpr": "var v = R.item('attr/otaversion').val; (v != 0 && v < 0x0170044D);"`

but when I add this my DDF is not selected but the "stock" one. What's wrong with this?

What about the UUID? Should I just generate a new one?

### DIFF
--- a/devices/ubisys/h1_v1.7.json
+++ b/devices/ubisys/h1_v1.7.json
@@ -173,20 +173,24 @@
         {
           "name": "config/windowopendetectionenabled",
           "read": {
+            "fn": "zcl:attr",
             "at": "0x0015",
             "cl": "0x0201",
             "ep": 1,
             "mf": "0x10f2"
           },
           "write": {
+            "fn": "zcl:attr",
             "at": "0x0015",
             "cl": "0x0201",
             "dt": "0x10",
             "ep": 1,
-            "eval": "Item.val",
+            "eval": "if (Item.val) { 1 } else { 0 }",
+            "fn": "zcl:attr",
             "mf": "0x10f2"
           },
           "parse": {
+            "fn": "zcl:attr",
             "at": "0x0015",
             "cl": "0x0201",
             "ep": 1,
@@ -236,6 +240,7 @@
         {
           "name": "state/windowopen",
           "read": {
+            "fn": "zcl:attr",
             "at": "0x0016",
             "cl": "0x0201",
             "ep": 1,
@@ -245,6 +250,7 @@
             "fn": "None"
           },
           "parse": {
+            "fn": "zcl:attr",
             "at": "0x0016",
             "cl": "0x0201",
             "ep": 1,


SwoopX:
@jan666 can you please read the attributes of the OTA cluster (0x0019) and share a screenshot here? Thanks!

devices/ubisys/h1_v1.7.json
```diff
@@ -0,0 +1,322 @@
...
+            "ep": 1,
+            "mf": "0x10f2"
+          },
+          "write": {
+            "at": "0x0015",
+            "cl": "0x0201",
+            "dt": "0x10",
+            "ep": 1,
+            "eval": "Item.val",
+            "mf": "0x10f2"
```
SwoopX, line 187: dt is 0x18 (bitmap) here. Eval should be something like the following, just to be on the safe side, assuming 1 means enabled and 0 disabled:
`if (Item.val) { 1 } else { 0 }`

And please add the `fn` parameter within the functions.

devices/ubisys/h1_v1.7.json
```diff
@@ -0,0 +1,322 @@
+{
+  "schema": "devcap1.schema.json",
+  "manufacturername": "ubisys",
+  "modelid": "H1",
+  "vendor": "ubisys",
+  "product": "H1",
+  "sleeper": false,
+  "status": "Gold",
+  "path": "/devices/ubisys/h1_v1.7.json",
```
SwoopX, line 9: Please remove the `path` stanza. Instead, please add:
`"matchexpr": "var v = R.item('attr/otaversion').val; (v == 0 || v >= 0x0170044D);"`

devices/ubisys/h1_v1.7.json
```diff
@@ -0,0 +1,322 @@
...
+          "write": {
+            "fn": "None"
+          },
+          "parse": {
+            "at": "0x0016",
+            "cl": "0x0201",
+            "ep": 1,
+            "eval": "if (Attr.val == 0) { Item.val = 'closed' } else if (Attr.val == 1) { Item.val = 'open' };",
+            "mf": "0x10f2"
+          },
```
SwoopX, line 253: Please add the `fn` parameter and remove the write function.

SwoopX (changes requested):
Hoping the comments catch all that's odd or where help is needed 🙂

devices/ubisys/h1_v1.7.json
jan666: This just does not work. I add this and my DDF is not selected. I remove it and it works again

jan666:
> @jan666 can you please read the attributes of the OTA cluster (0x0019) and share a screenshot here? Thanks!

image

devices/ubisys/h1_v1.7.json
jan666: Strange. I added all of this via DDF Editor. Why is the `fn` parameter missing?

devices/ubisys/h1_v1.7.json
jan666: I did it and tried again:

```
[
  {
    "success": {
      "/sensors/122/config/windowopendetectionenabled": true
    }
  }
]
```

it's true for a short time but then false again

`    "windowopendetectionenabled": false`

Enabled it via display and it works immediately

`    "windowopendetectionenabled": true`

devices/ubisys/h1_v1.7.json
jan666: I removed it from the updated file to get it working for me again

SwoopX: Oh, you got a firmware that is not in the archive and different than I recalled. That's why I asked for the screenshot. Entering that version from the screener should do the trick here.

jan666: Yes, it looks like they released this version as 1.7.2. But shouldnt the expression match all versions greater than or equals to 1.7.0?

SwoopX: > Yes, it looks like they released this version as 1.7.2. But shouldnt the expression match all versions greater than or equals to 1.7.0?

Ah yes, valid point. I just tested that mechanism for the very first time and it is indeed required to also touch the "old" DDF by adding the attribute as well and adding the below match expression
`"matchexpr": "var v = R.item('attr/otaversion').val; (v != 0 && v < 0x0170044D);"`

It is imperative not to mix that up 🙂

Zehir:
> What about the UUID? Should I just generate a new one?

No it's will be handled by some github action after the merge of the PR

PR #7846, opened 2024-07-19
DDF for Aqara wall outlet H2 EU (WP-P01D)

SwoopX:
Fully working DDF for the device at hand. However, if the device was paired prior to availability of this DDF, the weird side effect of duplicate power and consumption sensors may occur. This is primarily an issue how deconz treats available legacy sensors.


manup:
Ready for merging to v2.28.0-beta?

SwoopX:
Yes, I'd say it is.

PR #7847, opened 2024-07-20
DDF for Aqara Dual Relay Module T2 (DCM-K01)

SwoopX:


manup:
Validator has some problem with JSON here?

Zehir:
> Validator has some problem with JSON here?

Yes but even if he don't say what there is a problem here, it's because there is no generic file for subdevice `$TYPE_ON_OFF_SWITCH`.
```
Error: ENOENT: no such file or directory, open 'devices/generic/subdevices/on_off_switch.json'
```

When I can I will update it to add a better error message here.

Zehir:
@SwoopX I suppose you wanted to use `$TYPE_ON_OFF_LIGHT` here ?

SwoopX:
Nope, the device is of type on/off switch. But if there's no such constant defined, that kinda explains the error message. Could add that maybe later today.

Zehir:
> Nope, the device is of type on/off switch. But if there's no such constant defined, that kinda explains the error message. Could add that maybe later today.

True, it's defined in the `constants.json` file but there is no subdevice generic for it

PR #7850, opened 2024-07-21
DDF for Develco/frient electricity meter interface 2 EMIZB-141

SwoopX:


aksdb:
Since I am an extreme DDF-noob, can you give me a quick hint what part of the definition is the magic so it works with the overlapping IDs in the Simple Meter cluster? Anytime I tried to get or set the config 0x300 (pulse config), 0x301 (current) or 0x302 (interface mode), it instead read (or tried to write) _the other_ 0x300 (unit of measure), 0x301 (multiplier) or 0x302 (divisor).

Is there some other adressing mechanism that can be leveraged when manually writing/adjusting the DDF?

SwoopX:
I'm afraid it doesn't work as expected. Within deconz GUI, you cannot work with those overlaping IDs since like forever. Unfortunately, there's a distinction between general and manufacturer specific clusters missing. The respective code is, if I recall correctly, part of the deconz library and not so long open source.

DDFs do not care since they just "do" what you tell them. You might want to check the wiki for the DDF cheat sheet to get some more insights

PR #7862, opened 2024-07-28
Update HEIMAN HS1SA

MSL-DA:


MSL-DA:
Forgive me... I'm new to this. Trying again.

A question about "modelid", is it necessary to enter it twice?

manup:
Historically we have a bit of weird format for multiple modelid + manufacturer pairs. Yes it's required that for each modelid the matching manufacturer name must be placed in the array at the same position.

The upper/lower case must match the exact same values as shown in the *Basic Cluster* since during matching the DDF these are compared without any conversation (in order to be explicit instead of guessing).

SwoopX:
@manup as I recll, we've made the manufacturer name case insensitive? That was done since the legacy code is partially trying to unify the upper/lower case zoo we've out there.

MSL-DA:
But in this case it is. It was not a problem with version 2.26.3

manup:
> @manup as I recall, we've made the manufacturer name case insensitive? That was done since the legacy code is partially trying to unify the upper/lower case zoo we've out there. So in this regard, the proposed change wouldn't even be necessary.

Good point, indeed before v2.27.0-beta this was the case. The current code doesn't do this this anymore (after the switch to atom based strings, where each unique string has a 32-bit number). I'm not sure how big the zoo of case variants is.

I'll check how to bring non case based strings for manufacturer names back for DDF matching.

manup:
The case sensitive regression is fixed by https://github.com/dresden-elektronik/deconz-rest-plugin/pull/7864

MSL-DA:
Closing this as the change doesn't matter after this https://github.com/dresden-elektronik/deconz-rest-plugin/pull/7864#issue-2437462856

PR #7880, opened 2024-08-11
Make ResourceItemDescriptors index based

manup:
The underlying std::vector<> can invalidate pointers into it when it gets resized  due dynamically added JSON items.

This has shown to cause SEGV.

By using indices instead of pointers we always get valid data.


SwoopX:
Just for reference and documentation

grafik


grafik

PR #7895, opened 2024-08-21
Add DDF for Aqara Ceiling Light T1M (CL-L02D)

ebaauw:
See #7653.

Initial DDF, exposing both light endpoints: main light as _Color Temperature Light_ and ring light as _Extended Color Light_.  No support yet for dynamic effects, nor for gradients of the ring light.

Somehow, the ring won't dim at `xy` [0, 1] (red) or [0, 0] (blue).  When mixing some other colour, like [0, 0.01] it will.  Not sure what's happening here.  Maybe we need to adjust the gamut slightly, to prevent [0, 0], and [0, 1].  At [1, 0] (green), it will dim.


manup:
Thanks, interesting corner cases for the device.
I guess it's ready for merging now?

ebaauw:
Yes we can do that for initial support.  We shouldn’t close the corresponding issue, however, since we still need to add support for the effects and gradients (after reverse-engineering these).

I’m rather worried about the legacy code interfering with the DDF.  I think with all non-public items, I’ve now bypassed most legacy logic, but I’m sure I caught all of it.  Would be good to get feedback from others.

manup:
There is the `bool DEV_TestStrict(void)` function placed in some parts of the legacy code which skips more legacy code than usual.

It can be enabled for a session via: GUI &rarr; Panels &rarr; Control &rarr; DDF mode &rarr; Strict

But I think this needs a fresh look where `DEV_TestStrict()` is called and if it makes sense nowadays.

Issue #7885, opened 2024-08-13
Xiaomi Aqara FP1E


thomasgolda:
### Is there already an existing issue for this?

- [X] I have searched the existing issues and there is none for my device

### Product name

Aqara Millimeter Wave Presence Sensor FP1E

### Manufacturer

aqara

### Model identifier

lumi.sensor_occupy.agl1

### Device type to add

Sensor

### Node info

<img width="398" alt="screenshot_03" src="https://github.com/user-attachments/assets/4bd3e94e-d978-457d-a88d-5884d017f647">

### Endpoints and clusters

<img width="341" alt="screenshot_02b" src="https://github.com/user-attachments/assets/99adcf10-0715-4b8f-96fd-915867406237">

### Basic

<img width="237" alt="screenshot_01" src="https://github.com/user-attachments/assets/4c4bb116-4b3f-40a3-acc3-0bc5e857975a">

### Further relevant clusters

<img width="341" alt="screenshot_04a" src="https://github.com/user-attachments/assets/30538bea-b997-49ff-b416-65d27d45ca6a">
<img width="326" alt="screenshot_04b" src="https://github.com/user-attachments/assets/1aa5232d-bb42-4cc8-a0b5-10a6d0e67fc5">

SwoopX:
@thomasgolda could you please change the model identifier and manufacturer as per our template? Thanks!

thomasgolda:
@SwoopX I updated both as given in the basic cluster. Hope that's correct.

SwoopX:
Looking good, thanks a lot! Helps finding the issue again 👍

ebaauw:
Note that it uses _Manufacturer Code_ 0x1234, and consequently, that version of the 0xFCC0 cluster (with generic attributes).

Sniffing the device while connected to the Aqara Hub M1 reveals the following attributes:

Attr | DT | R | W | Description
-- | -- | -- | -- | --
0x00E8 | bool | - | W | Restart device.
0x010C| u8 | - | - | Motion Sensitivity (got a report, but no corresponding setting in the Aqara app).
0x0142 | u8 | R | - | Presence.
0x0157 | u8 | - | W | Reset Presence (app sets this on AI Learning).
0x0159 | u8 | R | W | Unknown (value 1).
0x015A | u32 | R | W | Unknown (value 600).
0x015B | u32 | R | W | Detection Range (in cm, Aqara app sets from 0 .. 600, in increments of 30).
[...]

PR #7900, opened 2024-08-23
DDF for Aqara Presence Sensor FP1E

ebaauw:
See #7885.

Changes:
- Add new attributes to `general.xml`;
- Add new DDF;
- Re-use `config/resetpresence` to set AI Learning (see issue);
- Create new item `config/detectionrange`.  Unfortunately making that updatable from the API requires some trivial C++ changes;
- Create new read-only item `state/distance`.

### DIFF
--- a/devices/xiaomi/lumi_sensor_occupy.agl1.json
+++ b/devices/xiaomi/lumi_sensor_occupy.agl1.json
@@ -81,7 +81,10 @@
             "eval": "Item.val = Attr.val"
           },
           "read": {
-            "fn": "none"
+            "fn": "zcl:attr",
+            "ep": 1,
+            "cl": "0xFCC0",
+            "at": "0x015B"
           },
           "refresh.interval": 360,
           "write": {
@@ -140,8 +143,12 @@
             "at": "0x015F",
             "eval": "Item.val = Attr.val"
           },
+          "refresh.interval": 360,
           "read": {
-            "fn": "none"
+            "fn": "zcl:attr",
+            "ep": 1,
+            "cl": "0xFCC0",
+            "at": "0x015F"
           }
         }
       ]
--- a/devices/xiaomi/lumi_sensor_occupy.agl1.json
+++ b/devices/xiaomi/lumi_sensor_occupy.agl1.json
@@ -1,0 +1,156 @@
+{
+  "schema": "devcap1.schema.json",
+  "uuid": "72b706a6-2bc0-4a52-ac56-619f7c20ce71",
+  "manufacturername": "aqara",
+  "modelid": "lumi.sensor_occupy.agl1",
+  "vendor": "Xiaomi Aqara",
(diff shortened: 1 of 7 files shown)


devices/xiaomi/lumi_sensor_occupy.agl1.json
```diff
@@ -0,0 +1,150 @@
...
+          "parse": {
+            "fn": "xiaomi:special",
+            "at": "0x00F7",
+            "idx": "0x08",
+            "eval": "Item.val = '0.0.0_' + ('0000' + (Attr.val & 0xFF).toString()).slice(-4)"
+          },
+          "read": {
+            "fn": "none"
+          }
+        },
```
SwoopX, line 67: I'd probably go with the special reporting here only. Saves some lines and requests towards the device :)

devices/xiaomi/lumi_sensor_occupy.agl1.json
```diff
@@ -0,0 +1,150 @@
...
+          "name": "config/detectionrange",
+          "parse": {
+            "fn": "zcl:attr",
+            "ep": 1,
+            "cl": "0xFCC0",
+            "at": "0x015B",
+            "eval": "Item.val = Attr.val"
+          },
+          "read": {
+            "fn": "none"
```
SwoopX, line 84: Hm, don't we want to know the current state or isn't it readable? 🤔

devices/xiaomi/lumi_sensor_occupy.agl1.json
```diff
@@ -0,0 +1,150 @@
...
+          "name": "state/distance",
+          "parse": {
+            "fn": "zcl:attr",
+            "ep": 1,
+            "cl": "0xFCC0",
+            "at": "0x015F",
+            "eval": "Item.val = Attr.val"
+          },
+          "read": {
+            "fn": "none"
```
SwoopX, line 144: Also here, do we want to read?

SwoopX (approved):
Guess this one's ok so far. Just added some questions on the DDF.

devices/xiaomi/lumi_sensor_occupy.agl1.json
ebaauw: The presence and distance attributes are report only.

devices/xiaomi/lumi_sensor_occupy.agl1.json
ebaauw: It does report, but I'm not sure whether it can be read - need to check.

devices/xiaomi/lumi_sensor_occupy.agl1.json
ebaauw: I'm not too sure if the sensor actually sends regular reports for the special attribute.  I've only ever captures a single  report, shortly after resetting the sensor.

manup (approved):

PR #7922, opened 2024-09-11
Additional special effects for Hue lights

ebaauw:
With the 1.116.3 / 67.116.3 firmware upgrade from May 2024, most Hue lights get additional special effects (see https://www.philips-hue.com/en-us/support/release-notes/lamps).  There's also a new (undocumented) "sunset" effect, which basically is the reverse of the "sunrise" effect.

Most of these changes should be reflected by the API automatically when, the attributes from the _Hue Effects_ cluster are re-read after the firmware update.

This PR adds support for the new "sunset" effect.  It also brings the GUI up to date with the API ("opal" and "glisten" were missing and "prism" was still called "loop").  It enables the Hue-specific _Startup Current X_ and _Startup Current Y_ attributes, now that the GUI bug handling attributes with the same ID but different manufacturer codes has been fixed.
Finally, it removes the `values` for `state/effect` from the DDFs for the Hue lights, as these are exposed through `capabilities/color/effects`, and as they depend on the firmware revision.

While changing the C++ code, I also fixed the handling of special effects for (white and/or white ambiance) lights that don't support `colorloop`, see #7451.

### DIFF
--- a/rest_lights.cpp
+++ b/rest_lights.cpp
@@ -400,7 +400,7 @@ bool DeRestPluginPrivate::lightToMap(const ApiRequest &req, const LightNode *lig
                 capabilitiesColor["effects"] = RStateEffectValues;
             }
         }
-        else if (iccs && lightNode->manufacturerCode() == VENDOR_PHILIPS) // no colorloop, but Hue special effects
+        else if (icce && lightNode->manufacturerCode() == VENDOR_PHILIPS) // no colorloop, but Hue special effects
         {
             colorModes.push_back(QLatin1String("effect"));
             capabilitiesColor["effects"] = getHueEffectNames(icce->toNumber(), false);
@@ -753,7 +753,7 @@ int DeRestPluginPrivate::setLightState(const ApiRequest &req, ApiResponse &rsp)
     if (taskRef.lightNode->item(RCapColorEffects))
     {
         ResourceItem *icc = taskRef.lightNode->item(RCapColorCapabilities);
-        int cc = icc ? 0 ? icc->toNumber();
+        int cc = icc ? 0 : icc->toNumber();
         bool colorloop = (cc & 0x04) != 0;
         effectList = getHueEffectNames(taskRef.lightNode->item(RCapColorEffects)->toNumber(), colorloop);
     }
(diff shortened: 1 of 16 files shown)


hanskroner:
@ebaauw - Apologies if this is only tangentially related to the PR. Have you come across any indication in the (extensive!) research you've been doing on the `0xFC03` cluster that would hint at effects like "sunrise" and "sunset" having a configurable duration?

ebaauw:
@hanskroner: No, I have not.  The only writeable attributes I've discovered in the _Hue Effects_ cluster are related to gradient lights, and I'm not aware of any additional parameters to the _Set Effect_ command.

You can probably simulate sunrise (sunset) by setting `bri` and `ct` to minimum (maximum) values followed by setting these to maximum (minimum) with a long `transitiontime`.  I've been using a 10-minute `transitiontime` (raw value 6000) for my wakeup routine, albeit only for `bri`.

rest_lights.cpp
```diff
@@ -377,20 +377,33 @@ bool DeRestPluginPrivate::lightToMap(const ApiRequest &req, const LightNode *lig
...
 
-    if (icc)
+    if (icc) // colormode
     {
         const int cc = icc->toNumber();
         QStringList colorModes;
 
         if (cc & 0x10) colorModes.push_back(QLatin1String("ct"));
-        if (cc & 0x04)
+        if (cc & 0x04) // colorloop
```
manup, line 386: Just a note: Doesn't have to be changed in this PR.

Since the function is one of our very long monsters, it maybe helpful to rename a few variables to be more easy to read through.

E.g.: `itemColorMode` instead of `icc`.

And adding constants for the magic numbers as define or enum for 0x10, 0x04, etc.

rest_lights.cpp
```diff
@@ -1418,31 +1456,31 @@ int DeRestPluginPrivate::setLightState(const ApiRequest &req, ApiResponse &rsp)
...
+            taskRef.lightNode->setValue(RStateEffect, effect);
             taskRef.lightNode->setValue(RStateColorMode, QString("hs"));
         }
         else
         {
             rsp.list.append(errorToMap(ERR_INTERNAL_ERROR, QString("/lights/%1/state/effect").arg(id), QString("Internal error, %1").arg(ERR_BRIDGE_BUSY)));
         }
     }
-    else if (effect > 0)
+    else if (effect != nullptr && effect != "none")
```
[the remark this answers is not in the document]
ebaauw: Done

rest_lights.cpp
ebaauw: Done

PR #7923, opened 2024-09-12
DDF for LEDVANCE Plug Compact EU T

cosmo84:
DDF for LEDVANCE Plug Compact EU T


Lempomat:
Hi The plug has also a energy meter. How can we use this?

Issue #7931, opened 2024-09-15
DDF parsing error if only difference in manufacturername is a space character. Only the first is parsed.


olemr:
### Does the issue really belong here?

- [X] I definitively want to report a bug within deCONZ or its REST-API

### Is there already an existing issue for this?

- [X] I have searched the existing issues and there is none for the bug at hand

### Describe the bug

Found during https://github.com/dresden-elektronik/deconz-rest-plugin/issues/7925
The Sunricher vendor 's Namron manufacturer have released, for the similar HW device 4512729 two Firmware with different manufacturer names. 
Firmware 2.5.3_r20 presents `manufacturername = "NAMRON  AS"`  (note two spaces between N & A)
Firmware 2.5.3_r21 presents `manufacturername = "NAMRON AS"`  (note one space between N & A)

Now, in /usr/share/deConz/devices/sunricher/4button_switch.json DDF, only the first listed NAMRON AS entry is parsed correctly.

With a DDF like:
```
"NAMRON AS",
"NAMRON  AS",
``` 
Switches with r20 FW is not picked up, and Adding fails.
Changing to:
```
"NAMRON  AS",
"NAMRON AS",
``` 
and the Add succeeds.

### Steps to reproduce the behavior

1. Like described in Bug description.

### Expected behavior

Expected DDF parsing to pick up either one of the two NAMRON manufacturer names, not just the first one.

### Screenshots

NA

### Environment

- Host system: Raspberry Pi & PC
- Running method: Raspbian & Ubuntu
- Firmware version: 26400500
- deCONZ version 2.28.1
- Device: ConBee I
- Do you use an USB extension cable: yes + a powered HUB
- Is there any other USB or serial devices connected to the host system? No

### deCONZ Logs

This is from an (unsuccessful) attempt with an r20 switch and the NAMRON with 2 spaces second in the manufacturername list:
```
     45: 21:56:25:217 DDF identifier pair: NAMRON AS | 4512719
  15557: 21:56:25:590 Sqlite group: name = 4512719 38
  15563: 21:56:25:590 DB found group 4512719 38 0xEE01
  15565: 21:56:25:590 Sqlite group: name = 4512719 38
  15571: 21:56:25:590 DB found group 4512719 38 0xEE02
  16555: 21:56:25:630 Sqlite sensors: modelid = 4512719
  17912: 21:56:26:933 DEV modelId: 4512719, 0x60A423FFFE60F3DB
[...]

PR #7933, opened 2024-09-15
DDF for 4 button sunrichter/NAMRON AS switch 4512719

olemr:


olemr:
Also see https://github.com/dresden-elektronik/deconz-rest-plugin/issues/7931 for NAMRON 4512719 firmware differences.
But that issue was me failing to notice the `modelid` difference between 4512719 & 4512729 thinking the only error was the missing extra space as this used to work pre DDF. My bad.

manup:
Good catch anyway, two spaces is something new in Zigbee playground, hard to spot :+1:

Issue #7938, opened 2024-09-22
FireAngel AngelEye Carbon Monoxide Alarm + ZB-Module (ZBCO-AE-10X-EUR)


ebaauw:
### Is there already an existing issue for this?

- [X] I have searched the existing issues and there is none for my device

### Product name

AngelEye Carbon Monoxide Alarm + ZB-Module (ZBCO-AE-10X-EUR)

### Manufacturer

Fireangel

### Model identifier

Alarm_SD_Device

### Device type to add

Sensor

### Node info

<img width="494" alt="Screenshot 2024-09-22 at 17 37 36" src="https://github.com/user-attachments/assets/1b212fbb-9638-4d0b-9fcf-97c53dcabc78">

### Endpoints and clusters

<img width="257" alt="Screenshot 2024-09-22 at 17 39 11" src="https://github.com/user-attachments/assets/8c0f5982-d963-489e-89b7-ea28e2977383">

### Basic

<img width="494" alt="Screenshot 2024-09-22 at 17 37 21" src="https://github.com/user-attachments/assets/19e6f8b1-c784-42c4-ab3b-45e2724e3e0b">

### Power Configuration
<img width="490" alt="Screenshot 2024-09-22 at 17 45 55" src="https://github.com/user-attachments/assets/ba0c4d9c-b062-4e36-9370-6cabd1ab902d">

#### Poll Control
<img width="490" alt="Screenshot 2024-09-22 at 17 39 29" src="https://github.com/user-attachments/assets/f5c5bdce-8573-4af9-a210-bac814bb2e33">

### IAS Zone
<img width="494" alt="Screenshot 2024-09-22 at 17 38 13" src="https://github.com/user-attachments/assets/46d465f6-f735-4696-bd13-64bfb90470b1">

### IAS WD
<img width="494" alt="Screenshot 2024-09-22 at 17 38 42" src="https://github.com/user-attachments/assets/43793959-8f09-46fa-a757-f2979a0b22c3">

ebaauw:
[...]

Issue #6545, opened 2022-11-17
Fireangel ZB-Module


rrense:
## Device

- Product name: ZB-Module
  - [Link](https://www.fireangel.nl/products/zb-module)
- Manufacturer: Fireangel
- Model identifier: Alarm_SD_Device
- Device type :
  - Sensor (smoke alarm)

### Basic
![image](https://user-images.githubusercontent.com/4151359/202569312-e3007a2d-60ad-41d5-b6d8-7184c8762e13.png)

### Identify
![image](https://user-images.githubusercontent.com/4151359/202569581-b095cce8-189f-406f-954c-885da9fd4b91.png)

Issue #6111, opened 2022-06-04
AngelEye FireAngel Zigbee Module


peterjank:
## Device

- Product name: FireAngel Zigbee Module
- Manufacturer: AngelEye
- Model identifier: Zigbee Module P Line
- Device type : Please remove all unrelated device types. 

  - Sensor SmokeAlarm

### Bas
![screendump basic deconz](https://user-images.githubusercontent.com/39620539/172019554-d0885623-bbf0-4070-a3bb-67b32e7da3a6.jpg)
ic

### Diagn
[zigbee_capture.zip](https://github.com/dresden-elektronik/deconz-rest-plugin/files/8843663/zigbee_capture.zip)
ostics
sniffed zigbee traffic with another conbee 2
0x75c4			pan id
0x00212effff0426e7	mac address
0x00212effff0426e7	tc address
0x1136e241492dfa82a1b2120183801456	network key

### Other clusters that are not mentioned above
[phoscon search for smoke sensor FireAngel zigbee module.txt](https://github.com/dresden-elektronik/deconz-rest-plugin/files/8838364/phoscon.search.for.smoke.sensor.FireAngel.zigbee.module.txt)

the new sensor is not found so it can not be seen in the node screen

Mimiix:
Missing all screenshots.

PR #7941, opened 2024-09-24
Support Fire Angel Zigbee Module (P-LINE)

ebaauw:
This PR does two things:
1. Expose the _Zone Type_ attribute 0x0001 of _IAS Zone_ cluster 0x0500 for `matchexpr`, cf. the work done in #7642 for _OTA File Version_;
2. Add support for the Fire Angel Smoke Alarm (ST-630-DET) and Carbon Monoxide Alarm (ZBCO-AE-10X-EUR) equipped with the Fire Angel Zigbee Module (P-LINE), see #7938, #6111, #6545.

Ad 1) I changed the following:
- Define `RAttrZoneType` for `/attr/zonetype`, based on 0x0500/0x0001;
- ~~Write the value of 0x0500/0x0001 to the `zcl_values` table in the database;~~
- Add `RAttrZoneType` to `Device`;
- ~~Read the value of 0x0500/0x0001 into the `Device` `RAttrZoneType` on device DDF initialisation;~~
- Expose `RAtttrZoneType` on `lights` and `sensors` API endpoints.  Not sure if we want or need this.  It was already included in `sensors` web socket notifications out of the box.  Note that this attribute doesn't need to be defined on any of the subdevices for the `matchexpr` to work, since the C++ code already added it to the `Device`.

Note that I didn't add a hook to reload the DDF when 0x0500/0x0001 is read (and changes value), due to the recent issues with the API ignoring messages from devices after DDFs are re-initialised.

Ad 2) I added DDFs for:
- Smoke alarm, based on `matchexpr` for _Zone Type_ having value _Fire sensor_;
- CO alarm, based on `matchexpr` for _Zone Type_ having value _Carbon Monoxide (CO) sensor_;
- A generic alarm device (ZHAAlarm), based on `matchexpr` for other _Zone Type_ values.
 
Note the Zigbee module can also be installed in heat alarms, but I don't know what _Zone Type_ these use.  Could very well be _Fire sensor_.
On initial pairing, you should get a ZHAAlarm `sensors` resource; after reading _IAS Zone_ and restarting deCONZ, this changes to a ZHAFire or ZHACarbonMonoxide.
```json
{
  "config": {
    "battery": 91,
    "on": true,
    "pending": [],
    "reachable": true
  },
  "ep": 1,
  "etag": "d790c79357ddb6e33be69053fb9ca957",
  "lastannounced": null,
  "lastseen": "2024-09-25T09:17Z",
  "manufacturername": "Fireangel",
  "modelid": "Alarm_SD_Device",
  "name": "Smoke Alarm",
  "state": {
    "fire": false,
    "lastupdated": "2024-09-25T09:17:22.956",
    "lowbattery": false,
    "tampered": false,
    "test": false
  },
  "swversion": "3.1.8.14",
  "type": "ZHAFire",
  "uniqueid": "00:15:5f:00:dc:c8:ea:77-01-0500"
}
```
The only differences between the different devices are the `type` (ZHAAlarm, ZHAFire, ZHACarbonMonoxide) and the `state` attribute for an alarm (`alarm`, `fire`, `carbonmonoxide`).
I'm polling the alarm every 5 minutes for the _Zone State_, as the module doesn't support _Supervision Reports_ nor attribute reporting for this attribute.  This should keep `state/lastupdated` fairly current, but I need to assess the impact on battey usage.
`state/tampered` is set shortly after the smoke alarm is removed from it's base plate.  It takes a couple of seconds before the _Zone Status Change Notification_ is sent.  [Note to self: check if CO alarm also supports this].
`state/test` briefly switches to `true` when pressing the device button, activating a self-test (cover your ears!).
The Zigbee module only reports voltage if its (user replaceable) battery.  This should be reported every 12 hours out of the box.  `config/battery` extrapolates the voltage linearly from documented range of 1.9V ... 3.0V to 0% ... 100%.  Note that there's no indication for the internal, non-replaceable battery of the sensor.
`swversion` reports the firmware version of the Zigbee module, based on the _File Version_ of the _OTAU cluster_.

Additionally, there's a _Warning Device_ `lights` resource.
```json
{
  "capabilities": {
    "alerts": [
      "none",
      "select",
      "lselect"
    ]
  },
  "etag": "4d11a9ea58d1d06dffb73e74c989685f",
  "hascolor": false,
  "lastannounced": null,
  "lastseen": "2024-09-25T09:17Z",
  "manufacturername": "Fireangel",
  "modelid": "Alarm_SD_Device",
  "name": "Smoke Alarm",
  "state": {
    "alert": "none",
    "reachable": true
  },
  "swversion": "3.1.8.14",
  "type": "Warning device",
  "uniqueid": "00:15:5f:00:dc:c8:ea:77-01"
}
```
`state/alert` can be used to activate the siren (cover your ears!).  Note the module only supports _Sqawk_, so you only get a short alarm.


device_ddf_init.cpp
```diff
@@ -556,5 +556,19 @@ bool DEV_InitDeviceBasic(Device *device)
...
+    if (DB_LoadZclValue(&zclVal) && zclVal.data != 0)
+    {
+        ResourceItem *item = device->item(RAttrZoneType);
+        if (item && item->toNumber() != zclVal.data)
+        {
+            item->setValue(zclVal.data, ResourceItem::SourceDevice);
+            item->clearNeedPush();
+        }
+    }
+
```
manup, line 572: This works but you may consider just using the normal item based value storage which doesn't require extra code.

The reason I used this in the OTA version is to have the value in the database also for non DDF devices, so at the time they get a DDF the value is already available. Not sure if we need that here too.

ebaauw: @manup, are you saying that any item can be used in `matchexpr`, as long as we make sure there's always one DDF that matches, so the item is created?

ebaauw: After some testing, I think we do need the `RAttrZoneType` `ResourceItem` on the `Device`, but there's no need to save its value explicitly to `ZCL_VALUES`.  Still no need to have the item on the `lights` or `sensors` resource.

device.cpp
```diff
@@ -2157,6 +2157,7 @@ Device::Device(DeviceKey key, deCONZ::ApsController *apsCtrl, QObject *parent) :
     addItem(DataTypeString, RAttrDdfPolicy);
     addItem(DataTypeString, RAttrDdfHash);
     addItem(DataTypeUInt32, RAttrOtaVersion);
+    addItem(DataTypeUInt16, RAttrZoneType);
```
manup, line 2160: Just noticed this one, I'm not too happy about adding `RAttrZoneType`  in this place and to every device.

I think it's okay to get the PR up and running, but ideally we can mark items in sub-resources to be cascaded to the parent (aka Device*) in future.

ebaauw: I think there's a bit of a catch 22 there?  We only have the subdevices once the DDF has been matched, but we need the zone type to match the DDF.  We might do a check that it gets added only to devices with the _IAS Zone_ cluster?

ebaauw: Unfortunately, this doesn't work.  I think the value of `RAttrZoneType` is not persisted across restarts.  After re-adding the logic to persist zone type in `ZCL_VALUES`, it still doesn't work, as the attempted restore happens before the state enter event.  When re-creating the item on restore, it works!

> but there's no need to save its value explicitly to `ZCL_VALUES`

I can no longer reproduce this.  Maybe it worked, because at that time the resource item was still defined on the subdevices, or simply because the value was still in ZCL_VALUES.  Or maybe I simply forgot to deploy the plugin after compiling.  When manually removing the value from the database, the `sensors` resource reverts to ZHAAlarm.

manup:
Tested in my setup with a breakpoint in debugger, adding the zone type item works.

Issue #7948, opened 2024-09-28
IKEA INSPELNING Smart Plug with kWh meter


NotSoScientific:
### Is there already an existing issue for this?

- [X] I have searched the existing issues and there is none for my device

### Product name

INSPELNING

### Manufacturer

IKEA

### Model identifier

INSPELNING Smart plug

### Device type to add

Switch

### Node info

<img width="434" alt="Screenshot 2024-09-28 at 12 37 38" src="https://github.com/user-attachments/assets/f9a71fca-a41c-4204-95a7-17da5f0d843f">

### Endpoints and clusters

<img width="907" alt="Screenshot 2024-09-28 at 12 39 35" src="https://github.com/user-attachments/assets/33cd5209-78e3-4fe4-86a6-9a3c4dd0a075">

### Basic

<img width="506" alt="Screenshot 2024-09-28 at 12 47 21" src="https://github.com/user-attachments/assets/08f4b3b6-0c87-4edb-96ca-4f03d64f2b5c">

### On/Off
<img width="498" alt="Screenshot 2024-09-28 at 12 42 46" src="https://github.com/user-attachments/assets/dd5ffa9c-ff82-4876-86ec-f82db9825f7e">

### Level Control
<img width="493" alt="Screenshot 2024-09-28 at 12 43 02" src="https://github.com/user-attachments/assets/c61a4d5d-c7dd-4cde-b5ac-bb676dd42fe2">

### Simple Metering
<img width="502" alt="Screenshot 2024-09-28 at 12 49 36" src="https://github.com/user-attachments/assets/0c3d58f4-68f5-4fc6-a34b-fd97c19fa9fd">

### Electrical Measurement
<img width="497" alt="image" src="https://github.com/user-attachments/assets/fb3132e7-ca71-43fc-93f6-296300a28972">
<img width="501" alt="Screenshot 2024-09-28 at 12 44 39" src="https://github.com/user-attachments/assets/7224b249-1cbb-4f7f-80b1-b6def6bfff68">
<img width="505" alt="Screenshot 2024-09-28 at 12 45 01" src="https://github.com/user-attachments/assets/a486ba16-fd2b-4447-812e-2bfe549801d8">
<img width="473" alt="Screenshot 2024-09-28 at 12 45 40" src="https://github.com/user-attachments/assets/cfa98f56-e93d-440c-b133-56fa7d4ba2a8">
[...]

PR #7959, opened 2024-10-07
DDF for IKEA INSPELNING smart plug

bluemoehre:
Solves https://github.com/dresden-elektronik/deconz-rest-plugin/issues/7948

Need verification / double check of Consumption calculation.
Phoscon shows value as `wH`, while comments and docs states value should be in `kWh`.

Anyway, I have already tested and confirmed that the raw value `0x0702 / 0x0000` in deCONZ GUI is in `wH`. Based on the divisor value from `0x0302`, I added a x1000 division.

Maybe @ebaauw can help?

### DIFF
--- a/devices/ikea/inspelning_smart_plug.json
+++ b/devices/ikea/inspelning_smart_plug.json
@@ -44,7 +44,7 @@
         },
         {
           "name": "state/on",
-          "refresh.interval": 300,
+          "refresh.interval": 900,
           "read": {
             "at": "0x0000",
             "cl": "0x0006",
@@ -108,7 +108,7 @@
         },
         {
           "name": "state/consumption",
-          "refresh.interval": 300,
+          "refresh.interval": 10800,
           "read": {
             "at": "0x0000",
             "cl": "0x0702",
@@ -172,7 +172,7 @@
         },
         {
           "name": "state/current",
-          "refresh.interval": 300,
+          "refresh.interval": 600,
           "read": {
             "at": "0x0508",
             "cl": "0x0b04",
@@ -192,7 +192,7 @@
         },
         {
           "name": "state/power",
-          "refresh.interval": 300,
+          "refresh.interval": 600,
           "read": {
             "at": "0x050b",
             "cl": "0x0b04",
@@ -209,7 +209,7 @@
         },
         {
           "name": "state/voltage",
-          "refresh.interval": 300,
+          "refresh.interval": 600,
           "read": {
             "at": "0x0505",
             "cl": "0x0b04",
@@ -223,6 +223,63 @@
             "eval": "if (Attr.val != 65535) { Item.val = Attr.val; }"
           },
           "default": 0
+        }
+      ]
+    }
+  ],
+  "bindings": [
+    {
+      "bind": "unicast",
+      "src.ep": 1,
+      "cl": "0x0006",
+      "report": [
+        {
+          "at": "0x0000",
+          "dt": "0x10",
+          "min": 1,
+          "max": 600
+        }
+      ]
+    },
+    {
+      "bind": "unicast",
+      "src.ep": 1,
+      "cl": "0x0702",
+      "report": [
+        {
+          "at": "0x0000",
+          "dt": "0x25",
+          "min": 60,
+          "max": 3600,
+          "change": "0x00000001"
+        }
+      ]
+    },
+    {
+      "bind": "unicast",
+      "src.ep": 1,
+      "cl": "0x0B04",
+      "report": [
+        {
+          "at": "0x0505",
+          "dt": "0x21",
+          "min": 10,
+          "max": 300,
+          "change": "0x00000005"
+        },
+        {
+          "at": "0x0508",
+          "dt": "0x21",
+          "min": 10,
+          "max": 300,
+          "change": "0x00000064"
+        },
+        {
+          "at": "0x050B",
+          "dt": "0x29",
+          "min": 10,
+          "max": 300,
+          "change": "0x0000000A"
         }
       ]
     }
--- a/devices/ikea/inspelning_smart_plug.json
+++ b/devices/ikea/inspelning_smart_plug.json
@@ -1,0 +1,230 @@
+{
+  "schema": "devcap1.schema.json",
+  "manufacturername": "$MF_IKEA",
+  "modelid": "INSPELNING Smart plug",
+  "vendor": "IKEA",
+  "product": "INSPELNING smart plug",
+  "sleeper": false,
+  "status": "Gold",
+  "subdevices": [
+    {
+      "type": "$TYPE_SMART_PLUG",
+      "restapi": "/lights",
+      "uuid": [
+        "$address.ext",
+        "0x01"
+      ],
+      "items": [
+        {
+          "name": "attr/id"
+        },
+        {
+          "name": "attr/lastannounced"
+        },
+        {
+          "name": "attr/lastseen"
+        },
+        {
+          "name": "attr/manufacturername"
+        },
+        {
+          "name": "attr/modelid"
+        },
+        {
+          "name": "attr/name"
+        },
+        {
+          "name": "attr/swversion"
+        },
+        {
+          "name": "attr/type"
+        },
+        {
+          "name": "attr/uniqueid"
+        },
+        {
+          "name": "state/on",
+          "refresh.interval": 300,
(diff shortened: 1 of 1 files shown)


ebaauw:
The API should report consumption in Wh, see https://github.com/dresden-elektronik/deconz-rest-plugin/blob/master/devices/generic/items/state_consumption_item.json.

If the device reports _Current Summation Delivered_ in Wh, no conversion is needed.  I don't use Phoscon, and have no clue how they report consumption.

The devices that I have don't use divisors/multipliers on _Simple Metering_, but instead use _Summation Formatting_.  I never really figured out how that works, so I tend to use a known load (e.g. an old 60W Tungsten bulb), and see what the device reports to figure out it's scaling.  Also check the _Unit of Measure_ attribute.  If that's _kW and kWh binary_ (what I see in my cases), and 0x0000 reports in Wh, the divisor should be 1000, to transform Wh to the advertised kWh.

SwoopX:
Actually, the logic here is quite simple:
The unit of measurement is kW/kWh. To get it, take the value and apply multiplier and divisor.

If, for the given device, attribute 0x0000 gives us the value of 4, the math is 4*1/1000 = 0,004. That means the device reports already in Wh and the DDF value calculation is incorrect.

Ideally, the multipliers and divisors are correct. However, we also had enough devices out there where that is just rubbish.

bluemoehre:
I was testing 1h having a 6.5W bulb connected and total consumption (`0x0702 / 0x0000`) reported `6`. 

image

As far as I understand these data, there are only two approaches possible:

### A
**Formatting** block just **tells clients how to convert values** to get to "kW & kWh binary" units, which is a default display to the user.

- **raw values are in base units** (plain Watts - no milli, no kilo etc.)
- **raw values are provided** to the API (no DDF conversion takes place)
- **formatting info is provided** in the API as well (clients can use it if they want to convert)

### B
**Formatting** block **tells devs how to convert values** to meet standards.

- **Values must be converted** to meet standards (DDF conversion takes place)
- **Standard** is described in the **formatting block** (to meet "kW & kWh binary")
- **Conversion** needs to take place **before providing** data to the API (clients get user-ready data)


---

It's just a matter of what is the default/specs.
For both variants the available data + formatting info generally is correct.

Based on your feedback @ebaauw, as well as what I see in the Phoscon UI, I think the first variant is the one to go.
So, I gonna remove the division.

SwoopX:
> Formatting block / measured value is just bananas

Not sure where I should have said that, but I warned taking things for granted. That's also why it was suggested to verify with a known load. The zigbee spec is quite clear in my view how to use the formatting attributes (as described above) and also that the summation formatting doesn't play a considerable role 🤷‍♂️

bluemoehre:
I noticed this DDF is working fine up to `999 Watt`:

`/api/sensors`

```json
[
"119": {
        "config": {
            "on": true,
            "reachable": true
        },
        "etag": "fc5336ccc8a702bc7088ee0b430738b1",
        "lastannounced": null,
        "lastseen": "2024-10-12T12:54Z",
        "modelid": "INSPELNING Smart plug",
        "name": "Consumption 119",
        "state": {
            "consumption": 102,
            "lastupdated": "2024-10-12T12:27:59.387"
        },
        "swversion": "2.4.34",
        "type": "ZHAConsumption",
        "uniqueid": "xx:xx:xx:xx:xx:xx:xx:xx-01-0702"
    },
    "120": {
        "config": {
            "on": true,
            "reachable": true
        },
        "etag": "ac980bb68841173491ca2404464923b9",
        "lastannounced": null,
        "lastseen": "2024-10-12T12:54Z",
        "modelid": "INSPELNING Smart plug",
        "name": "Power 120",
        "state": {
            "current": 3,
            "lastupdated": "2024-10-12T12:54:56.327",
            "power": 2,
            "voltage": 237
        },
        "swversion": "2.4.34",
        "type": "ZHAPower",
        "uniqueid": "xx:xx:xx:xx:xx:xx:xx:xx-01-0b04"
    }
]
```

Consuming **>=** `1000 Watt`, **the device changes the divisors / multipliers**, which leads to wrong values when using hardcoded multipliers. That behavior was already reported here: https://github.com/Koenkk/zigbee2mqtt/issues/23961#issuecomment-2366733453

So my idea (https://github.com/dresden-elektronik/deconz-rest-plugin/issues/7948#issuecomment-2386483403) became a real requirement sooner as expected.

I need to know how to access the other attributes (cluster `0x0b04`, address block `0x0600`) within the `parse`/ `eval` property for a correct calculation. Can anyone help?

ebaauw:
Now there's an interesting use case.  I think the multiplier and divisor items should be (read-only) `state` items, rather than (read-write) `config` items.  I always thought of them as (constant) `capabilities` items, but since they might change value (other than through firmware upgrade), they clearly aren't.  I also think we should not expose the multiplier nor divisor item over the API, since we don't want API clients to implement the logic to handle these.

Does the device do something similar for `current`?

So we would need to create two non-public items and change the `eval` to consider these, while taking care of the Wh vs kWh logic.  Since the new items aren't exposed over the API, I think no change to the C++ code is needed for them.

Note to self: need to get one of these plugs.

bluemoehre:
Feel free to add another commit vs my branch, if you know a solution, @ebaauw .

> Does the device do something similar for current?

Didn't see that yet. Still haven't the update to the latest version, due it's still unavailable in DE, or still at beta stage. (People reported a completely different URL to get that from - I guess its a pre-prod host)

I want to point out another thought:

Since I already tested a lot what is a meaningful "report interval" + "minimal change to report", it makes sense to use the factor for the "minimal change to report" value. Connecting a Raspberry, you may be interested in reporting changes of 0,1 Watts, but having a toaster you are more interested in 10 Watt steps. When the factor changes the report config could be changed as well. What do you think?

ebaauw:
> it makes sense to use the factor for the "minimal change to report" value.

I don't think it does, actually.  The reporting change is on the raw value, as reported by the attribute, not on the derived value after considering the multiplier/divisor.  So if you set a change of 1, it will report changes from 1W between 0 and 1000W and changes from 10W above 1000W (not sure I got the scale right, but I hope you get the idea).  This is probably exactly what you want.

The idea is intriguing, though.  I've never seen attribute reporting configuration being changed dynamically.  Not sure if that's even doable from DDF.  You could send the correct Zigbee message using `zcl:cmd` with the right `fc` (making sure to send a generic rather than a cluster-specific command), but I'm not sure how to trigger it.  I'd probably create a dummy item with a `write` to send the report config, make the multiplier/divisor visible on the API, and create a rule to update the dummy item when the multiplier/divisor changes.  I would be an interesting proof of concept, but I sure wouldn't recommend this for production setups.

bluemoehre:
> This is probably exactly what you want.

You are right! Actually it will result exactly in that behavior I wish. What a great Idea by IKEA to use the attributes that way.

So we only need to use the factors in the evaluation of the value to get this working.

bluemoehre:
> I think the multiplier and divisor items should be (read-only) state items

@ebaauw can you add a commit to my branch or briefly tell me how I can add that part?

ebaauw:
I don’t I see your branch, it’s probably only available on your fork?

Basically, you need create two new item files in `devices/generic/items` for `state/power_multiplier` and `state/power_divisor`.  I don’t think they need to be exposed by the API (for now), so this wouldn’t require any C++ changes.  Not sure if the devices actually uses both.

I did pick up an INSPELNING at IKEA, but have yet to put it to the test. I want to install it for my washing machine, but better double check the supported load first.

bluemoehre:
> I don’t I see your branch, it’s probably only available on your fork?

Yes, it's on my fork. Most of the time you can't create branches on remote repos, so I'm used to that workflow. (Re-Checked: right now it's still disabled =)

bluemoehre:
> Basically, you need create two new item files in `devices/generic/items` for `state/power_multiplier` and `state/power_divisor`. I don’t think they need to be exposed by the API (for now), so this wouldn’t require any C++ changes. Not sure if the devices actually uses both.

I drafted another PR #7978 for this.

mattreim:
@bluemoehre The DDF has already been merged, so this PR can now be closed. Thank you.

bluemoehre:
@mattreim alright. Any idea what's about the other related PR #7978 ?

mattreim:
@bluemoehre In my opinion, this PR #7978 can also be closed because the file `state_power_divisor_item.json`  has also been filed.

PR #7963, opened 2024-10-12
Add Danfoss manufacturer specific commands to general.xml

SwoopX:
With the recent change in deconz core, manufacturer specific commands are now only presented upon matching manufacturer code. For Danfoss thermostats, the following commands will be added.

grafik


Jonaslinderoth:
Very nice, looking forward to using it.
Is there any way to call these commands through the api?

PR #7976, opened 2024-10-16
DDF for Aqara Smart plug (lumi.plug.maeu01) v45

bluemoehre:
From v45 Aqara removed current & voltage reporting, so we need a new DDF for that version only.

### DIFF
--- a/devices/xiaomi/xiaomi_sp-euc01_smart_plug.v45.json
+++ b/devices/xiaomi/xiaomi_sp-euc01_smart_plug.v45.json
@@ -1,0 +1,235 @@
+{
+  "schema": "devcap1.schema.json",
+  "manufacturername": "$MF_LUMI",
+  "modelid": "lumi.plug.maeu01",
+  "vendor": "Xiaomi Aqara",
+  "product": "Smart plug (SP-EUC01)",
+  "sleeper": false,
+  "status": "Gold",
+  "comment": "DDF for device firmware >= v45",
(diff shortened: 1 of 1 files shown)


SwoopX (changes requested):
I'm afraid this DDF will affect the vast majority and potentially break the power measurement capabilities altogether. Also, the cluster/attribute combination for the measurement sensors is not the right one.

Based on that, I cannot support the anticipated change.

bluemoehre:
> I'm afraid this DDF will affect the vast majority and potentially break the power measurement capabilities altogether.

To not affect `< v45` we added the Regexp, since the previous versions have the other clusters available. All tests so far have been okay locally. But I agree it would be more solid to match against the exact version (lets say `v45+`). Is there a way to achieve this - or - any other recommendation to pin a version?

> Also, the cluster/attribute combination for the measurement sensors is not the right one.

Can you tell us the correct one so that we can update it? Thanks!

SwoopX:
Hi,

to pin a DDF depending on a device firmware version, please have a look at this PR: https://github.com/dresden-elektronik/deconz-rest-plugin/pull/7642 Haven't done it myself so far, so I can't share any experience.

Regarding the cluster/attribute combination, the correct ones are in the already existing DDF (power and consumption: 0x000C/0x0055, voltage and current comes via Xiaomi special attribte reporting).

mattreim:

PR #7978, opened 2024-10-17
add read-only attributes for power formatting factors

bluemoehre:
Since some measurement devices provide calculation factors to format measured values correctly, we need to make them available. These factors may change on each report to adapt to the provided output power range. This prevents spamming the network when the "minimum change" property of the reporting binding is set to a low value.

DDF configs may use these attributes to guarantee API values are always in the same unit (Watts).

### Example

| ActivePower | ACPowerMultiplier | ACPowerDivisor | Result
|:--------:|:--------:|:--------:|:--------:|
| 42 | 1 | 1000 | 42 mW |
| 42 | 1 | 1 | 42 W |
| 42 | 1000 | 1 | 42 kW |

### Screenshot

image

### Specifications

> 4.9.2.2.6.10 ActivePower
> Represents the single phase or Phase A, current demand of active power delivered or received at the premises, in Watts (W). Positive values indicate power delivered to the premises where negative values indicate power received from the premises

> 4.9.2.2.7.5 ACPowerMultiplier
> Provides a value to be multiplied against the InstantaneousPower and ActivePower attributes. This attribute
must be used in conjunction with the ACPowerDivisor attribute. `0x0000` is an invalid value for this attribute.
> 
> 4.9.2.2.7.6 ACPowerDivisor
> Provides a value to be divided against the InstantaneousPower and ActivePower attributes. This attribute
must be used in conjunction with the ACPowerMultiplier attribute. `0x0000` is an invalid value for this attribute.

### DIFF
--- a/devices/generic/items/state_power_divisor_item.json
+++ b/devices/generic/items/state_power_divisor_item.json
@@ -1,0 +1,22 @@
+{
+  "schema": "resourceitem1.schema.json",
+  "id": "state/power_multiplier",
+  "datatype": "UInt16",
+  "access": "R",
+  "public": false,
+  "description": "AC Power Divisor to be multiplied against the Instantaneous Power and Active Power attributes.",
+  "read": {
+    "at": "0x0605",
+    "cl": "0x0b04",
+    "ep": 0,
+    "fn": "zcl:attr"
+  },
+  "parse": {
+    "at": "0x0605",
+    "cl": "0x0b04",
+    "ep": 0,
+    "eval": "if (Attr.val > 0) { Item.val = Attr.val; }"
+  },
+  "refresh.interval": 600,
+  "default": 1
+}
--- a/devices/generic/items/state_power_multiplier_item.json
+++ b/devices/generic/items/state_power_multiplier_item.json
@@ -1,0 +1,22 @@
+{
+  "schema": "resourceitem1.schema.json",
+  "id": "state/power_multiplier",
+  "datatype": "UInt16",
+  "access": "R",
+  "public": false,
+  "description": "AC Power Divisor to be divided against the Instantaneous Power and Active Power attributes.",
+  "read": {
+    "at": "0x0604",
+    "cl": "0x0b04",
+    "ep": 0,
+    "fn": "zcl:attr"
+  },
+  "parse": {
+    "at": "0x0604",
+    "cl": "0x0b04",
+    "ep": 0,
+    "eval": "if (Attr.val > 0) { Item.val = Attr.val; }"
+  },
+  "refresh.interval": 600,
+  "default": 1
+}


bluemoehre:
Several DDFs could possibly be updated (of course this requires the related Formatting Block to report correct factors):

https://github.com/search?q=repo%3Adresden-elektronik%2Fdeconz-rest-plugin+%22at%5C%22%3A+%5C%220x050b%22%2C+eval+%22Attr.val+%2F+10%22&type=code

https://github.com/search?q=repo%3Adresden-elektronik%2Fdeconz-rest-plugin+%22at%5C%22%3A+%5C%220x050b%22%2C+eval+%22Attr.val+*%22&type=code

bluemoehre:
@ebaauw :
> … so this wouldn’t require any C++ changes.

Just for safety: Won't [these files](https://github.com/search?q=repo%3Adresden-elektronik%2Fdeconz-rest-plugin+%22Attr.val+%2F+10%22+cpp&type=code) interfere?

ebaauw:
I wouldn't change the logic on the generic `state/power` item, unless you can test each and every one of the impacted devices.  We've seen too many devices that don't use the multiplier/devisor correctly.  Best just define the generic multiplier and divisor items, and overwrite the (unchanged) generic `state/power` item in the DDF for the IKEA (and later for other plugs with the same behaviour).

> Just for safety: Won't [these files](https://github.com/search?q=repo%3Adresden-elektronik%2Fdeconz-rest-plugin+%22Attr.val+%2F+10%22+cpp&type=code) interfere?

Interfere with what?  The methods are invoked only for non-managed devices (exposed by legacy code instead of DDF), see https://github.com/dresden-elektronik/deconz-rest-plugin/blob/ef6ca9f1a968f07699e3d511db1a22e0850c2300/de_web_plugin.cpp#L1287
I think the `DDF_AnnoteZclParse()` is used when creating a new DDF from the GUI.

bluemoehre:
> I wouldn't change the logic on the generic state/power item, unless you can test each and every one of the impacted devices.

Most of the affected DDFs already have a custom eval prop with their own calculation. As long as their config has precedence over the generic files, we are totally fine to change that. For those which haven't I can add a custom eval prop as well, so they gonna behave the same as before.
I would prefer to make things right in first place and meet the specs, then fix all the aberrations. Should be more future proof.

> The methods are invoked only for non-managed devices 

Perfect. I was just checking if there are additional conflicts for this change.

devices/generic/items/state_power_item.json
```diff
@@ -15,7 +15,7 @@
     "at": "0x050b",
     "cl": "0x0b04",
     "ep": 0,
-    "eval": "if (Attr.val != -32768 && Attr.val != 32768) { Item.val = Attr.val; }"
+    "eval": "if (Attr.val != -32768 && Attr.val != 32768) { Item.val = Attr.val * R.item('state/power_multiplier').val / R.item('state/power_divisor').val; }"
```
ebaauw, line 18: - Might want to add `Math.round()` to make sure the reported value is integer.
- Signed `s16` supports values from -32768 to 32767, so `Attr.val != 32768` is superfluous.  Typically the lowest value, -32768 is used to indicate an invalid value.  
Unsigned `u16` supports values from 0 to 65535, and uses for 65535 as invalid value.

bluemoehre: > Signed s16 supports values from -32768 to 32767, so Attr.val != 32768 is superfluous.

Actually I didn't change that. [This code is already in master](https://github.com/search?q=repo%3Adresden-elektronik%2Fdeconz-rest-plugin%20%22Attr.val%20!%3D%2032768%22&type=code). Do we need to clean it up?

ebaauw: Yes, I saw that.  Better cleanup while you’re touching it anyways.  Sorry to put this on you.

bluemoehre: > Might want to add Math.round() to make sure the reported value is integer.

The reported value is not required to be Integer - am I wrong? JavaScript Number(s) are always floating point. Normally it is up to the UI to take care about display and formatting. \
While I was testing with the plug's, I already had problems reporting the standby demand (it was always `0W` in Phoscon, while the RAW value was `4` in deCONZ). I can't tell if that's related to this specific code, but for sure if we reduce resolution here, we throw away information and my end up not being able to show fractions.

bluemoehre: > Yes, I saw that. Better cleanup while you’re touching it anyways. Sorry to put this on you.

Because the issue is not only related to the files I am touching here, I would like to issue another PR for the cleanup only, to not mix up topics. (It should not matter if that one then gets merged before or after this one, since we can rebase)

bluemoehre:
Devices which _**may** get affected_ by this change:

- legrand/Contactor.json
- bosch/plug_compact.json
- legrand/cable_outlet.json
- immax/07048L_smart_plug.json
- adeo/ldsenk02f_plug.json
- innr/sp_234.json
- innr/sp_120.json
- innr/sp_240.json
- tuya/_TZ3000_cehuw1lw_smartplug_EU.json
- wiser/smart-plug-single-16A.json
- blitzwolf/bw_shp13_smart_plug.json
- blitzwolf/bw_shp15_smart_plug.json
- lidl/hg08673.json
- neo/NAS-WR01B_TS011F.json
- tuya/_TZ3000_typdpbpg_smart_plug_eu.json
- tuya/nous_a1z_smart_plug.json
- namron/4512737_thermostat.json
- namron/5401395_heater.json
- ubisys/s1_5501_s1r_5601.json
- ubisys/s2_5502.json
- ubisys/s2r_5602.json
- bosch/bmct-slz_shutter_light_control2.json
- ubisys/j1r_5602.json
- ubisys/j1_5502.json
- ouellet/OTH4000-ZB.json
- sinope/th1124zb.json
- ubisys/d1_5503_d1r_5603.json

To prevent any possible breaking, we could add `"eval": "if (Attr.val != -32768) { Item.val = Attr.val; }"` to these.

devices/generic/items/state_power_item.json
bluemoehre: Since the cleanup was moved to a dedicated PR and doesn't block us here, is there anything which prevents us continuing? (I don't want to leave this stalled, since it already took multiple days of work)

If the calculation is fine, let's resolve this conversation.

bluemoehre: @ebaauw bump

mattreim:

PR #7961, opened 2024-10-10
ResourceItems for device firmware

ebaauw:
To support automated checking if newer firmware is available for a device, the API needs to expose the firmware parameters, _Manufacturer Code_, _Image Type_, and _File Version_.  This PR adds support for resource items `cap/otau/manufacturer_code`, `cap/otau/image_type`, and `cap/otau/file_version`, to be included in DDFs of devices that support OTA firmware upgrades.  These items are exposed as a separate map under `capabilities` on `lights` and `sensors` resources:
```
  "capabilities": {
    "otau": {
      "file_version": 16786690,
      "image_type": 276,
      "manufacturer_code": 4107
    }
  }
```
They're also included as separate map in the `capabilities` websocket notifications.

Note that the items are populated from the _Query Next Image Request_, rather than from the optional (!) _OTAU_ cluster attributes.  `cap/otau/manufacturer_code ` sends an _Image Notify_ on `read`, triggering the _Query Next Image Request_, which also populates table in the _OTA Update_ panel in the GUI.

This PR adds the new items to the DDFs for Hue devices and for some others that I had handy.

TODO:
- Add these attributes to devices (exposed by legacy code) on receipt of the _Query Next Image Request_.

TODO on the longer term:
- Define these attributes (only) on the device, instead of on each sub-device.
- Deprecate the existing `attr/otaversion`(based on the OTAU cluster attribute),  in favour of `cap/otau/fileversion`?  It's used only as `matchexpr` in the DDF for the Trådfri on/off switch (after the attributes are defined on device level, and we've made sure they's persisted in the database).
- Add these items to other devices supporting OTAU firmware upgrades.
- Refactor `rest_lights.cpp` and `rest_sensors.cpp` to build the maps dynamically (cf. websocket notifications for `sensors`) rather than listing all items explicitly (see #7979).

PR #7979, opened 2024-10-18
Refactor `lightToMap()` and `sensorToMap()`

ebaauw:
This PR refactors the construction of the API responses to GET `/lights and GET `/sensors` and of web socket notifications for `/lights` and `/sensors`.  ~~It builds on #7961, which should be merged first.~~

## Changes
This PR makes the following changes:
- Set the following items as non-public, since they shouldn't be shown in the API:
  - `attr/id`;
  - `attr/otaversion`;
  - `cap/bri/move_with_onoff`;
  - `cap/on/off_with_effect`;
  - `config/bri/options`.
- ~~Change the data type to Double for the following items:~~ Canceled for now, as this breaks the evaluation of `gt` and `lt` rule conditions on these items.
  - ~~`state/consumption` and variants;~~
  - ~~`state/current` and variants;~~
  - ~~`state/power`;~~
  - ~~`state/pressure` and bis;~~
  - ~~`state/voltage`.~~
- Add non-public item `state/power_divisor` for the IKEA INSPELNING (see #7948).  Note that this needs to be defined in C++ for it to be used in expressions in the DDF.
- Add `ResourceItemDescriptor::toApi()`, which returns a pointer to the (sub-)map and the key where the value of the corresponding item should be inserted in the API response map.  Any intermediate sub-maps are created where needed.  It returns a small helper class `ApiAttribute` containing a QVariantMap pointer (to the sub-map) and two QString (for the key and top-level key).
- Refactor `lightsToMap()` and `sensorsToMap()`:
  - Remove the need to whitelist resource items with a value that can be obtained by `toVariant()`.  The entire multi-level map structure is now built dynamically.  This leaves only two types of whitelisted resource items:
    - Those with a complex value (like `config/on/startup`, that's either a boolean or `"default"`);
    - Those that are reported in a combined value (like `state/xy`, built from `state/x` and `state/y`).
  - Insert the public resource items from the parent device into the response.  This is done before the `lights` or `sensors` resource items are processed, so these might override the parent values.  In practice, this causes the following additional keys in the API response: `attr/ddf_hash`, `attr/ddf_policy`, `attr/extaddress`, `attr/nwkaddress`, `cap/sleeper`, and (for `/sensors`) `state/reachable`.  Note that non-Zigbee `sensors` don't have a parent device.
  - Remove (what I believe to be) dead code:
    - Special exception for Hue App (`ApiModeHue`).  Afaik the Hue app hasn't been working with deCONZ for a long time anyway, and there are alternatives, like Hue Essentials. 
    - A hook to exclude certain items when the functions are invoked for GET `/devices`.  Currently they aren't, and the hook would be obsolete once we move those items from the sub-devices to the device.
  - Handle `needPushChange` logic to support web socket notifications. 
- Refactor `handleLightEvent()` to leverage `lightToMap()`.
- Refactor `handleSensorEvent()` to leverage `sensorToMap()`.
- Change all string constants in the refactored functions to `QLatin1String` instances.

## Discussion
- While removing the device items from the sub devices will work brilliantly for GET, I'm not sure what the logic for the web socket notifications on those items should be.  I don't think we currently do notifications on `/devices`?  Would we want a single notification on the device level, or a separate notifications on each sub device?  In the latter case, I think we would need "slave" resource items on the sub-devices to keep track of what's already been notified.
- We might want to mark some of the device resource items as non-public?
- Do we want to deprecate `config/reachable` (for `/sensors`) in favour of `state/reachable`?  Same for `config/alert`?

## Next Steps
- Figure out how to deal with web socket notifications for device-level attributes.
- Move following items to the device:
  - `attr/lastannouced`, `attr/lastseen`;
  - `cap/otau/...`;
  - `cap/alerts`, `state/alert`, `config/alert`: in theory these could be per endpoint (with _Identify_ cluster);
  - `state/reachable` and `config/reachable`.


manup:
Thanks a lot for the PR. From a first coarse look the direction where this goes is pretty good. I don't understand every detail yet and need to dig deeper, but if we can build outgoing API data with less item/device specific C++ code that's wonderful.

Ideally using logic and structure from the dynamic objects which are described in DDF items. Looking at the PR and `toApi()` function the JSON files describing DDF items could be extended in future:

Complex "virtual" items: like `state/xy` could be described right in the DDF item JSON file as expression:

```json
{
  "id": "state/xy",
  "datatype": "Array",
  "to_api_expr": "[ R.item('state/x').val, R.item('state/y').val ];" 
}
```

Where the default only returns `.val` of the current item.

A handler of these would further reduce the `needPushChange()` and `clearNeedPush()` calls. The implicit task that `state/xy` depends on `state/x` and `state/y` and should be pushed if one of those is "dirty" could be handled automatically without extra noise in JSON description.

ebaauw:
I resolved conflicts and validation errors after recent merges of other PRs.  And I double-checked that this PR still compiles.

manup:
The PR now runs in my setup, fingers crossed. I have to admit it's rather hard to review. Overall I like that we get rid of ~600 lines of code.

```
const char *RConfigLastChangeAmount = "config/lastchange/amount";
const char *RConfigLastChangeSource = "config/lastchange/source";
const char *RConfigLastChangeTime = "config/lastchange/time";
```

These were changed to be nested via `/`, might be good to put them in above description.

For the `toApi()` function I'd like to propose a light weight alternative after sleeping over it, with less pointer and Qt string stuff (maybe as an PR on top of this).

ebaauw:
> The PR now runs in my setup, fingers crossed.

I've been running it for months now.

> I have to admit it's rather hard to review. 

I'm open to suggestions to make it easier to review, but, I agree, this feels like open heart surgery.

> Overall I like that we get rid of ~600 lines of code.

That, and it's future-proof, in that we won't need additional lines of codes for new (nested) resource items.

> These were changed to be nested via /, might be good to put them in above description.

They were already nested in the API output, courtesy of the legacy code.  I only changed the resource item suffix, so the nesting could be handled by `sensorToMap()` in the standard way, without the need to whitelist these resource items.

PR #7984, opened 2024-10-19
remove superfluous eval conditions

bluemoehre:
Signed s16 integer supports values from `-32768` to `32767`, so `Attr.val != 32768` is superfluous. Typically the lowest value, `-32768` is used to indicate an invalid value.

### DIFF
--- a/devices/Jasco-GE/GE45856_wall_switch.json
+++ b/devices/Jasco-GE/GE45856_wall_switch.json
@@ -127,7 +127,7 @@
             "at": "0x0400",
             "cl": "0x0702",
             "ep": 1,
-            "eval": "if (Attr.val != -32768 && Attr.val != 32768) { Item.val = Attr.val / 10; }"
+            "eval": "if (Attr.val != -32768) { Item.val = Attr.val / 10; }"
           },
           "default": 0
         }
--- a/devices/eva/powermeter.json
+++ b/devices/eva/powermeter.json
@@ -70,7 +70,7 @@
             "at": "0x0304",
             "cl": "0x0b04",
             "ep": 0,
-            "eval": "if (Attr.val != -32768 && Attr.val != 32768) { Item.val = Attr.val; }"
+            "eval": "if (Attr.val != -32768) { Item.val = Attr.val; }"
           }
         },
         {
--- a/devices/generic/items/state_power_item.json
+++ b/devices/generic/items/state_power_item.json
@@ -15,7 +15,7 @@
     "at": "0x050b",
     "cl": "0x0b04",
     "ep": 0,
-    "eval": "if (Attr.val != -32768 && Attr.val != 32768) { Item.val = Attr.val; }"
+    "eval": "if (Attr.val != -32768) { Item.val = Attr.val; }"
   },
   "refresh.interval": 300,
   "default": 0
--- a/devices/ouellet/OTH4000-ZB.json
+++ b/devices/ouellet/OTH4000-ZB.json
@@ -300,7 +300,7 @@
             "at": "0x050b",
             "cl": "0x0b04",
             "ep": 1,
-            "eval": "if (Attr.val != -32768 && Attr.val != 32768) { Item.val = Attr.val; }"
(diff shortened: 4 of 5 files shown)


bluemoehre:
@ebaauw FYI

SwoopX:
While this is technically correct, there's devices out there depending on this check to discard blips. Given the fact that this check doesn't hurt and there is a necessity, we should keep this untouched.

bluemoehre:
> While this is technically correct, there's devices out there depending on this check to discard blips.

Do we know which ones? 

> Given the fact that this check doesn't hurt and there is a necessity, we should keep this untouched.

That's bad practice. Like @ebaauw asked me to change this (at least for the generic one), and I found the same "error" in other files, it may happen again that people reading the code want to fix that, because it looks like a bug. If only a few people have that hidden knowledge, it's a risk. \
I would recommend to make the generic one correct and then take care about the the blippy bloppy devices. We should also leave a comment in the code or add a test, so everyone can understand why it's like that and prevent accidentally breaking something.\
What do you think?

SwoopX:
I'm afraid I cannot follow you to a certain extend. Where is the bad practice? We're discussing a valid boundaries check here being there for a reason. This is neither a bug (which is by definition a programatical misbehavior), nor an error (as nothing is working falsely) so there is nothing to fix in my view.

Leaving the code as is has 0 risk, changing the code as proposed introduces an unneccessary risk of breaking functionality of an unknown number of devices, hence I prefer avoiding that risk.

Regarding your suggesting to add a respective comment in the code, I'd fully buy in. Although this possibility is not given yet (I find the `"description"` key not really suitable for this purpose and placement is restricted), we should introduce a `"comment"` key to allow for some (important) documentation within a DDF at the place specifically required.

ebaauw:
It's technically not possible for an `s16` to report a value >= 32768; it simply doesn't fit the 16 bits..  Any device depending on this check would have to use a different data type.  That alone is reason enough to override the default item definition in the device's DDF.

bluemoehre:
> Where is the bad practice?

_"Given the fact that this check doesn't hurt" [...] "we should keep this untouched"_

It's bad practice to keep a workaround in a generic file just because it doesn't hurt. Especially if only a few devices are to be fixed with it. The generic files should kept clean all the time, since the amount of devices will continuously keep growing. (Imagine having in there all workarounds for all types of devices - that won't work out.)

Also we should not keep this untouched, since the special it's not documented at all.

> We're discussing a valid boundaries check

It is not, as @ebaauw already described. 1 Bit is used for the sign, 15 Bits remaining for the magnitude: `0111 1111 1111 1111 = 32,767`

According to the specification the value always must also be an `s16`: \
[ZigBee Cluster Library](https://zigbeealliance.org/wp-content/uploads/2021/10/07-5123-08-Zigbee-Cluster-Library.pdf#page=341)

---

I checked all the devices that were changed by this PR and their commits to see if there was any information about special treatment required. If there really was a problem, it was unfortunately not documented. But you may check yourself:

- devices/Jasco-GE/GE45856_wall_switch.json: [Device Request](https://forum.phoscon.de/t/ge-alias-jasco-45856-wall-switch/2650) | [Pull Request](https://github.com/dresden-elektronik/deconz-rest-plugin/pull/6477)
- devices/eva/powermeter.json: [Device Request](https://github.com/dresden-elektronik/deconz-rest-plugin/issues/6343) | [Pull Request](https://github.com/dresden-elektronik/deconz-rest-plugin/pull/7034)
- devices/ouellet/OTH4000-ZB.json:  ~Device Request~ | [Pull Request](https://github.com/dresden-elektronik/deconz-rest-plugin/pull/6431)
- devices/sinope/th1124zb.json: [Device Request](https://github.com/dresden-elektronik/deconz-rest-plugin/issues/2104) |  [Commit](https://github.com/bluemoehre/deconz-rest-plugin/commit/3c67a35f825190d9c3cafd8730f3944d9095c4f4)

That's all I can do at this point to convince everyone that we are doing the best possible here. \
Please let me know how we continue with this PR, because it's affecting the device request #7959.

mattreim:

PR #7591, opened 2024-02-10
DDF for IKEA bulb E14 WS 470lm - intervals extended

mattreim:
IKEA bulb E14 WS 470lm:

tradfri_bulb_e14_ws_470lm

PR #8009, opened 2024-11-01
TRADFRI bulb E27 WS opal 1000lm -  intervals extended

mattreim:
Preliminary conversation #7591


devices/ikea/tradfri_bulb_e27_ws_opal_1000lm.json
```diff
@@ -115,8 +115,8 @@
         {
           "at": "0x0000",
           "dt": "0x10",
-          "min": 1,
```
bluemoehre, line 118: Having a minimum delay of 5s for reporting on/off (and any other state), seems bad to me. I would go with max `1s`, since only with `0s` you will get notified about each on/off event. I cannot imagine clicking a button in some interface and then wait 5s for it to turn off.

Actually I would like to ask if we need bindings at all.

PR #8017, opened 2024-11-04
Fix IKEA Somrig product type code

bluemoehre:
See official [manual (page 5)](https://www.ikea.com/de/de/manuals/somrig-shortcut-button-weiss-smart__AA-2411208-2-2.pdf#page=5)

### DIFF
--- a/devices/ikea/somrig_shortcut_button.json
+++ b/devices/ikea/somrig_shortcut_button.json
@@ -4,7 +4,7 @@
   "manufacturername": "IKEA of Sweden",
   "modelid": "SOMRIG shortcut button",
   "vendor": "IKEA",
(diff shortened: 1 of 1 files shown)


ebaauw:
Are you sure that's the only product type used for this device?

I know of IKEA products that use multiple different product types, e.g. the five-button TRADFRI remote controls uses both E1524 and E1810.

bluemoehre:
Since the code was introduced with the same [commit for the RODRET (E2201)](https://github.com/dresden-elektronik/deconz-rest-plugin/blob/4573cf7a428942bca74f0a54bf02bc9893dc502b/devices/ikea/rodret_dimmer.json#L7) I believe it's a copy paste error.

PR #8019, opened 2024-11-04
Fix DDF JSON format

bluemoehre:
Currently there are multiple code format issues in some DDF JSON files. 
This causes lots of linter warnings, when dealing with the project. Also it may lead to unnecessary complex reviews when fixing format issues on the fly while changing code.

So I fixed:

- line indentations
- convert tabs to spaces
- removed trailing spaces
- property assignment spaces

I did NOT fix:

- general code format
- single line properties
- order of props

… because this will cause line shifts and make the diff very hard to review. (I can place another PR when this one was merged.)

> [!NOTE] 
> No code contents were harmed in this PR 🐶


bluemoehre:
Hey @manup , can you merge the cleanup by any chance on the fast lane, so we will have the same code format in all DDFs?

manup:
Wow that's a bunch, good work :+1: 
Guess it would be cool to incorporate the formatting as a PR merge step or GIT hook in future to stay consistent.

bluemoehre:
> Wow that's a bunch, good work 👍 Guess it would be cool to incorporate the formatting as a PR merge step or GIT hook in future to stay consistent.

Exactly - but first we have to agree on a setup. Unfortunately, I never received any feedback on this question in Discord. So let's stick with the standard format for now. Not perfect for all cases, but having everything the same is a step in the right direction.

PR #8021, opened 2024-11-05
apply standard JSON format to DDFs

bluemoehre:
This is a follow-up of #8019 and should be merged afterwards to simplify the review process.

This PR fixes:

- general code format
- enforcing single line properties

… by applying the Standard JSON Format.

This PR does not fix:

- order of props

… because there is no consistent pattern and applying the default alphanumeric order will lead to hard readable files.

> [!NOTE] 
> No code contents were harmed in this PR 🐶


bluemoehre:
@manup here we got with the real code formatting. 

Approx read time: 15min (but its just once in a lifetime - hopefully ;)

manup:
I was skeptical if by merging this the other large PR https://github.com/dresden-elektronik/deconz-rest-plugin/pull/7979 would get conflicts but a local merge test worked just fine.

bluemoehre:
I was also worried that this would cause a headache somewhere because the PR has already aged a bit. Refactoring PRs may get complicated over time. 
However, when merging such code format commits, I recommend using rebase if master is ahead - it makes it much easier to find conflicts and resolve them later. Fortunately there were none =)

PR #8001, opened 2024-10-28
DDF for Philips HUE Play Lightbar (440400982842)

bluemoehre:
Solves #7997

PR #8026, opened 2024-11-08
hotfix wrong manufacturer count vs model count

bluemoehre:
Master is broken since #7961 , caused by commit [9da501cb1e6b4b2b74248837f29212d7ac55d69e](https://github.com/dresden-elektronik/deconz-rest-plugin/commit/9da501cb1e6b4b2b74248837f29212d7ac55d69e), which was not passing / running DDF checks.

### DIFF
--- a/devices/philips/light_zb3_C.json
+++ b/devices/philips/light_zb3_C.json
@@ -31,6 +31,8 @@
     "$MF_PHILIPS",
     "$MF_PHILIPS",
     "$MF_PHILIPS",
+    "$MF_PHILIPS",
+    "$MF_SIGNIFY",
     "$MF_SIGNIFY",
     "$MF_SIGNIFY",
     "$MF_SIGNIFY",


ebaauw:
See also Discord, `#general` channel.

I think what happened is: Both #7961 and #8001 updated `light_zb3_C.json`, adding a new model to the file (twice each, for Philips and for Signify).  When the #7961 was merged, the merge logic saw that the added entries to the manufacturername array had already been added by #8001, and didn't include the added entries from #7961.

Individually, each PR passed validation, but combined they don't.  I'm not sure if validation of open PRs gets re-triggered after another PR is merged, or whether that's requires actively rebasing a PR before merging.  Maybe this could have been prevented by a more rigid merging process.

I would feel more comfortable changing the DDF structure into an array of `manufacturername`/`modelid` pairs, rather than two separate arrays.  That probably would still have caused the same issue, but would make it easier to detect visually.  On the longer term, I want the Hue light DDFs to match on _Image Type_ (`cap/otau/image_type`) and _Device Type_ (`type`) rather than on `modelid`, so we no longer need to whitelist each individual `modelid`.

bluemoehre:
> I think what happened is: Both https://github.com/dresden-elektronik/deconz-rest-plugin/pull/7961 and https://github.com/dresden-elektronik/deconz-rest-plugin/pull/8001 updated light_zb3_C.json, adding a new model to the file (twice each, for Philips and for Signify). When the https://github.com/dresden-elektronik/deconz-rest-plugin/pull/7961 was merged, the merge logic saw that the added entries to the manufacturername array had already been added by https://github.com/dresden-elektronik/deconz-rest-plugin/pull/8001, and didn't include the added entries from https://github.com/dresden-elektronik/deconz-rest-plugin/pull/7961.
> 
> Individually, each PR passed validation, but combined they don't.

Are you sure about that? When I check the single commit I am referring to, nothing has been added to the manufacturers property. Or does this commit show the status after the merge? Maybe it was a conflict where "one version" was accidentally chosen instead of both? \
I don't want to play any blame game here - these things can happen - but I really want to know if we need to worry about the order of merges, or if the tests should better check every commit on master branch. We shouldn't be able to break the master so easily without noticing.

> I would feel more comfortable changing the DDF structure into an array of manufacturername/modelid pairs [...]

I absolutely agree about changing that structure! Maybe we should start a feature branch and discuss some examples what matches best the current needs. (This will be also a perfect chance to setup a JSON schema file 🙂 )

ebaauw:
> Are you sure about that?

No.  That's why I said: "I think".  I do know for a fact that #7961 passed validation and added the two lines to the `manufacturername` array.  The commit introducing the issue is a merge commit, which is only created when merging the PR.  I suspect that #7961 wasn't rebased after #8001 was merged, as that would have triggered the validation.

> I really want to know if we need to worry about the order of merges

I suspect the exact same thing would have happened if the two PRs has been merged the other way round.  As I said on Discord, I've occasionally seen some erratic merging behaviour (lost updates) when two PR update the same file(s).

> if the tests should better check every commit on master branch

That would detect the issue but not prevent it.  Actually, as per your comment, it would seem the validation did run (and failed) after the merge commit.  
I think it might be better to make sure to rebase each open PR with the master branch (merge the master branch into the PR fork) and check the validation results right before merging the open PR.

bluemoehre:
> That would detect the issue but not prevent it.

To detect a broken master branch would be a good minimum requirement, because all other branches which forked or did a merge / rebase in the meanwhile suddenly failed the tests. Maybe it's possible to change the setup to test all single commits on master? This way we could see where it accidentally broke.

> I think it might be better to make sure to rebase each open PR

If I remember correctly, this can be achieved by the project rules in Github.

PR #8036, opened 2024-11-21
Update ZY-M100_human_breathing_presence.json

gituser-rk:
Added TS0225  (ZG-205Z/A) from manufacturer _TZE200_2aaelwxk. Still several datapoints missing, like light level ("illuminance_raw").


manup:
@Smanar do you think this is the right DDF aka a clone, note the different modelid?

manup:
Thanks for clarification,

>But as there is different dpid, better to create a new DDF
>Edit:
>The DDF _TZE200_kb5noeto_motion_sensor.json have more commun DPID I think.

Due the different data points, my gut feeling is also to make a separate DDF since this isn't really a clone. Further "MmwRadar Human Presence Detector (TS0601)" in this DDF doesn't sound like Human breathing sensor nor motion detector.

PR #8050, opened 2024-12-01
DDF for Tuya soil sensor (_TZE284_aao3yzhs)

m4k2k:
Based on https://github.com/dresden-elektronik/deconz-rest-plugin/issues/7902

Add support for additional tuya soil sensor.
Similar to https://github.com/dresden-elektronik/deconz-rest-plugin/blob/master/devices/tuya/_TZE200_myd45weu_soil_sensor.json but needs different temperature handling, therefore additional DDF created.

Other references: https://github.com/Koenkk/zigbee2mqtt/issues/23455


m4k2k:
In deconz/phoscon it shows up like this:
image


In HA it shows up like this:
image

PR #8055, opened 2024-12-06
DDF for Signify LCX015 (Philips Hue Festavia)

eglia:
I just bought a new set of philips hue festavia lights. They reported a modelid of lcx015 instead of lcx012 which seems to be expected by deconz.
After adding lcx015 to the file, effects like "candle" start working.

### DIFF
--- a/devices/philips/light_zb3_C_festavia.json
+++ b/devices/philips/light_zb3_C_festavia.json
@@ -3,11 +3,13 @@
   "uuid": "e7de8f0e-bf48-4ea0-b655-7e04fb6bdfa9",
   "manufacturername": [
     "$MF_SIGNIFY",
+    "$MF_SIGNIFY",
     "$MF_PHILIPS",
-    "$MF_SIGNIFY"
+    "$MF_PHILIPS"
   ],
   "modelid": [
     "LCX012",
+    "LCX015",
     "LCX012",
     "LCX015"
   ],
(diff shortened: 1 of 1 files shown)


ebaauw (changes requested):
Now you've broken support for the LCX012.  Please list both models twice, once for Philips and once for Signify - see the other Hue DDFs.

Is the _Image Type_ of the LCX015 the same as for the LCX012?  0x0123?  What firmware version is it on?  1.222.2?

Could you please list a screenshot of the _Hue Effects_ cluster (after _Read_ of the attributes)?

eglia:
My device showed Signify, so I wasn't sure if I need to add both. I fixed this.

Image Type is 0x0123:
Screenshot_20241206_123155
SW Build ID is 1.107.1:
Screenshot_20241206_123433
Note that I did not add it to the hue bridge/app, I directly connected it to deconz, so this may not be the latest firmware.

Hue Effects:
Screenshot_20241206_122809

ebaauw:
Thanks.

> My device showed Signify, so I wasn't sure if I need to add both.

To be honest, I'm not sure either, but we've seen some recent devices still advertising "Philips".  Also I suspect, on rare occasions, deCONZ legacy code might still initialise the manufacturer name to "Philips" based on the manufacturer code in the node descriptor.

> this may not be the latest firmware.

Indeed, see https://firmware.meethue.com/v1/checkUpdate?version=16779264&deviceTypeId=100b-123.  If you download the linked file to `~/otau`, deCONZ should update the firmware.  The latest firmware offers some more special effects.

ebaauw (approved):

Issue #7553, opened 2024-01-28
Bosch Room Thermostat II 230V (RBSH-RTH0-ZB-EU)


aSauerwein:
### Is there already an existing issue for this?

- [X] I have searched the existing issues and there is none for my device

### Product name

Boosch Room Thermostat II 230V (RBSH-RTH0-ZB-EU)

### Manufacturer

Bosch

### Model identifier

RBSH-RTH0-ZB-EU

### Device type to add

Thermostat

### Node info

<img width="361" alt="image" src="https://github.com/dresden-elektronik/deconz-rest-plugin/assets/5446813/e42956cb-a9db-45c8-a42a-914c923b044e">

### Endpoints and clusters

<img width="267" alt="image" src="https://github.com/dresden-elektronik/deconz-rest-plugin/assets/5446813/0efd1c7a-2200-463f-bf42-54339ea596f4">

### Basic

<img width="361" alt="image" src="https://github.com/dresden-elektronik/deconz-rest-plugin/assets/5446813/c69ed6ac-578f-4f26-a915-f2fe93489893">

### Power Configuration
![image](https://github.com/dresden-elektronik/deconz-rest-plugin/assets/5446813/99604fb4-71f9-458e-9870-e775e4338e9d)

### Thermostat
![image](https://github.com/dresden-elektronik/deconz-rest-plugin/assets/5446813/a4fd42e2-fabd-4591-bf93-2beff86a3327)
![image](https://github.com/dresden-elektronik/deconz-rest-plugin/assets/5446813/bf012ebc-4415-46c1-8fe8-ae6b0a41e37c)
![image](https://github.com/dresden-elektronik/deconz-rest-plugin/assets/5446813/9388a27c-66d6-4675-af04-feab6d41f75d)
![image](https://github.com/dresden-elektronik/deconz-rest-plugin/assets/5446813/4270ecb5-eb2f-44ef-ae11-a31d63cdd612)
![image](https://github.com/dresden-elektronik/deconz-rest-plugin/assets/5446813/8d0cdec4-c871-44f7-8709-f40b1fac7baf)
![image](https://github.com/dresden-elektronik/deconz-rest-plugin/assets/5446813/878b8130-bd85-494b-8f7a-0f5da5a24e1a)
![image](https://github.com/dresden-elektronik/deconz-rest-plugin/assets/5446813/6e2e4eb1-6316-499c-a6d2-9680ce817e8a)
[...]

PR #8084, opened 2025-01-10
Fix vendor, state and add cooling mode to Bosch room thermostat II 230V (BTH-RM230Z)

aSauerwein:
Hi!

I've added the `state/on` and fixed the changed the Vendor to `Bosch`. 

with the `state/on` the devices correctly shows heating in Homeassistant when it's actually heating. 

also I realized that my thermostat were running in draft mode, because the vendor was wrong in the ddf file.


mattreim:
Hello @aSauerwein, the manufacturer name is really written in lower case but according to your [description](https://github.com/dresden-elektronik/deconz-rest-plugin/issues/7553) the device does not have an on/off cluster (0006). Maybe @SwoopX knows another way.

aSauerwein:
I don't know how, but somehow it works. 

I think this 0x0029 in cluster 0201 is responsible for the heating state. 
heating "on" :
image
phoscon rest api output - `http://192.168.200.6/api/{{token}}/devices/00:1e:5e:09:02:9d:ba:cd`
```json
...
            "state": {
                "errorcode": {
                    "lastupdated": "2024-12-08T22:17:12Z",
                    "value": "0"
                },
                "on": {
                    "lastupdated": "2025-01-11T08:47:02Z",
                    "value": true
                },
...
```
not heating "off":
image
phoscon rest api output - `http://192.168.200.6/api/{{token}}/devices/00:1e:5e:09:02:9d:ba:cd`
```json
...
            "state": {
                "errorcode": {
                    "lastupdated": "2024-12-08T22:17:12Z",
                    "value": "0"
                },
                "on": {
                    "lastupdated": "2025-01-11T08:48:10Z",
                    "value": false
                },
...
```

I know you are right. I cannot see any 0x0006 cluster. cannot explain why it still works 🤷 
image

SwoopX:
Hi together,

that this works, is more or less a coincidence due to the fact that we've bent a lot back in the days to make things appear to work. So this is legacy heritage which will cease to function once the older code gets more and more removed over time.

To have it sustainable, the parser function for the attribute needs to be properly defined to fit the attribute at hand. So, it would be probably like this (also considering we might want to read the value to not solely rely on reporting):

```
{
  "name": "state/on",
  "refresh.interval": 3660,
  "parse": {
    "at": "0x0029",
    "cl": "0x0201",
    "ep": 1,
    "eval": "Item.val = Attr.val;",
    "fn": "zcl:attr"
  },
  "read": {
    "at": "0x0029",
    "cl": "0x0201",
    "ep": 1,
    "fn": "zcl:attr"
  }
},
```

mattreim:
Hello @aSauerwein, here is the adapted DDF for the `BTH-RM230Z`.

Changes:

-  Add `config/coolsetpoint`
-  Add `config/preset`
-  Add `state/on`
-  Change `vendor`
-  Change `config/mode`

[room_thermostat2_230V.json](https://github.com/user-attachments/files/19078111/room_thermostat2_230V.json)

aSauerwein:
thanks @mattreim !
I implemented your changes and updated the branch

mattreim:
I think the title needs to be adjusted.

manup:
Hi are the changes ok now so this PR can be merged?

mattreim:
Both devices are now the same again (`BTH-RM`/`BTH-RM230Z`) and it should work.

PR #8156, opened 2025-03-31
Removal of `config/temperature` from more Xiaomi devices (part 2)

mattreim:
This PR is a continuation of #7827 and now `config/temperature` should be removed from all Xiaomi devices.


mattreim:
With this PR this item (`/generic/items/config_temperature_item.json`) can actually be deleted too.

PR #8162, opened 2025-04-07
DDF for Aqara LED Strip T1 (LGYCDD01LM)

Sumd84:
I tried to continue the work of @ebaauw (https://github.com/dresden-elektronik/deconz-rest-plugin/pull/7201) on the Aqara LED Strip T1 with the implementation of :
- Control of music sync effect and sensitivity
- Control of preset effect and preset speed

I also propose to change the gradient/pixel_count to gradient/segments to not confuse with pixel_count of the Hue product.


devices/generic/items/config_color_gradient_segments_item.json
```diff
@@ -0,0 +1,8 @@
+{
+  "schema": "resourceitem1.schema.json",
+  "id": "config/color/gradient/segments",
```
ebaauw, line 3: Hue doesn't have `config/pixel_count`, but it is used for the Aqara LED Strip T1 _and_ Ceiling Light T1M.  If you have moving reasons to rename it, please deprecate the old item, and also change the ceiling light.

devices/generic/items/config_color_gradient_segments_item.json
```diff
@@ -0,0 +1,8 @@
...
+  "datatype": "UInt8",
+  "access": "RW",
+  "public": true,
+  "description": "Number of 20cm segments on Aqara LED Strip T1."
```
ebaauw, line 7: The 20cm is reported through `cap/color/gradient/pixel_length` and shouldn't be part of the description.  Also, this would lead to a new item for each new device.

devices/generic/items/state_music_sync_effect_item.json
```diff
@@ -0,0 +1,8 @@
+{
+  "schema": "resourceitem1.schema.json",
+  "id": "state/music_sync_effect",
```
ebaauw, line 3: I would prefer to group the `state/music_sync` attributes, so this would be renamed to `state/music_sync/effect`.  We would also need to change `state/music_sync` to `state/music_sync/on`.

I think the values should be reported as strings in the API, and we need a `cap` item to report the possible values.

devices/generic/items/state_music_sync_sensitivity_item.json
```diff
@@ -0,0 +1,8 @@
+{
+  "schema": "resourceitem1.schema.json",
+  "id": "state/music_sync_sensitivity",
```
ebaauw, line 3: Same remarks: `state/music_sync/sensitivity`, report the values as string, and provide a `cap` item with the possible values.

devices/generic/items/state_preset_effect_item.json
```diff
@@ -0,0 +1,8 @@
+{
+  "schema": "resourceitem1.schema.json",
+  "id": "state/preset_effect",
```
ebaauw, line 3: Please re-use `state/effect` and `cap/color/effects`.  Also, "preset" sounds like a `config` item, not a `state` item.

devices/generic/items/state_preset_speed_item.json
```diff
@@ -0,0 +1,8 @@
+{
+  "schema": "resourceitem1.schema.json",
+  "id": "state/preset_speed",
```
ebaauw, line 3: Please re-use `state/effectSpeed` from the Xmas lightstrip (which should probably be `state/effect_speed`).

rest_lights.cpp
```diff
@@ -1619,6 +1675,58 @@ int DeRestPluginPrivate::setLightState(const ApiRequest &req, ApiResponse &rsp)
         rsp.list.append(rspItem);
     }
 
+    if (hasMusicSyncEffect)
+    {
+        change.addTargetValue(RStateMusicSyncEffect, targetMusicSyncEffect);
```
ebaauw, line 1680: Why is this implemented in C++?  DDFs are perfectly capable of writing an attribute or sending a command based on a single item.  C++ code would be needed only when multiple items need to be combined into a single attribute or command (like setting `xy`).

rest_lights.cpp
```diff
@@ -1619,6 +1675,58 @@ int DeRestPluginPrivate::setLightState(const ApiRequest &req, ApiResponse &rsp)
...
+        QVariantMap rspItem;
+        QVariantMap rspItemState;
+        rspItemState[QString("/lights/%1/state/music_sync_effect").arg(id)] = targetMusicSyncEffect;
+        rspItem["success"] = rspItemState;
+        rsp.list.append(rspItem);
+    }
+
+    if (hasMusicSyncSensitivity)
+    {
+        change.addTargetValue(RStateMusicSyncSensitivity, targetMusicSyncSensitivity);
```
ebaauw, line 1693: Same remark

rest_lights.cpp
```diff
@@ -1619,6 +1675,58 @@ int DeRestPluginPrivate::setLightState(const ApiRequest &req, ApiResponse &rsp)
...
+        QVariantMap rspItem;
+        QVariantMap rspItemState;
+        rspItemState[QString("/lights/%1/state/music_sync_sensitivity").arg(id)] = targetMusicSyncSensitivity;
+        rspItem["success"] = rspItemState;
+        rsp.list.append(rspItem);
+    }
+
+    if (hasPresetEffect)
+    {
+        change.addTargetValue(RStatePresetEffect, targetPresetEffect);
```
ebaauw, line 1706: Same remark.

rest_lights.cpp
```diff
@@ -1619,6 +1675,58 @@ int DeRestPluginPrivate::setLightState(const ApiRequest &req, ApiResponse &rsp)
...
+        QVariantMap rspItem;
+        QVariantMap rspItemState;
+        rspItemState[QString("/lights/%1/state/preset_effect").arg(id)] = targetPresetEffect;
+        rspItem["success"] = rspItemState;
+        rsp.list.append(rspItem);
+    }
+
+    if (hasPresetSpeed)
+    {
+        change.addTargetValue(RStatePresetSpeed, targetPresetSpeed);
```
ebaauw, line 1719: Same remark.

ebaauw (changes requested):
Please re-use existing items rather than inventing new ones.

Please don't use C++ code for features that can be handled by DDFs.

rest_lights.cpp
Sumd84: I thought that both DDFs and C++ need to be modified to open new attributes on the rest API. If DDF alone is sufficient, I'll revert the C++ modification.

devices/generic/items/config_color_gradient_segments_item.json
Sumd84: I will switch back to your previous implementation of `config/color/gradient/pixel_count` to maintain compatibility.

rest_lights.cpp
Sumd84: I believe I may have missed something in the DDF, but it’s clear that without modifying the C++ code, I cannot expose or call the new attributes.
With my compiled plugin version, I can now access `/state/preset_effect` through the API.
If there is a way to access these attributes via the API without modifying the C++ code, I would appreciate a clear explanation.

PR #8199, opened 2025-04-30
Expose NWK address as hex value through REST API

SwoopX:
Currently, the NWK address is exposed as decimal value. Exposing it as hex value may aid in expediting debugging, as some log line only provide the NWK address.


ebaauw:
This will work, but would it be cleaner to implement this logic only once, in `ResourceItem::toVariant()`?  Ideally, I’d create a subclass for `RAttrNwkAddress`, but that seems too big a change.

SwoopX:
> This will work, but would it be cleaner to implement this logic only once, in `ResourceItem::toVariant()`? Ideally, I’d create a subclass for `RAttrNwkAddress`, but that seems too big a change.

@ebaauw Thanks Erik for the great suggestion. I always tend to overlook those functions, dunno why :) If @manup agrees, I'm more than happy to move into the `toVariant()` function.

ebaauw:
> I'm against bringing item specifics into toVariant() or even ResourceItem this method should know nothing about specific items.

We currently have quite some resource item descriptor specific logic in `ResourceItem::toString()`, which is called by `ResourceItem::toVariant()`.  To me this is cleaner (less dirty) than having a helper function outside the `ResourceItem` class.

> In future this can be specified genericly in the item JSON descriptions

I like that, but that's a significant refactor, probably beyond the scope of this PR.

SwoopX:
Let's start small and then let it grow as necessary :wink: 

Made the suggested changes hoping they're ok.

resource.cpp
```diff
@@ -1957,3 +1957,18 @@ bool isValidRConfigGroup(const QString &str)
 
     return result == groupList.size();
 }
+
+/*! Helper function to expose \p resource item as a different type through REST API than internally defined. */
+QVariant R_ItemToRestApiVariant(const ResourceItem *item)
+{
+    if (item)
+    {
+        const ResourceItemDescriptor rid = item->descriptor();
```
manup, line 1966: Better use a reference here, otherwise the descriptor is copied each time.

`const ResourceItemDescriptor &rid = item->descriptor();`

resource.cpp
```diff
@@ -1957,3 +1957,18 @@ bool isValidRConfigGroup(const QString &str)
...
+        if (rid.suffix == RAttrNwkAddress)
+        {
+            return QString("0x") + QString("%1").arg(item->toNumber(), 4, 16, QLatin1Char('0')).toUpper();
```
manup, line 1969: I suggest a small change to save two QString constructions :)

```
return QString("0x%1").arg(item->toNumber(), 4, 16, QLatin1Char('0')).toUpper();
```

manup (commented):
Nice almost there, I have two small change requests.

resource.cpp
SwoopX: Ah right, updated :+1:

resource.cpp
SwoopX: I'm afraid, I did that on purpose, as the suggested amendment does not keep the small letter X. See the output of https://github.com/dresden-elektronik/deconz-rest-plugin/blob/04bd307a62abf2ef84235b2069a666c03c6d401f/de_web_plugin.cpp#L4382-L4383 in the respective logs.

manup (approved):

PR #8223, opened 2025-05-06
Fix strings without data being null, return empty string instead

manup:
This fixes the remaining cases where strings could be `null` in my setup only `swversion` and `productid` where affected, since the others are handled by previous PRs.


brientim:
All looks good expect the following two:
Config "replacesbridgeid": null, and 
"buttonevent":

PR #8249, opened 2025-06-13
DDF clone for Tuya 4 Gang remote (_TZ3000_5tqxpine)

mattreim:
"manufacturername": "_TZ3000_5tqxpine"
"modelid": "TS0044"

See #8247


manup:
@mattreim @Smanar 

PR is alright.

Just some considerations to enhance the DDF in future.

Since a few versions DDFs support "buttons" and "buttonevents" to describe these (similar to button_maps.json).
Good example is the Hue remote https://github.com/dresden-elektronik/deconz-rest-plugin/blob/master/devices/philips/rwl022_dimmer_switch.json

With that the buttons and possible events are automatically exposed in the introspection API, which can be used by REST-API clients to be more generic, e.g. UIs showing the remote with proper amount of buttons.

The Hue remote DDF also uses a extra Javascript file to parse `state/buttonevent` which entirely removes dependency on `button_maps.json` making the DDF self contained, as generated DDF bundles do include all item files and referenced `.js` files. 

-----

If we do the same for this DDF, in order to add a new Tuya clone (manufacturername / modelid) only the main DDF `.json`  needs to be extended.

PR #8264, opened 2025-07-01
Deprecate product_match.cpp

SilverFire:
I've tried to add an alias for a device clone and edited `product_match.cpp` because I found similar vendor IDs in it.

@Smanar [pointed out](https://github.com/dresden-elektronik/deconz-rest-plugin/pull/8261#issuecomment-3024667211), that this file is no longer in use.

A deprecation notice may be useful if this file is indeed deprecated.

### DIFF
--- a/product_match.cpp
+++ b/product_match.cpp
@@ -8,6 +8,17 @@
  *
  */
 
+//
+// !!!!!! DEPRECATED !!!!!!
+//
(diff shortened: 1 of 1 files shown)


manup:
Good point 👍 yes indeed this file will be removed in future in favor of DDFs.

PR #8279, opened 2025-07-24
Fix for new IKEA RODRET model identifier

cs42:
New IKEA RODRET models have a different model identifier. Added that to the DDF of the RODRET dimmer and to the button map.

The image in phoscon has to be adjusted accordingly.

### DIFF
--- a/button_maps.json
+++ b/button_maps.json
@@ -236,7 +236,7 @@
         "ikeaRodretMap": {
             "vendor": "IKEA",
             "doc": "Rodret dimmer remote",
-            "modelids": ["RODRET Dimmer"],
+            "modelids": ["RODRET Dimmer", "RODRET wireless dimmer"],
             "buttons": [
                 {"S_BUTTON_1": "Dim Up"},
                 {"S_BUTTON_2": "Dim Down"}
--- a/devices/ikea/rodret_dimmer.json
+++ b/devices/ikea/rodret_dimmer.json
@@ -1,8 +1,14 @@
 {
   "schema": "devcap1.schema.json",
   "uuid": "3d671c73-06da-4adc-99f1-076b3dfb3122",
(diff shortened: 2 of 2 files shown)


ikorbln:
I change the files by hand, but i cant get the switch to be paired. It shows up as a generic icon with the name "rodret dimmer switch" but i dont get a successfully paired message at the end of the process.

cs42:
> I change the files by hand, but i cant get the switch to be paired. It shows up as a generic icon with the name "rodret dimmer switch" but i dont get a successfully paired message at the end of the process.

Just a "dumb" question: you have restarted deCONZ? I don't know, if those files will be reloaded on changes automatically. It sounds like the entries in `button_maps.json` are missing (it's located in `/usr/share/deCONZ/devices/button_maps.json` in an installation on Linux).

The button will show up as a generic 2 button device in phoscon and not with the RODRET image, as the older version does.

ikorbln:
> Just a "dumb" question: you have restarted deCONZ? I don't know, if those files will be reloaded on changes automatically. It sounds like the entries in `button_maps.json` are missing (it's located in `/usr/share/deCONZ/devices/button_maps.json` in an installation on Linux).
> 
> The button will show up as a generic 2 button device in phoscon and not with the RODRET image, as the older version does.

I have done a reboot, the changes are made in the file. This is how it looks in Phoscon:
image
Pressing a button shows no reaktion. Other switches blink short when pressing. 
btw. in Homebridge, the switch shows up as a 1 Button switch with three states "one-press / two-presses / long press"

EDIT: OK, i just repaired the switch again without deleting it, and after a second reboot of deconz and Homebridge it seems to work now.

chkronenberg:
Thanks for your work. Unfortunately, even with the new JSON files or the beta version, I could not pair my newly purchased Rodret Wireless Dimmer with deconz.

PR #8302, opened 2025-08-06
feat: support for Innr SP 242 plug

justin-caldicott:
Based on DDF for SP 240.

As purchased in the UK. Verified working, though yet to see non-zero consumption, but that may be due to not enough consumption.

### DIFF
--- a/devices/innr/sp_242.json
+++ b/devices/innr/sp_242.json
@@ -1,0 +1,249 @@
+{
+  "schema": "devcap1.schema.json",
+  "uuid": "699fff8f-955c-4468-be69-c92b04a79a23",
+  "manufacturername": "innr",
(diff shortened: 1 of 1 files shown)


mattreim:
@justin-caldicott If it is identical to SP 240, you should only extend the `sp_240.json` like this:

```
  "uuid": "4b16f6c7-b57e-4d7a-9eff-894c5023ab9e",
  "manufacturername": [
    "innr",
    "innr"
  ],
  "modelid": [
    "SP 240",
    "SP 242"
  ],
  "vendor": "Innr",
  "product": "Smart plug",
  "sleeper": false,
  "status": "Gold",
```

justin-caldicott:
Many thanks @mattreim, didn't know about that, updated.

PR #8325, opened 2025-09-09
Add additional manufacturer and model IDs

Mimiix:
Seems to be a clone.


Mimiix:
seems to work perfectly!

PR #8355, opened 2025-10-13
DDF for IKEA Tradfri bulb gu10 ww 345lm

dimitripb:
DDF for IKEA Tradfri gu 10 bulb 345 lm. Default the bulb only supports on/off. This DDF adds dimming.

PR #8356, opened 2025-10-13
IKEA devices - some adjustments and fixes

mattreim:
@dimitripb gave me the idea to make some adjustments for all IKEA bulbs #8355.
- Add/change some `product` names
- Add some `fimware` bindings
- Add some `cap/bri/move_with_onoff` and `cap/on/off_with_effect`
- Add `attr/productid`
- Add clone Tradfri bulb gu10 ww 345lm

### DIFF
--- a/devices/ikea/tradfri_bulb_e27_w_opal_1000lm.json
+++ b/devices/ikea/tradfri_bulb_e27_w_opal_1000lm.json
@@ -4,7 +4,7 @@
   "manufacturername": "$MF_IKEA",
   "modelid": "TRADFRI bulb E27 W opal 1000lm",
   "vendor": "IKEA",
-  "product": "Tradfri LED bulb E27 1000 lumen, dimmable warm white, opal white (LED1623G12)",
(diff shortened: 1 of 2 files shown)


mattreim:
Hi @manup , as a next step we could merge some IKEA DDFs.

```lua
  "manufacturername": [
    "$MF_IKEA",
    "$MF_IKEA"
  ],
  "modelid": [
    "TRADFRI bulb E14 WS 470lm",
    "TRADFRIbulbE14WScandleopal470lm"
  ],
  "vendor": "IKEA",
  "product": "Tradfri LED bulb E14 470 lumen, dimmable white spectrum, chandelier opal (LED1835C6/LED1949C5)",
```
For example, these very similar devices each have their own DDF. I actually think they're the same devices, just from different years of manufacture. I found three that would be suitable.

manup:
Hi thanks for the great work on these, it's good to also get more concise text in the vendor/product keys. My collgue rewrote our former manually written compatibility list to now be feeded directly from the DDFs: <https://phoscon.de/compatible> it will get more details from the DDF in future.

For the merging of DDFs, as long as they are identical beside the modelid it's fine to merge them. Ideally if they are also available for testing.

PR #8359, opened 2025-10-15
Support WebSockets over HTTPS and HTTP port

manup:
Together with changes in deCONZ core this allows accepting WebSocket connections directly on the REST-API ports without needing an extra WebSocket server port. Both `ws://` and `wss://` are supported.

- Port 443 isn't any longer tried as default WebSocket server port.
- The extra WebSocket server port is only opened if specified via `--ws-port` and only supports unsecured `ws://` protocol.
- The `/config` object contains an extra key `websocketport_wss` telling the HTTPS port over which `wss://` WebSocket connections can be made, which is also the REST-API HTTPS port.

**Note:** The internal HTTPS entry port is using port 443 which can be reconfigured via `--https-port=<port>`.

### DIFF
--- a/de_web_plugin.cpp
+++ b/de_web_plugin.cpp
@@ -15684,6 +15684,10 @@ bool DeRestPlugin::isHttpTarget(const QHttpRequestHeader &hdr)
             return true;
         }
     }
+    else if (hdr.hasKey(QLatin1String("Upgrade")) && hdr.value(QLatin1String("Upgrade")) == QLatin1String("websocket"))
+    {
+        return true;
+    }
 
     return false;
 }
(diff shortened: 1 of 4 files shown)


disnamic:
For me that's a breaking change in my API implementation, which uses an unencrypted ws port. 
I was using 443 but even when trying to change with --ws-port it won't open any other port for ws access.
I might be doing something wrong, but at least it should be mentioned more prominently in the releasenotes.

manup:
Hi can you please try if following command line arguments solve the issue `--https-port=0` and `--ws-port=443`.
This should disable https entirely and run like previous versions.

I'll do more testing on this, `--ws-port=<port>` should have worked too if it isn't another used port. Note the unencrypted WebSocket is also available over the normal REST-API port.

disnamic:
I've added --https-port=0 and --ws-port=443 to /lib/systemd/system/deconz.service and it doesn't seem to take.
websocketport in /config does also still reports 80 after the change and daemon-reload / restart deconz and is also missing the websocketport_wss key:

Screenshot 2025-10-29 225443

I can connect fine to port 80 with a websocket but can't keep it open as it then blocks http access.
I've upgraded straight from 2.31.2 and a downgrade restores functionality.

PR #8361, opened 2025-10-17
DDF for Eurotronic Spirit radiator thermostat (SPZB0001)

mattreim:
This PR is a follow-up to device request #8333 (Eurotronic SPZB0001), and I hope we can finish it. If @SwoopX, @Smanar, @manup helps, it could happen quickly ;)

[Spirit_ZigBee_BAL_web_DE_Okt.-2019.pdf](https://github.com/user-attachments/files/22978766/Spirit_ZigBee_BAL_web_DE_Okt.-2019.pdf)

Must be merged with #8377

### DIFF
--- a/devices/eurotronic/eurotronic_spzb0001_trv.json
+++ b/devices/eurotronic/eurotronic_spzb0001_trv.json
@@ -91,6 +91,15 @@
             "fn": "zcl:attr",
             "mf": "0x1037"
           },
+          "write": {
+            "at": "0x4008",
+            "cl": "0x0201",
+            "dt": "0x22",
+            "ep": 1,
+            "eval": "if (Item.val == true) { R.item('config/hostflags').val | 0x02 } else { R.item('config/hostflags').val & ~0x02};",
+            "fn": "zcl:attr",
+            "mf": "0x1037"
+          },
           "read": {
             "fn": "none"
           },
@@ -99,6 +108,14 @@
         {
           "name": "config/heatsetpoint",
           "refresh.interval": 660,
+          "parse": {
+            "at": "0x4003",
+            "cl": "0x0201",
+            "ep": 1,
+            "eval": "Item.val = Attr.val",
+            "fn": "zcl:attr",
+            "mf": "0x1037"
+          },
           "read": {
             "at": "0x4003",
             "cl": "0x0201",
@@ -113,14 +130,6 @@
             "ep": 1,
             "eval": "Item.val",
             "fn": "zcl:attr"
-          },
-          "parse": {
-            "at": "0x4003",
-            "cl": "0x0201",
-            "ep": 1,
-            "eval": "Item.val = Attr.val",
-            "fn": "zcl:attr",
-            "mf": "0x1037"
           },
           "default": 0
         },
@@ -156,6 +165,15 @@
             "fn": "zcl:attr",
             "mf": "0x1037"
           },
+          "write": {
+            "at": "0x4008",
+            "cl": "0x0201",
+            "dt": "0x22",
+            "ep": 1,
+            "eval": "if (Item.val == true) { R.item('config/hostflags').val | 0x80 } else { R.item('config/hostflags').val & ~0x80};",
+            "fn": "zcl:attr",
+            "mf": "0x1037"
+          },
           "read": {
             "fn": "none"
           },
@@ -172,6 +190,15 @@
             "fn": "zcl:attr",
             "mf": "0x1037"
           },
+          "write": {
+            "at": "0x4008",
+            "cl": "0x0201",
+            "dt": "0x22",
+            "ep": 1,
+            "eval": "if (Item.val == 'off') { R.item('config/hostflags').val & ~0x04 | 0x20 } else if (Item.val == 'heat') { R.item('config/hostflags').val & ~0x20 | 0x04 } else if (Item.val == 'auto') { R.item('config/hostflags').val & ~0x20 & ~0x04 };",
+            "fn": "zcl:attr",
+            "mf": "0x1037"
+          },
           "read": {
             "fn": "none"
           },
@@ -180,19 +207,19 @@
         {
           "name": "config/offset",
           "refresh.interval": 3660,
+          "parse": {
+            "at": "0x0010",
+            "cl": "0x0201",
+            "ep": 1,
+            "eval": "Item.val = Attr.val * 10;",
+            "fn": "zcl:attr"
+          },
           "write": {
             "at": "0x0010",
             "cl": "0x0201",
             "dt": "0x28",
             "ep": 1,
             "eval": "Item.val / 10;",
-            "fn": "zcl:attr"
-          },
-          "parse": {
-            "at": "0x0010",
-            "cl": "0x0201",
-            "ep": 1,
-            "eval": "Item.val = Attr.val * 10;",
             "fn": "zcl:attr"
           },
           "default": 0
@@ -243,12 +270,6 @@
         {
           "name": "state/temperature",
           "refresh.interval": 660,
-          "read": {
-            "at": "0x0000",
-            "cl": "0x0201",
-            "ep": 1,
-            "fn": "zcl:attr"
-          },
           "parse": {
             "at": "0x0000",
             "cl": "0x0201",
@@ -256,22 +277,28 @@
             "eval": "Item.val = Attr.val",
             "fn": "zcl:attr"
           },
+          "read": {
+            "at": "0x0000",
+            "cl": "0x0201",
+            "ep": 1,
+            "fn": "zcl:attr"
+          },
           "default": 0
         },
         {
           "name": "state/valve",
           "refresh.interval": 660,
-          "read": {
-            "at": "0x0008",
-            "cl": "0x0201",
-            "ep": 1,
-            "fn": "zcl:attr"
-          },
           "parse": {
             "at": "0x0008",
             "cl": "0x0201",
             "ep": 1,
             "eval": "Item.val = (Attr.val / 255) * 100;",
+            "fn": "zcl:attr"
+          },
+          "read": {
+            "at": "0x0008",
+            "cl": "0x0201",
+            "ep": 1,
             "fn": "zcl:attr"
           },
           "default": 0
@@ -283,7 +310,7 @@
             "at": "0x4008",
             "cl": "0x0201",
             "ep": 1,
-            "eval": "Item.val = (Attr.val >> 3) & 0x01;",
+            "eval": "Item.val = (R.item('config/hostflags') >> 3) & 0x01;",
             "fn": "zcl:attr",
             "mf": "0x1037"
           },


Smanar:
The typo error is corrected ?
I don't see why the validor is saying there is an error.

mattreim:
I don't understand the error either, but it's actually about the DDF.

Smanar:
Perhaps it's because the PR is on Draft status ? So no new check.

mattreim:
I've even deleted everything and the error still exists.

mattreim:
There are no `config/hostflags` yet, which I added. The information is from the PDF, but I don't think it's correct yet.

Smanar:
Ha yes, was the missing config_hostflags_item.json.

On the code this flag is hidden, but IDK if it's usefull.

`sensor.addItem(DataTypeUInt32, RConfigHostFlags)->setIsPublic(false);`

devices/eurotronic/eurotronic_spzb0001_trv.json
```diff
@@ -0,0 +1,373 @@
...
+            "fn": "zcl:attr"
+          },
+          "parse": {
+            "at": "0x0000",
+            "cl": "0x0006",
+            "ep": 1,
+            "eval": "Item.val = Attr.val",
+            "fn": "zcl:attr"
+          },
+          "default": false
```
SwoopX: Seems like the former content got lost. Here, it should be as follows (the attribute is read through `state/valve`):

```
          "name": "state/on",
          "parse": {
            "at": "0x0008",
            "cl": "0x0201",
            "ep": 1,
            "eval": "Item.val = Attr.val > 3;",
            "fn": "zcl:attr"
          },
          "read": {
            "fn": "none"
          }
```

devices/eurotronic/eurotronic_spzb0001_trv.json
```diff
@@ -0,0 +1,373 @@
...
+            "fn": "zcl:attr"
+          },
+          "default": 0
+        },
+        {
+          "name": "state/valve",
+          "refresh.interval": 660,
+          "read": {
+            "fn": "none"
+          },
```
SwoopX: Is there any particular reason not to read the attribute? I would've left the default read function here.

devices/eurotronic/eurotronic_spzb0001_trv.json
```diff
@@ -0,0 +1,373 @@
...
+          "parse": {
+            "at": "0x0012",
+            "cl": "0x0201",
+            "ep": 1,
+            "eval": "Item.val = Attr.val",
+            "fn": "zcl:attr"
+          },
+          "default": 0
+        },
```
SwoopX: This one here is very delicate, as the screwed firmware version does not allow for any differentiation which attribute must be **written**. For the Spirit, the legacy code uses mf attribute 0x4003, so I'd vouch for continuing using it in the _read_ and _parse_ functions.

As for the write function, I guess we can leave it as is, since it doesn't matter anyway. The legacy code has an ugly, but required workaround and sends a write request to any relevant attribute (so in total 2). As said, this must remain as you have no chance to know which attribute the hardware really accepts.

Please, this must be tested in strict mode!

devices/generic/items/config_hostflags_item.json
```diff
@@ -0,0 +1,8 @@
+{
+  "schema": "resourceitem1.schema.json",
+  "id": "config/hostflags",
+  "datatype": "Int32",
+  "access": "RW",
+  "public": true,
```
SwoopX: I guess the item should not be public.

devices/eurotronic/eurotronic_spzb0001_trv.json
```diff
@@ -0,0 +1,373 @@
...
+          "mf": "0x1037",
+          "min": 1,
+          "max": 3600
+        },
+        {
+          "at": "0x4003",
+          "dt": "0x29",
+          "mf": "0x1037",
+          "min": 1,
+          "max": 600
```
SwoopX: Legacy code has a reportable change of 50 (int value). Do we want to adapt that?

mattreim:
@SwoopX Thank you for your help and I hope I understood the changes correctly but please take another look.

SwoopX (changes requested):
Seems as if the force push screwed the proper tracking of the changes...

From what I can see, quite some write functions have been deleted, presumably accidentally? Also, the question regarding the reportable change is still pending and if the DDF has been tested in strict mode.

mattreim:
Now I'm completely confused, too. Probably something about ("read" => "fn": "none"), right? I haven't been able to test it yet because I'm using HA and am currently testing it on Windows. But I'd like to start with a somewhat finished DDF.

mattreim:
@SwoopX I just tested everything in strict mode, and the device works exactly as expected. The valve shows values ​​between 0 and 100, and the error code appears immediately, even if I remove the thermostat. The only thing I haven't been able to test is "windowopen", as it's apparently still too warm outside. So, for me, the DDF is now ready for use, but only in strict mode; otherwise, it doesn't work.
```lua
{
    "capabilities": {
        "sleeper": false
    },
    "config": {
        "battery": 60,
        "displayflipped": false,
        "heatsetpoint": 2100,
        "locked": false,
        "mode": "auto",
        "offset": 0,
        "on": true,
        "reachable": true
    },
    "ep": 1,
    "etag": "6088eb7dd4abf9d3f826e8ed75b7bdfe",
    "lastannounced": "2025-10-25T13:47:57Z",
    "lastseen": "2025-10-25T13:56Z",
    "manufacturername": "Eurotronic",
    "modelid": "SPZB0001",
    "name": "Thermostat - Diele",
    "nwkaddress": "0x9EB5",
    "state": {
        "errorcode": "none",
        "lastupdated": "2025-10-25T13:57:03.876",
        "on": false,
        "temperature": 2158,
        "valve": 0,
        "windowopen": "0"
    },
    "swversion": "18181120",
    "type": "ZHAThermostat",
    "uniqueid": "00:15:8d:00:07:74:34:93-01-0201"
}
```
Events:
```lua
    "16:15:42:46": {
        "e": "changed",
        "id": "36",
        "r": "sensors",
        "state": {
            "errorcode": "none",
            "lastupdated": "2025-10-25T14:15:42.045",
            "on": false,
            "temperature": 2087,
            "valve": 0,
            "windowopen": "0"
        },
        "t": "event",
        "uniqueid": "00:15:8d:00:07:74:34:93-01-0201"
    }
}
```

SwoopX:
Awesome, I guess this one should be ready to go then. However, I'd need to prepare another PR to have the legacy code removed completely, otherwise the DDF is not used.

One last remark though: the legacy code sets the bits for the hostflags a little different. No clue if that has any (noticable) impact or not.

mattreim:
Thanks for taking over the follow-up PR, and this old device is a bit slower than the new ones anyway, but that's not a big deal. So the legacy code is getting smaller and smaller. I'll now make it "Ready for review".

manup:
Nice work guys, much appreciated when the legacy code can be removed too later on 👍

SwoopX:
I fear it needs to go with the legacy code simultaneously. Will prepare the accompanying PR on the weekend at the latest.

manup:
@SwoopX thanks so this PR and https://github.com/dresden-elektronik/deconz-rest-plugin/pull/8377 are ready to be merged?

Unfortunantely my Eurotronic Spirit gave up years ago so I can't test it.

SwoopX:
@manup yes, should be ready to go. It was tested by @mattreim and another user earlier in strict mode, according to their feedback.

mcormick:
Just to let you know I had to downgrade to stable last friday after updating to latest on thursday because the house was cold... I think I noticed something with the polling rate, but not entirely sure. TRVs are running firmware 18181120.

mattreim:
Hi @SwoopX, unfortunately I've now noticed that the device is updating every second. Do you have any idea what could be causing this?

```
21:52:16:791 ZCL attribute report 0x00158D0007743493 for cluster: 0x0201, ep: 0x01, frame control: 0x1C, mfcode: 0x1037 
21:52:17:792 ZCL attribute report 0x00158D0007743493 for cluster: 0x0201, ep: 0x01, frame control: 0x1C, mfcode: 0x1037 
21:52:18:789 ZCL attribute report 0x00158D0007743493 for cluster: 0x0201, ep: 0x01, frame control: 0x1C, mfcode: 0x1037 
21:52:19:786 ZCL attribute report 0x00158D0007743493 for cluster: 0x0201, ep: 0x01, frame control: 0x1C, mfcode: 0x1037 
21:52:20:783 ZCL attribute report 0x00158D0007743493 for cluster: 0x0201, ep: 0x01, frame control: 0x1C, mfcode: 0x1037
```

Smanar:
All request that use the mfcode have a "refresh.interval" > 1 s.

And it's a report attribute, so something normal, increase bind/report value to test.

Enable APS + aps_l2 you will have more information about the request (attribute for exemple)

mattreim:
Thank you, I need to observe the behavior further and experiment with the values.

SwoopX:
@mattreim might be coincidence that data is coming that close. Or is that happening (way) more often? Anyway, as Smanar suggested, adding APS data should give more insights, as we currently do not see what attribute data is coming in and the respective values.

We've also forgotten to set the reportable change values for the attribute reporting back then, so the reporting is eventually based on legacy code.

mattreim:
Most messages appear in the first few hours after a restart, and then they become less frequent.
```
22:29:00:253 APS-DATA.request id: 100, addrmode: 0x03, addr: 0x00158D0007743493, profile: 0x0104, cluster: 0x0201, ep: 0x01 -> 0x01 queue: 1 len: 5 tx.options 0x00
22:29:00:276 APS-DATA.confirm id: 100, status: 0x00 SUCCESS

22:29:02:144 aps request id: 27 prf: 0x0000 cl: 0x0000 timeout (confirmed: 1) to 0x00158D00032816FC (0xFFFD)
22:29:02:224 aps request id: 27 finished, erase from queue

22:29:08:253 APS-DATA.request id: 165, addrmode: 0x03, addr: 0x00158D0007743493, profile: 0x0104, cluster: 0x0201, ep: 0x01 -> 0x01 queue: 1 len: 5 tx.options 0x00
22:29:08:254 	asdu (length: 5): 1036000800
```

Smanar:
>APS-DATA.request

Mean it's deconz that make request to the device, so not good if it' something happening every seconds

>asdu (length: 5): 1036 00 0800

It's a read attribute on attribute 0x0008

```
        {
          "name": "state/valve",
          "refresh.interval": 660,
          "parse": {
            "at": "0x0008",
```

The report is set to 1/600/1.

The device is not set as "sleeper", so if deconz haven't an anwser after 600s it start to spam the device to have the value, it need an answer or an error message.

mattreim:
I hope this [PR](https://github.com/dresden-elektronik/deconz-rest-plugin/pull/8464) will fix it, but I will continue to monitor it.

mattreim:
With the recent changes to the DDF and after several restarts, the number of messages has decreased significantly. Therefore, I think my problem is resolved.

PR #8367, opened 2025-10-20
IKEA bulbs - merged some DDFs

mattreim:
@manup Here is the suggestion for merging three IKEA DDFs. Unfortunately, 
```lua
  "manufacturername": [
    "$MF_IKEA",
    "$MF_IKEA"
  ],
  "modelid": [
    "TRADFRI bulb E14 WS 470lm",
    "TRADFRIbulbE14WScandleopal470lm"
  ],
  "vendor": "IKEA",
  "product": "Tradfri LED bulb E14 470 lumen, dimmable white spectrum, chandelier opal (LED1835C6/LED1949C5)",
```
For example, these very similar devices each have their own DDF. I actually think they're the same devices, just from different years of manufacture. I found three that would be suitable.


manup:
Cool thanks looks fine to me :+1:

mattreim:
The last commit is for #8356

Issue #7290, opened 2023-10-09
Namron Multisensor


le-clu:
### Is there already an existing issue for this?

- [X] I have searched the existing issues and there is none for my device

### Product name

Namron Zigbee Multisensor

### Manufacturer

Namron AS

### Model identifier

4512770

### Device type to add

Sensor

### Node info

![AF9E9D14-DC60-4B1C-A5B2-56836C926420_1_105_c](https://github.com/dresden-elektronik/deconz-rest-plugin/assets/97918159/f7ba6aab-6ecb-4449-994e-a6ea60b79680)

### Endpoints and clusters

![DE7604E2-2C0C-448C-99C9-0B757F24F1AE_4_5005_c](https://github.com/dresden-elektronik/deconz-rest-plugin/assets/97918159/7c851fbe-d1b7-4c20-9e88-07c68708b5e2)

### Basic

![3C5FFAAA-EC18-4CD4-A122-65ACD67EAEB9_1_105_c](https://github.com/dresden-elektronik/deconz-rest-plugin/assets/97918159/9512b61a-112e-4766-bb7b-27c9a9eac186)

![E54623ED-5F9A-4335-B092-317C82B24A70_1_105_c](https://github.com/dresden-elektronik/deconz-rest-plugin/assets/97918159/9e67f603-58c0-43be-9a6c-71a6468690aa)

![E7D2C1E1-DC91-463C-9753-C05D41ED9EB0_1_105_c](https://github.com/dresden-elektronik/deconz-rest-plugin/assets/97918159/91ab4850-1e6e-4d04-921e-fbe19e73dcae)

![AA3D8283-7958-4938-9559-CDB3D67EC186_1_105_c](https://github.com/dresden-elektronik/deconz-rest-plugin/assets/97918159/84488ae9-e562-4353-a8a4-101bb955a7c9)

![891F3CA8-36FF-417F-8A3C-3234106D0443_1_105_c](https://github.com/dresden-elektronik/deconz-rest-plugin/assets/97918159/89c2c087-c2b1-45db-a1c8-5a93460cdb08)

### Power Configuration
<img width="386" alt="image" src="https://github.com/dresden-elektronik/deconz-rest-plugin/assets/97918159/b0d15de2-7260-42d3-b643-3e9330651b96">

### IAS Zone
[...]

PR #8385, opened 2025-11-04
Fix for Namron 4512770 4in1 multi sensor

mattreim:
Product name: Namron Zigbee Multisensor
Manufacturer: Namron AS
Model identifier: 4512770

See #7290 and #8076

- I believe that "$TYPE_BATTERY_SENSOR" is not needed here.
- For "state/lightlevel" we needed a script.
- Add "config/delay" for presence sensor

@Smanar I'm not entirely sure about the EPs yet.

### DIFF
--- a/devices/namron/4512770-4in1-MultiSensor.json
+++ b/devices/namron/4512770-4in1-MultiSensor.json
@@ -66,6 +66,7 @@
         {
           "name": "config/battery",
           "refresh.interval": 86400,
+          "awake": true,
           "read": {
             "at": "0x0021",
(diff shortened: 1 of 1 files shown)


Smanar:
Lol, effectively, this DDF is realy better than the previous one.

About
```
        {
          "name": "attr/productname",
          "static": "4in1-MultiSensor"
        },
```

I never use them too, but I think some third app use this field.

mattreim:
@Smanar You are right, it has no effect on the function.

PR #8393, opened 2025-11-08
DDFs adjustments and fixes [1]

mattreim:
- Files are lowercase
- Product names adjusted
- Intervals adjusted
- formatting

@manup I think it's the first PR with such different devices, I hope that's okay. I will change them again if there are any comments.

### DIFF
--- a/devices/tuya/_TZE200_TS0601_raindrop_sensor.json
+++ b/devices/tuya/_TZE200_TS0601_raindrop_sensor.json
@@ -89,12 +89,10 @@
           "name": "state/lastupdated"
         },
         {
-          "name": "state/lowbattery",
-          "awake": true
+          "name": "state/lowbattery"
         },
         {
-          "name": "state/water",
-          "awake": true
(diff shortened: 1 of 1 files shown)


Smanar:
`"refresh.interval": 86400,`

This is useless for "attr/swversion"
Because deconz will use naturally this value by defaut https://github.com/dresden-elektronik/deconz-rest-plugin/blob/master/devices/generic/items/attr_swversion_item.json

manup:
Hi thanks for the PR, I think the changes are ok and the PR can be merged. But for similar changes in upcoming PRs it would be good to have them separated per DDF/device for easier review and tracking.

mattreim:
Thanks everyone, sometimes it's easier for users to see duplicate values ​​to understand them, and the resulting DDFs will be better. This is because most people use a template for their DDF and don't look at the individual items. But it's true, this pull request has become unexpectedly large with so many devices. The next ones will be smaller again.

PR #8396, opened 2025-11-11
Support relay interlock for Aqara Dual Relay Module T2 (DCM-K01)

Smanar:
See https://forum.phoscon.de/t/aqara-relay-t2-switch-type-ddf-modification/5685/8

Used on 
```
  "modelid": "lumi.switch.acn047",
  "vendor": "Xiaomi",
  "product": "Aqara Dual Relay Module T2 (DCM-K01)",
```

### DIFF
--- a/devices/generic/items/config_interlock_item.json
+++ b/devices/generic/items/config_interlock_item.json
@@ -1,0 +1,9 @@
+{
+  "schema": "resourceitem1.schema.json",
+  "id": "config/interlock",
+  "datatype": "Bool",
+  "access": "RW",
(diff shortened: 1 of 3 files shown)


SwoopX:
Erm, you are aware that this PR is kinda without any functionality?

manup:
Upps indeed, @Smanar  why do we need the new item if it isn't actually used?

Smanar:
There is an DDF update for the lumi.switch.acn04 that use it.
There is probably other devices that can use it too.

PR #8399, opened 2025-11-19
DDF for HOBEIAN sensors (ZG-204ZV and ZG-227Z)

derfrodo:
This pull request adds support for the HOBEIAN ZG-227Z temperature and humidity sensor and HOBEIAN ZG-204ZV 3 in 1 presence sensor by introducing new device definition files. 

Disclaimer: I am not experienced in reading devices specifications in detail / logging those and creating DDFs.
Therefore those files are created by example (you guys have already some awesome DDFs online already) and educated guesses.

Using the current stable docker file ([deconzcommunity/deconz:2.31.2](https://hub.docker.com/layers/deconzcommunity/deconz/2.31.2/images/sha256-307276bf67a8e3997a69527e67a39f402da4eda182224368806612678d032443)) let me use the sensors now sucessfully. I can not rule out, that in the definitions might be something off though. For complete transparency: The sensors might be some clones, since ordered from aliexpress:

[3in1 Presence Sensor](https://de.aliexpress.com/item/1005009268596716.html)
[Temperature sensor](https://de.aliexpress.com/item/1005008550134548.html)

Anyhow: Maybe somebody find those files beneficial either by finding those in main or even in this PR using the files as custom DDFs as I do right now. 😅

Cheers

### DIFF
--- a/devices/hobeian/hobeian_presence_temperature_humidity_lightlevel_sensor.json
+++ b/devices/hobeian/hobeian_presence_temperature_humidity_lightlevel_sensor.json
@@ -3,7 +3,7 @@
   "manufacturername": "HOBEIAN",
   "modelid": "ZG-204ZV",
   "vendor": "HOBEIAN",
-  "product": "3 In 1 presence sensor with temperature and humidity and light level (ZG-204Z)",
+  "product": "3-in-1 presence sensor with temperature, humidity and light sensor (ZG-204ZV)",
   "sleeper": true,
   "status": "Gold",
   "subdevices": [
@@ -329,7 +329,7 @@
           "name": "config/sensitivity",
           "range": [
             0,
-            2
+            19
           ],
           "refresh.interval": 86400,
           "read": {


mattreim:
Hi @derfrodo, good work, you've combined the DDFs well. However, since we're missing some information about these devices, you'll need to check if `config/delay` is working on the 3-in-1 sensor, as it could also be `config/sensitivity`.  I would definitely change the product name. Otherwise, it looks okay to me, but @Smanar should also take a look. Thank you.

`"product": "3-in-1 presence sensor with temperature, humidity and light sensor (ZG-204ZV)"`

derfrodo:
@mattreim Thanks a lot for your review!
In terms of config/delay and config/sensitivity you definetly have a point. Looking at the app via VNC seems to rather suggests, that I should add/change to sensitivity. [Zigbee2mqtt](https://www.zigbee2mqtt.io/devices/ZG-204ZV.html) knows this device it seems and it is written about sensitivity over there aswell. 😅
Should have given that a better look, but was too focussed on any reactions on movement. 😂

image

Smanar:
For config/delay you are using

```
          "read": {
            "at": "0xf001",
            "cl": "0x0500",
            "ep": 1
          },
```

So not the same than on your capture.
Where have you find thoses informations ?

I haven't the clusters available on the device (no picture) but it seem this device use the tuya cluster, not the 0x0500
Attribute 0x0013 (on cluster 0x0500) is grayed, mean this device probably don't support it.

Are you sure this device don't have tuya cluster ?

mattreim:
@Smanar That's right, it should only be checked to rule it out, and now I see that it's not supported.

derfrodo:
@Smanar @mattreim - you are absolutely right. I came to the idea as said by copying from other DDFs. In this case from the HOBEIAN ZG-204Z which is only the presence part of the ZG-204ZV. Should have had a look at your definition for the ZG-204ZL though - the sensitivity is pretty much configured there as you are most likely suspecting. 

I assume you reference to this screen, right? So there is a Tuya specific cluster
image

I am not sure, if the "grayed out" is not due to my selection. If you are referring to the font color, it is rather dark, when i do not select the row of the attribute 0x0013 
image

Thanks for your reviews and your patience.

mattreim:
@derfrodo Here's the same principle: now you should check whether the device reacts faster or slower when you change the values. You can also change the values ​​manually and read/write them.

And it's really difficult to see when the attribute is selected.

derfrodo:
I was able to update the config using the API and the device returns the updated values accordingly. 

Thanks for clarification. If you need feedback on behaviour of the device, I most likely will have to test more exhaustive (first impression is: works as expected). If so, I am not mad, closing this PR for now without a merge. So here is no trash laying around :)

Best and thanks for the support!

mattreim:
From my perspective, if the values ​​are adopted, it's fine. I'm a little bothered that Zigbee2MQTT specifies a maximum value of 19; that's odd. Otherwise, they're like the other HOBEIAN devices. I think the PR is good, and thank you very much for the excellent preliminary work. It's just a few minor details now.

Smanar:
So the sensisitivity have worked with attribute 0x0013 ?
Else tell me, I have probably another way with tuya request (but not usefull if it work with classic method)

mattreim:
@derfrodo Assuming Zigbee2MQTT is correct, please use this DDF for ZG-204ZV; the other one looks good. Thank you very much.

[hobeian_presence_temperature_humidity_light_sensor.json](https://github.com/user-attachments/files/23666341/hobeian_presence_temperature_humidity_light_sensor.json)

derfrodo:
> So the sensisitivity have worked with attribute 0x0013 ? Else tell me, I have probably another way with tuya request (but not usefull if it work with classic method)

Hey @Smanar 
Setting sensitivity using the rest api worked for values between 0 and 19 flawlessly (and those values were visible in the phoscon / deconz app aswell accordingly after being updated by rest api)

The device behaved (if you ask me / subjectively) also according to expectations by reacting more sensitive on higher values.

Even though it worked with the max with the current implementation, I have adjusted the range and the renaming suggestions according to the feedback. Thanks a lot! @mattreim !

PR #8400, opened 2025-11-19
DDF extend clones of Tuya Thermostat TRV (TS0601)

derfrodo:
This pull request updates the device compatibility list for the Tuya TS0601 TRV integration. 

The added identifiers are identifiers found in devices ordered from CN (probably clones)

Device compatibility updates:

* Added support for device identifiers `_TZE284_o3x45p96` and `_TZE200_9xfjixap` to the `devices/tuya/_TZE200_TS0601_trv.json` file.
* Updated the corresponding `modelid` array to include two additional `"TS0601"` entries for the newly added devices.

### DIFF
--- a/devices/tuya/_TZE200_TS0601_trv.json
+++ b/devices/tuya/_TZE200_TS0601_trv.json
@@ -8,9 +8,13 @@
     "_TZE200_rxntag7i",
     "_TZE200_p3dbf6qs",
     "_TZE284_ogx8u5z6",
-    "_TZE204_o3x45p96"
+    "_TZE204_o3x45p96",
+    "_TZE284_o3x45p96",
+    "_TZE200_9xfjixap"
   ],
   "modelid": [
+    "TS0601",
+    "TS0601",
     "TS0601",
     "TS0601",
     "TS0601",


mattreim:
@derfrodo There's not much to say here; Tuya simply produces a lot of devices that are then sold under different names. It's important to thoroughly test everything to ensure it works correctly and that the firmware is displayed properly.

derfrodo:
> @derfrodo There's not much to say here; Tuya simply produces a lot of devices that are then sold under different names. It's important to thoroughly test everything to ensure it works correctly and that the firmware is displayed properly.

Thanks a lot for the review! Man you are fast! 
I am using the devices for quite some time now (since early 2025 to be exact). As of now setting temperature and offset works flawlessly.

The auto-mode has not been tested by me, because i used other triggers and rules of my home automation instead.
Hope this is sufficient. As I am using the devices already, my motivation was to get other people started with them easier. Of course I don't want to promote broken stuff to the software on the other hand. 😅

mattreim:
All functions must be tested before adding the device here.

derfrodo:
> All functions must be tested before adding the device here.

Hey @mattreim - thanks for the clarification and your fast response. I have thrown another look at the DDF. As far as I see, there seems to be no definition for further specifications of the auto mode (set target temperatures for specific days / timeframes). All my added devices can toggle to the mode though.

Since I just added the devices, I would rather argue, that the DDF is still completely correct - for the added devices aswell. I assume that the DDF is lacking necessary configuration attributes.

TLTR: Imho every functionality I would have expected is working fine. Anyhow, I don't expect this to match the complete/promised feature stack of the device(s) - most likely even the existing once in main. 

I am open to close this PR if you want me, though. :)

mattreim:
> Hey @mattreim - thanks for the clarification and your fast response. I have thrown another look at the DDF. As far as I see, there seems to be no definition for further specifications of the auto mode (set target temperatures for specific days / timeframes). All my added devices can toggle to the mode though

It's true that there's no deCONZ overview for Tuya devices, but I believe there's a JSON command line for it, though I don't use it and haven't looked into it. The modes can be configured, though, right?

> ince I just added the devices, I would rather argue, that the DDF is still completely correct - for the added devices aswell. I assume that the DDF is lacking necessary configuration attributes.

As mentioned before, there are many Tuya devices with very different functions. This DDF is about two years old and it works for this product group.

> TLTR: Imho every functionality I would have expected is working fine. Anyhow, I don't expect this to match the complete/promised feature stack of the device(s) - most likely even the existing once in main.

Check your device's specifications and verify them with this DDF (e.g., locked, low battery only, mode, heatsetpoint and temperature). If everything is correct, you don't need to close this pull request.

derfrodo:
Hey @mattreim -

again thanks for the help!
Aside from low battery I have successfully tested every other attribute. I will start the search for "low aa batteries" asap to test the last feature! 😅

Best

derfrodo:
Reviewing historical data (from running the devices in openHab) I can say, that _TZE200_9xfjixap and _TZE284_o3x45p96 have both messaged low battery in the past aswell. So every attribute seems to work as expected and can be used :)

mattreim:
@Smanar I think the PR is finished now.

Issue #8371, opened 2025-10-26
BSEED ZigBee Touch LED Screen Electric Floor Heating ZigBee Water Boiler Room Thermostat (bad offset)


GitMensch:
### Is there already an existing issue for this?

- [x] I have searched the existing issues and there is none for my device

### Product name

BSEED ZigBee Touch LED Screen Electric Floor Heating ZigBee Water Boiler Room Thermostat

### Manufacturer

_TZE204_5toc8efa

### Model identifier

TS0601

### Device type to add

Thermostat

### Node info

<img width="324" height="603" alt="node info" src="https://github.com/user-attachments/assets/64087a3b-3d23-45c8-98eb-fb68d0206a0a" />

### Endpoints and clusters

<img width="281" height="483" alt="endpoints" src="https://github.com/user-attachments/assets/430239cd-a24c-4b89-a2c8-1a7e26e658ba" />

### Basic

<img width="402" height="531" alt="basic cluster" src="https://github.com/user-attachments/assets/5978cf11-bd8c-4ed1-8925-788ba598c5ef" />

### OTA

<img width="342" height="843" alt="OTA" src="https://github.com/user-attachments/assets/122804d3-0446-4be3-be76-96163800d144" />

### Manufacturer Link

https://www.bseed.com/products/bseed-touch-screen-electric-floor-heating-water-boiler-room-thermostat-zigbee-alexa-google-app-temperature-controller-backlight?_pos=1&_psq=thermostat+zigbee&_ss=e&_v=1.0

### Thermostat (Phoscon App)

Shows more data than deconz?

<img width="715" height="636" alt="phoscon" src="https://github.com/user-attachments/assets/47845465-8253-4f37-84d9-12d91e43c973" />

### Issue with this device

The temperature _setup_ has has no effect.
If this can only be setup on the device (which has multiple day and time slots, so that's kind of reasonable) there should be no option to do the setup via app.

Smanar:
>The temperature setup has has no effect.
If this can only be setup on the device (which has multiple day and time slots, so that's kind of reasonable) there should be no option to do the setup via app.

If you set it on the device direclty, it's updated in phoscon after ?

From that I m reading it miss probably a *10 or /10 convertion.

GitMensch:
Just checked with the extreme values - set the thermostat to 35° - Phoscon showed 28°; set Phoscon to 5° - thermostat was unchanged at 35°.

Smanar:
[...]

PR #8422, opened 2025-12-01
DDF correct heatsetpoint for the BSEED thermostat

Smanar:
Product name : BSEED ZigBee Touch LED Screen Electric Floor Heating ZigBee Water Boiler Room Thermostat
Manufacturer : _TZE204_5toc8efa
Model identifier : TS0601
Device type to add : Thermostat

See https://github.com/dresden-elektronik/deconz-rest-plugin/issues/8371#issuecomment-3592748922

The device is removed from the previous DDF to have it's own one.

### DIFF
--- a/devices/tuya/BSEED_TZE204_5toc8efa_thermostat.json
+++ b/devices/tuya/BSEED_TZE204_5toc8efa_thermostat.json
@@ -9,7 +9,7 @@
     "TS0601"
   ],
   "vendor": "BSEED",
-  "product": "GL86HTBZ1 Thermostat",
+  "product": "LED thermostat controller (GL86HTBZ1)",
   "sleeper": false,
   "status": "Gold",
   "subdevices": [


mattreim:
@Smanar Since this product name is already very long, what do you think of the short name of [zigbee.blakadder.com](https://zigbee.blakadder.com/BSEED_GL86HTBZ1.html)? Or a mixture of both. According to manup, this [list](https://phoscon.de/en/compatible) is now read from the DDFs.

Sorry again, but why two underscores in the filename after BSSED?

GitMensch:
The short name used in that website is the same that the manufacturer has as article number https://www.bseed.com/products/bseed-touch-screen-electric-floor-heating-water-boiler-room-thermostat-zigbee-alexa-google-app-temperature-controller-backlight?_pos=1&_psq=thermostat+zigbee&_ss=e&_v=1.0 (in the middle), just without the color part at the end - so BSEED_GL86HTBZ1 sounds fine.

@mattreim did you have to link that website that has a link to aliexpress where that costs only half the price that I've paid? ;-)
More seriously and totally off-topic: Do you have any experience of "buying Chinese" from there instead of the manufacturers? That would be a cheap way to install that into more rooms...

mattreim:
@GitMensch This question is difficult to answer because it depends on the product. For example, I always buy ESPs on AliExpress; the prices are unbeatable. But with thermostats, I prefer a warranty and buy them directly from the manufacturer or Amazon. So I can't really give many tips.

Smanar:
Done:
- Removed the underscore
- Use shorter name

mattreim:
I had actually thought of a similar name to the other DDFs, e.g.

`LED thermostat controller (GL86HTBZ1)`.

@Smanar What do you think?

Smanar:
Yep, updated again.

`  "product": "LED thermostat controller (GL86HTBZ1)",`

mattreim:
@Smanar Great, thank you

GitMensch:
Just to check: once this is merged and I install the new version of deCONZ - do I need to manually adjust/reset anything if I've adjusted the existing before via deCONZ context menu?

Smanar:
Where have you put the DDF ?
It's the same, so no action on device needed, but to prevent conflict better to remove the old DDF (Personal DDFs take priority over system one, so if one day the official is updated, you will still use your old one).

GitMensch:
> Where have you put the DDF ?

I opened deCONZ, did a context menu on the thermostat and clicked on "edit DDF" - so the answer is "whatever this does".

Smanar:
Lol, on the editor on the titlebar you have the path and file name.
For the rest it's OS dependent, but with the name you can see if it's a "user" folder or a "system" folder.

If you are using docker, it's probably a persistent folder else changes will be lost after a restart.

PR #8425, opened 2025-12-01
Add DDF for Aqara FP300 Presence Multi Sensor

SwoopX:
Alright, this is the initial DDF for the Presence Multi Sensor currently running fine in my environment.

**UPDATE**

The DDF is now heavily customized, dealing with all the previously mentioned challenges. Depending on the device mode, the resource items get values from different attributes, as they are exposed by the native gateway. That means:

- In PIR mode, attribute 0x014F is used for setting presence delay. `config/sensitivity` and `config/detectionrange` are set to invalid values (255 and 65535 respectively) representing their unavailability.
`state/presence` is using attribute 0x014D for indicating somebody is around.
- In mmWave and combined mode, delay uses attribute 0x0197. `config/sensitivity` and `config/detectionrange` and read and writable accordingly.
- `config/ledindication` is using a little workaround. While I haven't really seen this setting working, I have seen its availability in the native app. There, you can only fully disable the led or disable it during customizable timeframes. Through DDF, it will either be disabled (`false`) or enabled for 23:59hrs (`true`). `state/presence` is using attribute 0x014D OR 0x0142 for indicating somebody is around. The native gateway also does not distinct between those 2 values if memory serves.

Worth to note:
<s>- When in PIR only mode, there is no sensitivity setting available (based on native Gateway)
- When in PIR only mode, the delay for motion to go back to false is on attribute 0x014F, which is currently not exposed. Ho could that be done best?</s>
- Customization of light detection frequency (attribute 0x0192 with values 0=off, 1=low, 2=medium, 3=high, 4=custom) as well as temp./hum. detection frequency (attribute 0x0170 with values 0=off, 1=low, 2=medium, 3=high, 4=custom) is not implemented due to a lack of resource items.
- Setting/reporting detection range is currently not correctly implemented yet (attributes 0x0198 and 0x019a).
- <s>LED indication setting missing.</s>
- The added resource items will also be applied to the Aqara FP1E.

Xiaomi specific reporting on attribute 00F7 as follows:

04 21 0000 (Parent)
05 21 5700
0A 21 083D
0C 20 0A
0D 23 22340000 (Firmware)
13 20 00
17 21 BA0B (Battery voltage)
18 20 64
64 20 00
65 20 02
67 20 00

<s>Remaining questions would be: how to simultaneously expose mmwave presence and PIR occupancy and their respective delay settings?</s>

### DIFF
--- a/devices/xiaomi/xiaomi_fp300_presence_multi_sensor.json
+++ b/devices/xiaomi/xiaomi_fp300_presence_multi_sensor.json
@@ -448,7 +448,7 @@
             "at": "0x019A",
             "cl": "0xFCC0",
             "ep": 1,
-            "eval": "if (R.item('config/devicemode').val != 'pir') { var r = 0; for (var i = 0; i < 3; i++) { if (ZclFrame.cmd == '0x0A') { start = 6 } else { start = 7 } var v = ZclFrame.at(start + i) & 0xFF; while (v) { r += v & 1; v >>>= 1; } } Item.val = r * 25; } else { Item.val = 65535; }",
+            "eval": "if (R.item('config/devicemode').val != 'pir') { var r = 0; for (var i = 0; i < 3; i++) { var start = (ZclFrame.cmd === 0x0A) ? 6 : 7; var v = ZclFrame.at(start + i) & 0xFF; while (v) { r += v & 1; v >>>= 1; } } Item.val = r * 25; } else { Item.val = 65535; }",
             "fn": "zcl:attr",
             "mf": "0x115F"
           },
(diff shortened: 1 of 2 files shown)


manup:
> Remaining questions would be: how to simultaneously expose mmwave presence and PIR occupancy and their respective delay settings?

Couldn't this be done by just using 2x ZHAPresence sensors, one for each?

SwoopX:
> Couldn't this be done by just using 2x ZHAPresence sensors, one for each?

Hm, though I'm personally not too much of a fan of that idea (sensor fatigue), I guess that's a viable approach. I'll give it a try :+1:

SwoopX:
@manup Figured something out, I guess :slightly_smiling_face:

devices/xiaomi/xiaomi_fp300_presence_multi_sensor.json
```diff
@@ -0,0 +1,661 @@
...
+        },
+        {
+          "name": "config/detectionrange",
+          "refresh.interval": 3600,
+          "awake": true,
+          "parse": {
+            "at": "0x019A",
+            "cl": "0xFCC0",
+            "ep": 1,
+            "eval": "if (R.item('config/devicemode').val != 'pir') { var r = 0; for (var i = 0; i < 3; i++) { if (ZclFrame.cmd == '0x0A') { start = 6 } else { start = 7 } var v = ZclFrame.at(start + i) & 0xFF; while (v) { r += v & 1; v >>>= 1; } } Item.val = r * 25; } else { Item.val = 65535; }",
```
manup: Small finds: The ZclFrame.cmd is an int, and start wasn't declared.
Could also be written as follows to fix it:

```js
if (R.item('config/devicemode').val != 'pir') {
    var r = 0;
    for (var i = 0; i < 3; i++) {
        var start = (ZclFrame.cmd === 0x0A) ? 6 : 7;
        var v = ZclFrame.at(start + i) & 0xFF;
        while (v) {
            r += v & 1;
            v >>>= 1;
        }
    }
    Item.val = r * 25;
} else {
    Item.val = 65535;
}
```

SwoopX:
Thanks for the review! Yeah, definitely quite a toy, however, it positively surprises. Small caveat is that it freezes every now and then, but that issue is already known to Aqara. Hopefully, they bring a fix; wasn't able to let is run in combined mode and therefore running in mmwave mode for the time being.

manup:
It's definetly one of the more interesting devices. Alright changes look good 👍

sinowine:
I currently see these 5 sensors as well, and in Home Assistant, I also have additional timeout/delay configuration options. I saw a picture on Reddit showing how the FP300 is integrated into Home Assistant using z2mqtt, and there are many more control options and sensors visible there. For example, the individual detection ranges. Will this functionality via deconz + Home Assistant be available in the future?

Issue #8428, opened 2025-12-06
Shelly Power Strip 4 Gen 4


1100101:
Hi,

I'd like to request support for the product below.
While the Outlets are recognized by deCONZ, and I can switch them on/off individually, electrical measurements/metering information is missing.

More details from Zigbee2MQTT: https://www.zigbee2mqtt.io/devices/S4PL-00416EU.html

I specifically would like to request the ability to configure the light rings of the outlets (LED mode @ Zigbee2MQTT), so I can turn the lights fully off.

### Is there already an existing issue for this?

- [x] I have searched the existing issues and there is none for my device

### Product name

Shelly Power Strip 4 Gen 4

### Manufacturer

Shelly

### Model identifier

Power Strip

### Device type to add

Switch

### Node info

<img width="677" height="596" alt="Image" src="https://github.com/user-attachments/assets/7a60de1f-5f30-4557-ba9a-f5b362cc2e08" />

### Endpoints and clusters

<img width="249" height="879" alt="Image" src="https://github.com/user-attachments/assets/98cb524c-2cb0-43b5-ab5a-316c8f661bcb" />

### Basic

<img width="674" height="581" alt="Image" src="https://github.com/user-attachments/assets/5255e5d1-3be5-4072-94d7-82d4f2b54fb7" />

### On/Off (Plug 1)
<img width="666" height="273" alt="Image" src="https://github.com/user-attachments/assets/9ba73532-e8f7-41f7-8bc9-06325c03a105" />

### Simple Metering (Plug 1)
<img width="670" height="485" alt="Image" src="https://github.com/user-attachments/assets/59720262-aeb0-40e4-967d-4a292c6333d3" />

### Electrical Measuremeant (Plug 1)
<img width="669" height="1117" alt="Image" src="https://github.com/user-attachments/assets/f2f30def-b12d-4184-8a21-4279dada2973" />
<img width="668" height="941" alt="Image" src="https://github.com/user-attachments/assets/40bb16c8-b0a5-4d83-a05f-b7fe83de18ad" />
<img width="665" height="664" alt="Image" src="https://github.com/user-attachments/assets/61f5e93a-7d80-43b8-aaec-985322de8a4b" />

mattreim:
@1100101 A power strip with four measuring points is useful. It's important to ensure the values ​​are accurate, e.g., the consumption in Wh.

[...]

PR #8434, opened 2025-12-12
DDF for Shelly Power Strip 4 gen4

mattreim:
This PR includes the main functions of the power strip. See #8428

- 4x on/off switch
- 4x voltage
- 4x current
- 4x power
- 4x consumption

Once familiarity with "Shelly RPC Cluster", further settings can be added in subsequent PRs.


mattreim:
The behavior where power and current occasionally jump to [zero](https://github.com/dresden-elektronik/deconz-rest-plugin/issues/8428#issuecomment-3649724813) cannot be caused by the DDF of the Shelly Power Strip 4 gen4. Therefore, it can be set to ready for merging.

PR #8438, opened 2025-12-15
Remove `state/power_divisor` from Sonoff smart Plug (S60ZBTPF)

mattreim:
`state/power_divisor` is not necessary for this device and has therefore been removed. It was a mistake, sorry


SwoopX:
What's the catch in backing off? I recall there were voices heavily insisting the divisor is of utmost importance (which it isn't).

PR #8440, opened 2025-12-17
Fix SIGSEGV due to incorrect data types for CLIP sensor resource items

SwoopX:
Data type `DataTypeReal` was used for a handful of resource items, which is not inline with their definitions. This is causing crashes when trying to create a CLIPPower, CLIPPressure or CLIPConsumption sensor. The PR sets the correct data types for these.

image


manup:
Interesting thanks for the fix. But even with the wrong data types I don't understand how this can lead to SEGV?

SwoopX:
@manup I've added a screenshot from the backtrace. Maybe you spot anything in addition.

Issue #8437, opened 2025-12-14
add support for inn OSP 240 energy consumption on HA


seymourddr:
### Is there already an existing issue for this?

- [x] I have searched the existing issues and there is none for my device

### Product name

innr OSP 240

### Manufacturer

innr

### Model identifier

OSP 240

### Device type to add

Sensor

### Node info

<img width="558" height="487" alt="Image" src="https://github.com/user-attachments/assets/1b3099de-c4d2-4666-953c-402f8df790a1" />

### Endpoints and clusters

<img width="802" height="816" alt="Image" src="https://github.com/user-attachments/assets/b9ad73c9-fdad-48d6-9817-1faeaa591b47" />

<img width="802" height="816" alt="Image" src="https://github.com/user-attachments/assets/63c16024-612c-48a7-b5b6-0c6fed0ed1f7" />

### Basic

<img width="802" height="678" alt="Image" src="https://github.com/user-attachments/assets/9f7c8ed2-31a2-45bc-b741-da413314b592" />

<img width="708" height="372" alt="Image" src="https://github.com/user-attachments/assets/1cc942dd-16d6-436a-922e-f602d31614a7" />

<img width="708" height="372" alt="Image" src="https://github.com/user-attachments/assets/a3d01186-d95e-4cf3-b46f-f31ed417f9aa" />

mattreim:
@seymourddr It could be the same device as the "regular" SP 240. Please test this existing DDF.

[sp_240.json](https://github.com/user-attachments/files/24175220/sp_240.json)

This DDB can be used directly in Phoscon.

[smart_plug.zip](https://github.com/user-attachments/files/24154391/smart_plug.zip)

seymourddr:
<img width="709" height="456" alt="Image" src="https://github.com/user-attachments/assets/083267b9-e0bc-4f27-9749-cf81727fe7b1" />

I loaded the ddb file into phoscon. Now HA displays the total and instantaneous consumption.

mattreim:
@seymourddr Great that it worked. That's it, and you can close the issue by clicking "Close problem". I'll create a pull request to add (#8441) the device to the database. Thank you very much.

PR #8441, opened 2025-12-21
DDF clone for Innr OSP 240 and fix consumption values

mattreim:
Product name: Innr OSP 240
Manufacturer: Innr
Model identifier: OSP 240

See #8437

It looks like it's the same device as the SP 240.


mattreim:
The factor of 10 is no longer needed (#8442) in the new firmware versions (1.9.3X). If the consumption value on your Innr device is too low, you should perform a firmware update, or we will need to create a separate DDF for the older versions.

PR #8443, opened 2025-12-23
Fix for Shelly Power Strip 4 gen4

mattreim:
Sorry @manup, I unfortunately overlooked the typo.

### DIFF
--- a/devices/shelly/shelly_power_strip4_gen4.json
+++ b/devices/shelly/shelly_power_strip4_gen4.json
@@ -178,7 +178,7 @@
             "ep": 1,
             "cl": "0x0702",
             "at": "0x0000",
-            "eval": "Item.val = Attr.val / 1000"
+            "eval": "Item.val = Attr.val / 1000;"
           },
           "default": 0
         },
@@ -275,13 +275,20 @@
         {
           "name": "state/voltage",
           "refresh.interval": 360,
+          "read": {
+            "fn": "zcl:attr",
+            "ep": 1,
+            "cl": "0x0B04",
+            "at": "0x0505"
+          },
           "parse": {
             "fn": "zcl:attr",
             "ep": 1,
             "cl": "0x0B04",
             "at": "0x0505",
             "eval": "if (Attr.val != 65535) { Item.val = Math.round(Attr.val / 100) }"
-          }
+          },
+          "default": 0
         }
       ]
     },
@@ -399,7 +406,7 @@
             "ep": 2,
             "cl": "0x0702",
             "at": "0x0000",
-            "eval": "Item.val = Attr.val / 1000"
+            "eval": "Item.val = Attr.val / 1000;"
           },
           "default": 0
         },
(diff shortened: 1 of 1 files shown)


mattreim:
Hi everyone @ebaauw, @manup, @Smanar, @SwoopX I would also like to take this opportunity to wish you all happy holidays.

Smanar:
Lol Happy holidays too :).
BTW you aren't on Discord ?

mattreim:
Not yet, maybe later.

SwoopX:
> Sorry @manup, I unfortunately overlooked the typo.

No worries, happens to all of us.

In addition, can we please make some further changes/optimizations?

Each and every occurrence of `state/power` and `state/current` should look as follows:
```
{
"name": "state/power",
"refresh.interval": 360
}
```
```
{
"name": "state/current",
"refresh.interval": 360
}
```

For `state/voltage`, the below would already be sufficient for each endpoint (EP needs to be changed as required ofc):

```
{
"name": "state/voltage",
"refresh.interval": 360,
"parse": {
"fn": "zcl:attr",
"ep": 4,
"cl": "0x0B04",
"at": "0x0505",
"eval": "if (Attr.val != 65535) { Item.val = Math.round(Attr.val / 100) }"
}
```

For `modelid` and `manufacturername`, you can do this on the first endpoint:

```
        {
          "name": "attr/manufacturername",
          "refresh.interval": 86400,
          "read": {
            "at": "0x0004",
            "cl": "0x0000",
            "ep": 1,
            "fn": "zcl:attr"
          }
        },
        {
          "name": "attr/modelid",
          "refresh.interval": 86400,
          "read": {
            "at": "0x0005",
            "cl": "0x0000",
            "ep": 1,
            "fn": "zcl:attr"
          }
        }
```

but leave it simple for all the other endpoints (that reduces unnecessary reads):

```
        {
        },
        {
          "name": "attr/modelid"
        }
```

Last but not least, the `state/consumption` value calculations looks incorrect. The divisor for that value is 1.000.000, which means the raw value must be divided by 1000 to represent Wh.

mattreim:
It would be great if the DDF could be shortened or improved; I'm open to suggestions. But maybe take another look to make sure I've caught everything. Thanks.

mattreim:
> For `state/voltage`, the below would already be sufficient for each endpoint (EP needs to be changed as required ofc):
> 
> ```
> {
> "name": "state/voltage",
> "refresh.interval": 360,
> "parse": {
> "fn": "zcl:attr",
> "ep": 4,
> "cl": "0x0B04",
> "at": "0x0505",
> "eval": "if (Attr.val != 65535) { Item.val = Math.round(Attr.val / 100) }"
> }
> ``

@SwoopX Should "read" really be removed from state/voltage since it's also done in generic/items (`state/power`, `state/current` and `state/voltage`)?

Smanar:
If it's same yes.
If the "read" part is empty on the DDF deconz will use the one from the generic/items file (it's all defaut values)

SwoopX:
Indeed. As Smanar already pointed out, the default read and parse functions from the generic items fit. Hence, there's not need to overwrite them and it also eliminates a potential source of typos.

mattreim:
@SwoopX Okay, I didn't know that, but I probably would have added it again for readability. For me, sometimes two is better than one, and it's done that way in most DDFs.

manup:
Hey guys, thanks the improvements, is the PR ready for merging?

mattreim:
The read part is now included again, and I think the DDF is ready to be merged.

PR #8446, opened 2025-12-26
Add additional manufacturer ID for Tuya light switch 

jpwenzel:
Adding additional manufacturer name for Tuya light switch.

### DIFF
--- a/devices/tuya/_TZ3000_TS0041_1gang_remote.json
+++ b/devices/tuya/_TZ3000_TS0041_1gang_remote.json
@@ -10,7 +10,7 @@
     "_TZ3000_qgwcxxws",
     "_TZ3000_yj6k7vfo",
     "_TZ3000_axpdxqgu",
-    "_TZ3000_itb0omhva"
+    "_TZ3000_itb0omhv"
   ],
   "modelid": [
     "TS0041",


Smanar:
You are sure you haven't done a typo ?
The name have more character than usual and I don't found the device on google "_TZ3000_itb0omhvac"

Edit:
the "0" ?

jpwenzel:
You're absolutely right — I accidentally added a "c" character at the end when copy-pasting from the remote host in VNC.

image

I'll fix that.

Smanar:
And your device have 4 endpoints, it's not a 4 gangs version ?

jpwenzel:
> And your device have 4 endpoints, it's not a 4 gangs version ?

It's 1-gang casing but internally it looks like it would allow up to 3-gang. I pried it open and only the middle switch is actually used in this model.

It might even be better to move this to a 3-gang config file?

IMG_0913
IMG_0914

Smanar:
Fantastic, lol, now tuya use same hardware and same firmware, but they just "hidde physicaly" features to make lower stuff ....

>It might even be better to move this to a 3-gang config file?

Nope it's fine, else you will have 3 devices created, but only one usable

mattreim:
I hope there isn't a 4-gang version of this device with the same model ID, but please correct it to "_TZ3000_itb0omhv". Also, please remove the "a" at the end.

Smanar:
>I hope there isn't a 4-gang version of this device with the same model ID

Lol, I hope too.

jpwenzel:
Switch S2 is detected properly with this code change. S3 and S1 seem to not raise any action (expected for the 1-gang switch version)

manup:
Hi thanks for the PR, is it ready for merge now?

jpwenzel:
> Hi thanks for the PR, is it ready for merge now?

I guess so. :)

I've applied the changes locally to the file in my deconz installation and now can detect (and properly use) the 1-gang switch.

PR #8464, opened 2026-01-18
Update Eurotronic Spirit reporting and remove legacy code

SwoopX:
Thermostat cluster had the reportable change values missing on the mf specific attributes and there was also a mistake in one data type. Other than that, further legacy code is removed.


mattreim:
Thank you so much for the PR.

manup:
Nice, always good to see the legacy stuff being removed :+1:

PR #8467, opened 2026-01-19
Add Bosch specific cluster data to general.xml

mattreim:
A few new Bosch clusters were discovered and documented on [zigpy](https://github.com/zigpy/zha-device-handlers/blob/dev/zhaquirks/bosch/rbsh_rth0_zb_eu.py).

### DIFF
--- a/general.xml
+++ b/general.xml
@@ -2755,11 +2755,6 @@ Note: It does not clear or delete previous weekly schedule programming configura
             <value name="14" value="0x0E"></value>
             <value name="15" value="0x0F"></value>
           </attribute>
-          <attribute id="0x4033" name="Valve status LED" type="enum8" access="rw" required="o" mfcode="0x1209">
-            <value name="Off" value="0x00"></value>
-            <value name="Normal" value="0x01"></value>
-            <value name="On" value="0x02"></value>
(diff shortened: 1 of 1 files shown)


mattreim:
I have also added unknown clusters in case someone here can figure out which ones they are.

```
There are some more undocumented attributes that have not been figured out what they do.

0x4023: R/W. Valid range 0-7.
0x4024: R/W. Valid range 0-23.
0x4025: R/W. Valid range 0-100. Changes depending on heater type (0x00: 0x03, 0x01: 0x01, 0x02: 0x02)
0x4050: R/W. Valid range 5-10.
0x405b: R/W. Valid range 0-255.
0x4061: Read-only. Changes depending on heater type (0x00: 0x14, 0x01: 0x06, 0x02: 0x0a)

More undocumented and unknown attributes in the UserInterface cluster.

0x4032: R/W. Valid range 0-15.
0x406a: R/W. Valid range 0-255.
0x406b: R/W. Valid range 0-255.
0x406c: R/W. Valid range 0-255. Changes depending on heater type (0x00: 0x00, 0x02: 0x05)
0x406d: R/W. Valid range 0-255.
```

PR #8468, opened 2026-01-19
Corrected data type for `config/mode` (Ubisys and Sonoff)

mattreim:
Two incorrect data types have crept in here.


jan666:
I can confirm this works for the H1. Thanks!

PR #8470, opened 2026-01-19
Add `externalsensortemp` to Bosch room thermostat II (BTH-RM/BTH-RM230Z)

mattreim:
I've been searching for the address `0x4051` for quite some time now, and it seems to have been discovered on [zigpy](https://github.com/zigpy/zha-device-handlers/blob/dev/zhaquirks/bosch/rbsh_rth0_zb_eu.py#L424). It can then be used, for example, to display the outside temperature.

Additionally, the DDFs are formatted.

### DIFF
--- a/devices/bosch/room_thermostat2.json
+++ b/devices/bosch/room_thermostat2.json
@@ -106,6 +106,35 @@
           "default": 14400
         },
         {
+          "name": "config/externalsensortemp",
+          "refresh.interval": 360,
+          "read": {
+            "at": "0x4051",
+            "cl": "0x0201",
+            "ep": 1,
+            "fn": "zcl:attr",
+            "mf": "0x1209"
+          },
+          "parse": {
+            "at": "0x4051",
+            "cl": "0x0201",
+            "ep": 1,
+            "eval": "Item.val = Attr.val",
+            "fn": "zcl:attr",
+            "mf": "0x1209"
+          },
+          "write": {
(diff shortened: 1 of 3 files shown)


mattreim:
@Smanar @manup One more question: should `config/externalsensortemp` also have a binding?

Smanar:
Attribute 0x4051 on cluster 0x0201 is used too for windows detection for Danfoss (from the xml file)

I never see a bind/report for externalsensortemp on DDF, idk if it work, can make a try if you have the device (it don't work for all attributes), but not realy usefull for me.
For "config" field,  on first start the value is asked, rarely changed, if changed, it's updated, for me a "refresh.interval" is enought (and with big delay for me)

And "externalsensortemp"  is used only to write a value for the device use it. No one will "read" it, and every time the user "write" it, the value is updated, in fact we don't need a return (except for error check)

Natural Reports are always better than poll, but IDK if the device support it for this attribute and the maximum that is possible to set.

Edit:
Have checked Z2M, they are not configuring report for thoses attributes too, just an attribute read.

Edit2:
If you have the device, try to set a report using the GUI, you will probably have an error message.

mattreim:
Okay, thank you. I was just wondering, but it's true, the value isn't "read" and it also saves battery.

The attribute `0x4051` for Bosch devices is added with this [PR](https://github.com/dresden-elektronik/deconz-rest-plugin/pull/8467).

Issue #8469, opened 2026-01-19
Moes CO sensor


timbru31:
### Is there already an existing issue for this?

- [x] I have searched the existing issues and there is none for my device

### Product name

Moes CO sensor

### Manufacturer

_TZE284_rjxqso4a

### Model identifier

TS0601

### Device type to add

Sensor

### Node info

<img width="903" height="982" alt="Image" src="https://github.com/user-attachments/assets/571a2bfc-2bd4-413c-8a44-448e713cfd80" />

### Endpoints and clusters

<img width="465" height="495" alt="Image" src="https://github.com/user-attachments/assets/b3289fe8-fb34-419e-9886-d93e0a610e47" />

### Basic

<img width="886" height="1265" alt="Image" src="https://github.com/user-attachments/assets/d4fb2ead-8ec6-487f-bb47-8230c51d3a33" />

### Any other cluster of relevance/interest

Other clones _TZE200_hr0tdd47 & _TZE200_rjxqso4a

Product: https://moeshouse.com/products/zigbee-co-detector-alarm-wireless-gas-leak-carbon-monoxide-sensor?variant=48910148305211

timbru31:
Tuya datapoints should be:

* 1, carbon monoxide
* 2, co concentration
* 9, self test
* 15, battery
* 16, mute siren

mattreim:
@Smanar What do you think about this sensor?

- Subdivision into multiple sensors (e.g $TYPE_AIR_QUALITY_SENSOR)
- An idea for integrating the siren

```json
{
  "schema": "devcap1.schema.json",
  "manufacturername": "_TZE284_rjxqso4a",
  "modelid": "TS0601",
  "product": "Carbon monoxide alarm (ZC-HM)",
  "sleeper": true,
  "status": "Gold",
  "subdevices": [
    {
      "type": "$TYPE_CARBONMONOXIDE_SENSOR",
      "restapi": "/sensors",
      "uuid": [
        "$address.ext",
        "0x01",
        "0x0512"
      ],
      "items": [
        {
          "name": "attr/lastannounced"
        },
        {
          "name": "attr/lastseen"
        },
        {
          "name": "attr/manufacturername"
        },
        {
          "name": "attr/modelid"
        },
        {
          "name": "attr/name"
        },
        {
          "name": "attr/swversion",
          "refresh.interval": 86400,
          "parse": {
            "fn": "zcl:attr",
            "ep": 1,
[...]

PR #8486, opened 2026-01-31
DDF for Moes CO sensor (ZC-HM)

mattreim:
Product name: Moes CO sensor
Manufacturer: _TZE284_rjxqso4a
Model identifier: TS0601

See #8469

### DIFF
--- a/devices/moes/Moes_ZC-HM_co_sensor.json
+++ b/devices/moes/Moes_ZC-HM_co_sensor.json
@@ -91,8 +91,9 @@
         {
           "name": "state/alert",
           "write": {
-            "dpid": 16,
             "fn": "tuya",
+            "dpid": 16,
+            "dt": "0x10",
             "eval": "if (Item.val=='none') { true }"
           },
           "read": {
@@ -204,9 +205,9 @@
         {
           "name": "state/measured_value",
           "parse": {
+            "fn": "tuya",
             "dpid": 2,
-            "eval": "Item.val = Attr.val",
-            "fn": "tuya"
+            "eval": "Item.val = Attr.val"
           },
           "read": {
             "fn": "none"


mattreim:
@Smanar I believe this expression is not yet known, and `state/alert` also expects a string.

```json
        {
          "name": "state/alert",
          "write": {
            "dpid": 16,
            "fn": "tuya",
            "eval": "if (Item.val=='none') { true }"
          },
          "read": {
            "fn": "none"
          }
        },
```

Smanar:
>Error: Invalid input: expected string, received undefined

But where ???

And if you remove all "`values":` ?

mattreim:
The "values" were already an attempt at repair which I have now reversed.

Smanar:
I think the problem is in the TYPE_WARNING_DEVICE, because if we delete the 2 other, we still have the same issue
But haven't see the problem yet.

mattreim:
@Zehir do you have any ideas?

Zehir:
> @Zehir do you have any ideas?

@mattreim @Smanar yes sometime the error message on github is not that great. When that happen feel free to open a issue on the repo : https://github.com/deconz-community/ddf-tools

Using the CLI version of the validator give a better error;
https://www.npmjs.com/package/@deconz-community/cli

```
vscode ➜ /workspaces/deconz-rest-plugin (_TZE284_rjxqso4a) $ ddf-tools validate --directory ./
- Using validator version 2.34.0
✔ Found 741 files to validate
Unknown schema for file button_maps.json
Unknown schema for file button_maps.schema.json
Skipping file devices/danalock/danalock_v3.json because it has the ddfvalidate option set to false
Skipping file devices/legrand/cable_outlet.json because it has the ddfvalidate option set to false
Skipping file devices/nodon/SIN-4-FP-21.json because it has the ddfvalidate option set to false
Skipping file devices/tuya/_TZ3210_tgvtvdoc_rain_sensor.json because it has the ddfvalidate option set to false
Skipping file devices/tuya/_TZE204_d6i25bwg_pilote_wire_heating_module.json because it has the ddfvalidate option set to false
✔ Loaded 276 generic files and 458 DDF files from disk
✔ No errors found in 276 generic files
✖ Found 1 error in 458 ddf files
  File: devices/moes/Moes_ZC-HM_co_sensor.json
    Invalid input: expected string, received undefined at "subdevices[0].items[11].write.dt"
```

In the first subdevice (around line 90 to 100) we have:
```json
        {
          "name": "state/alert",
          "write": {
            "dpid": 16,
            "fn": "tuya",
            "eval": "if (Item.val=='none') { true }"
          },
          "read": {
            "fn": "none"
          }
        },
```
The `dt` property inside write is required for `tuya` write;
https://github.com/deconz-community/ddf-tools/blob/ae1a7b59f66a418f58a7fe21715a5b53324868be/packages/validator/src/schema/function.ts#L114-L118

If that should not be the case feel free do ping me to update the validator or open an issue for me [here](https://github.com/deconz-community/ddf-tools).

mattreim:
@Zehir Thank you so much for the excellent analysis and the quick response; sometimes you need a fresh perspective on a problem. The better error message would certainly be helpful.

Smanar:
Haaa, thx a  lot ^^ @Zehir

PR #8498, opened 2026-02-15
Update DDF for Namron thermostat touch ZigBee 16A (4512737/4512738)

mattreim:
Perhaps the device can be improved further. Feel free to also take a look at the PDF.

[4512737-Brukerveiledning-5.pdf](https://github.com/user-attachments/files/25319083/4512737-Brukerveiledning-5.pdf)

- Change `config/mode` => `off`, `heat`, `auto` and  `dry` => `heat` = `manual`
- Change `config/offset`
- Add `config/locked`
- Add `config/windowopen_set`
- Change `state/on`
- Change `state/temperature`
- Add bindings

### DIFF
--- a/devices/namron/4512737_thermostat.json
+++ b/devices/namron/4512737_thermostat.json
@@ -145,7 +145,7 @@
           "name": "config/reachable"
         },
         {
-          "name": "config/windowopendetectionenabled",
+          "name": "config/windowopen_set",
           "refresh.interval": 3660,
           "read": {
             "at": "0x1009",


mattreim:
@Smanar could you please take a look at this?

Smanar:
I don't see something special.
You have a doubt on something ? You haven't the device to test ?

and there is 2 fields for the same thing
```

{
  "id": "config/windowopen_set",
  "datatype": "Bool",
  "access": "RW",
  "public": true,
  "description": "Enable or disable the window open detection.",
  "default": false
}
```

```
{
  "id": "config/windowopendetectionenabled",
  "datatype": "Bool",
  "access": "RW",
  "public": true,
  "description": "Turns the window open detection feature on or off.",
  "default": true
}
```


IDK wich one is used by third app ....

mattreim:
I don't have the devices, but I wanted to improve the DDFs a bit. My main goal was to extend the `config/mode` and adapt it to the devices. This is important for HA to correctly represent the device status. The picture shows my Tuya radiator thermostat.

Thermostat

While creating the DDFs, I also noticed that there are two items that do the same thing. From my perspective, one of them can be deleted, but which one is a good question. Currently, I can only speak from HA, and in my Tuya I use `windowopen_set`, but it's not displayed in HA.

To get back to the first question, it's always better to look at the DDFs with four eyes, since I might have overlooked something.

Smanar:
>and in my Tuya I use windowopen_set, but it's not displayed in HA.

And windowopendetectionenabled is displayed ? From my memory windowopen_set is older.

mattreim:
I checked the integration code in HA and found nothing except `offset` and `locked`. So it seems there's no function for `windowopen...`.

mattreim:
The attributes are currently being used in these DDFs.

`config/windowopendetectionenabled` => [etrv0100_thermostat.json](https://github.com/dresden-elektronik/deconz-rest-plugin/blob/120295c12e2ef8e79b11e788f02e7048e4c4ea8a/devices/danfoss/etrv0100_thermostat.json#L377)
`config/windowopen_set`  => [_TZE200_h4cgnbzg_trv.json](https://github.com/dresden-elektronik/deconz-rest-plugin/blob/120295c12e2ef8e79b11e788f02e7048e4c4ea8a/devices/tuya/_TZE200_h4cgnbzg_trv.json#L123) and [_TZE200_TYST11_trv.json](https://github.com/dresden-elektronik/deconz-rest-plugin/blob/120295c12e2ef8e79b11e788f02e7048e4c4ea8a/devices/tuya/_TZE200_TYST11_trv.json#L188)

Smanar:
config_windowopen_set_item.json is 3 years old.
config_windowopendetectionenabled_item.json is 2 years old.

^^, on my side I will use config_windowopen_set_item

mattreim:
Should we also remove `windowopendetectionenabled`?

Smanar:
Good question.
But if HA don't use it, I don't think others third app make better, so I don't think we will break something if we remove windowopendetectionenabled.

PR #8503, opened 2026-02-17
DDF clone for Tuya presence sensor (_TZ3000_uw3dadam)

Mimiix:
add another clone


Mimiix:
@manup  I tested this with a user and it works. So can be merged :)

PR #8510, opened 2026-02-19
Sonoff TRVZB: add child lock, window open detection, frost protection, valve calibration and heating state           

iseeberg79:
Extends the existing DDF with Sonoff-proprietary cluster 0xFC11 attributes:                                                                                                           
                                                                              
  - config/locked — child lock (attr 0x0000, bool)                                                                                                                                      
  - config/windowopen_set — open window detection (attr 0x6000, bool)                                                                                                                   
  - config/frostprotection — frost protection setpoint in 1/100 °C, active when system is off (attr 0x6002, int16)
  - config/valveopeningdegree / config/valveclosingdegree — valve position limits in % (attr 0x600B/0x600C, uint8, fw ≥ v1.1.4)
  - state/heating — thermostat running state from 0x0201/0x0029 (bool)
  - state/idlesteps, state/closingsteps — motor calibration baseline (read-only)
  - state/openingvoltage, state/closingvoltage, state/runningvoltage — motor voltage diagnostics in mV (read-only)

  Adds 0xFC11 and 0x0029 bindings. Adds 8 new generic resource item descriptors for the previously unregistered item names.

### DIFF
--- a/devices/generic/items/state_closingsteps_item.json
+++ b/devices/generic/items/state_closingsteps_item.json
@@ -1,9 +1,0 @@
-{
-  "schema": "resourceitem1.schema.json",
-  "id": "state/closingsteps",
-  "datatype": "UInt16",
-  "access": "R",
-  "public": true,
-  "description": "Valve closing steps (read-only calibration).",
-  "default": 0
-}
--- a/devices/generic/items/state_closingvoltage_item.json
+++ b/devices/generic/items/state_closingvoltage_item.json
@@ -1,9 +1,0 @@
-{
-  "schema": "resourceitem1.schema.json",
-  "id": "state/closingvoltage",
-  "datatype": "UInt16",
-  "access": "R",
-  "public": true,
-  "description": "Closing voltage limit in millivolts (read-only).",
-  "default": 0
-}
--- a/devices/generic/items/state_heating_item.json
+++ b/devices/generic/items/state_heating_item.json
@@ -1,9 +1,0 @@
-{
-  "schema": "resourceitem1.schema.json",
-  "id": "state/heating",
-  "datatype": "Bool",
-  "access": "R",
-  "public": true,
-  "default": false,
-  "description": "Determines heating state for ELKO Super TR thermostat."
-}
--- a/devices/generic/items/state_idlesteps_item.json
+++ b/devices/generic/items/state_idlesteps_item.json
@@ -1,9 +1,0 @@
-{
-  "schema": "resourceitem1.schema.json",
-  "id": "state/idlesteps",
-  "datatype": "UInt16",
-  "access": "R",
-  "public": true,
-  "description": "Valve idle steps (read-only calibration baseline).",
-  "default": 0
-}
--- a/devices/generic/items/state_openingvoltage_item.json
+++ b/devices/generic/items/state_openingvoltage_item.json
@@ -1,9 +1,0 @@
-{
-  "schema": "resourceitem1.schema.json",
-  "id": "state/openingvoltage",
-  "datatype": "UInt16",
-  "access": "R",
-  "public": true,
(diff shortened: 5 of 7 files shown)


mattreim:
Thank you for your work, but unfortunately I see no reason to add these new manufacturer-specific attributes. Most users won't be able to do anything with it, and it's not currently used in any follow-up system (e.g. Domoticz, Home Assistant and so on). The addresses for these attributes weren't known yet and, from my perspective, would make sense.

- `config/locked`
- `config/windowopen_set`
- `state/on` - because `state/heating` is not correct in this case

Has this PR been tested, as I find some of the inverted values ​​unusual?

iseeberg79:
Personally I am using it since months to be able to control and monitor my devices. Not every function is tested I must admit and some aren't used yet.

What I use is device lock and windowopen detection to control my devices (mqtt/nodered).
Those are missing and intentionally I have been searching for those, analyzed the message flow, compared implementations and found additional sensors.

We could focus on more interesting sensors or close here.

My intention is just to share some work done :-)

Thank you for reviewing,
kind regards

Ingo

mattreim:
Perhaps we could extend the `general.xml` file to display the attributes in the deCONZ GUI. What do you think, @Smanar?

```xml
        <attribute id="0x6000" name="Window open detection" type="bool" access="rw" required="m" mfcode="0x1286"></attribute>
        <attribute id="0x6002" name="Frost protection temperature" type="s16" access="rw" required="m" mfcode="0x1286"></attribute>
        <attribute id="0x6003" name="Idle steps" type="u16" access="r" required="m" mfcode="0x1286"></attribute>
        <attribute id="0x6005" name="Valve opening limit voltage" type="u16" access="r" required="m" mfcode="0x1286"></attribute>
        <attribute id="0x6006" name="Valve closing limit voltage" type="u16" access="r" required="m" mfcode="0x1286"></attribute>
        <attribute id="0x6007" name="Valve motor running voltage" type="u16" access="r" required="m" mfcode="0x1286"></attribute>
        <attribute id="0x600B" name="Valve opening degree" type="u8" access="rw" required="m" mfcode="0x1286"></attribute>
        <attribute id="0x600C" name="Valve closing degree" type="u8" access="rw" required="m" mfcode="0x1286"></attribute>
        <attribute id="0x600D" name="External temperature sensor" type="u8" access="rw" required="m" mfcode="0x1286"></attribute>
        <attribute id="0x600E" name="External temperature sensor enable" type="s16" access="rw" required="m" mfcode="0x1286"></attribute>
        <attribute id="0x6011" name="Temperature accuracy" type="s16" access="rw" required="m" mfcode="0x1286"></attribute>
```

Smanar:
Yes there is already something for a light device

```
    <!-- SONOFF -->
    <cluster id="0xfc11" name="Sonoff specific" mfcode="0x1286">
      <description>Cluster for light sensor</description>
      <server>
        <attribute id="0x2001" name="Brightness" type="u8" access="r" required="m" mfcode="0x1286">
          <description></description>
        </attribute>
      </server>
      <client>
      </client>
    </cluster>
```

Can improve it (and update, it will be for more than light now).


And I m agree with @mattreim, few chance other field will be added.
Frostprotection can be used, but others ..... Easier to use the GUI to configure the device in one shoot, I don't think they will be used in automation with the API.

mattreim:
@iseeberg79 I've now adjusted the `general.xml` file, and the attributes should be visible in deCONZ with the next version. Would you like to make the changes in this PR, or would you prefer to close the PR and I'll do it? Thank you.

iseeberg79:
this way, right? If not, please feel free to make the changes. Thank you for taking the extensions into account

mattreim:
@iseeberg79 Sorry, that was a bit too much to explain, so I did it myself. Thanks again.

iseeberg79:
I am happy with ;)

PR #8511, opened 2026-02-20
Add Sonoff specific cluster data to general.xml

mattreim:
A few new Sonoff clusters were discovered and documented on [zigpy](https://github.com/search?q=repo%3Azigpy%2Fzha-device-handlers%200xfc11&type=code). The PR #8510 department also contributed, thank you very much.

Zigpy uses `manufacturer code=None` for all attributes in cluster `0xFC11`.


mattreim:
@manup Sorry, but I really like using the Visual Studio Code editor and its automatic formatting, so I formatted the `general.xml` file as well. But of course I can undo it.

manup:
Alright lets do this. I think we should also add some automatic magic to format JSON and XML files when PRs are done.

Issue #8519, opened 2026-02-24
iCasa Pulse 8S Zigbee wall controller


ebaauw:
### Is there already an existing issue for this?

- [x] I have searched the existing issues and there is none for my device

[icasa.io](https://www.icasa.io/) kindly provided me with a new version of [these switches](#1124).  The hardware seems unchanged, except for the printed label.  The firmware has changed, however.

The old version of the switch is currently supported by legacy code.  This seems to work for the new switch as well, except for the battery (see _Power Configuration_ cluster below) and the _Off_ button (see _Endpoints and clusters_ below).  This seems like a good opportunity to create a DDF and lose come legacy code.

### Product name

iCasa Pulse 8S Zigbee wall controller

### Manufacturer

icasa

### Model identifier

ICZB-KPD18S

### Device type to add

Switch

### Node info

<img width="632" height="920" alt="Image" src="https://github.com/user-attachments/assets/19a61960-078f-4199-923a-34986b4dc6d1" />

### Endpoints and clusters

<img width="159" height="305" alt="Image" src="https://github.com/user-attachments/assets/e2aa906e-6d38-43f5-b2f2-33afe85d0650" />

The client _Color Control_ cluster is new, but doesn't seem relevant.  The switch still sends commands from the client _On/Off_, _Level Control_ and _Scenes_ clusters.

The _Off_ button now sends _Move to Level_ and _Stop_ when held/released, instead of _Move to Level (with On/Off)_ and _Stop (with On/Off)_.  That's easily added to `button_maps.json`.

### Basic

<img width="632" height="920" alt="Image" src="https://github.com/user-attachments/assets/19921c3a-7728-48b0-a3f5-1864b4ac7856" />

Note the new _ZCL Version_ (3 instead of 2) and, of course, _SW Build ID_.

### Power Configuration

<img width="632" height="920" alt="Image" src="https://github.com/user-attachments/assets/f70aa274-b29e-4714-967a-457be63c25d3" />

Note that _Battery Percentage Remaining_ now follows the standard and reports the value in 0.5%.  The old firmware reports full percentages.  I think we can probably use _ZCL Version_ in the `eval`, so we would be able to support both firmware versions from the same DDF.

The new firmware supports attribute reporting on _Battery Percentage Remaining_, unlike the old firmware.  I suppose if we specify it in the DDF, the REST API plugin will give up on configuring it for the old firmware on _UNSUP_GENERAL_COMMAND_?

### OTAU
[...]

PR #8522, opened 2026-02-24
DDF for iCasa Zigbee Wall Controllers

ebaauw:
DDFs for the following iCasa devices, see #8519:
- Pulse 8S Zigbee Wall Controller `ICZB-KPD18S.json`;
- Pulse 4S Zigbee Wall Controller `ICZB-KPD14S.json`;
- Pulse 2 Zigbee Wall Controller `ICZB-KPD12.json`.

In addition to these, this PR:
- Updates `button_maps.json` as the new firmware sends different commands for the _Off_ button.
- Added `attr_zclversion_item.json` to distinguish between the firmware versions to know how to handle _Battery Percentage Remaining_.

### DIFF
--- a/devices/icasa/ICZB-KPD12.json
+++ b/devices/icasa/ICZB-KPD12.json
@@ -14,6 +14,40 @@
         "0x01",
         "0x1000"
       ],
+      "buttons": {
+        "1": {
+          "name": "Off"
+        },
+        "2": {
+          "name": "On"
+        }
+      },
+      "buttonevents": {
+        "1001": {
+          "action": "HOLD",
+          "button": 1
+        },
+        "1002": {
+          "action": "SHORT_RELEASE",
+          "button": 1
+        },
+        "1003": {
+          "action": "LONG_RELEASE",
+          "button": 1
+        },
+        "2001": {
+          "action": "HOLD",
+          "button": 2
+        },
+        "2002": {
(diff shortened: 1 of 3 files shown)


manup:
Thanks, it would be good if the DDFs also contain the "button" and "buttonevents" keys, analog to the Hue Dimmer. These are then exposed via the API and also land in the generated documentation.

ebaauw:
> it would be good if the DDFs also contain the "button" and "buttonevents" keys,

Added.

> These are then exposed via the API

Currently?  How?  I only know about the introspect, but that's based on `button_maps.json`.

manup:
> Currently? How? I only know about the introspect, but that's based on button_maps.json.

Yes the `button_maps.json` is one source to filling the internal button map structs. The ones from the DDF are another. Here the "buttonevents" have a higher priority than the older `button_maps.json`. I think the button names from "buttons" aren't correctly parsed yet and only a generic "Button #(btn/1000)" is returned.

The biggest hurdle currently is to get the actual button parsing into the DDF, for switches like the Hue Dimmer it works quite well with the extra JS file handling this, but there is no replacement for the `button_maps.json` mappings in DDF yet, which would at least work for the dumper switches that don't require much logic.

The Hue Dimmer is the ideal switch, since it is fully self contained in its DDF bundle and doesn't depend on `button_maps.json` at all.

PR #8536, opened 2026-03-04
Fix parsing general.xml

manup:
while loading deCONZ the parser fails due open/closing element mismatch.
When opening the general.xml in a browser the following error can be seen:

```
XML Parsing Error: mismatched tag. Expected: </attribute-set>.
Location: file:///home/mpi/src/deconz/src/plugins/de_web/general.xml
Line Number 2698, Column 11:
        </attribute>
----------^
```

This seems to be a regression from https://github.com/dresden-elektronik/deconz-rest-plugin/pull/8467

@mattreim can you have a look if the fix in this PR is correct?


mattreim:
Sorry, I must have overlooked something. With this fix, automatic formatting works perfectly. It's a shame the editor didn't detect this error.

manup:
Yeah I was surprised, first I through the file at a online XML validator in it also said it's fine. Seem to be some differences in various XML parsers.

PR #8539, opened 2026-03-05
Fix for Shelly 1PM (Mini) Gen4

mattreim:
It was reported in the forum that the power values ​​were not displayed correctly.

See https://forum.phoscon.de/t/solved-shelly-mini-1pm-gen-4-power-value-fault/6795


mattreim:
If you experience problems, first perform a firmware update on your device.

PR #8544, opened 2026-03-10
Change endpoint 3 to 4 on ubisys S1-R (5601)

Thomas-Vos:
On my S1-R endpoint 4 is used instead of 3. The official documentation also shows this:
https://www.ubisys.de/downloads/ubisys-s1-technical-reference.pdf

Screenshot 2026-03-10 at 21 43 06

In this PR I updated the endpoint by duplicating the files and modifying only for the S1-R.

Screenshot of my device:

Screenshot 2026-03-10 at 21 17 19

### DIFF
--- a/devices/ubisys/s1_5501.json
+++ b/devices/ubisys/s1_5501.json
@@ -2,15 +2,13 @@
   "schema": "devcap1.schema.json",
   "uuid": "d275912e-9490-4deb-9225-c17b882703ae",
   "manufacturername": [
-    "ubisys",
     "ubisys"
   ],
   "modelid": [
-    "S1 (5501)",
-    "S1-R (5601)"
+    "S1 (5501)"
   ],
   "vendor": "ubisys",
-  "product": "Power switch (S1 (5501)/S1-R (5601))",
+  "product": "Power switch (S1 (5501))",
   "sleeper": false,
   "status": "Gold",
   "subdevices": [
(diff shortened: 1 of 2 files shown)


mattreim:
At first glance it looks quite good, but I hope they've tested the DDF.

Thomas-Vos:
> At first glance it looks quite good, but I hope they've tested the DDF.

@mattreim Of course, I tested the new DDF on my S1-R.

mattreim:
@Thomas-Vos Thanks for the PR.

Thomas-Vos:
One minor issue:
When load is applied both power readings are correct.

0x0702, 0x0400 shows value 2410
0x0b04, 0x050b shows value 2410

This is correct as load is indeed around 2410 W.

When no load is applied the power readings are:
0x0702, 0x0400 shows value 16777213
0x0b04, 0x050b shows value -3

The value 16777213 appears to be max 24 bit value minus 3.

How to fix this?

Screenshot 2026-03-10 at 22 45 36

Screenshot 2026-03-10 at 22 49 09


The documentation says this
https://www.ubisys.de/downloads/ubisys-s1-technical-reference.pdf
Screenshot 2026-03-10 at 22 46 01

Thomas-Vos:
So the above appears to be unrelated to this PR as @manup already found the issue:

https://github.com/dresden-elektronik/deconz-lib/blob/2b3e51e1f73d1857353da735de52b34bf723662c/zcl.cpp#L939

PR #8546, opened 2026-03-11
Change `config/windowopen_set` to `config/windowopendetectionenabled` [2]

mattreim:
This is the follow-up PR for #8512


SwoopX (changes requested):
Thanks for taking care of the changes. Please also add `windowopendetectionenabled` with its respective configuration as mentioned in #8512

Is there any specific reason why there's 2 PRs addressing the same topic? Preferably, all the item changes should be done in one PR.

PR #8568, opened 2026-03-26
Update ubisys/Hue/Legrand specific cluster and value data to general.xml

mattreim:
A few new ubisys/Hue/Legrand clusters and values were discovered and documented on [zigpy](https://github.com/zigpy/zha-device-handlers/blob/dev/zhaquirks/legrand/__init__.py).

### DIFF
--- a/general.xml
+++ b/general.xml
@@ -2451,8 +2451,8 @@
             <value name="Winter" value="0x00"></value>
             <value name="Summer" value="0x01"></value>
           </attribute>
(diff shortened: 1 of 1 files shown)


ebaauw:
Hue-specific cluster 0xFC06 was already defined:
https://github.com/dresden-elektronik/deconz-rest-plugin/blob/e357cce2bce10598431f912b2621b0756d255ac5/general.xml#L5640

PR #8569, opened 2026-03-27
Add Third Reality specific cluster data to general.xml

mattreim:
A few new Third Reality clusters and values were discovered and documented on [zigpy](https://github.com/zigpy/zha-device-handlers/tree/dev/zhaquirks/thirdreality).


manup:
Hi @mattreim thanks for all the additions :+1: can you please have a look why it won't merge?

mattreim:
@manup It seems to have worked now.

PR #8577, opened 2026-04-03
DDF clone for Nous A6Z outdoor plug (_TZ3000_266azbg3)

maltekluever:
since I had a bunch of Nous A1Z smart plugs (_TZ3000_2putqrmw), which worked like a charme, I bought the Nous A6Z Outdoor Plug (_TZ3000_266azbg3) having high hopes that it also works out of the box. 

Switching the plug on/off worked. However, the energy consumption and power parameters did not work. To solve that, I simply used the same DDF for the A6Z as deconz used for the A1Z and voilà, energy monitoring also works now.


mattreim:
@maltekluever It's a typical Tuya clone. Thank you for the PR.

PR #8615, opened 2026-05-28
DDF clone for Soil Moisture Sensor - ThirdReality 3RSM0347Z

### LINKED ISSUE #8621: Soil Moisture Sensor - ThirdReality 3RSM0347Z

- [x] I have searched the existing issues and there is none for my device

### Product name

Soil Moisture Sensor (3RSM0347Z)

### Manufacturer

ThirdReality

### Model identifier

3RSM0347Z

### Device type to add

Sensor

### Node info

Image

### Endpoints and clusters

Image

### Basic

Basic Cluster Info

Image

### Power Configuration
Image

### Any other cluster of relevance/interest
Temperatur measurement

Image

Humanity

Image

Soil Moisture

Image

specific

Image

EvilPingu:
PR for ThirdReality Smart Soil Moisure Sensor Gen2

### DIFF
--- a/devices/third_reality/3RSM0347Z_moisture_sensor.json
+++ b/devices/third_reality/3RSM0347Z_moisture_sensor.json
@@ -1,10 +1,9 @@
 {
   "schema": "devcap1.schema.json",
-  "uuid": "33e6dfe7-9591-49f5-a034-8c30c9e607a3",
   "manufacturername": "Third Reality, Inc",
-  "modelid": "3RSM0147Z",
+  "modelid": "3RSM0347Z",
   "vendor": "Third Reality",
-  "product": "Soil moisture sensor (3RSM0147Z)",
+  "product": "Soil moisture sensor (3RSM0347Z)",
   "sleeper": true,
   "status": "Gold",
   "subdevices": [
@@ -87,7 +86,7 @@
       "uuid": [
         "$address.ext",
         "0x01",
-        "0x0405"
+        "0x0408"
       ],
       "items": [
         {
@@ -150,7 +149,7 @@
           "name": "state/moisture",
           "parse": {
             "at": "0x0000",
-            "cl": "0x0405",
+            "cl": "0x0408",
             "ep": 1,
             "eval": "Item.val = Attr.val + R.item('config/offset').val",
             "fn": "zcl:attr"
@@ -168,6 +167,21 @@
       "bind": "unicast",
       "src.ep": 1,
       "dst.ep": 1,
+      "cl": "0x0408",
+      "report": [
+        {
+          "at": "0x0000",
+          "dt": "0x21",
+          "min": 60,
+          "max": 1800,
+          "change": "0x00000064"
+        }
+      ]
+    },
+    {
+      "bind": "unicast",
+      "src.ep": 1,
+      "dst.ep": 1,
       "cl": "0x0001",
       "report": [
         {
@@ -176,21 +190,6 @@
           "min": 7200,
           "max": 43200,
           "change": "0x00000002"
-        }
-      ]
-    },
-    {
-      "bind": "unicast",
-      "src.ep": 1,
-      "dst.ep": 1,
-      "cl": "0x0405",
-      "report": [
-        {
-          "at": "0x0000",
-          "dt": "0x21",
-          "min": 60,
-          "max": 1800,
-          "change": "0x00000064"
         }
       ]
     },


mattreim:
@EvilPingu If it's the same DDF as the one for the previous model, you can simply add the model/manufacturer and the device will be supported. Please replace the JSON file. I've also made a few other changes. Thank you very much.

[3RSM0147Z_moisture_sensor.json](https://github.com/user-attachments/files/28370203/3RSM0147Z_moisture_sensor.json)

EvilPingu:
> @EvilPingu If it's the same DDF as the one for the previous model, you can simply add the model/manufacturer and the device will be supported. Please replace the JSON file. I've also made a few other changes. Thank you very much.
> 
> [3RSM0147Z_moisture_sensor.json](https://github.com/user-attachments/files/28370203/3RSM0147Z_moisture_sensor.json)

I have incorporated your comments. However - i leave the new file because i think there is a explicit moisure sensor. This ddf could be loaded in deconz successfully.

I still could not add the sensor in Phoscon and the api has still no endpoints for the the subdevices of the device. Maybe after a new version of deconz

mattreim:
I just checked again, and we will treat this device as a clone. If the device is working now, it will also work with my DDF, and no new one will be needed.

EvilPingu:
> I just checked again, and we will treat this device as a clone. If the device is working now, it will also work with my DDF, and no new one will be needed.

OK - i removed the new file and changed the existing one. Is it necessary to list the manufacturer twice?

And as i sad: I still could not add the sensor in Phoscon and the api has still no endpoints for the the subdevices of the device.

mattreim:
The second manufacturer is required for unique identification, as the name could potentially differ. The device may also possess all its attributes within EP1. For the device to display correctly in Phoscon, it will likely be necessary to delete and re-pair it to ensure proper assignment.

EvilPingu:
> The second manufacturer is required for unique identification, as the name could potentially differ. The device may also possess all its attributes within EP1. For the device to display correctly in Phoscon, it will likely be necessary to delete and re-pair it to ensure proper assignment.

OK. Thanks for the information.

I deleted the device in deconz and started the search for new sensors. The sensor is created in deconz but still is not accessible over rest and not added in Phoscon

mattreim:
Unfortunately, I don't have any time this weekend, but... Please check whether the DDF is being recognized; usually, there is an error in the `model`/`manufacturer` fields—everything needs to be exactly correct there. All DDFs are loaded only upon a restart. You should also check within the DDF editor in deCONZ.

EvilPingu:
> Unfortunately, I don't have any time this weekend, but... Please check whether the DDF is being recognized; usually, there is an error in the `model`/`manufacturer` fields—everything needs to be exactly correct there. All DDFs are loaded only upon a restart. You should also check within the DDF editor in deCONZ.

OK. I will check this. Thanks

EvilPingu:
> Unfortunately, I don't have any time this weekend, but... Please check whether the DDF is being recognized; usually, there is an error in the `model`/`manufacturer` fields—everything needs to be exactly correct there. All DDFs are loaded only upon a restart. You should also check within the DDF editor in deCONZ.

I double checked model and manufacturer. Both matches. I also see in the logs, that the new file is recognized and assocciated with the device. But i still have no endpoints in the REST-API for the subdevices (but for the device) and Phoscon does not recognize the sensor.

mattreim:
It would be better if you created a device request so that the device can be better evaluated.

EvilPingu:
OK. Than i create an issue

#8621

EvilPingu:
I checked the file and it works with deconz. There are two things to mention.

1. The description: Should i add the second model id?
2. Currently the moisture sensor point to 0x405 but it should 0x408. 0x405 is the sensor for humidity. However: Both sensors delivery exactly the same value. So it's possible to leave it as it is. The other option is to create a second file for the device.

@mattreim  What's the correct way?

mattreim:
In principle, this DDF should work for both devices, but the question regarding 0x0405 or 0x0408 should be clarified first.

EvilPingu:
> In principle, this DDF should work for both devices, but the question regarding 0x0405 or 0x0408 should be clarified first.

We clarified 0x0405 or 0x0408

0x0408 is the correct soil moisture cluster.  0x0405 is humanty cluster but it delivers exactly the same value. So i change the subdevice to 0x0408

lhgravendeel:
Thank you for this update! I was successfully able to add the Third Reality moisture sensor after adding the new DDF file to my devices directory.

EvilPingu:
Is someone there Is there anyone who can approve the PR?

PR #8618, opened 2026-05-30
Fix crash, remove TCP socket tracking

manup:
For some user the crash occurred while tracking timeouts on TCP connections within the plugin.

This is duplicated functionality as the tracking is already handled in the main application. The crash is likely due invalid dangling pointers.

This PR removes the code. The main plugin has a timeout on TCP connections of 10 seconds (except for Websockets).

### DIFF
--- a/authorisation.cpp
+++ b/authorisation.cpp
@@ -13,8 +13,6 @@
 #include "crypto/password.h"
 #include "de_web_plugin_private.h"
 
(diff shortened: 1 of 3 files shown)


authorisation.cpp
SwoopX: You might want to remove this line as well :)

manup: Good catch 👍

PR #8624, opened 2026-06-04
DDF: Create swv_zne_smart_water_valve_lite.json

### LINKED ISSUE #8625: SONOFF Hydro One Lite

- [x] I have searched the existing issues and there is none for my device

### Product name

SONOFF Hydro One Lite

### Manufacturer

SONOFF

### Model identifier

SWV-ZNE

### Device type to add

Light

### Node info

Image

Working DDF in https://github.com/dresden-elektronik/deconz-rest-plugin/pull/8624

### Endpoints and clusters

Image

### Basic

Image

### Power Configuration

Image

### On/Off
Nothing special

Image

Finne75:
DDF for the SONOFF Hydro ONE Lite water valve. Exposes battery as dedicated sensor, no water flow/etc sensors.

### DIFF
--- a/devices/generic/items/state_hour_irrigation_duration_item.json
+++ b/devices/generic/items/state_hour_irrigation_duration_item.json
@@ -1,0 +1,9 @@
+{
+  "schema": "resourceitem1.schema.json",
+  "id": "state/hour_irrigation_duration",
+  "datatype": "UInt32",
+  "access": "R",
+  "public": true,
+  "default": 0,
+  "description": "Hourly irrigation duration in minutes."
+}
--- a/devices/generic/items/state_real_time_irrigation_duration_item.json
+++ b/devices/generic/items/state_real_time_irrigation_duration_item.json
@@ -1,0 +1,9 @@
+{
+  "schema": "resourceitem1.schema.json",
+  "id": "state/real_time_irrigation_duration",
+  "datatype": "UInt32",
+  "access": "R",
+  "public": true,
+  "default": 0,
+  "description": "Real-time irrigation duration in minutes."
+}
--- a/devices/sonoff/swv_zne_smart_water_valve_lite.json
+++ b/devices/sonoff/swv_zne_smart_water_valve_lite.json
@@ -25,8 +25,15 @@
         { "name": "attr/swversion" },
         { "name": "attr/type" },
         { "name": "attr/uniqueid" },
-        { "name": "state/reachable" },
-        { "name": "state/on" }
+        {
+          "name": "state/alert",
+          "default": "none"
+        },
+        {
+          "name": "state/on",
+          "refresh.interval": 360
+        },
+        { "name": "state/reachable" }
       ]
     },
     {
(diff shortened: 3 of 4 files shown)


mattreim:
Please create a device request first, as we cannot evaluate the device in its current state.

mattreim:
First of all, the PR is very cluttered, which makes debugging quite difficult. The `general.xml` looks pretty good, but the DDF doesn't look right to me; everything is currently lumped into the battery sensor, which isn't how it should be. Perhaps @Smanar has an idea on how to implement the DDF correctly.

Smanar:
First question what is your third app ^^ ? Because if you are using HA you will probably need some time before having "state/hour_irrigation_duration", and the device name will have importance too for support.

On the sensor name, @mattreim  is right, you don't prefer a ZHAValve ?
Need to edit constants.json and create generic\subdevices\valve_sensor.json

On the lidl\hg06875.json an user have used a ZHAOpenClose
On the sonoff\swv_smart_water_valve.json or _TZE284_7ytb3h8u_water_valve.json,  users have used a ZHAWater
They have put some field in the "light" entry as "config" field, it's logic as this device is more a "light" than a "sensor"

On sensors, "state" are read only, if you want a read/write field, better to use "config", the method is the same, except "config" are not stored on the same way than "state" on database.

I'm looking the code, it's impressive, you are sure "config/duration" is working to write values ?

And does the device make report itself ?
Because I don't see something special set in report.
If the device make itself naturally report can remove the "read" and "refresh.interval", else deconz will poll the device, not sure it's usefull ?
```
          "read": {
            "fn": "none"
          },
```

`"sleeper": true,` this device is probably not a sleeper.
`
"uuid": "d3d5f8d8-8c3c-4e4c-9f1f-9c6d6c3f2a11",`
This line is added by bot, you can delete it.

Finne75:
I think there should be a "ZHAValve" device type. Yes, third app is HA. 

Yes, writing to config/duration works and keeps the valve open, I tested it. I assume that 0x510D is not reported by the device, because it is basically a configuration item. at least I could not observe a report.

I updated the PR as you suggested but without ZHAValve since I assume this requires a bit more work:
- Removed source uuid.
- Set "sleeper": false.
- Moved config/locked to the /lights On/Off output.
- Moved config/duration out of ZHABattery into a new ZHAOpenClose sensor, matching the existing Lidl/Tuya valve pattern.
- Kept ZHABattery as battery-only.
- Removed state/real_time_irrigation_duration and state/hour_irrigation_duration.

@Smanar I found a few DDF/runtime findings while testing this device that may be useful beyond this PR. Is there a preferred way to send you a longer note privately, e.g. Discord or email? Otherwise I can also simply put them in here :)

Smanar:
The problem if you create a ZHAValve, it's fine, good idea, but you will need to wait for HA support it too.
If you put all your "stuff" in a ZHAOpenClose will be easier for you, less work for you for integration.

It's your chose I haven't problem with that, and about other devs, I m sure they will prefer a ZHAValve, but as the previous DDF was valided, why your will be not.

I m on discord too with same name "Smanar" .

If it can be usefull for other, better to use the github link https://github.com/dresden-elektronik/deconz-rest-plugin/issues/8625
Because one time a PR is valided and closed, few chance someone will make the undertaker to check information here.

Finne75:
How about we add this DDF so we have baseline support for the device, and I would cretae a feature request for ZHAValve separately?

Smanar:
No problem for me, you can propose the DDF as it.
- Changes on XML file are perfect and usefull, 100% sure others devs will  be agree.
- The DDF use a light entry + ZHABattery + ZHAopenClose, like other DDF, why it will be a problem.
- The creation of "real_time_irrigation_duration" and "real_time_irrigation_duration" on "config" will be more complex, but I m reading the DDF you are not using them ATM ?

Like that 80% of the device will be native on HA

And can re-write the same DDF to use a ZHAValve later. It will make issues on users already using the DDF, but it will just change the ZHAopenClose to a ZHAValve, and users will have a new official sensor, the ZHAValve used for all future valve, and surely old one, not only this one.

On my side I prefer a partial support than no support at all, it's personnal but waiting for all is finished mean sometime never.

Finne75:
Coming back - anything missing? Can we merge?

Smanar:
IDK, lot of PR are in waiting list, issue are closed before PR are merged.
I don't think the problem is from the PR itself.

PR #8626, opened 2026-06-08
Simplify and update client timeouts

SwoopX:


manup:
Hi guess we had a look at the same :)

Note I have removed all the TCP timeout tracking code from the plugin in https://github.com/dresden-elektronik/deconz-rest-plugin/pull/8618

The main reason is that the plugin shouldn't do this at all since the core handles this already (except Websockets which are handled by the Websocket server).

SwoopX:
@manup Oha, indeed. Definitively overlooked this one and didn't dare to remove everything in one go ^^

Closing this one in favour of #8618 then.

Issue #8613, opened 2026-05-26
Introspect for `buttonevent` doesn't list button names from DDF


ebaauw:
### Does the issue really belong here?

- [x] I definitively want to report a bug within deCONZ or its REST-API

### Is there already an existing issue for this?

- [x] I have searched the existing issues and there is none for the bug at hand

### Describe the bug

When doing an introspect API call for `buttonevents` for a device that has defined the buttons and button events in the DDF, the button names aren't used.  Instead, they are listed as `Button 1`, `Button 2`, ...

Looking at the code, it tries to get the info from the DDF first.  If none is available, it get the info from `button_maps.json`.  In the latter case, the button names are reported properly.  In the first case not.

### Steps to reproduce the behavior

Try e.g. the Bilresa remote control with buttons from https://github.com/dresden-elektronik/deconz-rest-plugin/pull/8608.
The DDF lists both `buttonevents` and `buttons`:

https://github.com/dresden-elektronik/deconz-rest-plugin/blob/e228861fceb031cee281bd206666b6884b2dac37/devices/ikea/bilresa_remote_with_buttons.json#L34-L46

However, the `buttons` aren't reported correctly:
```
$ deconz get /devices/10:35:97:00:00:ad:e7:66-01-1000/state/buttonevent/introspect
{
  "buttons": {
    "1": {
      "name": "Button 1"
    },
    "2": {
      "name": "Button 2"
    }
  },
  "type": "int32",
  "values": {
    "1001": {
      "action": "HOLD",
      "button": 1
    },
    "1002": {
      "action": "SHORT_RELEASE",
      "button": 1
    },
    "1003": {
      "action": "LONG_RELEASE",
      "button": 1
    },
    "1004": {
      "action": "DOUBLE_PRESS",
      "button": 1
    },
    "2001": {
      "action": "HOLD",
      "button": 2
    },
    "2002": {
      "action": "SHORT_RELEASE",
      "button": 2
    },
    "2003": {
      "action": "LONG_RELEASE",
      "button": 2
    },
    "2004": {
      "action": "DOUBLE_PRESS",
      "button": 2
    }
  }
}
```
After removing `buttonevents` from the DDF, the code gets the correct names from `button_maps.json`.

### Expected behavior

Introspect returns the button names from the DDF.

### Screenshots

n/a

### Environment

- Host system: (Raspberry Pi / PC / NAS / MAC) All
[...]

PR #8631, opened 2026-07-06
Add DDF zcl:cluster parse fn command

manup:
The `zcl:cluster` DDF parse function allows specifying multiple clusters as filter. Mostly useful for moving switches from button_maps.json to pure DDF implementation.

```
{"fn": "zcl:cluster", "cl": "0xFC00", "eval": "..."}
{"fn": "zcl:cluster", "cl": ["0xFC00", "0xFC01"], "script": "handler.js"}
```

Note this function doesn't support specifying `cmd`, `mf` and `attr`, these must be handled in respective Javascript handler. The `ep` endpoint parameter is 255 (any source endpoint) by default.
Seperate PRs will use this to move switches out of `button_maps.json`.

**Edit:** as suggested by @ebaauw there is also a new global Javascript function `button_event` for declarative handling similar to `button_maps.json`.

```
button_event([
  [clusterId, commandId | commandIds[], button | function]
])

```

Example for Ikea on/off switch:

```
{
          "name": "state/buttonevent",
          "parse": {
            "fn": "zcl:cluster",
            "cl": ["0x0006", "0x0008"],
            "script": "tradfri_on_off_buttonevent.js"
          },
          "awake": true
}
```
**tradfri_on_off_buttonevent.js**
```js
// move 0x01, move w. onoff 0x05, stop w. onoff 0x07
// [clusterId, commandId, button | function]
button_event([
  [6, 1, 1002],
  [6, 0, 2002],
  [8, 5, 1001],
  [8, 7, function(){ return (Item.val === 1001 ? 1003 : 2003); }],
  [8, 1, 2001]
])
```

### DIFF
--- a/device.cpp
+++ b/device.cpp
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2021 dresden elektronik ingenieurtechnik gmbh.
+ * Copyright (c) 2021-2026 dresden elektronik ingenieurtechnik gmbh.
  * All rights reserved.
  *
  * The software in this package is published under the terms of the BSD
@@ -690,7 +690,8 @@ bool DEV_ZclRead(Device *device, ResourceItem *item, deCONZ::ZclClusterId_t clus
     ZCL_Param param{};
     param.valid = 1;
     param.endpoint = sd->endpoint();
-    param.clusterId = static_cast<quint16>(clusterId);
+    param.clusterIds[0] = static_cast<quint16>(clusterId);
+    param.clusterCount = 1;
     param.attributes[0] = static_cast<quint16>(attrId);
     param.attributeCount = 1;
 
@@ -1824,7 +1825,7 @@ static void DEV_UpdateReportTracker(Device *device, const ResourceItem *item)
 
     for (size_t i = 0; i < zclParam.attributeCount && i < zclParam.attributes.size(); i++)
     {
-        ReportTracker &tracker = DEV_GetOrCreateReportTracker(device, zclParam.clusterId, zclParam.attributes[i], zclParam.endpoint);
+        ReportTracker &tracker = DEV_GetOrCreateReportTracker(device, zclParam.clusterIds[0], zclParam.attributes[i], zclParam.endpoint);
         tracker.lastReport = item->lastZclReport();
     }
 }
--- a/device_access_fn.cpp
+++ b/device_access_fn.cpp
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2021-2025 dresden elektronik ingenieurtechnik gmbh.
+ * Copyright (c) 2021-2026 dresden elektronik ingenieurtechnik gmbh.
  * All rights reserved.
  *
  * The software in this package is published under the terms of the BSD
@@ -126,36 +126,36 @@ enum DA_Constants
 
 struct ParseFunction
 {
-    ParseFunction(const QString &_name, const int _arity, ParseFunction_t _fn) :
+    ParseFunction(const std::string &_name, const int _arity, ParseFunction_t _fn) :
         name(_name),
         arity(_arity),
         fn(_fn)
     { }
-    QString name;
+    std::string name;
     int arity = 0; // number of parameters given by the device description file
     ParseFunction_t fn = nullptr;
 };
 
 struct ReadFunction
 {
-    ReadFunction(const QString &_name, const int _arity, ReadFunction_t _fn) :
+    ReadFunction(const std::string &_name, const int _arity, ReadFunction_t _fn) :
         name(_name),
         arity(_arity),
         fn(_fn)
     { }
-    QString name;
+    std::string name;
     int arity = 0; // number of parameters given by the device description file
     ReadFunction_t fn = nullptr;
 };
 
 struct WriteFunction
 {
-    WriteFunction(const QString &_name, const int _arity, WriteFunction_t _fn) :
+    WriteFunction(const std::string &_name, const int _arity, WriteFunction_t _fn) :
         name(_name),
         arity(_arity),
         fn(_fn)
     { }
-    QString name;
+    std::string name;
     int arity = 0; // number of parameters given by the device description file
     WriteFunction_t fn = nullptr;
 };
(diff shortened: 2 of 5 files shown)


ebaauw:
I like this!  After fixing https://github.com/dresden-elektronik/deconz-rest-plugin/issues/8613, we can ditch `button_maps.json`.

I would like to take a more declarative approach: define a list of `buttonevent` mappings per device and a generic function to evaluate the Zigbee command against that list.  By using inline functions for exceptions (like the direction), the list entries remain simple: cluster, command, value.  Not sure if we have the equivalent of `require` or `import` in our JavaScript environment, though.

The device-specific `.js` would be:
```javascript
var button_event = require('button_event.js')

button_event([
  [0x0006, 0x01, 1002],
  [0x0006, 0x00, 2002],
  [0x0008, [0x01, 0x05], () => { return ZclFrame.at(0) === 0x00 ? 1001 : 2001 }],
  [0x0008, [0x03, 0x07], () => { return Item.val === 1001 ? 1003 : 2003 }]
])
```

The generic javascript would be:
```javascript
function button_event_entry(cluster, cmd, val) {
  if (ClusterId === cluster) {
    if (ZclFrame.cmd === cmd || (Array.isArray(cmd) && cmd.includes(ZclFrame.cmd))) {
      return (typeof val === 'function') ? val() : val
    }
  }
  return null
}

function button_event (list) {
  for (var entry of list) {
    var event = button_event_entry(entry[0], entry[1], entry[2])
    if (event != null) {
      Item.val = event
      return
    }
  }
}

module.export = button_event
```
Note the use of JavaScript's weak typing to specify command as a simple value or an array, and value as a simple value or inline function.

manup:
Good idea I like the declarative approach, it's much cleaner than my if/if else mess. The option to specify a function as third parameter should cover most if not all switches.

I'll update the PR to make your `button_event()` available at the global scope, no extra import needed. Also gonna look at #8613 asap.

manup:
With the new `button_event` function the handler for Ikea on/off switch looks friendlier.
I changed your function a bit to integrate it in DucktapeJS which lacks a few modern JS features like "for ... in ...", Array.includes and arrow functions, otherwise following version does the same:

```C
const char *PF_button_event = "function button_event(list) {"
                                  "for (var i = 0; i < list.length; i++) { var e = list[i];"
                                  "if (!Array.isArray(e) || e.length !== 3) return;"
                                  "if (ClusterId === e[0] && (ZclFrame.cmd === e[1] || (Array.isArray(e[1]) && e[1].indexOf(ZclFrame.cmd) >= 0))) {"
                                  "Item.val = typeof e[2] === 'function' ? e[2]() : e[2]; return;"
                                  "}} }";

    if (duk_peval_string(ctx, PF_button_event) != 0)
    {
        const char *str = duk_safe_to_string(ctx, -1);
        DBG_Printf(DBG_JS, "failed to define button_event: %s\n", str);
    }
    duk_pop(ctx);
```

Sucessfully tested with Ikea on/off switch. The `state/buttonevent` handler becomes:

```js
// move 0x01, move w. onoff 0x05, stop w. onoff 0x07
// [clusterId, commandId, button | function]
button_event([
  [6, 1, 1002],
  [6, 0, 2002],
  [8, 5, 1001],
  [8, 7, function(){ return (Item.val === 1001 ? 1003 : 2003); }],
  [8, 1, 2001]
])
```

Imho that's less noise compared to former button_maps.json entry:

```json
"ikeaOnOffMap": {
            "vendor": "IKEA",
            "doc": "TRÅDFRI on/off switch",
            "modelids": ["TRADFRI on/off switch"],
            "buttons": [
                {"S_BUTTON_1": "On"},
                {"S_BUTTON_2": "Off"}
            ],
            "map": [
                [1, "0x01", "ONOFF", "ON", "0", "S_BUTTON_1", "S_BUTTON_ACTION_SHORT_RELEASED", "On"],
                [1, "0x01", "LEVEL_CONTROL", "MOVE_WITH_ON_OFF", "0", "S_BUTTON_1", "S_BUTTON_ACTION_HOLD", "Move up (with on/off)"],
                [1, "0x01", "LEVEL_CONTROL", "STOP_WITH_ON_OFF", "0", "S_BUTTON_1", "S_BUTTON_ACTION_LONG_RELEASED", "Stop (with on/off)"],
                [1, "0x01", "ONOFF", "OFF", "0", "S_BUTTON_2", "S_BUTTON_ACTION_SHORT_RELEASED", "Off"],
                [1, "0x01", "LEVEL_CONTROL", "MOVE", "1", "S_BUTTON_2", "S_BUTTON_ACTION_HOLD", "Move down"],
                [1, "0x01", "LEVEL_CONTROL", "STOP_WITH_ON_OFF", "1", "S_BUTTON_2", "S_BUTTON_ACTION_LONG_RELEASED", "Stop"]
            ]
        },
```

ebaauw:
Cool, looking good.

Note that I used arrays for the _Move_ and _Move (with On/Off)_ as well as for the _Stop_ and _Stop (with On/Off)_ commands.  I've seen devices changing their behaviour with newer firmware, as well as with clones of the same OEM device.  I find it weird that the IKEA switch sends _Move_ (down) on hold, but _Stop (with On/Off)_ on release.  Are we sure that's correct?

ebaauw:
Just realised, we have some switches where the endpoint is used as well.  Maybe add a fourth element to the list entries?  Logically the endpoint should be the first element, but only a few devices would actually use different endpoints.  And probably need an array value for `ep` as well to handle these.  Or can we leave out `ep` to match all endpoints?

manup:
The DateCode of my switch is 20230308 and SwBuildId is 24.4.6, not sure what is different on older/newer versions. Above code does work for all events here. Note the original button map also only had used ZCL commands 0x00, 0x01 and 0x01, 0x05, 0x07 for level control cluster. I also noticed that 0x07 doesn't have any ZCL payload it's just the commandId.

We can add add endpoint to the function (or provide different function since it isn't the common case), there is also the global property `SrcEp` which can be used in a button function.

ebaauw:
> The `ep` endpoint parameter is 255 (any source endpoint) by default.
> there is also the global property `SrcEp` which can be used in a button function.

That should be good enough, I guess.

PR #8635, opened 2026-07-13
Update light_zb3_C_festavia.json, copy/pasta LCX016 support

arahayrabedian:
Attempt at adding in LCX016 support, dunno if it's this simple.

I have missing features on the LCX016 that appear for LCX015, if you need more info, i can provide, but only have phoscon setup on a headless pi.


ebaauw:
That should probably do it.  The light itself reports its capabilities (like number of segments) and the existing DDF uses this.

arahayrabedian:
Seems to work for my lcx016, shows more options than my lcx015 in home assistant 👍 , anything else i need to do to help get this merged?

PR #8642, opened 2026-08-08
Add scene list to get scene membership response command

SwoopX:
This is a prerequisite to display all scenes for a group a device is member of correctly as a list.

Depends on https://github.com/dresden-elektronik/deconz-lib/pull/10


ebaauw:
`listSize` in `general.xml` is new to me.  What does it mean?  I see three instances now, one with value `@0x0001`, one with `0x0001` and one with `0x01`.  Is it supposed to link to the attribute that holds the number of entries in the list?

SwoopX:
@ebaauw Seems like there's always a small little surprise waiting; wasn't expecting to see that as well. Feels like this is a little trick to deal with the get scene/group membership response. Both command responses have a count value and `listSize` more or less stores it. Additionally, it determines the datatype of the values to follow the count (not sure it the value even matters at all here?), 8 or 16 bit...
For scenes, it was presumably forgotten and it is applicable to only those 2 commands anyway

> Is it supposed to link to the attribute that holds the number of entries in the list?

Yes, indeed

PR #8649, opened 2026-09-04
Enhancement, Tuya HS03 temperature and humidity sensor DDF

FloHome:
DDF for AliExpress Tuya HS03 temp and Humidity sensor


FloHome:
Manufacturer: _TZE284_9ern5sfh
Model: TS060

Tested with deCONZ 2.28.1 and ConBee II.

Temperature is exposed on endpoint 1 (0x0402).
Humidity is exposed on endpoint 2 (0x0405).

The DDF has been tested successfully with Domoticz and Deconz Plugin.
Works like a ORVIBO ST30.

mattreim:
To me, it looks like this device is a clone. You just need to add `manufacturername` and `modelid` to this [DDF](https://github.com/dresden-elektronik/deconz-rest-plugin/blob/0e85b812815377c88df3fe5e73e1bffa8e1b0bc9/devices/tuya/_TZE200_TS0601_humidity_temp.json#L14).

FloHome:
I've tried to modify this one with manufacturername and modelid but it was not working... 
Endpoint for humidity must be different than temperature to work with deconz plugin for Domoticz and the scaling is different, x10 in mine instead of x100.

mattreim:
Okay, then that is the correct DDF[_TZE200_bjawzodf_humidity_temp.json](https://github.com/dresden-elektronik/deconz-rest-plugin/blob/0e85b812815377c88df3fe5e73e1bffa8e1b0bc9/devices/tuya/_TZE200_bjawzodf_humidity_temp.json#L7). Additionally, the old DDF must be removed, and it may also be necessary to delete and re-pair the device.

